PR #16, opened 2017-08-12
Fixes Datadogs API quirk described in #3

### LINKED ISSUE #3: Datadog monitors - 'metric alert' versus 'query alert'
_This issue was originally opened by @kmshultz as hashicorp/terraform#13784. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
$ terraform version
Terraform v0.9.2
```

### Affected Resource(s)
- datadog_monitor

```hcl
resource "datadog_monitor" "outliermonitor" {
  name = "test outlier monitor"
  type = "metric alert"
  query = "avg(last_5m):outliers(avg:chef.resources.updated{env:prod} by {host}, 'DBSCAN', 3.0) > 0"
  message = "chef resources outlier alert"
}
```

### Debug Output

https://gist.github.com/kmshultz/23bf92aea315a1596d21920092da1c11

### Expected Behavior

Terraform ought not to update the `datadog_monitor` on every `apply`.

### Actual Behavior

Terraform updated the `datadog_monitor` because while it passed 'metric alert' to the Datadog API on creation, the API categorized it — because of its more complex `query` — as 'query alert'.

This is a quirk on Datadog's side, **not a bug in Terraform**. When a client POSTs or PUTs a Datadog monitor with type 'metric alert', the Datadog API inspects the query and sometimes recategorizes it as 'query alert' (e.g. for APM monitors, Outlier monitors, monitors that do arithmetic on two or more metrics). However, clients are NOT advised to pass 'query alert' in POST/PUTs — it's up to Datadog to decide when a 'metric alert' should be recategorized. 'query alert' is only important for the Datadog monitoring backend to know about, but the type is sort of exposed to clients right now, and this isn't likely to change any time soon.

So, it's hacky, but the `datadog_monitor` resource should probably look for 'query alert' in its Read function and override it to 'metric alert'. Without this override, `terraform apply` will always try to PUT a 'query alert' monitor back to 'metric alert', fail to do so (though Datadog will still return 2xx), and keep trying to set it to 'metric alert' every time.

### Steps to Reproduce

See Debugging output gist above.

MrMarvin:
This fixes #3 / https://github.com/hashicorp/terraform/issues/13784.

Fixed as initially suggested by @kmshultz, thanks for bringing this up!

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -51,6 +51,14 @@ func resourceDatadogMonitor() *schema.Resource {
 			"type": {
 				Type:     schema.TypeString,
 				Required: true,
+				// Datadog API quirk, see https://github.com/hashicorp/terraform/issues/13784
+				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+					if oldVal == "query alert" && newVal == "metric alert" {
+						log.Printf("[DEBUG] Monitor '%s' got a '%s' response for an expected '%s' type. Suppressing change.", d.Get("name"), newVal, oldVal)
+						return true
+					}
+					return newVal == oldVal
+				},
 			},
 
 			// Options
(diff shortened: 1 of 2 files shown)


grubernaut:
Hey @MrMarvin, thanks for the contribution! 
We can, however, add this as a `DiffSuppressFunc` to this resources schema, instead of adding additional items to the Read function. 
https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L59-L66
https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L195-L200

This allows us to fully know that the suppression is happening, without having to find where it's happening inside of the Read function.

MrMarvin:
Thanks @grubernaut for the hint to `DiffSuppressFunc`. A much cleaner solution! 
I just pushed a second commit that reworks the modification to make use of such a function.

MrMarvin:
setting it to WIP for now, found another issue that needs investigation before merging.

MrMarvin:
removing 'WIP', double checked and its fine, from my point of view at least. 
Is there anything missing?

MrMarvin:
Hi @radeksimko, @grubernaut, is there anything I can do to help get this merged? Please let me know and thanks!

brendanjerwin:
Whats the release timeline/process? I'd love to get to use this fix!

brendanjerwin:
Awesome! thx. Much quieter now. :) 

(I think I found another case of a similar nature though, I'll make a new issue for it: ` graph.0.style.palette_flip:                                          "false" => "0"`

Issue #50, opened 2018-02-19
timeboard resource missing support for some field names in "marker" block


robinbowes:
I'm attempting to create DataDog timeboards from terraform.

I'm exporting the definitions using the DataDog API and parsing the output to generate terraform config.

It seems that the markers section in the graph definition block supports some field names that are not included in terraform, eg:

```json
            "markers": [
              {
                "max": null,
                "min": 1,
                "type": "error solid",
                "value": "y > 1.0",
                "dim": "y"
              },
              {
                "max": 1,
                "min": 0.8,
                "type": "warning solid",
                "value": "0.8 < y < 1.0",
                "dim": "y"
              },
              {
                "max": 0.8,
                "min": 0,
                "type": "ok dashed",
                "value": "0 < y < 0.8",
                "dim": "y"
              }
            ]
```

Is this just a case of the provider not keeping up with changes in DataDog capabilities?

tmack8001:
I'm also seeing similar when not specifying label values in the Datadog UI which then breaks our CI/CD pipelines for terraform changes.

```
* module.main-timeboard.datadog_timeboard.main-dashboard: datadog_timeboard.main-dashboard: json: cannot unmarshal bool into Go struct field GraphDefinitionMarker.label of type string
```

here is the example marker block as defined in the Datadog API for the graph that breaks terraform builds:

```
"markers": [
    {
      "val": "250000000",
      "value": "y = 250000000",
      "type": "error dashed",
      "label": true
    },
    {
      "val": "200000000",
      "value": "y = 200000000",
      "type": "warning dashed",
      "label": true
    }
  ]
```

This "label: true" is the syntax for when a string isn't provided in the "label" field in the UI / API. I'm guessing the go http library used by this provider is expecting this to always be a string and sometimes (when no text label is provided) breaks unmarshalling with the expected struct.

masci:
@robinbowes setting `dim`, `min` and `max` explicitly is not actually needed, to set a range between 0.8 and 1.0 on y, all you have to do is the following 
```hcl
    marker {
      label = "High Latency"
      type  = "error solid"
      value = "0.8 > y > 0"
    }
```

masci:
[...]

PR #51, opened 2018-02-19
Add additional timeboard graph request marker parameters

robinbowes:
This addresses #50.

I don't know if any of the previous parameters have been deprecated or not, but this change allows me to generate marker sections that match those generated using the DataDog web interface.


masci:
Hi @robinbowes thanks for the PR.
I have the same result by just defining the marker like this:
```hcl
    marker {
      label = "High Latency"
      type  = "error solid"
      value = "0.8 > y > 0"
    }
```
I'm not sure this change is needed, can you try and let me know?

robinbowes:
I'm not currently working on datadog, but I'll try and make time to give this a try.

Cheers,

R.

PR #55, opened 2018-03-14
Update go-datadog-api to include support for integrations/pagerduty api

rlhh:


rlhh:
Sorry, I actually intended to create this on my own fork. Will create a proper PR here once everything is done.

PR #59, opened 2018-04-04
document datadog_timeboard "id" attribute

jantman:
I was trying to figure out a way to find the URL to a timeboard created via this provider. By experimentation, I found that the Datadog timeboard ID is set as the resource ID and therefore exposed as an attribute, though this wasn't documented.


jantman:
Thanks, @masci !

PR #92, opened 2018-09-06
Fixes #19 by updating go-datadog-api lib version

### LINKED ISSUE #19: cannot unmarshal string into Go struct field Yaxis.min of type float64 after upgrading to 0.10.3
Running terraform plan or apply over existent terraform datadog timeboards resulted in a  ```json: cannot unmarshal string into Go struct field Yaxis.min of type float64 ``` error

Note that despite the error is happening with terraform v0.10.3, I was able to push changes to datadog with this version of terraform until 31/08. But since early September, it has started to fail with that error.

Deleting existent terraform.state and re running plan or apply using the same timeboard configuration works. But doing this will duplicate information in ddog as it will recreate all resources.

The following is how the graph looks atm:
```graph {
    title = "graph title"
    viz = "timeseries"
    autoscale = true
    precision = "3"
    request {
      q = "the-query"
      type = "area"
      style {
        palette = "grey"
      }
      aggregator = "sum"
    }
    marker {
      type = "error dashed"
      value = "y = 99.95"
      label = "99.95"
    }
    yaxis {
      min = "99"
      max = "100"
    }
  }
```
And in terraform.state it is shown as
```
"graph.1.yaxis.%": "2",
"graph.1.yaxis.max": "100",
"graph.1.yaxis.min": "99",
```

### Terraform Version
```v0.10.3```

### Affected Resource(s)
```datadog_timeboard```

### Expected Behavior
The state file should  have been modified with the new added configuration in the datadog timeboard

### Actual Behavior
Terraform plan or apply is failing. The error reads ``` json: cannot unmarshal string into Go struct field Yaxis.min of type float64```

### Steps to Reproduce
1. Have an existent terraform state containing datadog timeboards
2. Upgrade to the indicated version
3. Run terraform plan or apply

nyanshak:
* v2.13.0 of go-datadog-api library adds a custom unmarshaller for Yaxis
struct, which can properly handle the issue where strings (e.g. "99")
for yaxis.max would cause the tf provider to bail out.

### DIFF
--- a/vendor/github.com/zorkian/go-datadog-api/dashboards.go
+++ b/vendor/github.com/zorkian/go-datadog-api/dashboards.go
@@ -52,9 +52,56 @@ type GraphEvent struct {
 }
 
 type Yaxis struct {
-	Min   *float64 `json:"min,omitempty"`
-	Max   *float64 `json:"max,omitempty"`
-	Scale *string  `json:"scale,omitempty"`
+	Min     *float64 `json:"min,omitempty"`
+	AutoMin bool     `json:"-"`
+	Max     *float64 `json:"max,omitempty"`
+	AutoMax bool     `json:"-"`
+	Scale   *string  `json:"scale,omitempty"`
+}
+
+// UnmarshalJSON is a Custom Unmarshal for Yaxis.Min/Yaxis.Max. If the datadog API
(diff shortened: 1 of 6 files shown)


masci:
Hi @nyanshak we've the same change in #91 that's about to be merged but thanks for taking the time to address the issue!

PR #113, opened 2018-10-31
Use dates to set downtime interval

### LINKED ISSUE #69: datadog_downtime is unusable with start and end times
_This issue was originally opened by @ssonne as hashicorp/terraform#18081. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
$ terraform -v
Terraform v0.11.3
+ provider.datadog v1.0.3
+ provider.local v1.1.0
```

```
resource "datadog_downtime" "my_downtime" {
  scope = ["*"]
  start = 1526684400
  end   = 1526731200

  recurrence {
    type   = "days"
    period = 1
  }
  monitor_id  = "${datadog_monitor.my_monitor.id}"
}
```

### Debug Output

```
  ~ datadog_downtime.my_downtime
      active:                   "true" => "false"

...

* datadog_downtime.my_downtime: 1 error(s) occurred:

* datadog_downtime.my_downtime: error updating downtime: API error 400 Bad Request: {"errors":["Scheduled downtime start cannot be in the past"]}
```

### Expected Behavior

- `active` should not be considered changed - it changes daily from true to false as the downtime window passes, so it is getting updating all the time for no reason.
- `start` and `end` should be modified by stepping forward in time until they are in the future, or else there should be a way to write the time of day in relative terms. Configuring the window using POSIX timestamps makes it really hard to modify the downtime, because any change requires looking up the POSIX timestamp of the next start and end times of the downtime (to avoid the "Scheduled downtime start cannot be in the past" error). Combined with `active` being considered a change, the same config will succeed during part of the day and fail during the downtime.

### Actual Behavior

Unnecessary change to the downtime's status, and an error.

### Steps to Reproduce

1. Configure a recurring downtime using `start` and `end`.
2. `terraform init`
3. `terraform apply`
4. Wait until the downtime has started.
5. `terraform apply`

### References

https://github.com/hashicorp/terraform/pull/10994

masci:
Setting downtime intervals using Unix timestamps can be very annoying, this PR adds two new arguments `start_date` and `end_date` that serve the same purpose as `time` and `date` but let users specify a datetime string in RFC3339 format, like:
```
  start_date = "2018-10-31T11:11:00+01:00"
  end_date = "2018-10-31T21:00:00+01:00"
```

Also suppressed the diff for `active` when the downtime has a `recurrence` set. This is to avoid TF complaining about `active` changing if ran during downtime window. The implementation is a bit simplistic but should work for the average use case.

Fixes #69


datadog/resource_datadog_downtime_test.go
```diff
@@ -300,6 +300,49 @@ func testAccCheckDatadogDowntimeExists(n string) resource.TestCheckFunc {
 	}
 }
 
+func TestAccDatadogDowntimeDates(t *testing.T) {
```
nmuesch: 💜

website/docs/r/downtime.html.markdown
```diff
@@ -23,6 +23,12 @@ resource "datadog_downtime" "foo" {
     type   = "days"
     period = 1
   }
+
+  # Datadog API will reject dates in the past so let's ignore `start` and `end`
+  # arguments during lifecycle
+  lifecycle {
+    ignore_changes = ["start", "end"]
```
nmuesch: This is likely attached to my initial question around the diffsuppression but not totally clear to me why this needs suppressing?

nmuesch:
Would it also be possible/useful to test the case where both are present and ensure we get a conflict error? (Not sure the amount of effort required there)

datadog/resource_datadog_downtime.go
```diff
@@ -25,14 +27,42 @@ func resourceDatadogDowntime() *schema.Resource {
...
 			},
 			"disabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"start": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+					_, endDatePresent := d.GetOk("start_date")
```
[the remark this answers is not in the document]
masci: If you look at the state file `terraform.tfstate` after an apply, you'll see `start` is there even if you're using `start_date`. This means `plan` would always tell you `start` has changed even if you're only using `start_date` in your `.tf` file.

website/docs/r/downtime.html.markdown
masci: This is because Datadog API doesn't accept dates in the past at the moment you `POST` your request. If you don't ignore `start` after the first `apply`, terraform always tries to `POST` it making the request fail.

nmuesch: Discussed this offline. The easiest scenario for when one wants to update/extend a downtime would be to comment out this `ignore` section, apply a new time, and re place the `ignore` section. 

Maybe it'd be helpful to update the comment above with:

```
  # Datadog API will reject dates in the past so let's ignore `start` and `end`
  # arguments during lifecycle. To update or extend an existing downtime, temporarily
  # remove the `ignore` section, apply timestamp changes, and re-apply the `ignore` section.
```
or something similar?

PR #122, opened 2018-11-22
datadog_integration_aws

hryamzik:
Created `datadog_integration_aws`. Update operations are not supported so there's a workaround to recreate resource on each change.

I'm also not very confident on import operation, state file requires manual edit to put `external_id`.

### DIFF
--- a/datadog/resource_datadog_integration_aws.go
+++ b/datadog/resource_datadog_integration_aws.go
@@ -33,21 +33,25 @@ func resourceDatadogIntegrationAws() *schema.Resource {
 			"role_name": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true, // waits for update API call support
 			},
 			"filter_tags": {
 				Type:     schema.TypeList,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				ForceNew: true, // waits for update API call support
 			},
 			"host_tags": {
 				Type:     schema.TypeList,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				ForceNew: true, // waits for update API call support
 			},
 			"account_specific_namespace_rules": {
 				Type:     schema.TypeMap,
 				Optional: true,
 				Elem:     schema.TypeBool,
+				ForceNew: true, // waits for update API call support
 			},
 			"external_id": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


coretemp:
To the maintainer: 

Please do not merge anything which requires a *manual* edit or where the word "not confident" appears in the PR or where updates are not supported.

hryamzik:
Any ideas on providing `external_id` as an input parameter?

coretemp:
My suggestion is to work together with Datadog in order to make something that is workable. If Datadog is not willing to "play ball", then we can just abandon their services. I don't mind. If they don't build what the market wants, we do something else.

hryamzik:
@coretemp implemented import of `external_id` field with an env variable set, is it OK now?

website/docs/r/integration_aws.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+sidebar_current: "docs-datadog-resource-integration_aws"
+description: |-
+  Provides a Datadog - Amazon Web Services integration resource. This can be used to create and manage the integrations.
+---
+
+# datadog_integration_aws
+
+Provides a Datadog - Amazon Web Services integration resource. This can be used to create and manage Datadog - Amazon Web Services integration.
+
+Update operations are currently not supported with datadog API so any change forces a new resource.
```
masci: We should probably make this more evident and also explain the workaround of using `${account_id}_${role_name}` in the state in place of the real identifier.

website/docs/r/integration_aws.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `external_id` - AWS External ID
+
+## Import
+
+Amazon Web Services integrations can be imported using their account ID and role name, e.g.
```
masci: I'd say something on this line:
> ...can be imported using their account ID and role name to compose the integration id, while the `external_id` can be passed by setting an environment variable called `EXTERNAL_ID`

iancward:
@hryamzik Datadog normally provides the `external_id` as part of the create process for an integration. I think the best bet is to expose it as an attribute so that it can be consumed elsewhere.

datadog/resource_datadog_integration_aws.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	}
+	iaws.FilterTags = filterTags
+	iaws.HostTags = hostTags
+	iaws.AccountSpecificNamespaceRules = accountSpecificNamespaceRules
+	return iaws
+}
+
+func resourceDatadogIntegrationAwsCreate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[WARN] debugging logger")
```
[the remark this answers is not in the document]
hryamzik: My bad =)

hryamzik:
> @hryamzik Datadog normally provides the external_id as part of the create process for an integration. I think the best bet is to expose it as an attribute so that it can be consumed elsewhere.

@iancward that's already implemented. And it's set using env variable during import process as well.

website/docs/r/integration_aws.html.markdown
hryamzik: What do you mean by real identifier? There's no such an ID exposed via API. Should we manage IDs on our own in this case?

iancward:
That makes sense, yep. Sorry, I thought you were asking about how to support providing `external_id` to create the resource, which isn't supported.

website/docs/r/integration_aws.html.markdown
hryamzik: Basically combination of an account ID and a role name should be unique for an integration, I can't think of a better way to implement ID.

masci: My bad, I was somehow convinced we returned an id in the response payload, please ignore my comment and let's stick with your plan

datadog/resource_datadog_integration_aws.go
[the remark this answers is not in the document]
masci: 👍

website/docs/r/integration_aws.html.markdown
hryamzik: Great, I guess there's nothing preventing a merge now. =)

iancward: My only recommendation would be to use a `/` to separate account ID and Role name instead of `_`. That would be consistent with some of the multi-key import statements in the AWS provider. Example: https://www.terraform.io/docs/providers/aws/r/iam_role_policy_attachment.html#import

Well, on further investigation, it looks like they're anything but consistent...
https://github.com/terraform-providers/terraform-provider-aws/search?q=imported+separated+by+%2F&unscoped_q=imported+separated+by+%2F
If anything, I guess the `:` is the most common separator.

hryamzik: Any non-numeric char could work here, changing to a colon now.

hryamzik: Done.

hryamzik:
Rebased to get `make install` work. Is there anything else preventing a merge?

PR #123, opened 2018-11-23
Added install and uninstall make targets

hryamzik:


hryamzik:
@masci can I also ask you to review #122, please?

PR #144, opened 2019-01-11
Fix fill_min and fill_max panic issue

dabcoder:
Without this running `terraform apply` would result in a panic:

```
panic: interface conversion: interface {} is string, not float64
[...]
...datadog.setJSONNumberFromDict...
```

### DIFF
--- a/datadog/resource_datadog_screenboard.go
+++ b/datadog/resource_datadog_screenboard.go
@@ -598,7 +598,15 @@ func setBoolFromDict(dict map[string]interface{}, key string, field **bool) {
 // For setJSONNumberFromDict, dict[key] is expected to be a float64
 func setJSONNumberFromDict(dict map[string]interface{}, key string, field **json.Number) {
 	if v, ok := dict[key]; ok {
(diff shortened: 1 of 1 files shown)


datadog/resource_datadog_screenboard.go
```diff
@@ -598,7 +598,15 @@ func setBoolFromDict(dict map[string]interface{}, key string, field **bool) {
 // For setJSONNumberFromDict, dict[key] is expected to be a float64
 func setJSONNumberFromDict(dict map[string]interface{}, key string, field **json.Number) {
 	if v, ok := dict[key]; ok {
-		f := json.Number(strconv.FormatFloat(v.(float64), 'e', -1, 64))
+		// style fields can be numbers or strings so we need to handle both types
+		var number string
+
+		if val, ok := v.(float64); ok {
+			number = strconv.FormatFloat(val, 'e', -1, 64)
+		} else {
```
gzussa: I think you can remove the `else` part here and just set `number = v`

PR #152, opened 2019-02-19
Make the `metric_metadata` docs example adhere to `terraform fmt`

issyl0:
- This was the only example that wasn't `terraform fmt`ed so here's a
  fix. :-)


bkabrda:
LGTM, thanks! Merging.

PR #154, opened 2019-02-22
Change monitor element schema to TypeString

mnaboka:
The element type is set to `schema.TypeInt`, which does not allow us to
do interpolation. By changing this type to string all seem to work
correctly.

Our terraform configuration
```
monitor {
  id = "${module.poller_latency.id}"
}
```

The error we are seeing is this:
```
widget.0.monitor (id): cannot parse '' as int: strconv.ParseInt: parsing "${module.poller_latency.id}": invalid syntax
```

Feel free to close this PR if this is a not relevant / breaking change, i could be missing context here.

### DIFF
--- a/datadog/resource_datadog_screenboard.go
+++ b/datadog/resource_datadog_screenboard.go
@@ -499,7 +499,7 @@ func resourceDatadogScreenboard() *schema.Resource {
 				"monitor": {
 					Type:     schema.TypeMap,
 					Optional: true,
-					Elem:     &schema.Schema{Type: schema.TypeInt},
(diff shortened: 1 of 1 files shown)


bkabrda:
This PR seem to be correct. Testing locally without your PR, I can't even use a literal int as id, as the code in `buildWidgets` expects it to be a string and that causes Terraform to crash. I don't see how anyone could actually be using this, hence I don't see a possibility to break backwards compatibility. Using your PR makes everything work, so I'm merging it.

Thanks a lot for submitting!

PR #168, opened 2019-04-04
[dashboard] Add support for the new API

MLaureB:
This PR adds a new `datadog_dashboard` resource that uses the [new Dashboard API](https://docs.datadoghq.com/api/?lang=python#dashboards) in Datadog Terraform Provider.

Note that this work depends on [the PR that adds support for the new Dashboard API in the Go API library](https://github.com/MLaureB/go-datadog-api/pull/2).


bkabrda:
LGTM :+1:

bkabrda:
After some offline talking and evaluating whether the approach with JSON blobs is ok or not, we decided not to go forward with that. I am therefore closing this PR and we'll try to redo it in the future using a non-JSON blob approach (so fully providing access to all params in the resource).

PR #169, opened 2019-04-05
Support Synthetics Tests

gabsn:
This PR will add support for CRUD operations on API Synthetics Tests.

We need to wait for [this PR](https://github.com/zorkian/go-datadog-api/pull/228) to be merged before merging this one.

### TODO
- [x] support for API tests
- [x] support for Browser tests (step creation not included)
- [x] support for import
- [x] update TFP doc

## Testing guidelines

Either run `make testacc` to run acceptance tests or try the CLI tool:
```sh
# In terraform-provider-datadog
make build
cp ~/go/bin/terraform-provider-datadog ~/.terraform.d/plugins/

# In your infra repo
terraform init
terraform apply
```

With a basic `main.tf` terraform config file:
```tf
# Configure the Datadog provider
provider "datadog" {
  api_key = "XXX"
  app_key = "XXX"
}

resource "datadog_synthetics_test" "foo" {
  type = "api"

  request {
	  method = "GET"
	  url = "https://docs.datadoghq.com"
  }
  assertions = [
    {
      type = "statusCode"
      operator = "is"
      target = "200"
  	},
    {
      type = "responseTime"
      operator = "lessThan"
      target = "2000"
  	}
  ]

  locations = [ "aws:eu-central-1", "aws:ap-northeast-1" ]
  options {
	tick_every = 60
  }

  name = "name for synthetics test foo"
  message = "Notify @datadog.user"
  tags = ["foo:bar", "baz"]

  paused = true
}
```

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -85,7 +85,7 @@ func syntheticsTestRequest() *schema.Schema {
 			Schema: map[string]*schema.Schema{
 				"method": {
 					Type:     schema.TypeString,
-					Optional: true,
+					Required: true,
 				},
 				"url": {
 					Type:     schema.TypeString,
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -85,7 +85,7 @@ func syntheticsTestRequest() *schema.Schema {
 			Schema: map[string]*schema.Schema{
 				"method": {
 					Type:     schema.TypeString,
-					Required: true,
+					Optional: true,
 				},
 				"url": {
 					Type:     schema.TypeString,
@@ -98,7 +98,7 @@ func syntheticsTestRequest() *schema.Schema {
 				"timeout": {
 					Type:     schema.TypeInt,
 					Optional: true,
-					Default:  0,
+					Default:  60,
 				},
 			},
 		},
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -23,3 +24,5 @@ require (
 	golang.org/x/net v0.0.0-20181023162649-9b4f9f5ad519 // indirect
 	google.golang.org/genproto v0.0.0-20181016170114-94acd270e44e // indirect
 )
+
+replace github.com/zorkian/go-datadog-api => github.com/DataDog/go-datadog-api v0.0.0-20190403100613-64d00d08241a
```
gabsn: This should go away before merging the PR

datadog/resource_datadog_synthetics_check.go
```diff
@@ -0,0 +1,272 @@
...
+				assertion.Type = &assertionType
+			}
+			if v, ok := assertionMap["operator"]; ok {
+				assertionOperator := v.(string)
+				assertion.Operator = &assertionOperator
+			}
+			if v, ok := assertionMap["target"]; ok {
+				assertionTarget := v.(string)
+				t, _ := strconv.Atoi(assertionTarget)
+				assertion.Target = t
```
tanguylebarzic: target may not always be an int, right?

datadog/resource_datadog_synthetics_check.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+func syntheticsTestOptions() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeMap,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"tick_every": {
```
tanguylebarzic: Please add all options we support (I'm thinking about min_failure_duration and min_location_failed).

datadog/resource_datadog_synthetics_check.go
```diff
@@ -0,0 +1,272 @@
...
+		},
+	}
+}
+
+func syntheticsTestRequest() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeMap,
+		Required: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
```
tanguylebarzic: Please add all parameters we support.

irabinovitch:
@ojongerius Hope you're doing well. We'd love your input on our approach in this PR. Let us know if theres anyway we (the Datadog team) can get more involved in the project.

gzussa:
Let's update the API Go Client upstream.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -0,0 +1,432 @@
...
+		// Note that Id won't be set, so no state will be saved.
+		return fmt.Errorf("error creating synthetics test: %s", err.Error())
+	}
+
+	// If the Create callback returns with or without an error without an ID set using SetId,
+	// the resource is assumed to not be created, and no state is saved.
+	d.SetId(createdSyntheticsTest.GetPublicId())
+
+	// Call resume/pause webservice, because it is a dedicated endpoint apart from classical CRUD operations
+	updateSyntheticsTestLiveness(d, client)
```
bkabrda: The `updateSyntheticsTestLiveness` should have a return value for errors. It's ok to return errors from Create/Update functions as mentioned in [1], even when the resource is created/updated. This is called partial state and I think it fits nicely with what is done with the `paused` attribute being set via additional call.

[1] https://www.terraform.io/docs/extend/writing-custom-providers.html#error-handling-amp-partial-state

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -0,0 +1,432 @@
...
+		case bool:
+			if v {
+				valStr = "1"
+			} else {
+				valStr = "0"
+			}
+		case int:
+			valStr = strconv.Itoa(v)
+		case float64:
+			valStr = strconv.Itoa(int(v))
```
bkabrda: Are there really some float values returned by the API? Because if so, this line will truncate them and that's probably not what we want. It would be better to use `strconv.FormatFloat` in that case.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -0,0 +1,432 @@
...
+		case string:
+			valStr = v
+		default:
+			// Ignore value
```
bkabrda: I think it'd be better to do something like `valStr = string(v)`. Ignoring the default case could lead to some weird errors.

datadog/resource_datadog_synthetics_test_.go
btoueg: changed to `valStr = strconv.FormatFloat(v, 'f', -1, 64)`

datadog/resource_datadog_synthetics_test_.go
btoueg: Changed to:
```golang
// TODO: manage target for JSON body assertions
valStrr, err := json.Marshal(v)
if err == nil {
    valStr = string(valStrr)
}
```

datadog/resource_datadog_synthetics_test_.go
tanguylebarzic: We changed the API to accept directly the status live/paused, so it's not necessary anymore to do an extra call (whose semantics were hard to get right IMO).

bkabrda:
LGTM :+1:. Code is ok, acceptance tests are green and some hand testing with TF also looks good. (I don't have rights to merge though, so someone else needs to push the button)

PR #172, opened 2019-04-15
Update changelog with support of Synthetics resources

gabsn:
Update changelog with support of Synthetics resources

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ INTERNAL:
 
 * provider: Enable request/response logging in `>=DEBUG` mode [GH-153]
 
+IMPROVEMENTS:
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ INTERNAL:
 
 * provider: Enable request/response logging in `>=DEBUG` mode [GH-153]
 
+IMPROVEMENTS:
```
nmuesch: Lets bump the version header to 1.8.0 since we have an improvement

gabsn: sure

nmuesch:
There are some other changes we should include, but I'll take care of those in another PR since this is from a fork. Thanks for this!

Issue #142, opened 2019-01-08
Timeboard graph fields ignored


miguelaferreira:
I'm reporting reproducible behaviour of the provider that ignores configuration fields that are set in terraform and wants to set them over and over. Also defaults to certain fields that are not set in terraform keep being reported as changes to be applied with every plan.

It is likely that some of the things I'm reporting here have been reported in other issues, but I'll still keep them here together with the code to reproduce it.

### Terraform Version
```
Terraform v0.11.11
+ provider.datadog v1.6.0
```

### Affected Resource(s)
- datadog_timeboard

### Terraform Configuration Files
```hcl
variable "datadog_api_key" {
  default = ""
}

variable "datadog_app_key" {
  default = ""
}

provider "datadog" {
  api_key = "${var.datadog_api_key}"
  app_key = "${var.datadog_app_key}"
}

resource "datadog_timeboard" "request_type_ignored" {
  description = "Ignored fields in dashboard config DataDog"

  "graph" {
    "request" {
      q         = "avg:system.cpu.user{*}"
      type      = "line"
      extra_col = true
    }

    title = "Query value graph"
    viz   = "query_value"
  }

  "graph" {
    autoscale = true

    "request" {
      q         = "avg:system.cpu.user{*}"
      type      = "line"
    }

    title = "Top list graph"
    viz   = "toplist"
  }

  graph {
    autoscale = true

    "request" {
      q         = "avg:system.cpu.user{*}"
      type      = "bars"

      aggregator = "avg"

      style {
        palette      = "orange"
        palette_flip = false
      }
    }

    title = "Timeseries graph"
    viz   = "timeseries"
  }

  title       = "Ignored fields in dashboard config DataDog"
}
```

### Debug Output
https://gist.github.com/miguelaferreira/9397dce94aa9c0f8deb79d9833917c34

### Expected Behavior
At first run terraform should create the dashboard and subsequent terraform runs should not produce any changes.

### Actual Behavior
At first run terraform should create the dashboard and subsequent terraform runs always present changes to graph request `type`, `extra_col` and `precision`.
```
  ~ datadog_timeboard.request_type_ignored
      graph.0.precision:                    "2" => ""
      graph.0.request.0.extra_col:          "" => "1"
      graph.0.request.0.type:               "" => "line"
      graph.1.autoscale:                    "false" => "true"
      graph.1.request.0.type:               "" => "line"
      graph.2.autoscale:                    "false" => "true"
[...]

PR #173, opened 2019-04-15
Set the default for graph precision

therve:
This sets the default to "2" for the graph precision in dashboards. Otherwise,
the API returns a value different from our state, and we always see a
difference.

See issue #142.


bkabrda:
LGTM :+1:

therve:
This is the kind of graph reproducing the issue:

```
  graph {
    title = "Query value graph"
    viz   = "query_value"

    request {
      q = "avg:system.cpu.user{*}"
    }
  }
```

therve:
I think it would be better to use `computed` here.

PR #187, opened 2019-04-30
Fix various acceptance tests failures

bkabrda:
This PR fixes various minor issues in the acceptance test suite. Some of these were only tests forgotten to be updated, but some were real issues, one of which also required a fix in go-datadog-api (thus the dependency update).

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -11,6 +11,8 @@ import (
 	"github.com/zorkian/go-datadog-api"
 )
 
+const logAlertMonitorType = "log alert"
+
 func resourceDatadogMonitor() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceDatadogMonitorCreate,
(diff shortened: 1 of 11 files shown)


datadog/resource_datadog_monitor.go
```diff
@@ -150,10 +150,12 @@ func resourceDatadogMonitor() *schema.Resource {
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			// since this is only useful for "log alert" type, we don't set a default value
```
gzussa: 👍

datadog/resource_datadog_monitor.go
```diff
@@ -232,8 +234,12 @@ func buildMonitorStruct(d *schema.ResourceData) *datadog.Monitor {
 		Options: &o,
 	}
 
-	if attr, ok := d.GetOk("enable_logs_sample"); ok && m.GetType() == "log alert" {
-		o.SetEnableLogsSample(attr.(bool))
+	if m.GetType() == "log alert" {
```
gzussa: Should we use a constant variable?

datadog/resource_datadog_monitor.go
```diff
@@ -232,8 +234,12 @@ func buildMonitorStruct(d *schema.ResourceData) *datadog.Monitor {
...
+		if attr, ok := d.GetOk("enable_logs_sample"); ok {
```
gzussa: Same for this string

datadog/resource_datadog_monitor.go
bkabrda: Good point, will do.

datadog/resource_datadog_monitor.go
bkabrda: Since this is not a value, but an argument name, I'd actually prefer to keep it this way. The other argument names are used the same amount of times without having constants associated.

bkabrda:
Since this already got 2 approvals, I'm merging. Thanks for all the reviews!

PR #194, opened 2019-05-07
Update to Terraform 0.12.0-beta2

bkabrda:
Please review carefully.

I needed to do a lot of test/doc changes because of changes in strictness of HCL block vs attribute parsing [1]. After accepting this PR, the TF DD Provider would still stay compatible with TF 0.11 and users would only need to do similar changes if (or rather when) they actually migrated to TF 0.12 - while they use TF 0.11, they'll be able to keep everything as is.

[1] https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each


bkabrda:

PR #198, opened 2019-05-09
Update to Terraform 0.12.0-rc1

bkabrda:
This is just a result of `go get github.com/hashicorp/terraform@v0.12.0-rc1` and `go mod vendor`, no files were modified by hand and tests still pass.


bkabrda:

Issue #148, opened 2019-01-30
Doesn't support Log Alert Monitor


geekeren:
Hi there, I create two monitors, one is created manually, the other is by terraform.

### Terraform Version
Terraform v0.11.11
+ provider.datadog v1.6.0

### Terraform Configuration Files
```hcl
resource "datadog_monitor" "return_404s" {
  name               = "LB returning 404s"
  type               = "log alert"
  message            = <<EOT
The threshold on this may need to be tweaked!

@pagerduty-EdgeLoadBalancer
EOT
  query = "logs(\"source:nginx AND @http.status_code:404\").index(\"main\").rollup(\"count\").last(\"5m\") > 5"

  thresholds {
    critical = 5
  }
  silenced {
    "*" =  false ## unmute or not
  }
}
```

When I edit the monitor, I expected to see:
### Expected Behavior

![Expected Behavior](https://user-images.githubusercontent.com/13082375/51976129-e2dc8600-24be-11e9-8c12-c47d79948730.png)

### Actual Behavior
but I saw the search query is empty:

![image](https://user-images.githubusercontent.com/13082375/51976170-fbe53700-24be-11e9-9f48-f21f79b2374a.png)

When I go to view page, the two monitors are the same:
![image](https://user-images.githubusercontent.com/13082375/51976684-46b37e80-24c0-11e9-8473-77778dbe8cd5.png)

**I am not sure whether is UI issues of datadog, or the log alert doesn't work at all**

rsun-thoughtworks:
I have the same problem.
To find the root cause of it, I read the source code of terraform-provider and the POST message when we save a monitor in datadog page.

I found out that the "AND"s are in `queryConfig.queryString` and terraform-provider does not support this property.

I suggest that datadog official team can fix this bug.

Thanks for your team and work with love.

bkabrda:
Hi everyone, I’ve raised this issue internally to our monitors team and they’ve acknowledged this is a bug and will work to fix it. I'll post an update here when there's some development on this.

geekeren:
> Hi everyone, I’ve raised this issue internally to our monitors team and they’ve acknowledged this is a bug and will work to fix it. I'll post an update here when there's some development on this.

Can I ask your team will fix the datadog or terraform datadog provider? Because I am trying to fix this by changing code of terraform-datadog-provider then create a PR?

bkabrda:
[...]

PR #199, opened 2019-05-09
#148 Fix log monitor UI issue: generate queryConfig.queryString from query

geekeren:
Fix log monitor UI issue of the search query
-  generate queryConfig.queryString from the query attribute

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


bkabrda:
So overall this looks great, I'd just like to ask you to create a function for the piece of code that is (AFAICS) completely the same and used in two parts of your patch - the piece that starts with `matchStrs := ...` - let's avoid code duplication.

geekeren:
> So overall this looks great, I'd just like to ask you to create a function for the piece of code that is (AFAICS) completely the same and used in two parts of your patch - the piece that starts with `matchStrs := ...` - let's avoid code duplication.

You're right, I am going to do that.

geekeren:
@bkabrda I have refactored it, could you take a look?

bkabrda:
Yeah, LGTM now :+1:, merging. Thanks a lot for the PR!

gregsymons:
I've been waiting for this bug to get fixed. Any idea when the next release will be?

bkabrda:
@gregsymons hopefully within next two weeks, but it's definitely something that can change. Hope this is ok for you.

PR #201, opened 2019-05-12
docs(monitor): finish sentence for include_tags

JordanSussman:
The docs for `include_tags` seem to have been incomplete, so I grabbed the same wording from the [api docs](https://docs.datadoghq.com/api/?lang=python#get-a-monitor-downtime) to clarify usage.

### DIFF
--- a/website/docs/r/monitor.html.markdown
+++ b/website/docs/r/monitor.html.markdown
@@ -55,7 +55,7 @@ The following arguments are supported:
     * `event alert`
     * `query alert`
     * `composite`
-    * `log alert`    
+    * `log alert`
 * `name` - (Required) Name of Datadog monitor
 * `query` - (Required) The monitor query to notify on. Note this is not the same query you see in the UI and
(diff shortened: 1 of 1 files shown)


website/docs/r/monitor.html.markdown
```diff
@@ -107,7 +107,7 @@ The following arguments are supported:
     Defaults to false.
 * `timeout_h` (Optional) The number of hours of the monitor not reporting data before it will automatically resolve
     from a triggered state. Defaults to false.
-* `include_tags` (Optional) A boolean indicating whether notifications from this monitor will automatically insert its
+* `include_tags` (Optional) A boolean indicating whether notifications from this monitor automatically inserts its triggering tags into the title. Defaults to true.
```
bkabrda: "inserts" should actually be "insert", as it refers to plural "notifications", doesn't it?

JordanSussman: Just updated with a new commit.

PR #204, opened 2019-05-15
Downtime improvements

platinummonkey:
1. Adds the timezone parameter
1. Adds descriptions to parameters
1. Adds validation to monitor_tags
1. Ensures that the updated downtime id is correctly handled for future immutable downtimes

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -31,10 +31,12 @@ func resourceDatadogDowntime() *schema.Resource {
 					_, recurrencePresent := d.GetOk("recurrence")
 					return recurrencePresent
 				},
+				Description: "when true indicates this downtime is being actively applied",
 			},
 			"disabled": {
-				Type:     schema.TypeBool,
-				Optional: true,
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "When true indicates this downtime is not being applied",
 			},
 			"start": {
 				Type:     schema.TypeInt,
@@ -43,6 +45,7 @@ func resourceDatadogDowntime() *schema.Resource {
 					_, startDatePresent := d.GetOk("start_date")
 					return startDatePresent
 				},
+				Description: "Specify when this downtime should start",
 			},
 			"start_date": {
 				Type:          schema.TypeString,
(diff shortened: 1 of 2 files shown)


platinummonkey:
cc @bkabrda

datadog/resource_datadog_downtime.go
```diff
@@ -108,18 +121,23 @@ func resourceDatadogDowntime() *schema.Resource {
...
+				Required:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "specify the group scope to which this downtime applies. For everything use '*'",
 			},
 			"monitor_id": {
-				Type:     schema.TypeInt,
-				Optional: true,
+				Type:          schema.TypeInt,
+				Optional:      true,
+				ConflictsWith: []string{"monitor_tags"},
```
nmuesch: Is setting both not an available option here? Would the API return an error previously if both were being sent?

datadog/resource_datadog_downtime.go
```diff
@@ -305,6 +326,8 @@ func resourceDatadogDowntimeUpdate(d *schema.ResourceData, meta interface{}) err
 	if err = client.UpdateDowntime(dt); err != nil {
 		return fmt.Errorf("error updating downtime: %s", err.Error())
 	}
+	// handle the case when a downtime is replaced
+	d.SetId(strconv.Itoa(dt.GetId()))
```
nmuesch: Just to confirm, the idea here being that we would be updating the current downtime on each subsequent apply, rather than creating a new one?

platinummonkey: It just ensures the downtime ID is set. In the future when downtimes are immutable this ID can change on updates

datadog/resource_datadog_downtime.go
platinummonkey: while technically it works, it's one of those, `it's broken if the tags also don't match the monitor and won't apply` it'll be an API validation in the future. The matching logic today is: (pseudo-python): `if (downtime.monitor_id is None or downtime.monitor_id == monitor.id) and monitor.tags.issuperset(downtime.monitor_tags)`

nmuesch: Ok makes sense, thanks for the explanation :)

PR #207, opened 2019-05-17
Clarify the format of api_url provider argument

bkabrda:


bkabrda:
Thanks for the review, merging!

octplane:
Thanks, was away in :bus: so couldn't approve and all, but that's nice!

PR #209, opened 2019-05-17
Make sure the computed value for monitor's query_config gets stored on creation

bkabrda:
Without this, having a monitor that's not of type `log alert` would result in not saving computed value for `query_config` on creation and thus having non-empty plan after first apply.


bkabrda:

PR #212, opened 2019-05-20
Add default of 10m for monitor no data timeframe

### LINKED ISSUE #206: wrong default value for datadog_monitor.no_data_timeframe
### Terraform Version
`Terraform v0.11.13`

### Affected Resource(s)
- datadog_monitor

```hcl
resource "datadog_monitor" "agent_running" {
  name = "${local.prefix} Agent is not running"
  type = "metric alert"
  query = "min(last_1m):min:datadog.agent.running{${local.filters}} by {host} < 1"

  message = <<EOT
${var.alert_handles}
EOT
  notify_no_data = true
}
```

### Expected Behavior
According to the [docs](https://www.terraform.io/docs/providers/datadog/r/monitor.html#no_data_timeframe) the default value for `datadog_monitor.no_data_timeframe` should be 
> Default: 2x timeframe for metric alerts, 2 minutes for service checks. 

To be honest this default seems randomly chosen and I am not sure if it makes a lot of sense. 

### Actual Behavior

`datadog_monitor.no_data_timeframe` is set to 0 by default. This bring the strange side effect that editing a monitor created from terraform with `datadog_monitor.no_data_timeframe` equal to 0 in the web app will make the UI reset the value to the web app default (10 minutes).

nmuesch:
Adds a default of 10 minutes for the no_data_timeframe to be used instead of the 0 value the provider currently emits. 

Ultimately looks to fix #206

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -132,6 +132,7 @@ func resourceDatadogMonitor() *schema.Resource {
 			"no_data_timeframe": {
 				Type:     schema.TypeInt,
 				Optional: true,
+				Default:  10,
 			},
 			"renotify_interval": {
 				Type:     schema.TypeInt,
--- a/website/docs/r/monitor.html.markdown
+++ b/website/docs/r/monitor.html.markdown
@@ -55,7 +55,7 @@ The following arguments are supported:
     * `event alert`
     * `query alert`
     * `composite`
-    * `log alert`    
+    * `log alert`
 * `name` - (Required) Name of Datadog monitor
 * `query` - (Required) The monitor query to notify on. Note this is not the same query you see in the UI and
     the syntax is different depending on the monitor `type`, please see the [API Reference](https://docs.datadoghq.com/api/?lang=python#create-a-monitor) for details. **Warning:** `terraform plan` won't perform any validation of the query contents.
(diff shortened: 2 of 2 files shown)


datadog/resource_datadog_monitor.go
```diff
@@ -132,6 +132,7 @@ func resourceDatadogMonitor() *schema.Resource {
 			"no_data_timeframe": {
 				Type:     schema.TypeInt,
 				Optional: true,
+				Default:  10,
```
bkabrda: We should use `DefaultFunc` and only set this when `notify_no_data` is `true`, I think.

Also, we could actually try parsing out the evaluation timeframe from the query and set it to the recommended values from the docs [1] - let's fall back to 10 minutes if fail to parse out the evaluation timeframe.

[1] https://docs.datadoghq.com/api/?lang=python#monitors

nmuesch: As discussed offline a bit, lets go with a generically default 10 minutes and just add a note that it should be set to 2x the evaluation window. We can tackle something more complex if its needed later 😄

nmuesch: Actually, the docs have a pretty clear note that the timeframe, if specified, should be at least 2x the eval window. Do you think this is good to be merged?

bkabrda: Yeah, I'm ok with merging this :+1:

PR #213, opened 2019-05-20
Add link to documentation about monitor types

nmuesch:
Add a link to the Datadog Documentation in the monitor terraform docs that provides the mapping between API/Terraform monitor `types` and the ones that you see in the Datadog platform. 

Help make choosing a monitor type easier!

### DIFF
--- a/website/docs/r/monitor.html.markdown
+++ b/website/docs/r/monitor.html.markdown
@@ -49,7 +49,7 @@ resource "datadog_monitor" "foo" {
 
 The following arguments are supported:
 
-* `type` - (Required) The type of the monitor. The mapping from these types to the in app Monitor types can be found in our [documentation](https://docs.datadoghq.com/api/?lang=python#create-a-monitor) Available options to choose from are:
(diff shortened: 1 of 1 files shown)


website/docs/r/monitor.html.markdown
```diff
@@ -49,13 +49,13 @@ resource "datadog_monitor" "foo" {
 
 The following arguments are supported:
 
-* `type` - (Required) The type of the monitor, chosen from:
+* `type` - (Required) The type of the monitor. The mapping from these types to the in app Monitor types can be found in our [documentation](https://docs.datadoghq.com/api/?lang=python#create-a-monitor) Available options to choose from are:
```
bkabrda: Two notes here:

* Not a native speaker, but "to the in app Monitor types" doesn't make much sense to me. Did you mean to use "to the types in Monitor app"?
* "our documentation" is ambiguous here IMO. Let's use "Datadog API documentation"

bkabrda:
Note: this is related to https://github.com/terraform-providers/terraform-provider-datadog/issues/177 and to https://github.com/terraform-providers/terraform-provider-datadog/issues/208 which both were reported because of this missing link in docs. Therefore I believe we can close them whenever this is merged.

bkabrda:
LGTM now, merging. Thanks!

PR #215, opened 2019-05-22
Allow adding metadata aliases for widgets in screenboards and timeboards

### LINKED ISSUE #61: Datadog Timeboard: Metadata key
### Terraform Version

```
Terraform v0.11.7
+ provider.aws v1.8.0
+ provider.datadog v1.0.3
```

### Affected Resource(s)
- https://www.terraform.io/docs/providers/datadog/r/timeboard.html

Trying to take the JSON config from a dashboard I created in Datadog, and port it to HCL so that I can manage it with Terraform. It appears that the `metadata` key is not supported yet by this provider.

The JSON config for Datadog is:

```json
{
  "requests": [
    {
      "q": "sum:aws.trustedadvisor.green_checks, sum:aws.trustedadvisor.yellow_checks, sum:aws.trustedadvisor.red_checks",
      "type": "bars",
      "style": {
        "palette": "dog_classic",
        "type": "solid",
        "width": "normal"
      },
      "metadata": {
        "sum:aws.trustedadvisor.green_checks": {
          "alias": "OK"
        },
        "sum:aws.trustedadvisor.yellow_checks": {
          "alias": "WARN"
        },
        "sum:aws.trustedadvisor.red_checks": {
          "alias": "ERROR"
        }
      },
      "conditional_formats": []
    }
  ],
  "viz": "timeseries",
  "autoscale": true
}
```

bkabrda:
Fixes #61

### DIFF
--- a/datadog/resource_datadog_screenboard.go
+++ b/datadog/resource_datadog_screenboard.go
@@ -1,6 +1,7 @@
 package datadog
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -153,6 +154,11 @@ func resourceDatadogScreenboard() *schema.Resource {
 				"increase_good": {
 					Type:     schema.TypeBool,
 					Optional: true,
+				},
+				"metadata_json": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validateMetadataJSON,
 				},
 			},
 		},
@@ -582,6 +588,25 @@ func resourceDatadogScreenboard() *schema.Resource {
 // # Convenience functions to safely pass info from Terraform to the Datadog API wrapper #
 // #######################################################################################
 
+func getMetadataFromJSON(jsonBytes []byte, unmarshalled interface{}) error {
+	decoder := json.NewDecoder(bytes.NewReader(jsonBytes))
+	// make sure we return errors on attributes that we don't expect in metadata
+	decoder.DisallowUnknownFields()
+	err := decoder.Decode(unmarshalled)
+	if err != nil {
+		return fmt.Errorf("Failed to unmarshal metadata_json: %s", err)
+	}
+	return nil
+}
+
+func validateMetadataJSON(v interface{}, k string) (ws []string, errors []error) {
+	err := getMetadataFromJSON([]byte(v.(string)), &map[string]datadog.TileDefMetadata{})
+	if err != nil {
+		errors = append(errors, fmt.Errorf("%q contains an invalid JSON: %s", k, err))
+	}
+	return
+}
+
 func setStringFromDict(dict map[string]interface{}, key string, field **string) {
 	if v, ok := dict[key]; ok && v != nil {
 		*field = datadog.String(v.(string))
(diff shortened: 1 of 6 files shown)


datadog/resource_datadog_screenboard.go
```diff
@@ -582,6 +588,25 @@ func resourceDatadogScreenboard() *schema.Resource {
...
+	if err != nil {
+		return fmt.Errorf("Failed to unmarshal metadata_json: %s", err)
+	}
+	return nil
+}
+
+func validateMetadataJSON(v interface{}, k string) (ws []string, errors []error) {
+	err := getMetadataFromJSON([]byte(v.(string)), &map[string]datadog.TileDefMetadata{})
+	if err != nil {
+		errors = append(errors, fmt.Errorf("%q contains an invalid JSON: %s", k, err))
```
nmuesch: Should the append be reversed here? The named return seems to want `string, errors`, wouldn't this return `errors, string`?

datadog/resource_datadog_timeboard.go
```diff
@@ -579,7 +590,7 @@ func resourceDatadogTimeboardCreate(d *schema.ResourceData, meta interface{}) er
 		return fmt.Errorf("Failed to create timeboard using Datadog API: %s", err.Error())
 	}
 	d.SetId(strconv.Itoa(timeboard.GetId()))
-	return nil
+	return resourceDatadogTimeboardRead(d, meta)
```
nmuesch: Why this change?

datadog/resource_datadog_screenboard.go
bkabrda: This line actually only appends the current error to the slice of errors. I guess internally TF passes the `errors` slice around multiple validators and similar functions to collect all errors.

datadog/resource_datadog_timeboard.go
bkabrda: This is a best practice, as mentioned in [1]. It's necessary in this PR as line [2] actually creates an attribute on the TF schema that doesn't exist and thus persisting it in the state would make TF fail. This is precisely one of the reasons why calling the read function at the end of create/update is recommended, as it overwrites changes like this one.

[1] https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification
[2] https://github.com/terraform-providers/terraform-provider-datadog/pull/215/files#diff-22ba5b0fbe5d1b3cdfe783aeb61bf51aR755

datadog/resource_datadog_screenboard.go
nmuesch: Ahh ok, thanks 🙇

datadog/resource_datadog_timeboard.go
nmuesch: Ahh I see, thanks for the helpful explanation/links 🙇

website/docs/r/timeboard.html.markdown
```diff
@@ -26,6 +26,11 @@ resource "datadog_timeboard" "redis" {
     request {
       q    = "avg:redis.info.latency_ms{$host}"
       type = "bars"
+      metadata_json = jsonencode({
```
nmuesch: I can't seem to make this example work locally. Are you running TF 0.12? Is it possible there is a difference in syntax?

nmuesch: Actually I notice the way you call functions seems to be different between 0.11 and 0.12
I also see jsonencode called out in the 0.12 changelog, is this going to be breaking - https://github.com/hashicorp/terraform/releases

bkabrda: Good catch, I'll look into this and update the docs when I figure out how to make this work on TF < 0.12.

bkabrda: Ok, I figured out the way to do it and amended the documentation. Please give it a go. Thanks!

bkabrda:
Thanks for your review! I'm merging it.

PR #221, opened 2019-05-30
Deprecate Silence attribute and Fix multiple terraform runs

nmuesch:
This PR looks to accomplish a few things surrounding the `silence` attribute of monitor resources. 

* Add a deprecation notice that we'll stop supporting the `silenced` param soon. (In a couple Major versions of the Provider as outlined by the best practices doc - https://www.terraform.io/docs/extend/best-practices/deprecations.html)
* Add the ability to specify a `-1` on any scope of the silenced attribute to unmute that scope. This will make a separate API call to unmute those scopes (Only on updates, newly created monitors shouldn't be muted :) ) 
* Fix an issue where running terraform apply would detect a diff on the silence scope if the timestamp were in the past. This is due to the Datadog Monitor API not returning scopes with timestamps in the past in the API response. So the monitor resource will detect this and leave the state unchanged. 
* Add some tests 🎉

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -157,9 +157,10 @@ func resourceDatadogMonitor() *schema.Resource {
 				Optional: true,
 			},
 			"silenced": {
-				Type:     schema.TypeMap,
-				Optional: true,
-				Elem:     schema.TypeInt,
+				Type:       schema.TypeMap,
+				Optional:   true,
+				Elem:       schema.TypeInt,
+				Deprecated: "use Downtime Resource instead",
 			},
 			"include_tags": {
 				Type:     schema.TypeBool,
@@ -323,6 +324,19 @@ func resourceDatadogMonitorExists(d *schema.ResourceData, meta interface{}) (b b
 	return true, nil
 }
 
+func getUnmutedScopes(d *schema.ResourceData) []string {
+	var unmuteScopes []string
+	if attr, ok := d.GetOk("silenced"); ok {
+		for k, v := range attr.(map[string]interface{}) {
+			if v.(int) == -1 {
+				unmuteScopes = append(unmuteScopes, k)
+			}
+		}
+		log.Printf("[DEBUG] Unmute Scopes are: %v", unmuteScopes)
+	}
+	return unmuteScopes
+}
+
 func resourceDatadogMonitorCreate(d *schema.ResourceData, meta interface{}) error {
 
 	client := meta.(*datadog.Client)
@@ -398,7 +412,6 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error 
 	d.Set("notify_audit", m.Options.GetNotifyAudit())
 	d.Set("timeout_h", m.Options.GetTimeoutH())
 	d.Set("escalation_message", m.Options.GetEscalationMessage())
-	d.Set("silenced", m.Options.Silenced)
 	d.Set("include_tags", m.Options.GetIncludeTags())
 	d.Set("tags", tags)
 	d.Set("require_full_window", m.Options.GetRequireFullWindow()) // TODO Is this one of those options that we neeed to check?
@@ -511,6 +524,7 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 	if attr, ok := d.GetOk("escalation_message"); ok {
 		o.SetEscalationMessage(attr.(string))
 	}
+
 	silenced := false
 	if attr, ok := d.GetOk("silenced"); ok {
 		// TODO: this is not very defensive, test if we can fail non int input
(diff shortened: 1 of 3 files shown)


datadog/resource_datadog_monitor.go
```diff
@@ -545,17 +570,26 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	var retval error
-	if retval = resourceDatadogMonitorRead(d, meta); retval != nil {
-		return retval
-	}
 
+	// if the silenced section was removed from the config, we unmute it via the API
+	// The API mointor wouldn't automatically unmute if the config is just missing
```
bkabrda: Typo: `mointor`

datadog/resource_datadog_monitor.go
```diff
@@ -418,6 +431,17 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("query_config", queryConfig)
 	}
 
+	// Update the state with the contents of the API + whats in the config
```
bkabrda: I think there's one problem in this situation:

* create a downtime for some scope that’s a specific timestamp in the future
* unmute that scope in the ui
* run `terraform apply`
=> no change in plan even though the resource in the config is clearly different from current state

website/docs/r/monitor.html.markdown
```diff
@@ -118,7 +118,8 @@ The following arguments are supported:
 * `threshold_windows` (Optional) A mapping containing `recovery_window` and `trigger_window` values, e.g. `last_15m`. Can only be used for anomaly monitors.
   * `recovery_window` describes how long an anomalous metric must be normal before the alert recovers.
   * `trigger_window`  describes how long a metric must be anomalous before an alert triggers.
-* `silenced` (Optional) Each scope will be muted until the given POSIX timestamp or forever if the value is 0.
+* `silenced` (Optional) Each scope will be muted until the given POSIX timestamp or forever if the value is 0. Use `-1` if you want to un-silence the scope. **Deprecated** The `silenced` parameter is being deprecated in favor of the downtime resource. This will be removed in the next major version of the Terraform Provider.
```
bkabrda: One of the sentences has `...be muted until...` and the other `...to un-silence...` - let's just use "mute" or "silence" in both of them to make this less confusing (I'd prefer using "unmute" in the second sentence, because that's the term used in the UI).

nmuesch: Good call, done!

datadog/resource_datadog_monitor.go
```diff
@@ -418,6 +432,24 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
...
 	}
 
+	// The Datadog API doesn't return old timestamps or support a special value for unmuting scopes
+	// So we provide this functionality by saving values to the state
+	//If the user specified scope timestamp is -1 and its not in the API response, save it to the state
+	//If the user specified scope timestamp is in the past and its not our 0 value, remove it from the state (it won't be returned via the API)
+	//If the user specified value isn't in the API response and its not an above special case, remove it from the state so we re mute it.
+	apiSilenced := m.Options.Silenced
+	configSilenced := d.Get("silenced").(map[string]interface{})
+	for k, configuredVal := range configSilenced {
```
bkabrda: So unfortunately this now breaks the functionality from [1]. Consider this scenario:

* I create a monitor with one muted scope (could be more, that doesn't matter), where the timestamp is correct future timestamp.
* I remove the scope (=> unmute it), making the `silenced` parameter empty. This will result in the API not unmuting anything, as `silenced` is not even sent by the underlying client library.
* The logic in the condition in [1] relies on the `resourceDatadogMonitorRead` to correctly read all the values from API into `d`. If this happens correctly, the condition evaluates to `true` (because `ok` is `true`, as the scope is still muted on the API side, and `!silenced` is `true` as well, because `silenced` is not set.
* Your implementation only traverses through keys of `silenced` from the current configuration (which is empty), so it won't set the silenced scope on `d` and thus [1] will fail to evaluate correctly and the scope will remain muted.

You need to traverse all keys from both current config and the API response to get this right, I think.

All the other scenarios that I could think of work fine.

[1] https://github.com/terraform-providers/terraform-provider-datadog/commit/b4d5d17b2320ee7287f9b0f88a0c6240eb5061c1

datadog/resource_datadog_monitor.go
```diff
@@ -418,6 +432,32 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("query_config", queryConfig)
 	}
 
+	// The Datadog API doesn't return old timestamps or support a special value for unmuting scopes
+	// So we provide this functionality by saving values to the state
+	apiSilenced := m.Options.Silenced
+	configSilenced := d.Get("silenced").(map[string]interface{})
+	unmutedScopes := getUnmutedScopes(d)
+
+	// If the scope is in the API response but not in the config
```
bkabrda: I think this functionality should be in the update function, not read. The reason for that is that if it's in read, it will create confusing TF plans with the `-1` value that the user never configured anywhere. Does that make sense?

I *think* the above should be possible, but if not, this is still better than what we have currently.

nmuesch:
Good idea, just changed the test function names based on your suggestion. I'll do a `squash and merge` if thats alright with you.

bkabrda:
Yup, go ahead with squashing and merging 👍

gzussa:
Great stuff :)

PR #224, opened 2019-06-04
Add note about logs columns on facet attributes

nmuesch:
Add example for how to submit attributes into the `columns` section of a log screenboard widget. 

When using a facet made from an attribute, you need to use `log_` as the prefix instead of the `@` that the UI displays. Using a `@` directly in the config doesn't parse properly in the backend.


website/docs/r/screenboard.html.markdown
```diff
@@ -458,7 +458,7 @@ Nested `widget` blocks have the following structure:
 - `manage_status_title_text` - (Optional, only for widgets of type "manage_status") The title of the widget.
 - `manage_status_title_size` - (Optional, only for widgets of type "manage_status") The size of the widget's title.
 - `manage_status_title_align` - (Optional, only for widgets of type "manage_status") The alignment of the widget's title. One of "left", "center", or "right".
-- `columns` - (Optional, only for widgets of type "log_stream") Stringified list of columns to use. Example: `"[\"column1\",\"column2\",\"column3\"]"`
+- `columns` - (Optional, only for widgets of type "log_stream") Stringified list of columns to use. Example: `"[\"column1\",\"column2\",\"column3\"]"` Note: when creating a column on a facet from an attribute (i.e. one with `@`), prefix the attribute name with `log_` instead of `@`. For example `@http.status_code` would become `log_http.status_code` in your terraform config.
```
ruthnaebeck:
```suggestion
- `columns` - (Optional, only for widgets of type "log_stream") Stringified list of columns to use. Example: `"[\"column1\",\"column2\",\"column3\"]"` **Note**: When creating a column on a facet from an attribute prefixed with `@`, change the prefix to `log_`. For example `@http.status_code` would become `log_http.status_code` in your terraform config.
```

PR #234, opened 2019-06-12
Fix several integration tests and related code

bkabrda:
* Set datadog_downtime.timezone attribute in resourceDatadogDowntimeRead,
  handle it in import test
* Make datadog_downtime.tags a TypeSet to be able to properly ignore
  differences in ordering of its contents
* Fix two regressions caused by calling resourceDatadogMonitorRead from
  resourceDatadogMonitorUpdate and resourceDatadogMonitorCreate:
  * Make sure that API sometimes changing "metric alert" to "query alert"
    continues to be ignored
  * Fix tests for threshold values to always expect floats

The last change should be mentioned in the changelog, because that will actually create a non-empty plan for people who have used integers as values for thresholds previously - with next apply, these are all going to change to floats (one-time only). It's not really a regression, since the values will just change their type, but still noteworthy IMO.

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -55,14 +55,6 @@ func resourceDatadogMonitor() *schema.Resource {
 			"type": {
 				Type:     schema.TypeString,
 				Required: true,
-				// Datadog API quirk, see https://github.com/hashicorp/terraform/issues/13784
-				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
-					if oldVal == "query alert" && newVal == "metric alert" {
-						log.Printf("[DEBUG] Monitor '%s' got a '%s' response for an expected '%s' type. Suppressing change.", d.Get("name"), newVal, oldVal)
-						return true
-					}
-					return newVal == oldVal
-				},
 			},
 
 			// Options
@@ -168,7 +160,10 @@ func resourceDatadogMonitor() *schema.Resource {
 				Default:  true,
 			},
 			"tags": {
-				Type:     schema.TypeList,
+				// we use TypeSet to represent tags, paradoxically to be able to maintain them ordered;
+				// we order them explicitly in the read/create/update methods of this resource and using
+				// TypeSet makes Terraform ignore differences in order when creating a plan
+				Type:     schema.TypeSet,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
@@ -277,7 +272,7 @@ func buildMonitorStruct(d *schema.ResourceData) *datadog.Monitor {
 
 	if attr, ok := d.GetOk("tags"); ok {
 		tags := []string{}
-		for _, s := range attr.([]interface{}) {
+		for _, s := range attr.(*schema.Set).List() {
 			tags = append(tags, s.(string))
 		}
 		sort.Strings(tags)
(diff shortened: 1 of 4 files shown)


datadog/resource_datadog_monitor_test.go
```diff
@@ -49,11 +49,13 @@ func TestAccDatadogMonitor_Basic(t *testing.T) {
 						"datadog_monitor.foo", "require_full_window", "true"),
 					resource.TestCheckResourceAttr(
 						"datadog_monitor.foo", "locked", "false"),
-					// Tags are sorted
+					// Tags are a TypeSet => use a weird way to access members by their hash
 					resource.TestCheckResourceAttr(
-						"datadog_monitor.foo", "tags.0", "baz"),
+						"datadog_monitor.foo", "tags.#", "2"),
```
nmuesch: This just means there are two tags, right?

datadog/resource_datadog_monitor_test.go
```diff
@@ -318,11 +328,13 @@ func TestAccDatadogMonitor_UpdatedToRemoveTags(t *testing.T) {
...
 					resource.TestCheckResourceAttr(
-						"datadog_monitor.foo", "tags.1", "foo:bar"),
+						"datadog_monitor.foo", "tags.2644851163", "baz"),
```
nmuesch: Not sure I totally follow where these hashes come from?

datadog/resource_datadog_monitor.go
```diff
@@ -383,7 +386,14 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("name", m.GetName())
 	d.Set("message", m.GetMessage())
 	d.Set("query", m.GetQuery())
-	d.Set("type", m.GetType())
+	if typ, ok := m.GetTypeOk(); ok {
+		if d.Get("type").(string) == "metric alert" && typ == "query alert" {
```
nmuesch: If we do this here, do we still need the DiffSuppress in the schema?

datadog/resource_datadog_monitor_test.go
bkabrda: Correct, this checks that the number of tags is what we expect it to be, so two.

datadog/resource_datadog_monitor_test.go
bkabrda: TF `TypeSet` is internally represented as a map that maps computed hashes to actual values. Since the hashes are always the same for one value, this is the way to get them. I'll amend the comment to explain this.

datadog/resource_datadog_monitor.go
bkabrda: Good question, I'm not sure. I'll find out :)

datadog/resource_datadog_monitor_test.go
bkabrda: Comment added.

datadog/resource_datadog_monitor.go
bkabrda: You were right, `DiffSuppressFunc` is no longer necessary with this. I added a comment explaining it and removed `DiffSuppressFunc`.

bkabrda:
Yeah, you're probably right about that one-time diff, so let's not put it into the changelog.

datadog/resource_datadog_monitor.go
```diff
@@ -383,7 +378,21 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
...
+		/* Datadog API quirk, see https://github.com/hashicorp/terraform/issues/13784
+		 *
+		 * If current type of monitor is "metric alert" and the API is returning "query alert",
+		 * we want to keep "metric alert". We previously had this as DiffSuppressFunc on "type".
+		 * After adding a call to "resourceDatadogMonitorRead" in create/update methods, this
+		 * started creating the monitor as "query alert". To make sure that the behaviour stays
+		 * the same, we added this code (which made DiffSuppressFunc useless, so we removed it).
+		 */
+		if d.Get("type").(string) == "metric alert" && typ == "query alert" {
+			// (see the comment there for more info on this situation)
```
nmuesch: Should we remove this now?

bkabrda: Right, I missed that.

bkabrda:

PR #236, opened 2019-06-14
Mark datadog_monitor.type as ForceNew, since changing type is not actually possible

bkabrda:
To change a monitor type, one needs to actually delete the old one and create new, which is precisely what `ForceNew` does.


bkabrda:

PR #237, opened 2019-06-14
Make it possible to configure individual PD service objects independently

bkabrda:
This PR uses newly added Datadog API features (undocumented ATM, unfortunately) to make it possible to configure individual PD integration service objects independently.

The idea is to be able to use the current `datadog_integration_pagerduty` resource and allow adding number of independent service objects via `datadog_integration_pagerduty_service_object` resources.

Some notes:
* I wanted to make sure that we preserve backwards compatibility for the current `datadog_integration_pagerduty` and the way it works with services. To that end, I added `individual_services` attribute to it, which is mutually exclusive with `services`. If `services` is present, everything will work as it used to up until now. If `individual_services` is present, then this resource will not be storing list of service object into its state, meaning that creating `datadog_integration_pagerduty_service_object` resources won't cause TF to produce non-empty plans for it - it will pretty much remain as a configuration object for the whole integration.
* All the individual services defined by `datadog_integration_pagerduty_service_object` have to have `depends_on = ["datadog_integration_pagerduty.foo"]`, since they can't be created if the integration is deactivated.
* `datadog_integration_pagerduty.services` is now marked as deprecated because of this new functionality.

TODO:
* [x] Document all of the above
* [x] Write acceptance tests and make sure that it all actually works as expected
* [x] Figure out (for docs) how to force-update a service object; since the API isn't returning the actual service keys, we have no way of detecting a drift; we should find a way for users to explicitly update a service object even if drift is not detected
  * This can be done using `terraform taint`, which I documented.
* [x] Make sure importing the new service object resource works
  * This doesn't really work, as the API never returns `service_key`, so import wouldn't work well.

This supersedes https://github.com/terraform-providers/terraform-provider-datadog/pull/164, implements https://github.com/terraform-providers/terraform-provider-datadog/issues/75 and also fixes https://github.com/terraform-providers/terraform-provider-datadog/issues/160 thanks to updating the go-datadog-api dependency to latest master, which contains the fix for it: https://github.com/zorkian/go-datadog-api/commit/ddd27c7d09028b1a6aa7de014d83b1284acd121e


datadog/resource_datadog_integration_pagerduty.go
```diff
@@ -18,10 +18,16 @@ func resourceDatadogIntegrationPagerduty() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"individual_services": {
```
gzussa: Shoudln't `is_individual` be a better name?

datadog/resource_datadog_integration_pagerduty_service_object.go
```diff
@@ -0,0 +1,100 @@
...
+
+const maskedSecret = "*****"
+
+func resourceDatadogIntegrationPagerdutySO() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDatadogIntegrationPagerdutySOCreate,
+		Read:   resourceDatadogIntegrationPagerdutySORead,
+		Update: resourceDatadogIntegrationPagerdutySOUpdate,
+		Delete: resourceDatadogIntegrationPagerdutySODelete,
+		// TODO: importer
```
gzussa: ?

datadog/resource_datadog_integration_pagerduty.go
bkabrda: Well this would be a good name on the actual service, but this attribute expresses that this resource *is using* individual services, so I'd prefer to keep it.

datadog/resource_datadog_integration_pagerduty_service_object.go
bkabrda: Ah, yes, that's my WIP TODO that I will solve and remove... basically I need to make sure that `terraform import` works well with the service object resource.

bkabrda:
So the code and docs are finished and reviewable from my PoV. While writing the tests, I've discovered a race condition on the API side, so I'll need to wait for it to get fixed before this PR can be finalized.

datadog/resource_datadog_integration_pagerduty_service_object.go
bkabrda: Due to the fact that the Datadog API doesn't return service keys, it's not really meaningful to import this resource, so I explicitly disabled it by setting `Importer: nil`. I'm marking this discussion as resolved.

bkabrda:
So I was advised to workaround the race condition by serializing all operations that change individual service objects - I did that by using a mutex. I also added tests, so this PR is now complete from my side and ready to be fully reviewed.

datadog/resource_datadog_integration_pagerduty.go
gzussa: ok

nmuesch: Is the main reason we're enforcing this because if you do individual services you still need the main resource to configure global integration fields?

nmuesch: I was wondering if there were a way to not rely on this resource later, but I'm seeing that it makes sense since we'll always have some global components.

datadog/resource_datadog_integration_pagerduty_service_object.go
```diff
@@ -0,0 +1,112 @@
...
+func resourceDatadogIntegrationPagerdutySOCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*datadog.Client)
+	integrationPdSOMutex.Lock()
+	defer integrationPdSOMutex.Unlock()
+
+	so := buildIntegrationPagerdutySO(d)
+
+	if err := client.CreateIntegrationPDService(so); err != nil {
+		// TODO: warn user that PD integration must be enabled to be able to create service objects
+		return fmt.Errorf("Failed to create integration pagerduty using Datadog API: %s", err.Error())
```
nmuesch: Should we link to the documentation here?

nmuesch: And say that you need to install the tile first? (And/or show the TF page about enabling the integration globally?)

website/docs/r/integration_pagerduty.html.markdown
```diff
@@ -12,6 +12,38 @@ Provides a Datadog - PagerDuty resource. This can be used to create and manage D
 
 ## Example Usage
 
+*Note:* In the past, service objects under the `services` key were specified inside the `datadog_integration_pagerduty` resource. This was incompatible with multi-configuration-file setups, where users wanted to have individual service objects controlled from different Terraform configuration files. The recommended approach now is specifying service objects as individual resources using [datadog_integration_pagerduty_service_object](/docs/providers/datadog/r/integration_pagerduty_service_object.html) and adding `individual_services = true` to the `datadog_integration_pagerduty` object.
```
nmuesch: Could we specify a provider version here instead of `in the past`? (I assume we'll have to bump to 3.0.0 since we're deprecating something?

bkabrda:
Thanks for the review @nmuesch. I admit I haven't fully investigated the migration path, so that's something I need to look into and it will likely take some time to make sure it's done right. I'll update this issue when I figure out the best approach here.

datadog/resource_datadog_integration_pagerduty.go
bkabrda: Yeah, we'll always need it to be able to:

* activate the integrtation in the first place
* configure global settings of the integrtation

datadog/resource_datadog_integration_pagerduty_service_object.go
bkabrda: The error returned from API says `Pagerduty Integration not found` (which is printed with the error message) is IMO sufficient. I think this is self-explanatory enough. Do you need it needs to be improved? Because if we wanted to provide the docs just in this case, we'd need to scan precisely for the `Pagerduty Integration not found` string and that can be error-prone if that string changes a bit in the future...

website/docs/r/integration_pagerduty.html.markdown
bkabrda: Fixed.

datadog/resource_datadog_integration_pagerduty_service_object.go
nmuesch: Yeah thats totally valid. Lets keep as is then 🙂

bkabrda:
All right, I figured out a migration path, documented it and wrote an integration test for it. Ready for re-review.

jvitkauskas:
Any plans to release this?

bkabrda:
@jvitkauskas hey, this will be part of the next release. We're hoping to get it released either this Thursday/Friday or in the first half of next week (not guaranteed at this point, we may hold the release if we find some issues during pre-release testing).

PR #247, opened 2019-06-25
Use suppressdiff on monitor type

### LINKED ISSUE #241: Query/Metric alert type change causing delete+recreate
### Terraform Version
Seems after https://github.com/terraform-providers/terraform-provider-datadog/pull/236

### Affected Resource(s)
- monitor

N/A
define a standard `monitor` config for either `metric alert` type or `query alert` type but use the opposite type definition for the correct type. 

### Debug Output
This will delete the current monitor and create a new monitor.

### Panic Output
None

### Expected Behavior
update the monitor in place

### Actual Behavior
Deleted the monitor and created a new one.

### Steps to Reproduce
See above

### Important Factoids
Given that the API accepts either in this case and silently updates the type to the correct type. We should probably make an exception for https://github.com/terraform-providers/terraform-provider-datadog/pull/236 to disregard `metric/query alert` changes. 

### References
https://datadoghq.slack.com/archives/C3BM8SGKZ/p1561047720247900

nmuesch:
This PR takes us back to using the `suppressDiff` function for the case of monitor types being one of `metric alert` or `query alert`

The issue with the previous approach was that the diff is calculated based on what the state is refreshed as vs what the config has. Take this example:

User provides `metric alert` in the config. Runs terraform apply and the API successfully creates the monitor (but the monitor is really a `query alert`. Future API calls return `query alert` in the response. 
With older monitors (created before Provider 2.0.0), the state contained `query alert` because the diffsuppress ignored the diff and set the state to contain what the API returned. 
With 2.0.0 and 2.0.1, since this suppression was moved into the Read function, the state wasn't updated with the api response, but contained what the config had. This caused (expected) no diff to be created with new monitors, but any old monitor would be recreated. 

This should fix #241

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -56,6 +56,15 @@ func resourceDatadogMonitor() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
+				// Datadog API quirk, see https://github.com/hashicorp/terraform/issues/13784
+				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+					if (oldVal == "query alert" && newVal == "metric alert") ||
+						(oldVal == "metric alert" && newVal == "query alert") {
+						log.Printf("[DEBUG] Monitor '%s' got a '%s' response for an expected '%s' type. Suppressing change.", d.Get("name"), newVal, oldVal)
+						return true
+					}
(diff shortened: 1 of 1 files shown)


gzussa:
A quirk indeed 😄

nmuesch:
Thanks for the reviews! I've tested this one manually a few times by manually changing my state to have different types (not sure if we can do that in the testing framework) Based off this I'm going to merge.

Issue #259, opened 2019-07-13
Invalid query error with latest provider


heldersepu:
### Terraform Version
0.11.14

### Affected Resource(s)
- datadog_monitor

### Terraform Configuration Files
```hcl
provider "datadog" {
  api_key = "abc123"
  app_key = "def456"
}

resource "datadog_monitor" "test" {
  name    = "test"
  message = "test"
  query   = "avg(last_1h):anomalies(avg:aws.s3.5xx_errors{bucketname:abc}, 'basic', 2, direction='above') >= 1"
  type    = "query alert"

  evaluation_delay    = 900
  no_data_timeframe   = 120
  renotify_interval   = 0
  include_tags        = false
  require_full_window = false

  thresholds {
    critical          = 1
    critical_recovery = 0.9
    warning           = 0.95
    warning_recovery  = 0.85
  }
}
```

### Expected Behavior
The monitor should be created

### Actual Behavior
API error 400 Bad Request: {"errors":["The value provided for parameter 'query' is invalid"]}

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
This same code works fine on older versions of the provider
- Working version:
> https://github.com/terraform-providers/terraform-provider-datadog/tree/a650e8b784d84a26ebbe6e4f71e19b057e1fc13c

- Broken latest:
> https://github.com/terraform-providers/terraform-provider-datadog/tree/d9f14f8469ca3bf825ba0a39317975a3bd01d150

zioproto:
Hello @heldersepu ,
thanks for the report of the issue
Could you please specify the exact version numbers for the providers that are working and that are not working.
thank you

heldersepu:
> Hello @heldersepu ,
> thanks for the report of the issue
> Could you please specify the exact version numbers for the providers that are working and that are not working.
> thank you
Links added for the commits tested

heldersepu:
I managed to narrow down to:
https://github.com/terraform-providers/terraform-provider-datadog/commits/11c6e0071372241431e76f5edd83f6bcd5d3bcb5 
that is the commit that breaks the provider, @enbashi any idea what could have gone wrong?

heldersepu:
Looking at the json that is posted to the API, I notice that the broken version is sending: 
`"period": null` and in some other cases `"threshold_windows": {},`
The API does not like that!  maybe a problem with the go-datadog-api ?

enbashi:
[...]

PR #260, opened 2019-07-14
Correction to the threshold_windows

heldersepu:
This fixes issue #259
The `threshold_windows` should be added only if a window has been set

The Options structure already has the flag to omit it when it is empty:
```
type Options struct {
        ...
	EscalationMessage *string           `json:"escalation_message,omitempty"`
	Thresholds        *ThresholdCount   `json:"thresholds,omitempty"`
	ThresholdWindows  *ThresholdWindows `json:"threshold_windows,omitempty"`
	IncludeTags       *bool             `json:"include_tags,omitempty"`
        ...
}
```

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -223,10 +223,12 @@ func buildMonitorStruct(d *schema.ResourceData) *datadog.Monitor {
 
 	o := datadog.Options{
 		Thresholds:        &thresholds,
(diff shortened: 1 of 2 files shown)


bkabrda:
Looks perfect 👍 Thank you very much for the contribution! Merging.

heldersepu:
Wow! Thanks, @bkabrda that was quick!

PR #263, opened 2019-07-17
Adds Service Level Objective (SLO) support

platinummonkey:
Adds Service Level Objective support to this provider. These are not to be confused with the SLO widget currently available in dashboards (and soon to be migrated), but the newer centralized SLO management.  `/api/v1/slo`

Note: This required a bump in the `zorkian/go-datadog-api` client to leverage this change


datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+			// Common
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				StateFunc: func(val interface{}) string {
+					return strings.TrimSpace(val.(string))
```
nmuesch: Any specific reason for this? Not against it but curious 😄 
I guess a typo of "A gooddescription" -> "A good description" wouldn't get picked up in this case if I understand correctly?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+				},
+			},
+			"tags": {
+				// we use TypeSet to represent tags, paradoxically to be able to maintain them ordered;
+				// we order them explicitly in the read/create/update methods of this resource and using
+				// TypeSet makes Terraform ignore differences in order when creating a plan
+				Type:     schema.TypeSet,
```
nmuesch: Interesting, the primary goal here being that you can flip the order of the tags and no diff will be shown? 💜

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: ValidateServiceLevelObjectiveTypeString,
+			},
+
+			// Metric-Based SLO
+			"query": {
+				Type:          schema.TypeMap,
+				Optional:      true,
+				ConflictsWith: []string{"monitor_ids", "monitor_search"},
```
nmuesch: Should we be exhaustive here and conflict with `groups` as well?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	return
+}
+
+func buildServiceLevelObjectiveStruct(d *schema.ResourceData) *datadog.ServiceLevelObjective {
+
+	slo := datadog.ServiceLevelObjective{
+		Type:        datadog.String(d.Get("type").(string)),
+		Name:        datadog.String(d.Get("name").(string)),
+		Description: datadog.String(d.Get("description").(string)),
```
nmuesch: Since description is optional, should we check if it exists here?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+		sloThresholds := make(datadog.ServiceLevelObjectiveThresholds, 0)
+		for _, rawThreshold := range attr.([]interface{}) {
+			threshold := rawThreshold.(map[string]interface{})
+			t := datadog.ServiceLevelObjectiveThreshold{}
+			if tf, ok := threshold["timeframe"]; ok {
+				t.TimeFrame = datadog.String(tf.(string))
+			}
+
+			if targetValue, ok := threshold["target"]; ok {
+				t.Target = datadog.Float64(targetValue.(float64))
```
nmuesch: Is it safe to type assert without checking on these optional fields?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,433 @@
...
+		thresholds = append(thresholds, t)
+	}
+
+	tags := make([]string, 0)
+	for _, s := range slo.Tags {
+		tags = append(tags, s)
+	}
+	sort.Strings(tags)
+
+	log.Printf("[DEBUG] service level objective: %+v", slo)
```
nmuesch: Leftover?

website/docs/r/examples/service_level_objectives.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+The following attributes are exported:
+
+* `id` - ID of the Datadog service level objective
+
+## Import
+
+Service Level Objectives can be imported using their string ID, e.g.
+
+```
+$ terraform import datadog_service_level_objective.bytes_received_localhost "foo"
```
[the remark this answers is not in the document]
nmuesch: Based on this I think this should be the ID of the SLO https://github.com/zorkian/go-datadog-api/blob/93f907f168dda135d23d6e739ea11c3bf4677c01/service_level_objectives_test.go#L20

datadog/resource_datadog_service_level_objective.go
platinummonkey: https://golang.org/pkg/strings/#TrimSpace only trims leading and trailing space

datadog/resource_datadog_service_level_objective.go
platinummonkey: will check and fix

datadog/resource_datadog_service_level_objective.go
platinummonkey: yes

website/docs/r/examples/service_level_objectives.html.markdown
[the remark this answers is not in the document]
platinummonkey: fixed :D

datadog/resource_datadog_service_level_objective_test.go
```diff
@@ -0,0 +1,199 @@
...
+resource "datadog_service_level_objective" "foo" {
+  name = "name for metric SLO foo"
+  type = "metric"
+  description = "some description about foo SLO"
+  query = {
+	numerator = "sum:my.metric{type:good}.as_count()"
+	denominator = "sum:my.metric{*}.as_count()"
+  }
+
+  thresholds = [
```
nmuesch: Terraform 0.12 doesn't seem to support this syntax. It looks like TF 0.11 (at least 0.11.10) + 0.12 allow you to use 

```
thresholds {
}

thresholds {
}
```

Could we use that syntax to support more? (Also in the docs)

gzussa:
related https://github.com/terraform-providers/terraform-provider-datadog/issues/179

website/docs/r/service_level_objective.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  thresholds {
+    timeframe = "30d"
+    target = 99.9
+    warning = 99.99
+    target_display = "99.900"
+    warning_display = "99.990"      
+  }
+  ]
```
nmuesch: leftover char ?

website/docs/r/service_level_objective.html.markdown
```diff
@@ -0,0 +1,123 @@
...
+## Example Usage
+
+### Metric-Based SLO
+```hcl
+# Create a new Datadog service level objective
+resource "datadog_service_level_objective" "foo" {
+  name               = "Example Metric SLO"
+  type               = "metric"
+  description        = "My custom metric SLO"
+  query = {
```
nmuesch: Since `query` is a typeList now, we need to remove the `=`

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,489 @@
...
+	return []*schema.ResourceData{d}, nil
+}
+
+// Ignore any diff that results from the mix of *_display string values from the
+// DataDog API.
+func suppressDataDogSLODisplayValueDiff(k, old, new string, d *schema.ResourceData) bool {
+	sloType := d.Get("type")
+	if sloType == datadog.ServiceLevelObjectiveTypeMonitor {
+		// always suppress monitor type, this is controlled via API.
+		return false
```
nmuesch: should this be `return true`?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -0,0 +1,489 @@
...
+		for _, v := range attr.(*schema.Set).List() {
+			s = append(s, v.(string))
+		}
+		sort.Strings(s)
+		slo.Tags = s
+	}
+
+	if attr, ok := d.GetOk("thresholds"); ok {
+		sloThresholds := make(datadog.ServiceLevelObjectiveThresholds, 0)
+		thresholds := attr.([]map[string]interface{})
```
nmuesch: TF 0.11 gives me a panic here on subsequent updates. Likely related to my comment on the diffsuppress function.

website/docs/r/service_level_objective.html.markdown
nmuesch: I think this is the last thing then I think we're good to merge!

PR #264, opened 2019-07-19
Downtime fixes

bkabrda:
This PR includes 2 fixes:

* On updates to downtime attributes, `start` and `end` are no longer sent when unchanged. This makes it possible to update various downtime attributes without having to touch the timestamps.
* Fixes non-empty plan for monitors that have empty `monitor_tags`.
  * Note that this issue wasn't previously discovered because of a faulty test (that I fixed in the same commit). The test is faulty because it seems `ConflictsWith` doesn't work right when one of the attributes has its content from a variable - this seems to be a bug in Terraform and I'm currently discussing that with Hashicorp folks.

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -144,6 +144,11 @@ func resourceDatadogDowntime() *schema.Resource {
 }
 
 func buildDowntimeStruct(d *schema.ResourceData) *datadog.Downtime {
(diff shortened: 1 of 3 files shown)


bkabrda:
Update: looks like the issue with `ConflictsWith` is resolved in Terraform 0.12.5 - as soon as this PR is merged, I'll open another one which will do dependency update, so that we would be building against the latest TF SDK.

bkabrda:
Thanks for the review, merging!

PR #267, opened 2019-07-23
Make sure downtime resource properly fixes state drifts in start/end attributes

bkabrda:
This is a followup to #264, which made the provider ignore drifts in start/end attributes for resources. This PR makes sure that the drifts are detected and fixed properly.

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -152,18 +152,12 @@ func resourceDatadogDowntime() *schema.Resource {
 	}
 }
 
-// getDowntimeBoundaryTimestamp returns an int timestamp for start/end and whether or not
-// to apply this timestamp to the Downtime structure for POST/PUT request. Arguments:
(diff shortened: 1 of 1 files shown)


datadog/resource_datadog_downtime.go
```diff
@@ -152,26 +152,77 @@ func resourceDatadogDowntime() *schema.Resource {
...
+		}
+	} else {
+		// when creating, we always apply
+		apply = true
+	}
+
+	return ts, apply
+}
+
+func buildDowntimeStruct(d *schema.ResourceData, client *datadog.Client, updating bool) (*datadog.Downtime, error) {
```
gzussa: Let's split this method in 2:
- One that takes care of retrieving and transforming the correct date
- One that deals with understanding if we should update or not. Let's document this one as well a little bit more to better explain why we need this.

bkabrda: Done, please re-review. Thanks for the feedback!

bkabrda:
Thanks for your reviews! Merging.

PR #272, opened 2019-07-30
[dashboards] Add resources for log, apm and process query in legacy dashboards

jbenais:
- Fixes issue: https://github.com/terraform-providers/terraform-provider-datadog/issues/120
- Adds new tests for the new log, apm and process query params.
- Updates the documentation for dashboards, screenboards and timeboards resources.

*This functionality relies on a change in the Go client that is in a merged PR: zorkian/go-datadog-api#251*

### DIFF
--- a/datadog/resource_datadog_screenboard.go
+++ b/datadog/resource_datadog_screenboard.go
@@ -916,7 +916,7 @@ func buildTileDefRequestsApmOrLogQuery(source interface{}) *datadog.TileDefApmOr
 		d.Search = &s
 	}
 	// GroupBy
-	d.GroupBy = buildTileDefRequestsGroupBys(datadogQuery["groupBy"])
+	d.GroupBy = buildTileDefRequestsGroupBys(datadogQuery["group_by"])
 	return &d
 }
 func buildTileDefRequestsProcessQuery(source interface{}) *datadog.TileDefProcessQuery {
--- a/datadog/resource_datadog_screenboard.go
+++ b/datadog/resource_datadog_screenboard.go
@@ -872,13 +872,15 @@ func buildTileDefRequestsGroupBys(source interface{}) []datadog.TileDefApmOrLogQ
 			}})
 		if groupBySort, ok := groupByMap["sort"].(map[string]interface{}); ok {
 			s := datadog.TileDefApmOrLogQueryGroupBySort{}
-			batchSetFromDict(batch{
-				dict: groupBySort,
-				matches: []match{
-					{"aggregation", &s.Aggregation},
-					{"order", &s.Order},
-					{"facet", &s.Facet},
-				}})
+			if aggr, ok := groupBySort["aggregation"].(string); ok && len(aggr) > 0 {
+				s.Aggregation = datadog.String(aggr)
+			}
+			if order, ok := groupBySort["order"].(string); ok && len(order) > 0 {
+				s.Order = datadog.String(order)
+			}
+			if facet, ok := groupBySort["facet"].(string); ok && len(facet) > 0 {
+				s.Facet = datadog.String(facet)
+			}
 			d.Sort = &s
 		}
 		r = append(r, d)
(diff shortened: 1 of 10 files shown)


datadog/resource_datadog_screenboard.go
```diff
@@ -88,6 +88,115 @@ func resourceDatadogScreenboard() *schema.Resource {
...
+							"aggregation": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"facet": {
+								Type:     schema.TypeString,
+								Optional: true,
+							},
+							"interval": {
+								Type:     schema.TypeInt,
```
enbashi: This looks wrong for the legacy API. I think it should be a string. Try to import a dashboard created by the UI. You will get:
`Error: datadog_screenboard.import (import id:xxx-xxx-xxx): import datadog_screenboard.import (id: xxx-xxx-xxx): json: cannot unmarshal string into Go struct field TileDefApmOrLogQueryCompute.interval of type int

nmuesch:
Hey @jbenais thanks for the PR! The golang client was released + bumped in this provider. Can you rebase from master and we can work on getting this reviewed?

datadog/resource_datadog_screenboard.go
```diff
@@ -741,6 +853,96 @@ func buildTileDefRequestsConditionalFormats(source interface{}) []datadog.Condit
...
+	// Index
+	d := datadog.TileDefApmOrLogQuery{
+		Index: datadog.String(datadogQuery["index"].(string)),
+	}
+
+	// Compute
+	terraformCompute := datadogQuery["compute"].(map[string]interface{})
+	datadogCompute := datadog.TileDefApmOrLogQueryCompute{}
+	if aggr, ok := terraformCompute["aggregation"].(string); ok && len(aggr) != 0 {
+		datadogCompute.Aggregation = datadog.String(aggr)
```
nmuesch: Should we use the SetX methods from the datadog accessors for these?

nmuesch: Not sure the history, but I notice we aren't really using any of the GetX or SetX methods in these resources.

datadog/resource_datadog_timeboard.go
```diff
@@ -13,15 +13,127 @@ import (
 )
 
 func resourceDatadogTimeboard() *schema.Resource {
+	apmOrLogQuery := &schema.Schema{
```
nmuesch: Is this the same schema definition thats also in screenboards? If so do you think its worth it to find a way to refactor this so both reference the same definition?

datadog/resource_datadog_timeboard_test.go
```diff
@@ -17,25 +17,26 @@ resource "datadog_timeboard" "acceptance_test" {
   description = "Created using the Datadog provider in Terraform"
   read_only = true
   graph {
-    title = "Top System CPU by Docker container"
```
nmuesch: Looks like the indentation here got updated with tabs vs spaces.

datadog/resource_datadog_timeboard.go
jbenais: It's _almost_ the same, but some fields like `compute->interval` (that is an integer in timeboards, and a string in screenboards) are different.

website/docs/r/screenboard.html.markdown
```diff
@@ -592,6 +657,34 @@ The nested `style` blocks has the following structure:
 - `width` - (Optional) Line width. Possible values: "thin", "normal", "thick". Default: "normal".
 - `type` - (Optional) Type of line drawn. Possible values: "dashed", "solid", "dotted". Default: "solid".
 
+### Nested `widget` `tile_def` `request` `apm_query` and `log_query` blocks
+
+Nested `apm_query` and `log_query` blocks have the following structure (Visit the [ Graph Primer](https://docs.datadoghq.com/graphing/) for more information about these values):
+
+- `index` - (Required)
```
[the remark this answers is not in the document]
jbenais: I think it's already done above (i.e. lign 602 to 604)

datadog/resource_datadog_timeboard.go
```diff
@@ -371,15 +486,111 @@ func buildTemplateVariables(terraformTemplateVariables *[]interface{}) *[]datado
...
+	if v, ok := terraformQuery["metric"].(string); ok && len(v) != 0 {
+		datadogQuery.SetMetric(v)
+	}
+	if v, ok := terraformQuery["search_by"].(string); ok {
+		datadogQuery.SetSearchBy(v)
+	}
+
+	if terraformFilterBys, ok := terraformQuery["filter_by"].([]interface{}); ok {
+		datadogFilterbys := make([]string, len(terraformFilterBys))
+		for i, filtrBy := range terraformFilterBys {
```
nmuesch: filter?

PR #273, opened 2019-07-30
Add example Process Monitor

nmuesch:
Create an example of a process monitor in the Monitor Examples page of the TF Provider docs


nmuesch:
Good catch! I fixed and I'm going to merge this.

PR #277, opened 2019-08-02
Add support for style block in dashboard widgets

enbashi:
Add support for `style` block in dashboard widgets. Fixes https://github.com/terraform-providers/terraform-provider-datadog/issues/275.


enbashi:
While both the datadog API and Go client support `style` param in toplist widget, it's not offered as an option in our widget editor UI. I added the support for it in code, but I intentionally omitted it in the tests and code to minimize confusion.

website/docs/r/dashboard.html.markdown
```diff
@@ -683,6 +709,10 @@ Nested `widget` blocks have the following structure:
             - `log_query`: (Optional) The log query to use in the widget. The structure of this block is described [below](dashboard.html#nested-apm_query-and-log_query-blocks).
             - `process_query`: (Optional) The process query to use in the widget. The structure of this block is described [below](dashboard.html#nested-process_query-blocks).
             - `display_type` - (Optional) Type of display to use for the request. Available values are: `area`, `bars`, or `line`.
+            - `style` - (Optional) Style of the widget graph. One nested block is allowed with the following structure:
+              - `palette` - (Optional) Color palette to apply to the widget.
```
nmuesch: Do we have docs about what values these can be? Can we link to those here?

enbashi: Good catch. For some reason markdown link syntax was causing the URLs to disappear from the docs when added in a nested list item so I just added the link URL as text instead.

PR #278, opened 2019-08-02
Add support for metadata block in dashboard widgets

enbashi:
Add support for `metadata` block in dashboard widgets. Fixes https://github.com/terraform-providers/terraform-provider-datadog/issues/271. To go after https://github.com/terraform-providers/terraform-provider-datadog/pull/277.


website/docs/r/dashboard.html.markdown
```diff
@@ -683,6 +713,13 @@ Nested `widget` blocks have the following structure:
             - `log_query`: (Optional) The log query to use in the widget. The structure of this block is described [below](dashboard.html#nested-apm_query-and-log_query-blocks).
             - `process_query`: (Optional) The process query to use in the widget. The structure of this block is described [below](dashboard.html#nested-process_query-blocks).
             - `display_type` - (Optional) Type of display to use for the request. Available values are: `area`, `bars`, or `line`.
+            - `style` - (Optional) Style of the widget graph. One nested block is allowed with the following structure:
+              - `palette` - (Optional) Color palette to apply to the widget.
+              - `line_type` - (Optional) Type of lines displayed. Available values are: `dashed`, `dotted`, or `solid`.
+              - `line_width` - (Optional) Width of line displayed. Available values are: `normal`, `thick`, or `thin`.
+            - `metadata` - (Optional). Used to define expression aliases. Multiple nested blocks are allowed with the following structure:
+              - `expression` - (Required)
+              - `alias_name` - (Optional)
```
nmuesch: Is there a reason one would add the expression but not an alias for it?

datadog/resource_datadog_dashboard.go
```diff
@@ -2976,6 +3046,43 @@ func getTimeseriesRequestSchema() map[string]*schema.Schema {
...
+			},
+		},
+		"metadata": {
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"expression": {
+						Type:     schema.TypeString,
+						Optional: true,
```
nmuesch: Should this be required (the docs changes say this field is required)

datadog/resource_datadog_dashboard.go
```diff
@@ -3000,6 +3107,28 @@ func buildDatadogTimeseriesRequests(terraformRequests *[]interface{}) *[]datadog
...
+		if terraformMetadataList, ok := terraformRequest["metadata"].([]interface{}); ok && len(terraformMetadataList) > 0 {
+			datadogMetadataList := make([]datadog.WidgetMetadata, len(terraformMetadataList))
+			for i, _metadata := range terraformMetadataList {
+				metadata := _metadata.(map[string]interface{})
+				// Expression
+				datadogMetadata := datadog.WidgetMetadata{
+					Expression: datadog.String(metadata["expression"].(string)),
+				}
+				// AliasName
+				if v, ok := metadata["alias_name"].(string); ok && len(v) != 0 {
```
nmuesch: does this need the same `String` call as the one above?

website/docs/r/dashboard.html.markdown
enbashi: probably bot a typical use case, but the UI allows it, and so does the the API and Go client.

datadog/resource_datadog_dashboard.go
enbashi: You're absolutely correct. It is a required. Thanks for catching this.

PR #279, opened 2019-08-07
[Synthetics] Support SSL

btoueg:
- Add support for ssl tests
 - Add acceptance test
 - Update doc
 - Have subtype defaults to http for api tests
 - Make sure existing TF files won't diff

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -32,8 +32,17 @@ func resourceDatadogSyntheticsTest() *schema.Resource {
 				ValidateFunc: validation.StringInSlice(syntheticsTypes, false),
 			},
 			"subtype": {
-				Type:         schema.TypeString,
-				Optional:     true,
+				Type:     schema.TypeString,
+				Optional: true,
+				DiffSuppressFunc: func(key, old, new string, d *schema.ResourceData) bool {
+					if key == "subtype" {
+						if d.Get("type") == "api" && old == "http" && new == "" {
+							// defaults to http if type is api for retro-compatibility
+							return true
+						}
+					}
+					return old == new
+				},
 				ValidateFunc: validation.StringInSlice(syntheticsSubTypes, false),
 			},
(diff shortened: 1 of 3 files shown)


nmuesch:
Hey @btoueg  thanks for the PR! The golang client was released + bumped in this provider. Can you rebase from master and we can work on getting this reviewed?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -316,6 +359,9 @@ func newSyntheticsTestFromLocalState(d *schema.ResourceData) *datadog.Synthetics
 		minLocationFailed, _ := strconv.Atoi(attr.(string))
 		options.SetMinLocationFailed(minLocationFailed)
 	}
+	if attr, ok := d.GetOk("options.accept_self_signed"); ok {
+		options.SetAcceptSelfSigned(attr.(string) == "true")
```
nmuesch: Any reason not to just assign this to `attr.(string)` Should we also safely type assert this?

website/docs/r/synthetics.html.markdown
```diff
@@ -43,6 +45,38 @@ resource "datadog_synthetics_test" "foo" {
 }
 ```
 
+## Example Usage (Synthetics SSL test)
+
+Create a new Datadog Synthetics API/SSL test on example.org
+
+```hcl
+resource "datadog_synthetics_test" "foo" {
```
nmuesch: Lets name this something else to not clash with the other example named `foo`

datadog/resource_datadog_synthetics_test_.go
btoueg: Not sure, can't wrap my head around it. It's value is "1" or "0" in TF 0.11, even though is never set to such value

btoueg:
> two successive plans cause an un-empty diff

Indeed, that's the case in TF 0.11. I've handled it, let me know

nmuesch:
Thanks for the iterations here! Looks good to me, I'll go ahead and merge. Thanks for this 🎉🚀

PR #286, opened 2019-08-12
Enable ForceNew=true for dashboard.layout_type

usererror:
Datadog dashboard's `layout_type` parameter cannot be updated/modified through the API:

```
Error: Failed to update dashboard using Datadog API: API error 400 Bad Request: {"errors": ["'layout_type' is not editable for existing dashboards. This dashboard uses 'ordered' layout."]}
```

As the error helpfully points out, dashboards must be destroyed and recreated to modify the `layout_type`. Turning on `ForceNew` forces Terraform to follow this protocol when changes are made to `layout_type`.

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -39,6 +39,7 @@ func resourceDatadogDashboard() *schema.Resource {
 				Required:     true,
 				Description:  "The layout type of the dashboard, either 'free' or 'ordered'.",
 				ValidateFunc: validateDashboardLayoutType,
+				ForceNew:     true,
 			},
 			"description": {
 				Type:        schema.TypeString,


nmuesch:
Hey @usererror thanks for the contribution! This is an interesting change and I've brought it up internally to discuss. I'm a bit hesitant about this change and wanted to jot down my thoughts here. 

I believe the original reason we haven't set this was primarily due to the differences in the widgets/allowed fields between a `free` layout and an `ordered` one. So simply changing this field without making the proper modifications to the remaining config would result in the dashboard being deleted and the new dashboard failing to be created. 

There are a few options I think we could take:
* Leave the field without ForceNew and better document that users would have to delete and recreate the new dashboard
* Remove the forceNew and add some new logic to the `update` command that would first validate that the new config works (by hitting an api endpoint), continue if thats valid, otherwise fail with some meaningful error. 
* Add ForceNew and document this new behavior. - I want to try and avoid leaving users in a broken state if we can though 😄 

Mind sharing some context about your specific use case on this?

Issue #293, opened 2019-08-20
datadog_integration_aws: account_id attribute should be ForceNew


mlclmj:
### Terraform Version

```
$ terraform -v
Terraform v0.12.6
```

### Affected Resource
- datadog_integration_aws

### Terraform Configuration Files
```hcl
resource "datadog_integration_aws" "account-a" {
  account_id  = "1234"
  role_name   = "DatadogAWSIntegrationRole"
}
```

### Expected Behavior
Changing the `account_id` attribute of the AWS account integration should've updated the integration as stated in the `plan` output:

<img width="556" alt="Screen Shot 2019-08-20 at 4 44 13 PM" src="https://user-images.githubusercontent.com/17786500/63383483-e22c1400-c36a-11e9-9904-e79653f6aa4b.png">

### Actual Behavior
`terraform apply` failed:

<img width="569" alt="Screen Shot 2019-08-20 at 4 53 13 PM" src="https://user-images.githubusercontent.com/17786500/63383556-07208700-c36b-11e9-9a9f-010003ca03ab.png">

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Run `terraform apply` using the example config above.
2. Modify the `account_id` attribute.
3. Run `terraform apply` again.

nmuesch:
Hey @mlclmj Apologies for the delay here. With the latest provider versions (since 2.8.0 - https://github.com/terraform-providers/terraform-provider-datadog/blob/master/CHANGELOG.md#280-june-10-2020) The AWS resource supports a proper update call without the need for forceNew. This should allow you to properly change and update the fields of this schema. I'll go ahead and close this issue for now, but let us know if you continue to hit any issues with this.

PR #294, opened 2019-08-20
resource_datadog_integration_aws: Make account_id ForceNew and remove Update

mlclmj:
As I brought up in #293, `account_id` should be `ForceNew` as the Datadog API doesn't support updating this attribute. 

Setting `ForceNew` for the `account_id` requires removing the `Update` method of this resource since the resource no longer has any attributes that can be updated and Terraform doesn't allow this. However, I don't think this removal constitutes a breaking change as the the public API doesn't support updating this resource–it was simply attempting to create a new resource. The comments indicate that this functionality may be arriving at some point in the future, however that was some time ago and it seems best to me to just remove this functionality until Datadog's public API supports it.

### DIFF
--- a/datadog/resource_datadog_integration_aws.go
+++ b/datadog/resource_datadog_integration_aws.go
@@ -24,7 +24,6 @@ func resourceDatadogIntegrationAws() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceDatadogIntegrationAwsCreate,
 		Read:   resourceDatadogIntegrationAwsRead,
-		Update: resourceDatadogIntegrationAwsUpdate,
 		Delete: resourceDatadogIntegrationAwsDelete,
 		Exists: resourceDatadogIntegrationAwsExists,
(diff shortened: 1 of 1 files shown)


nmuesch:
Hey @mlclmj apologies for the delay here and thanks for the contribution! We actually have since introduced an `update` method for the AWS integration - https://docs.datadoghq.com/api/?lang=python#update-an-aws-integration (I think this may have come after your initial PR 😄)

The endpoint takes in query params (account_id and role_name) that reflect the AWS Account you're looking for as well as arguments to update individual fields of that account. This isn't something that is currently implemented in the underlying golang client yet so support would need to be added there first.

Given that this endpoint now exists, we'd have to remove `ForceNew` or add new attributes once the update method is properly introduced, so I think it'd be best not to include that for now. I don't have a current timeframe on when this would be worked on, but we're happy to provide any PRs should you like to take a look! Thanks again.

Issue #216, opened 2019-05-22
Provide data source for Datadog IP ranges


julianvmodesto:
Datadog provides the metadata endpoint https://ip-ranges.datadoghq.com/ for IP addresses.

We can create a data source for this like for the GitHub or AWS Providers:

https://www.terraform.io/docs/providers/github/d/ip_ranges.html
https://www.terraform.io/docs/providers/aws/d/ip_ranges.html

bkabrda:
Implemented in #298, will be part of the next release. Closing this issue.

PR #298, opened 2019-08-22
Add data source - IP ranges

dabcoder:
Attempts to address #216.


datadog/data_source_datadog_ip_ranges.go
```diff
@@ -0,0 +1,168 @@
...
+	client := meta.(*datadog.Client)
+
+	ipAddresses, err := client.GetIPRanges()
+
+	if err != nil {
+		return err
+	}
+
+	// v4
+	if len(ipAddresses.Agents["prefixes_ipv4"])+len(ipAddresses.API["prefixes_ipv4"])+len(ipAddresses.Apm["prefixes_ipv4"])+len(ipAddresses.Logs["prefixes_ipv4"])+len(ipAddresses.Process["prefixes_ipv4"])+len(ipAddresses.Synthetics["prefixes_ipv4"])+len(ipAddresses.Webhooks["prefixes_ipv4"]) > 0 {
```
bkabrda: So IIUC both IPv4 and IPv6 ranges could be returned in a single request to our API, so I think it should be sufficient to have a single condition checking len of all IPv4 and IPv6. Is there a reason why you did these two separately?

dabcoder: In case we would use one _and not_ the other for a given site. e.g. just IPv4 and not IPv6 we would just set an `id` for IPv4. Am I assessing this incorrectly?

bkabrda: Well there's only one Id of a given data source and the Id doesn't really matter to the users. If we only used IPv4 and not IPv6 for certain site, we would still have to save *some* id, as some of the lists wouldn't be empty. The value of the Id itself doesn't matter - what matters is whether or not it's set in this case.

PR #302, opened 2019-08-27
Update docs to explicitly note changing layout requires recreate

nmuesch:
Explicitly note that changing the layout type requires a destroy + re create of the dashboard.
Addresses - https://github.com/terraform-providers/terraform-provider-datadog/pull/286


bkabrda:
So maybe we should add a note that explains why we didn't make it `ForceNew`?

PR #304, opened 2019-08-28
Make sure PD services don't get removed by updating PD resource

### LINKED ISSUE #299: 404 error finding datadog_integration_pagerduty_service_object
Hello,

I am encountering an issue with the datadog pagerduty integration, specifically, the `datadog_integration_pagerduty_service_object` resource.

Initial creation of the resource works via `terraform plan` / `apply`.

Subsequent updates fail with:

```
Error: Error refreshing state: 1 error occurred:
	* module.main_monolith_dd_monitors.datadog_integration_pagerduty_service_object.datadog_pd_integration: 1 error occurred:
	* module.main_monolith_dd_monitors.datadog_integration_pagerduty_service_object.datadog_pd_integration: datadog_integration_pagerduty_service_object.datadog_pd_integration: API error 404 Not Found: {"errors": ["Service not found"]}
```
The configuration for the resource is:

```
resource "datadog_integration_pagerduty_service_object" "datadog_pd_integration" {
  depends_on   = ["datadog_integration_pagerduty.pd"]
  service_name = "${var.pd_service_name}"
  service_key  = "${var.pd_service_key}"
}
```

The variable `pd_service_name` or `pd_service_key` do not change between the initial creation and subsequent updates.

An odd workaround I've found is to remove the `datadog_integration_pagerduty_service_object` from state manually, and re-plan/applying.

### Terraform Version
➜  production git:(master) ✗ terraform -v
Terraform v0.11.14
+ provider.aws v1.60.0
+ provider.datadog v2.2.0
+ provider.grafana v1.5.0
+ provider.pagerduty v1.3.1
+ provider.random v2.2.0
+ provider.sentry (unversioned)
+ provider.template v2.1.2

### Affected Resource(s)
- `datadog_integration_pagerduty_service_object`

### Expected Behavior
I would expect my module to be updated.

### Actual Behavior
The module does not get updated, as the `datadog_integration_pagerduty_service_object` is not found.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
There are two modules at play here. 1 module creates the `pagerduty_service` and the `pagerduty_service_integration` to datadog.

The other module is the one that implements the `datadog_integration_pagerduty_service_object` as well as several monitors. The second module is where the failure is occurring.

bkabrda:
Previously, doing any update on the PD resource would delete the individual service objects. This PR fixes that (at the cost of doing the migration from inline services to individual service objects somewhat harder, but that's a one-time migration process, so shouldn't hurt a lot - I don't see a better way to solve this).

Fixes #299.

### DIFF
--- a/datadog/resource_datadog_integration_pagerduty.go
+++ b/datadog/resource_datadog_integration_pagerduty.go
@@ -167,15 +167,17 @@ func resourceDatadogIntegrationPagerdutyUpdate(d *schema.ResourceData, meta inte
 	// if there are none currently configured services, we actually
 	// have to remove them explicitly, otherwise the underlying API client
 	// would not send the "services" key at all and they wouldn't get deleted
-	currentServices := d.Get("services").([]interface{})
-	if len(currentServices) == 0 {
-		pd, err := client.GetIntegrationPD()
-		if err != nil {
-			return fmt.Errorf("Error while deleting Pagerduty integration service object: %v", err)
-		}
-		for _, service := range pd.Services {
-			if err := client.DeleteIntegrationPDService(*service.ServiceName); err != nil {
+	if value, ok := d.GetOk("individual_services"); !ok || !value.(bool) {
+		currentServices := d.Get("services").([]interface{})
+		if len(currentServices) == 0 {
+			pd, err := client.GetIntegrationPD()
+			if err != nil {
 				return fmt.Errorf("Error while deleting Pagerduty integration service object: %v", err)
+			}
+			for _, service := range pd.Services {
+				if err := client.DeleteIntegrationPDService(*service.ServiceName); err != nil {
+					return fmt.Errorf("Error while deleting Pagerduty integration service object: %v", err)
+				}
 			}
 		}
 	}
--- a/datadog/resource_datadog_integration_pagerduty_service_object_test.go
+++ b/datadog/resource_datadog_integration_pagerduty_service_object_test.go
@@ -1,6 +1,7 @@
 package datadog
 
 import (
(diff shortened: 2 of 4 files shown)


nmuesch:
Looks good overall. Just want to verify I understand the issue/solution:

* User has no services setup under the pd resource but under the service resource
* User runs TF Apply and everything is created, pd and services
* User runs TF Apply again, and there is a sensitive field `api_token` that the API doesn't return so TF assumes its changed and tries to update
* Since only the original PD resource updated, it would delete all services and not re apply all underlying services since they are part of a separate resource?

While this does fix that issue, I think the underlying issue is that the `api_token` isn't returned by the API, as expected, but terraform always thinks the pd resource needs to be updated. I think we  should add the `ignore_changes` block to the pagerduty resource docs. Similar to this section - https://www.terraform.io/docs/providers/datadog/r/monitor.html#silencing-by-hand-and-by-downtimes. What do you think?

bkabrda:
@nmuesch you're right in all your points, but the third point is not complete - the individual services would get deleted on *any* change to the pagerduty resource, not just change to the `api_token` (this is shown in the test, which actually changes `subdomain` instead of `api_key`).

I don't consider the `api_token` doc change to be strictly related to this, so I don't want to get it mixed in this PR, but I think what you're proposing for it makes sense.

nmuesch:
Alright thanks for the confirmation. The only reason I brought up `api_token` is that it's going to cause every plan to see an update, instead of the individual cases when another field is explicitly updated. I'm happy to do this in a separate PR though.

bkabrda:

Issue #217, opened 2019-05-24
Provide data source for Synthetics locations


julianvmodesto:
/api/v1/synthetics/locations returns data on available locations for Synthetics endpoints, and it requires auth.

This data source can be consumed by the Synthetics test resource.

https://docs.datadoghq.com/api/?lang=bash#get-available-locations

therve:
Fixed by #309 .

PR #309, opened 2019-09-02
Add synthetics locations data source

dabcoder:
Attempts to address #217.
related to https://github.com/terraform-providers/terraform-provider-datadog/issues/381

### DIFF
--- a/datadog/data_source_datadog_synthetics_locations.go
+++ b/datadog/data_source_datadog_synthetics_locations.go
@@ -1,60 +1,46 @@
 package datadog
 
 import (
-	"github.com/hashicorp/terraform/helper/schema"
-	datadog "github.com/zorkian/go-datadog-api"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
 func dataSourceDatadogSyntheticsLocations() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceDatadogSyntheticsLocationsRead,
 
-		// Locations are a list of string
+		// Locations are a map of IDs to names
 		Schema: map[string]*schema.Schema{
 			"locations": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeMap,
 				Computed: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 		},
 	}
 }
 
-type LocationMap map[string]string
-
 func dataSourceDatadogSyntheticsLocationsRead(d *schema.ResourceData, meta interface{}) error {
 
-	client := meta.(*datadog.Client)
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV1 := providerConf.DatadogClientV1
+	authV1 := providerConf.AuthV1
 
-	syntheticsLocations, err := client.GetSyntheticsLocations()
+	syntheticsLocations, _, err := datadogClientV1.SyntheticsApi.ListLocations(authV1).Execute()
 
 	if err != nil {
 		return err
 	}
 
-	// Create a list of location names to be used in the data source
-	var locationsList []string
+	locationsMap := make(map[string]string)
 
-	// Declare the list of maps
-	var LocationMapSlice []LocationMap
-
-	// Fill locationsList with the above map containing region, display name and name
-	for _, location := range syntheticsLocations {
-		// access the pointer of each struct element
-		lm := LocationMap{"region": location.GetRegion(), "display_name": location.GetDisplayName(), "name": location.GetName()}
-		LocationMapSlice = append(LocationMapSlice, lm)
+	for _, location := range syntheticsLocations.GetLocations() {
+		locationsMap[location.GetId()] = location.GetName()
 	}
 
-	for _, location := range LocationMapSlice {
-		// In order to create a Synthetics test, we only need a list of location "name"
-		locationsList = append(locationsList, location["name"])
-	}
-
-	if len(syntheticsLocations) > 0 {
+	if len(locationsMap) > 0 {
 		d.SetId("datadog-synthetics-location")
-		d.Set("locations", locationsList)
+		d.Set("locations", locationsMap)
 	} else {
-		d.Set("locations", []string{})
+		d.Set("locations", locationsMap)
 	}
 
 	return nil
--- a/datadog/data_source_datadog_synthetics_locations.go
+++ b/datadog/data_source_datadog_synthetics_locations.go
@@ -1,0 +1,50 @@
+package datadog
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	datadog "github.com/zorkian/go-datadog-api"
+)
+
+func dataSourceDatadogSyntheticsLocations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDatadogSyntheticsLocationsRead,
+
+		// Locations are a list of string
+		Schema: map[string]*schema.Schema{
+			"locations": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func dataSourceDatadogSyntheticsLocationsRead(d *schema.ResourceData, meta interface{}) error {
+
+	client := meta.(*datadog.Client)
+
+	syntheticsLocations, err := client.GetSyntheticsLocations()
+
+	if err != nil {
+		return err
+	}
+
+	// Create a list of location names to be used in the data source
+	var locationsList []string
+
+	// Fill locationsList with location names
+	for _, location := range syntheticsLocations {
+		// access the pointer of the struct element DisplayName
+		locationsList = append(locationsList, location.GetDisplayName())
+	}
+
+	if len(syntheticsLocations) > 0 {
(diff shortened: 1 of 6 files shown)


website/docs/d/synthetics_locations.html.markdown
```diff
@@ -0,0 +1,18 @@
...
+# datadog_synthetics_locations
+
+Use this data source to retrieve information about Datadog's Synthetics Locations.
+## Example Usage
+
+```hcl
+data "datadog_synthetics_locations" "test" {}
+```
+
+## Attributes Reference
```
nmuesch: This should include a description for `locations` right?

datadog/data_source_datadog_synthetics_locations.go
```diff
@@ -0,0 +1,52 @@
...
+	datadog "github.com/zorkian/go-datadog-api"
+)
+
+func dataSourceDatadogSyntheticsLocations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDatadogSyntheticsLocationsRead,
+
+		// Locations are a list of string
+		Schema: map[string]*schema.Schema{
+			"locations": {
```
nmuesch: Any use case for returning the rest of the values this API endpoint returns (id, region, etc) ? We can add new schemas here later but changing this `locations` attribute to include them later will be challenging if we want them.

dabcoder: I don't think so, in order to create a test (with this data source location), the "name" is the only attribute used: https://docs.datadoghq.com/api/?lang=bash#create-a-test. 
If we go that route we may want to update the attribute to `locations_name` maybe? 
If we want to fetch all attributes from the locations I am not sure how this would work with a list of maps perhaps (my testing isn't working in that case).

website/docs/d/synthetics_locations.html.markdown
dabcoder: Yes indeed, will update

datadog/data_source_datadog_synthetics_locations.go
```diff
@@ -0,0 +1,52 @@
...
+		// access the pointer of the struct element DisplayName
+		locationsList = append(locationsList, location.GetDisplayName())
+	}
+
+	if len(syntheticsLocations) > 0 {
+		d.SetId("datadog-synthetics-location")
+		d.Set("locations", locationsList)
+	} else {
+		d.Set("locations", []string{})
+	}
```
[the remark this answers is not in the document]
dabcoder: In case the list of locations returned by the Go client is nil, we still set a `locations` attribute but with no values. Similar to https://github.com/terraform-providers/terraform-provider-datadog/blob/master/datadog/data_source_datadog_ip_ranges.go#L113:L118

datadog/data_source_datadog_synthetics_locations.go
gzussa: @btoueg We need your input on this one 😄

datadog/data_source_datadog_synthetics_locations.go
```diff
@@ -0,0 +1,52 @@
...
+		return err
+	}
+
+	// Create a list of location names to be used in the data source
+	var locationsList []string
+
+	// Fill locationsList with location names
+	for _, location := range syntheticsLocations {
+		// access the pointer of the struct element DisplayName
+		locationsList = append(locationsList, location.GetDisplayName())
```
btoueg: If there's only one field to keep, it's `Name` rather than `DisplayName`
```suggestion
		locationsList = append(locationsList, location.GetName())
```

datadog/data_source_datadog_synthetics_locations.go
btoueg: I recommend we keep at least `region`, `display_name` and `name`, e.g.:

```javascript
[{"region":"Asia Pacific","display_name":"Tokyo (AWS)","name":"aws:ap-northeast-1"}, ...]
```

How is that data supposed to be used? Can we provide an example in the doc?

I'm afraid an array will be less usable than a map

dabcoder: Thanks for the feedback. 

My understanding is that it would be defined as a data source such as in the below:

```
data "synthetics_locations" "example" {}
```

... in order to fetch all active locations and use them when creating/editing a test using something along the lines of:

```
resource "datadog_synthetics_test" "test_api" {
  type = "api"
  subtype = "http"
  request = {
    method = "GET"
    url = "https://www.example.org"
  }
  request_headers = {
    Content-Type = "application/json"
    Authentication = "Token: 1234566789"
  }
  assertions = [
    {
      type = "statusCode"
      operator = "is"
      target = "200"
    }
  ]
  locations = data.synthetics_locations.example.location_names
...
}

Instead of hardcoding all active locations one by one.

Comment (dabcoder) [reply datadog/data_source_datadog_synthetics_locations.go]: > I'm afraid an array will be less usable than a map

According to https://www.terraform.io/docs/providers/datadog/r/synthetics.html#example-usage-synthetics-api-test-, locations needs to be a list, correct?

Comment (btoueg) [reply datadog/data_source_datadog_synthetics_locations.go]: Yes it does need to be a list. It all depends how you want to use it, for instance, it could be something like:

```
resource "datadog_synthetics_test" "test_api" {
  ...
  locations = [
    data.synthetics_locations["Tokyo"],
    data.synthetics_locations["Sidney"],
  ]
...
}
```

Example above is not syntactically sound but you get the idea

Comment (bkabrda) [reply datadog/data_source_datadog_synthetics_locations.go]: So assuming the `name` (e.g. `aws:us-west-2`) is unique for all the locations (CC @btoueg - is it guaranteed to be unique?), I think the optimal solution would be to have `locations` as map:

```
`"aws:us-west-2"`: {
    "id": 123,
    "display_name":"Oregon (AWS)",
    ... 
}
```

and then users would be able to set up locations for the `datadog_synthetics_test` such as:

```
resource "datadog_synthetics_test" "test_api" {
    locations = data.synthetics_locations.example.locations.keys
}
```

That seems reasonably easy to use and allows us to carry additional data if/when necessary. Assuming it's not guaranteed that location names are unique, we'll have to go with list of maps and figure out a clever oneliner to access all the names (which, thinking about it, wouldn't make much sense if names weren't unique, so they probably are...) :)

Comment (dabcoder) [reply datadog/data_source_datadog_synthetics_locations.go]: `name` should be unique indeed, AWS regions don't include duplicates. 
I've created a map here: https://github.com/terraform-providers/terraform-provider-datadog/pull/309/files#diff-247b9e616dd8b78804cde056db1f776aR23 and filled it with `region`, `display_name` and `name` so far. 
Is that the right approach you think? Should we fetch more keys such as the `id`s, `is_active`?

Comment (bkabrda) [reply datadog/data_source_datadog_synthetics_locations.go]: @dabcoder what I meant was to make the top-level `locations` attribute not a list, but a map - so that it would become more easily accessible. IOW, the current attribute you use to set `locations` is `locationsList` and is a `[]string` - I'm suggesting that you make this a `map[string]LocationMap`. Does that make sense?

File: website/docs/d/synthetics_locations.html.markdown [hunk 5]
```diff
@@ -0,0 +1,25 @@
...
+page_title: "Datadog: datadog_synthetics_locations"
+sidebar_current: "docs-datadog-datasource-synthetics-locations"
+description: |-
+  Get information on Datadog's Synthetics locations for Synthetics tests
+---
+
+# datadog_synthetics_locations
+
+Use this data source to retrieve Datadog's Synthetics Locations (to be used in Synthetics tests).
+## Example Usage
```
jirikuncar, line 12:
```suggestion

## Example Usage
```

website/docs/d/synthetics_locations.html.markdown
```diff
@@ -0,0 +1,25 @@
...
+
+resource "datadog_synthetics_test" "test_api" {
+  type = "api"
+  locations = keys(data.datadog_synthetics_locations.test.locations)
+}
+```
+
+## Attributes Reference
+
+ * `locations` - An map of available Synthetics location IDs to names for Synthetics tests.
```
zippolyte, line 25:
```suggestion
 * `locations` - A map of available Synthetics location IDs to names for Synthetics tests.
```

datadog/data_source_datadog_synthetics_locations.go
```diff
@@ -0,0 +1,47 @@
...
+	for _, location := range syntheticsLocations.GetLocations() {
+		locationsMap[location.GetId()] = location.GetName()
+	}
+
+	if len(locationsMap) > 0 {
+		d.SetId("datadog-synthetics-location")
+		d.Set("locations", locationsMap)
+	} else {
+		d.Set("locations", locationsMap)
+	}
```
zippolyte, line 44: What's this for ?

therve: We only set the ID if it's not empty, but we still set the map. I guess it should be outside the condition, let me fix it.

docs/data-sources/synthetics_locations.md
```diff
@@ -0,0 +1,22 @@
+---
+page_title: "Datadog: datadog_synthetics_locations"
+---
+
+# datadog_synthetics_locations Data source
```
jirikuncar, line 5: Data Source

PR #311, opened 2019-09-11
Fix integration tests (again)

bkabrda:
Looks like nobody ran these for a while and some failures accumulated due to several changes, so here's the fix to make all tests pass again (also on TF 0.12)


bkabrda:

PR #312, opened 2019-09-11
[Datadog] Add support for Logs Public config API

tt810:
Supporting logs Public config API pipeline and pipeline order. for more information please read [this](https://docs.datadoghq.com/api/?lang=python#logs-pipelines)


datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	return nil
+}
+
+func buildFilter(d *schema.ResourceData, ddFilter *datadog.FilterConfiguration) {
+	if v, ok := d.Get("filter").([]interface{}); ok && len(v) > 0 {
+		convertFilter(v[0], ddFilter)
+	}
+}
+
+func convertFilter(tfFilter interface{}, ddFilter *datadog.FilterConfiguration) {
```
ajacquemot: same remarks as above, we could return to the result and prefer `to` to `convert` as it's shorter and used broadly in most of the programming languages

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	if tfF, ok := tfFilter.(map[string]interface{}); ok {
+		if query, ok := tfF["query"].(string); ok {
+			ddFilter.Query = &query
+		}
+	}
+}
+
+func getPipelineSchema(isNested bool) map[string]*schema.Schema {
```
ajacquemot: I feel like this could be static, e.g:
```
var pipelineSchema = map[string]*schema.Schema{...}
```

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	return map[string]*schema.Schema{
+		"name":       {Type: schema.TypeString, Required: true},
+		"is_enabled": {Type: schema.TypeBool, Optional: true},
+		"sources":    {Type: schema.TypeList, Required: true, Elem: &schema.Schema{Type: schema.TypeString}},
+		"target":     {Type: schema.TypeString, Required: true},
+		"is_encoded": {Type: schema.TypeBool, Required: true},
+	}
+}
+
+const (
```
ajacquemot: I think we tend to define const at the header of the files, from the go library, `buffer` package: https://golang.org/src/bytes/buffer.go?s=14279:14313#L16

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
+package datadog
```
ajacquemot: There are multiple logic within this file, conversion, constants, schemas, do we want to split into smaller/responsibility files?

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	tfProcessors := d.Get("processor").([]interface{})
+	ddProcessors := make([]datadog.LogsProcessor, len(tfProcessors))
+	if err := convertTFProcessors(ddProcessors, tfProcessors); err != nil {
+		return ddProcessors, err
+	}
+	return ddProcessors, nil
+}
+
+func convertTFProcessors(ddProcessors []datadog.LogsProcessor, tfProcessors []interface{}) error {
+	for i, tfEntry := range tfProcessors {
```
ajacquemot: nits
```suggestion
	for i, tfProcessor := range tfProcessors {
```

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+		tfMap := tfEntry.(map[string]interface{})
+		for tfPType, ddPType := range tfProcessorTypes {
+			if v, ok := tfMap[tfPType].([]interface{}); ok && len(v) > 0 {
+				if tfP, ok := v[0].(map[string]interface{}); ok {
+					var ddProcessor = datadog.LogsProcessor{}
+					if err := convert(tfP, ddPType, &ddProcessor); err != nil {
+						return err
+					}
+					ddProcessors[i] = ddProcessor
```
ajacquemot: Here we have a good example of why it can be dangerous to create the object to mutate outside of the method, if the capacity of this array was smaller than the number of elements in `tfProcessors`, the code could crash.

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+}
+
+func convertTFProcessors(ddProcessors []datadog.LogsProcessor, tfProcessors []interface{}) error {
+	for i, tfEntry := range tfProcessors {
+		tfMap := tfEntry.(map[string]interface{})
+		for tfPType, ddPType := range tfProcessorTypes {
+			if v, ok := tfMap[tfPType].([]interface{}); ok && len(v) > 0 {
+				if tfP, ok := v[0].(map[string]interface{}); ok {
+					var ddProcessor = datadog.LogsProcessor{}
+					if err := convert(tfP, ddPType, &ddProcessor); err != nil {
```
ajacquemot: convert is not very friendly here, what about:
```
newDatadogProcessor(type string, tfProcessor {}interface)
```

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	ddProcessors := make([]datadog.LogsProcessor, len(tfProcessors))
+	if err := convertTFProcessors(ddProcessors, tfProcessors); err != nil {
+		return ddProcessors, err
+	}
+	return ddProcessors, nil
+}
+
+func convertTFProcessors(ddProcessors []datadog.LogsProcessor, tfProcessors []interface{}) error {
+	for i, tfEntry := range tfProcessors {
+		tfMap := tfEntry.(map[string]interface{})
```
ajacquemot: tfMap is too generic IMO, do we want to enforce the type of `tfProcessors` upstream to make sure we don't have to cast too much ? e.g.
```
tfProcessors := tfProcessors.([]map[string]interface{})
for i, tfProcessor := range tfProcessors {
   ...
}
```

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+		for tfPType, ddPType := range tfProcessorTypes {
+			if v, ok := tfMap[tfPType].([]interface{}); ok && len(v) > 0 {
```
ajacquemot: I am not sure I understand this data structure, how can a processor have multiple types ?

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+
+func convert(tfProcessor map[string]interface{}, processorType string, ddProcessor *datadog.LogsProcessor) error {
+	ddProcessor.Type = datadog.String(processorType)
+	if name, ok := tfProcessor["name"].(string); ok {
+		ddProcessor.Name = &name
+	}
+	if isEnabled, ok := tfProcessor["is_enabled"].(bool); ok {
+		ddProcessor.IsEnabled = &isEnabled
+	}
+	switch processorType {
```
ajacquemot: Do we want to create one method per case ?

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	return nil
+}
+
+func convertDDProcessors(tfProcessors []map[string]interface{}, processors []datadog.LogsProcessor) error {
+	for i, ddProcessor := range processors {
+		tfProcessor := make(map[string]interface{})
+		tfPDetails := make(map[string]interface{})
+		tfPDetails["name"] = *ddProcessor.Name
+		tfPDetails["is_enabled"] = *ddProcessor.IsEnabled
+		switch *ddProcessor.Type {
```
ajacquemot: same here do we want one method per case ?

datadog/resource_datadog_logs_pipeline.go
tt810: Check the existing files. It looks we are keeping a single file for each resource. cc @bkabrda

datadog/resource_datadog_logs_pipeline.go
tt810: I am not sure I understand what you mean. This function takes an argument for taking care of the nested pipeline case.

datadog/resource_datadog_logs_pipeline.go
tt810: will split.

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+	client := meta.(*datadog.Client)
+	if _, err := client.UpdateLogsPipeline(d.Id(), pipeline); err != nil {
+		return fmt.Errorf("error updating logs pipeline: (%s)", err.Error())
+	}
+	return resourceDatadogLogsPipelineRead(d, meta)
+}
+
+func resourceDatadogLogsPipelineDelete(d *schema.ResourceData, meta interface{}) error {
+	if err := meta.(*datadog.Client).DeleteLogsPipeline(d.Id()); err != nil {
+		if strings.Contains(err.Error(), "400 Bad Request") {
```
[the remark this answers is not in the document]
ajacquemot: what I meant is avoid code duplication:
```go
func isClientError(err error) bool {
  return strings.Contains(err.Error(), "400 Bad Request")
}
```

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,725 @@
+package datadog
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/zorkian/go-datadog-api"
+	"strings"
+)
+
+const (
```
ajacquemot: Can we group remappers and processors please ? It should ease readability

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,725 @@
...
+	tfMessageRemapperProcessor   = "message_remapper"
+	tfNestedPipelineProcessor    = "pipeline"
+	tfServiceRemapperProcessor   = "service_remapper"
+	tfStatusRemapperProcessor    = "status_remapper"
+	tfTraceIdRemapperProcessor   = "trace_id_remapper"
+	tfUrlParserProcessor         = "url_parser"
+	tfUserAgentParserProcessor   = "user_agent_parser"
+)
+
+var tfProcessorTypes = map[string]string{
```
ajacquemot: Same

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,725 @@
...
+	tfMessageRemapperProcessor:   datadog.MessageRemapperType,
+	tfNestedPipelineProcessor:    datadog.NestedPipelineType,
+	tfServiceRemapperProcessor:   datadog.ServiceRemapperType,
+	tfStatusRemapperProcessor:    datadog.StatusRemapperType,
+	tfTraceIdRemapperProcessor:   datadog.TraceIdRemapperType,
+	tfUrlParserProcessor:         datadog.UrlParserType,
+	tfUserAgentParserProcessor:   datadog.UserAgentParserType,
+}
+
+var ddProcessorTypes = map[string]string{
```
ajacquemot: Same

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,725 @@
...
+		Exists: resourceDatadogLogsPipelineExists,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: getPipelineSchema(false),
+	}
+}
+
+func resourceDatadogLogsPipelineCreate(d *schema.ResourceData, meta interface{}) error {
+	ddPipeline, err := buildDatadogPipeline(d)
```
ajacquemot: build looks/sounds much nice 👍

datadog/resource_datadog_logs_pipeline.go
ajacquemot: Ok for this method, what about https://github.com/terraform-providers/terraform-provider-datadog/pull/312/files#diff-8d432e003fcfd562bd1bd06805314206R637 for example ?

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+
+```hcl
+# Create a new Datadog logs pipeline
+resource "datadog_logs_pipeline" "sample_pipeline" {
+    name = "sample pipeline"
+    is_enabled = true
+    filter {
+        query = "source:foo"
+    }
+    processor {
```
ajacquemot: Why can't we have a `processors` list containing all the different processors ? It's a bit redundant to type processor all the time, WDYT ?

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+        "${datadog_logs_pipeline.my.id}"
+    ]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Your pipeline name.
```
ajacquemot: Why don't we group required parameters and put all the optional ones at the end ?

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+* `is_enabled` - (Optional, default = false) Boolean value to enable your pipeline.
+* `filter` - (Required) Defines your pipeline filter. Only logs that match the filter criteria are processed by this pipeline.
+  * `query` - (Required)
+* `processor` - (Optional) processors or nested pipelines. See [below](logs_pipeline.html#Processors) for more detailed descriptions.
+
+### Processors
```
ajacquemot: Same remark, why don't we group required and optional parameters ?

website/docs/r/logs_pipeline.html.markdown
tt810: it's a [recommended](https://www.terraform.io/docs/configuration/attr-as-blocks.html#defining-a-fixed-object-collection-value) way of defining list of objects in terraform.

website/docs/r/logs_pipeline.html.markdown
tt810: I will do that.

datadog/resource_datadog_logs_pipeline.go
tt810: I was sorting them alphabetically. It should be easier for the people who are not familiar with logs backend related logic? In addition, they are all called `processors`.

datadog/resource_datadog_logs_pipeline.go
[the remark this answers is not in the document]
tt810: Well, I prefer the way it is right now for a few reasons:
1. with `isClientError` function, we don't actually save any lines (we actually add one more function)
the code that is using this function won't change:
```
if strings.Contains(err.Error(), "400 Bad Request") { ==> if isClientError(err) {
	return nil
}
```
if strings.Contains(err.Error(), "400 Bad Request") { ==> if isClientError(err) {
	return false, nil
}
```

```
2. there are 2 places to check the error type. They happen to be both checking `400 Bad Request`, if API part change the return code for the Delete/Get method, this function wont make sense anymore.

website/docs/r/logs_pipeline.html.markdown
tt810: I write this part with the same idea. In addition, `name` and `is_enabled` always comes the first (common to all processors as well), followed by the distinct attributes for different processors.

datadog/resource_datadog_logs_pipeline.go
tt810: I updated to `static`.

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,705 @@
...
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: getProcessorSchema(isNested),
+			},
+		},
+	}
+}
+
+func getProcessorSchema(isNested bool) map[string]*schema.Schema {
```
bkabrda: I hope I'm not missing something, but it feels like you could make this function much simpler if you had a map such as `{tfArithmeticProcessor: arithmeticProcessor, ...}` and then just loop over it and add it's elements to `processorsSchema`. Does that make sense?

datadog/resource_datadog_logs_pipelineorder.go
```diff
@@ -0,0 +1,78 @@
+package datadog
```
bkabrda: Let's also rename this file to be `pipeline_order` to match the suggested new name of the resource.

datadog/resource_datadog_logs_pipelineorder.go
```diff
@@ -0,0 +1,78 @@
...
+
+// The deletion of pipeline order is not supported from config API.
+// This function simply delete the pipeline order resource from terraform state.
+func resourceDatadogLogsPipelineOrderDelete(d *schema.ResourceData, meta interface{}) error {
+
+	return nil
+}
+
+func resourceDatadogLogsPipelineOrderExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	if pipelineList, err := meta.(*datadog.Client).GetLogsPipelineList(); err == nil && len(pipelineList.PipelineIds) > 0 {
```
bkabrda: Several questions:
* Under what circumstances would the API return an error here?
* Why doesn't the order exist if there are 0 pipelines in it?

IIUC the pipeline order is a "singleton" resource that is always there, so this should always return true.

website/datadog.erb
```diff
@@ -13,9 +13,6 @@
         <li<%= sidebar_current("docs-datadog-resource") %>>
           <a href="#">Resources</a>
           <ul class="nav nav-visible">
-            <li<%= sidebar_current("docs-datadog-resource-dashboard-list") %>>
```
bkabrda: Why this removal?

website/datadog.erb
```diff
@@ -34,6 +31,9 @@
             <li<%= sidebar_current("docs-datadog-resource-integration_pagerduty_service_object") %>>
               <a href="/docs/providers/datadog/r/integration_pagerduty_service_object.html">datadog_integration_pagerduty_service_object</a>
             </li>
+            <li<%= sidebar_current("docs-datadog-resource-logs-pipeline") %>>
```
bkabrda: The pipeline order should have it's own documentation page, as it's a separate resource. The documentation of the pipeline should reference it where necessary.

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+## Import
+
+For the pipelines that are already created (including the **read-only** ones), you can include them into terraform by `import` operation.
+At the moment of writing this documentation, terraform requires to explicitly create resources that match the existing pipelines to be able to
+import them. You will need to do ```terraform import <resource.name> <pipelineID>``` for each existing pipeline.
+
+## Important Notes
+
+Each `datadog_logs_pipeline` resource defines a complete pipeline. The order of pipelines are maintained in resource `datadog_logs_pipelineorder`.
+There should be just one `datadog_logs_pipelineorder` resource. When creating a new pipeline, you need to **explicitly** add this pipeline to
```
bkabrda: I'd rephrase this to "There must be at most one", because otherwise Terraform would demonstrate all kinds of weird behavior.

datadog/import_datadog_logs_pipeline_test.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+}	
+`
+
+const pipelineorderConfigForImport = `
+resource "datadog_logs_pipelineorder" "pipelines" {
+	name = "pipelines"
+	pipelines = [
+		"TOYNsNfjTD6zTXVg8_ej1g",
```
bkabrda: Could you fix the indentation here please?

datadog/resource_datadog_logs_pipeline.go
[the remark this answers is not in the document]
bkabrda: I have to admit I'm a little torn here. As this is only used in 2 places right now, I'd say current way is ok, but if we ever need to use this in more places, I'd say we should go for the solution recommended by @ajacquemot.

datadog/resource_datadog_logs_pipeline.go
bkabrda: Since this is a comment on outdated diff, I didn't manage to find the current place of this code. @tt810 did you incorporate the comment somehow?

website/docs/r/logs_pipeline.html.markdown
bkabrda: I agree with @tt810 here, but I honestly don't have a strong opinion.

datadog/resource_datadog_logs_pipeline.go
tt810: Yes, I did rename the fields. I will resolve this comment. @bkabrda

website/datadog.erb
tt810: probably a mistake, I will put it back

datadog/resource_datadog_logs_pipelineorder.go
tt810: > * Under what circumstances would the API return an error here?

API will not return any other errors than server error and network error. 

> * Why doesn't the order exist if there are 0 pipelines in it?

It’s a decision I made for returning `not exist` when list len is 0. I will remove this check as I think it further now, if we don’t delete, it makes sense to always exists.

> IIUC the pipeline order is a "singleton" resource that is always there, so this should always return true.

As discussed with @bkabrda I will always return `true, nil` for this function, as the server error and network error can be taking care of by other functions.

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+description: |-
+  Provides a Datadog logs pipeline resource. This can be used to create and manage logs pipelines.
+---
+
+# datadog_logs_pipeline
+
+Provides a Datadog [Logs Pipeline API](https://docs.datadoghq.com/api/?lang=python#logs-pipelines) resource. This can be used to create and manage Datadog logs pipelines.
+
+
+## Example Usage: Datadog logs pipeline
```
gaetan-deputier: Example Usage: Create a Datadog logs pipeline

website/docs/r/logs_pipeline.html.markdown
gaetan-deputier: I'm more inclined to have required first and optional following as when you create the resource it's more natural to have the required elements first otherwise you'd have to scan all the attributes to ensure you did not miss any. You correctly have a below section with an anchor link that should help readers.

gaetan-deputier: Cf dashboard for example: https://www.terraform.io/docs/providers/datadog/r/dashboard.html

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+    ]
+    pipelines = [
+        "${datadog_logs_pipeline.sample_pipeline.id}"
+    ]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
```
gaetan-deputier: I'd emphasise on the fact by pipelines and processors are disabled by default

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+  * `target` - (Required) Final `attribute` or `tag` name to remap the sources to.
+  * `target_type` - (Required) Defines if the target is a log `attribute` or `tag`.
+  * `preserve_source` - (Optional, default = false) Remove or preserve the remapped source element.
+  * `override_on_conflict` - (Optional, default = false) Override or not the target element if already set.
+* category_processor
+  * `name` - (Optional) Name of the processor
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `target` - (Required) Name of the target attribute which value is defined by the matching category.
+  * `category` - (Required) List of filters to match or not a log and their corresponding name to assign a custom value to the log.
+    * `name` - (Required)
```
gaetan-deputier: Missing doc on name/query

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+* user_agent_parser
+  * `name` - (Optional) Name of the processor
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `sources` - (Required) List of source attributes.
+  * `target` - (Required) Name of the parent attribute that contains all the extracted details from the sources.
+  * `is_encoded` - (Optional, default = false) If the source attribute is url encoded or not.
+
+## Attributes Reference
+
+* `name` - (Required) The name attribute in resource `datadog_logs_pipelineorder` needs to be unique. It's better to set to the same value as the resource `NAME`.
```
gaetan-deputier: The name attribute in THE resource

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+            sources = ["user", "agent"]
+            target = "http_agent"
+            is_encoded = false
+        }
+    }
+}
+```
+## Example Usage: Datadog logs pipeline order
+
+```hcl
```
gaetan-deputier: Because it's another resource type, should not it be in another documentation file?

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+  * `name` - (Optional) Name of the processor
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `sources` - (Required) List of source attributes.
+  * `target` - (Required) Name of the parent attribute that contains all the extracted details from the sources.
+  * `is_encoded` - (Optional, default = false) If the source attribute is url encoded or not.
+
+## Attributes Reference
+
+* `name` - (Required) The name attribute in resource `datadog_logs_pipelineorder` needs to be unique. It's better to set to the same value as the resource `NAME`. 
+There is no related field can be found in  [Logs Pipeline API](https://docs.datadoghq.com/api/?lang=python#get-pipeline-order).
```
gaetan-deputier: > No related field can be found. What do you mean here?

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
...
+  * `target` - (Required) Name of the parent attribute that contains all the extracted details from the `sources`.
+  * `normalize_ending_slashes` - (Optional, default = false) If needs to normalize the ending slashes or not.
+* user_agent_parser
+  * `name` - (Optional) Name of the processor
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `sources` - (Required) List of source attributes.
+  * `target` - (Required) Name of the parent attribute that contains all the extracted details from the sources.
+  * `is_encoded` - (Optional, default = false) If the source attribute is url encoded or not.
+
+## Attributes Reference
```
gaetan-deputier: I'd put this argument reference for pipelineorder in a separate doc just for that resource type

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,254 @@
+---
+layout: "datadog"
+page_title: "Datadog: datadog_logs_pipeline"
+sidebar_current: "docs-datadog-resource-logs-pipeline"
+description: |-
+  Provides a Datadog logs pipeline resource. This can be used to create and manage logs pipelines.
```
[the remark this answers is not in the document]
tt810: Why do we need to explicitly state `processors`? Shouldn't it be a property of pipeline?

website/docs/r/logs_pipeline.html.markdown
tt810: This `name` field is unique to terraform resource. There is no representation on config API side of this field. The reason for it is to allow terraform to take it as an id as there is no way to refer to the resource name in state.

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,687 @@
...
+}
+
+func resourceDatadogLogsPipelineCreate(d *schema.ResourceData, meta interface{}) error {
+	pipeline, err := buildDatadogLogsPipeline(d)
+	if err != nil {
+		return err
+	}
+	createdPipeline, err := meta.(*datadog.Client).CreateLogsPipeline(pipeline)
+	if err != nil {
+		return fmt.Errorf("failed to create logs pipeline using Datadog API: %s", err.Error())
```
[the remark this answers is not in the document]
tt810: I think that information will be given as e.g. the `resource.name` error.... 
This format is used by the other resources.

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,721 @@
...
+			Elem: &schema.Resource{
+				Schema: getProcessorSchema(isNested),
+			},
+		},
+	}
+}
+
+func getProcessorSchema(isNested bool) map[string]*schema.Schema {
+	var processorsSchema = make(map[string]*schema.Schema)
+	if !isNested {
```
ajacquemot: Not sure I get this right, here what we are doing is if it's *NOT* nested, we are adding a nested pipeline ?

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,721 @@
...
+		processorsSchema[tfNestedPipelineProcessor] = &schema.Schema{
+			Type:     schema.TypeList,
+			MaxItems: 1,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: getPipelineSchema(!isNested),
```
ajacquemot: would it make sense to add `false` ?

website/docs/r/logs_pipeline_order.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+The following arguments are supported:
+
+* `name` - (Required) The name attribute in the resource `datadog_logs_pipeline_order` needs to be unique. It's better to set to the same value as the resource `NAME`. 
+There is no related field can be found in  [Logs Pipeline API](https://docs.datadoghq.com/api/?lang=python#get-pipeline-order).
+* `pipelines` - (Required) The pipeline ids list. The order of pipeline ids in this attribute defines the overall pipelines order for Logs.
+
+## Attributes Reference
+
+* `pipelines` - The `pipelines` list contains the ids of resources [datadog_logs_pipeline](logs_pipeline.html#datadog_logs_pipeline) that are created and imported.
+Update the order of pipelines in this list reflects the application order of the pipelines. You cannot delete or create pipelines by deleting or adding ids to this list.
```
ruthnaebeck:
```suggestion
Updating the order of pipelines in this list reflects the application order of the pipelines. You cannot delete or create pipelines by deleting or adding IDs to this list.
```

datadog/resource_datadog_logs_pipeline.go
tt810: yes. If the processor is not a nested processor, we can add nestedPipeline. 
so the layer will be:
```
pipeline {  // isNested=false pass down isNested to its processors
   processors {  // isNested == false, pass down !isNested to the nested pipeline.
       pipeline { // isNested == true, pass down isNested to its processors
           processors {   // isNested == true, can't contain nestedPipeline anymore.
           }
       }
   }
}
  
```

datadog/resource_datadog_logs_pipeline.go
tt810: you mean `getPipelineSchema(false)`?  See the above comment. isNested can be either `true` or `false` when pass to getPipelineSchema.

website/docs/r/logs_pipeline.html.markdown
```diff
@@ -0,0 +1,234 @@
...
+* message_remapper
+  * `name` - (Optional) Name of the processor.
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `sources` - (Required) List of source attributes.
+* pipeline
+  * `name` - (Optional) Name of the nested pipeline.
+  * `is_enabled` - (Optional, default = false) If the processor is enabled or not.
+  * `filter` - (Required) Defines the nested pipeline filter. Only logs that match the filter criteria are processed by this pipeline.
+    * `query` - (Required)
+  * `processor` - (Optional) [Processors](logs_pipeline.html#Processors). It can not be pipelines anymore.
```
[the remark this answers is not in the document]
tt810: I meant `Nested pipeline can't take any other nested pipeline as its processor.`

datadog/resource_datadog_logs_pipeline.go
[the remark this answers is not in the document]
bkabrda: Yeah, terraform will tell the user which resource is the problematic one and then print the error returned from here. This is absolutely ok.

website/docs/r/logs_pipeline.html.markdown
ajacquemot: Could we group required parameters and follow you convention, e.g. small fields first so that it makes everyone happy please ?

datadog/resource_datadog_logs_pipeline.go
ajacquemot: I don’t think so, it just grouping by suffix basically.

tt810:
```
const (
	tfArithmeticProcessor        = "arithmetic_processor"
	tfAttributeRemapperProcessor = "attribute_remapper"
	tfCategoryProcessor          = "category_processor"
	tfDateRemapperProcessor      = "date_remapper"
	tfGrokParserProcessor        = "grok_parser"
	tfMessageRemapperProcessor   = "message_remapper"
	tfNestedPipelineProcessor    = "pipeline"
	tfServiceRemapperProcessor   = "service_remapper"
	tfStatusRemapperProcessor    = "status_remapper"
	tfTraceIdRemapperProcessor   = "trace_id_remapper"
	tfUrlParserProcessor         = "url_parser"
	tfUserAgentParserProcessor   = "user_agent_parser"
)
```
To me the order is alphabetic:
```
tfArithmeticProcessor        
tfAttributeRemapperProcessor 
tfCategoryProcessor          
tfDateRemapperProcessor      
tfGrokParserProcessor        
tfMessageRemapperProcessor   
tfNestedPipelineProcessor    
tfServiceRemapperProcessor   
tfStatusRemapperProcessor    
tfTraceIdRemapperProcessor   
tfUrlParserProcessor         
tfUserAgentParserProcessor
```

website/docs/r/logs_pipeline.html.markdown
tt810: I update the order to have required before optional.

tt810:
Hi @ruthnaebeck, I added a recommendation at the end of `[website/docs/r/logs_pipeline.html.markdown](https://github.com/terraform-providers/terraform-provider-datadog/pull/312/files#diff-fcd3c4c2e640326e1c2accb1fafa86aeR237-R238)` could you check again please?

datadog/resource_datadog_logs_pipeline.go
```diff
@@ -0,0 +1,721 @@
...
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/zorkian/go-datadog-api"
+	"strings"
+)
+
+const (
+	tfArithmeticProcessor        = "arithmetic_processor"
+	tfAttributeRemapperProcessor = "attribute_remapper"
```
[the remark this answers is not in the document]
tt810: My understanding is processor is a general name which describes the processor properties of pipeline. Either a remapper or something else, they are all processors. I want to make it clear that they are all processors in TF.

datadog/resource_datadog_logs_pipeline.go
tt810: same as above.

datadog/resource_datadog_logs_pipeline.go
tt810: same as above.

ruthnaebeck:
> Hi @ruthnaebeck, I added a recommendation at the end of [website/docs/r/logs_pipeline.html.markdown](https://github.com/terraform-providers/terraform-provider-datadog/pull/312/files#diff-fcd3c4c2e640326e1c2accb1fafa86aeR237-R238) could you check again please?

Looks good :-)

PR #324, opened 2019-09-25
Add missing 'exists' methods to PD integration and PD service objects

bkabrda:
Without this, people would experience TF error out if the integration was manually deactivated outside TF (the Read functions would fail with 404 and TF wouldn't recover from that).


bkabrda:

PR #325, opened 2019-09-25
Fix parsing group_by.sort.facet for apm and log queries in dashboards

bkabrda:


bkabrda:

PR #326, opened 2019-09-25
Support Logs Index API

tt810:
* Resource for index and tests
* Resource for index order and tests
* Documentation


website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+sidebar_current: "docs-datadog-resource-logs-index"
+description: |-
+  Provides a Datadog logs index resource. This can be used to create and manage logs indexes.
+---
+
+# datadog_logs_index
+
+Provides a Datadog [Logs Index API](https://docs.datadoghq.com/api/?lang=python#logs-indexes) resource. This can be used to create and manage Datadog logs indexes.
+
+## Example Usage:
```
cswatt:
```suggestion
## Example Usage
```

looks like there isn't typically a colon after the Example Usage title

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+Create a Datadog logs index resource.
```
cswatt: This line says the sample is to create a Datadog logs index resource, but the sample says "update"—is this intentional?

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the index.
+* `filter` - (Required) Logs filter.
+  * `query` - (Required) Logs filter criteria. Only logs matching this filter criteria are considered for this index.
+* `exclusion_filter` - (Required) List of exclusion filter.
```
cswatt:
```suggestion
* `exclusion_filter` - (Required) List of exclusion filters.
```

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+  * `filter` - (Required)
+      * `query` - (Required) Only logs matching the filter criteria and the query of the parent index will be considered for this exclusion filter.
+      * `sample_rate` - (Optional, default = 0.0) the fraction of logs excluded by the exclusion filter, when active.
```
cswatt:
```suggestion
      * `sample_rate` - (Optional, default = 0.0) The fraction of logs excluded by the exclusion filter, when active.
```

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+  * `name` - (Optional) The name of the exclusion filter.
+  * `is_enabled` - (Optional, default = false) A boolean stating if the exclusion is active or not.
+
+## Import
+
+The current datadog terraform provider version does not support the creation and deletion of index.
```
cswatt:
```suggestion
The current Datadog Terraform provider version does not support the creation and deletion of indexes. 
```

Also: add the number of the version

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+  * `filter` - (Required)
+      * `query` - (Required) Only logs matching the filter criteria and the query of the parent index will be considered for this exclusion filter.
+      * `sample_rate` - (Optional, default = 0.0) the fraction of logs excluded by the exclusion filter, when active.
+  * `name` - (Optional) The name of the exclusion filter.
+  * `is_enabled` - (Optional, default = false) A boolean stating if the exclusion is active or not.
+
+## Import
+
+The current datadog terraform provider version does not support the creation and deletion of index. 
+To manage the existing indexes, do `terraform import <datadog_logs_index.name> <indexName>` to import them to terraform.
```
cswatt:
```suggestion
To manage the existing indexes, do `terraform import <datadog_logs_index.name> <indexName>` to import them to Terraform.
```

website/docs/r/logs_index_order.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+  Provides a Datadog logs index order resource. This can be used to manage the order of logs indexes.
+---
+
+# datadog_logs_index_order
+
+Provides a Datadog [Logs Index API](https://docs.datadoghq.com/api/?lang=python#logs-indexes) resource. This can be used to manage the order of Datadog logs indexes.
+
+## Example Usage
+
+Create a Datadog logs pipeline order resource
```
cswatt: Can we put this as a comment in the sample code rather than up here?

website/docs/r/logs_index_order.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The unique name of the index order resource. 
+* `indexes` - (Required) The index resource list. Logs are tested against the query filter of each index one by one following the order of the list.
+
+## Import
+
+The current datadog terraform provider version does not support the creation and deletion of index order.
```
cswatt:
```suggestion
The current Datadog Terraform provider version does not support the creation and deletion of index orders. 
```

Also add the version number.

website/docs/r/logs_index_order.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The unique name of the index order resource. 
+* `indexes` - (Required) The index resource list. Logs are tested against the query filter of each index one by one following the order of the list.
+
+## Import
+
+The current datadog terraform provider version does not support the creation and deletion of index order. 
+Do `terraform import <datadog_logs_index_order.name> <name>` to import index order to terraform. There must be at
```
cswatt:
```suggestion
Do `terraform import <datadog_logs_index_order.name> <name>` to import index order to Terraform. There must be at
```

website/docs/r/logs_index.html.markdown
tt810: @cswatt Thanks for the review. I wanted to say that by creating this resource we create a state in terraform, but this creation is going to UPDATE the index from backend.

website/docs/r/logs_index_order.html.markdown
tt810: I was trying to make it look similar to `logs_pipeline`. Do you think I should update both of them (pipeline and index)?

datadog/resource_datadog_logs_index.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if err = d.Set("exclusion_filter", buildTerraformExclusionFilters(ddIndex.ExclusionFilters)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func resourceDatadogLogsIndexUpdate(d *schema.ResourceData, meta interface{}) error {
+	ddIndex, err := buildDatadogIndex(d)
+	if err != nil {
```
bkabrda: It would be great if this could explicitly warn the user that creating new indexes is not allowed when 404 is returned here.

tt810: will do.

website/docs/r/logs_index.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+  Provides a Datadog logs index resource. This can be used to create and manage logs indexes.
+---
+
+# datadog_logs_index
+
+Provides a Datadog [Logs Index API](https://docs.datadoghq.com/api/?lang=python#logs-indexes) resource. This can be used to create and manage Datadog logs indexes.
+
+## Example Usage
+
+Create a Datadog logs index resource:
```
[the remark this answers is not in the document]
tt810: will do.

datadog/resource_datadog_logs_index.go
```diff
@@ -0,0 +1,168 @@
...
+
+var exclusionFilterSchema = map[string]*schema.Schema{
+	"name":       {Type: schema.TypeString, Optional: true},
+	"is_enabled": {Type: schema.TypeBool, Optional: true},
+	"filter": {
+		Type:     schema.TypeList,
+		Required: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"query":       {Type: schema.TypeString, Optional: true},
```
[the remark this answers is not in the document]
tt810: actually, this query is optional according to the documentation.

website/docs/r/logs_index_order.html.markdown
cswatt: Hmm might be best to update both!

datadog/resource_datadog_logs_index.go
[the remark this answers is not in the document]
bkabrda: Sounds good 👍

website/docs/r/logs_index_order.html.markdown
bkabrda: Yup, let's either put it in as a comment in the sample code or just remove it altogether (I don't think it's adding too much value TBH).

tt810: Sounds good. I will remove this line for both index_order and pipeline_order.

datadog/resource_datadog_logs_index_order.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+	return nil
+}
+
+func resourceDatadogLogsIndexOrderDelete(d *schema.ResourceData, meta interface{}) error {
+	return nil
+}
+
+func resourceDatadogLogsIndexOrderExists(d *schema.ResourceData, meta interface{}) (bool, error) {
```
ajacquemot: Would have been nice to explain with a comment why we always return true here

datadog/resource_datadog_logs_index_order.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+	if err := d.Set("indexes", ddIndexList.IndexNames); err != nil {
+		return err
+	}
+	return nil
+}
+
+func resourceDatadogLogsIndexOrderDelete(d *schema.ResourceData, meta interface{}) error {
```
ajacquemot: Same, a comment on why we don't need the delete method would have been nice.

datadog/resource_datadog_logs_index.go
```diff
@@ -0,0 +1,171 @@
...
+		if strings.Contains(err.Error(), "404 Not Found") {
+			return fmt.Errorf("logs index creation is not allowed, index_name: %s", tfName)
+		}
+		return fmt.Errorf("error updating logs index: (%s)", err.Error())
+	}
+	d.SetId(tfName)
+	return resourceDatadogLogsIndexRead(d, meta)
+}
+
+func resourceDatadogLogsIndexDelete(d *schema.ResourceData, meta interface{}) error {
```
ajacquemot: Why can't we delete an index ?

datadog/resource_datadog_logs_index.go
```diff
@@ -0,0 +1,171 @@
...
+
+func buildDatadogExclusionFilters(tfEFilters []interface{}) []datadog.ExclusionFilter {
+	ddEFilters := make([]datadog.ExclusionFilter, len(tfEFilters))
+	for i, tfEFilter := range tfEFilters {
+		ddEFilters[i] = buildDatadogExclusionFilter(tfEFilter.(map[string]interface{}))
+	}
+	return ddEFilters
+}
+
+func buildDatadogExclusionFilter(tfEFilter map[string]interface{}) datadog.ExclusionFilter {
```
ajacquemot: nits: as a general remark, I think `tfEFilter` is not really explicit and I'd rather use full name, e.g. `tfExclusionFiler`. Same for `tfFs`.

PR #341, opened 2019-10-16
add example to the Downtime resource documentation

sullis:
Motiviation:
- the existing documentation provides only one example for the Downtime resource
- the existing documentation provides an example that mutes **all** monitors
- Muting all monitors is not necessarily the best thing to do for large enterprises
- this PR is an attempt to show developers how to configure a Downtime for a single monitor


sullis:
please review @bkabrda @dabcoder

PR #347, opened 2019-10-24
Add Azure integration

dabcoder:
Relies on https://github.com/zorkian/go-datadog-api/pull/275/.

### DIFF
--- a/datadog/resource_datadog_integration_azure.go
+++ b/datadog/resource_datadog_integration_azure.go
@@ -1,0 +1,90 @@
+package datadog
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/zorkian/go-datadog-api"
+)
+
+func resourceDatadogIntegrationAzure() *schema.Resource {
+	return &schema.Resource{
+		Create:            resourceDatadogIntegrationAzureCreate,
+		Read:              resourceDatadogIntegrationAzureRead,
+		Update:            resourceDatadogIntegrationAzureUpdate,
+		UpdateHostFilters: resourceDatadogIntegrationAzureUpdateHostFilters,
+		Delete:            resourceDatadogIntegrationAzureDelete,
+		Exists:            resourceDatadogIntegrationAzureExists,
+		Importer: &schema.ResourceImporter{
+			State: resourceDatadogIntegrationAzureImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"tenant_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"client_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"client_secret": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+			"host_filters": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceDatadogIntegrationAzureRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*datadog.Client)
+
+	tenantName := d.Id()
+
+	integrations, err := client.ListIntegrationAzure()
+	if err != nil {
+		return err
+	}
+	for _, integration := range integrations {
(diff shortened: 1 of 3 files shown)


datadog/resource_datadog_integration_azure.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, integration := range integrations {
+		if integration.GetTenantName() == tenantName {
+			d.Set("tenant_name", integration.GetTenantName())
+			d.Set("client_id", integration.GetClientID())
+			d.Set("host_filters", integration.GetHostFilters())
+			return nil
+		}
+	}
+	return fmt.Errorf("error getting an Azure integration: tenant_name=%s", tenantName)
+}
```
Ricky-Thomas: Am I understanding correctly that this functionality is taking a specific configured Azure integration and returning those based fields based on a provided field? It's very possible I'm misunderstanding whats happening here. But the expected use case for the Azure List endpoint is that it does not take an argument and returns all integrated accounts.

datadog/resource_datadog_integration_azure_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, r := range s.RootModule().Resources {
+		tenantName := r.Primary.ID
+		for _, integration := range integrations {
+			if integration.GetTenantName() == tenantName {
+				return fmt.Errorf("The Azure integration still exist: tenantName=%s", tenantName)
+			}
+		}
+	}
+	return nil
+}
```
Ricky-Thomas: So I'm not very familiar with this repo, and you'll need to have someone from intg-tools-libs be the final say on this. 

However, for these tests, what are they testing exactly? The functions being added in the first file are named like `resourceDatadogIntegrationAzureRead` however I don't see any of those functions being tested here. I believe we should be testing those functions here versus the client calls however I'm not sure how it's done elsewhere.

Happy to hop on a call and chat if you want sometime!

Ricky-Thomas:
Looks like there's some tests failing the CI there as well so that will need to be addressed.

datadog/resource_datadog_integration_azure.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/zorkian/go-datadog-api"
+)
+
+func resourceDatadogIntegrationAzure() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDatadogIntegrationAzureCreate,
+		Read:   resourceDatadogIntegrationAzureRead,
+		Delete: resourceDatadogIntegrationAzureDelete,
+		Importer: &schema.ResourceImporter{
```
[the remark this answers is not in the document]
dabcoder: Yes, the GCP resource supports it so I've added it, thanks.

datadog/resource_datadog_integration_azure.go
```diff
@@ -0,0 +1,102 @@
...
+			},
+			"client_secret": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+			"host_filters": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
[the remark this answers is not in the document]
dabcoder: Done too.

datadog/resource_datadog_integration_azure.go
dabcoder: It should loop through all Azure integrations already set up in a given Datadog account and return the one that matches the provided `tenantName`, in order to resync the local state with the remote state (according to the Terraform docs): https://www.terraform.io/docs/plugins/provider.html#read. To be fair, I wrote something that's similar to the `resourceDatadogIntegrationGcpRead()` function, so I am also unsure if this is the right way to go here.

dabcoder:
> Looks like there's some tests failing the CI there as well so that will need to be addressed.

Indeed, that's because the `go-datadog-api` version needs to be bumped so that the provider can use the Client's Azure related functions.

nmuesch:
Discussed offline, going to close this and continue the work in https://github.com/terraform-providers/terraform-provider-datadog/pull/556

PR #352, opened 2019-10-28
Fix slo threshold warning value modified when storing the state

mzneos:
The value of the warning field in the threshold object was set to `0` instead of being omitted when the pointer value was nil.

This raised an error when updating a service level objective resource, as the value of `warning` must always be greater than the `slo`.
This error is not raised on creation of the resource because the warning field is only modified in the `resourceDatadogServiceLevelObjectiveRead` function, modifying the state after the creation of the resource.


tustvold:
Any chance of getting a patch release for this, it makes the optional field not so optional as it breaks updating the resource...

PR #358, opened 2019-11-08
monitor_search is not yet supported

unclebconnor:
removing `monitor_search` flag while unsupported

### DIFF
--- a/website/docs/r/service_level_objective.html.markdown
+++ b/website/docs/r/service_level_objective.html.markdown
@@ -105,7 +105,7 @@ query {
 ```
 
 * `monitor` type SLOs:
-    * `monitor_ids` - (Optional) A list of numeric monitor IDs for which to use as SLIs. Their tags will be auto-imported into `monitor_tags` field in the API resource.
+    * `monitor_ids` - (Required) A list of numeric monitor IDs for which to use as SLIs. Their tags will be auto-imported into `monitor_tags` field in the API resource.
     * `groups` - (Optional) A custom set of groups from the monitor(s) for which to use as the SLI instead of all the groups.
 
 
--- a/website/docs/r/service_level_objective.html.markdown
+++ b/website/docs/r/service_level_objective.html.markdown
@@ -29,7 +29,7 @@ resource "datadog_service_level_objective" "foo" {
     target = 99.9
     warning = 99.99
     target_display = "99.900"
-    warning_display = "99.990"      
+    warning_display = "99.990"
   }
 
   thresholds {
@@ -37,7 +37,7 @@ resource "datadog_service_level_objective" "foo" {
     target = 99.9
     warning = 99.99
     target_display = "99.900"
-    warning_display = "99.990"      
+    warning_display = "99.990"
   }
 
   tags = ["foo:bar", "baz"]
@@ -56,13 +56,13 @@ resource "datadog_service_level_objective" "bar" {
   thresholds {
     timeframe = "7d"
     target = 99.9
-    warning = 99.99      
+    warning = 99.99
   }
 
   thresholds {
     timeframe = "30d"
     target = 99.9
-    warning = 99.99      
+    warning = 99.99
   }
 
   tags = ["foo:bar", "baz"]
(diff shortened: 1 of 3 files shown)


website/docs/r/service_level_objective.html.markdown
```diff
@@ -96,17 +96,16 @@ The following options are specific to the `type` of service level objective:
...
     numerator   = "sum:my.custom.count.metric{type:good}.as_count()"
-    denominator = "sum:my.custom.count.metric{*}.as_count()" 
+    denominator = "sum:my.custom.count.metric{*}.as_count()"
 }
 ```
 
 * `monitor` type SLOs:
-    * `monitor_ids` - (Optional) A list of numeric monitor IDs for which to use as SLIs. Their tags will be auto-imported into `monitor_tags` field in the API resource. At least 1 of `monitor_ids` or `monitor_search` must be provided.
-    * `monitor_search` - (Optional) The monitor query search used on the monitor search API to add monitor_ids by searching. Their tags will be auto-imported into `monitor_tags` field in the API resource. At least 1 of `monitor_ids` or `monitor_search` must be provided.
+    * `monitor_ids` - (Optional) A list of numeric monitor IDs for which to use as SLIs. Their tags will be auto-imported into `monitor_tags` field in the API resource.
```
bkabrda: If one of `monitor_ids` and `monitor_search` must be provided and we are hiding `monitor_search`, shouldn't we then also make `monitor_ids` required for now?

unclebconnor: @bkabrda that makes sense to me.  I haven't actually updated any code here...should we just document that `monitor_ids` is required for now?

unclebconnor: Adjusted so `monitor_ids` is required

platinummonkey:
can you remove this from the schema as well?

website/docs/r/service_level_objective.html.markdown
unclebconnor: @bkabrda this should be ready for another review when you can

bkabrda:
Hmm, I'm honestly unsure about this. The Terraform deprecation policy [1] prevents us from just removing an attribute like this. OTOH this attribute never worked anyway, so I guess this should be fine...

I think after all I'm 👍 for doing this change, but would like to hear more opinions. @nmuesch thoughts?

[1] https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal

platinummonkey:
since it's not in use, I think it's safe to just drop.

unclebconnor:
@bkabrda @nmuesch any additional thoughts here?  ok to merge/drop the setting since it's not in use?  Other changes?

bkabrda:
One last question that I have here is if there's a specific timeframe on where `monitor_search` will be actually supported. IOW, is it worth removing it?

unclebconnor:
@bkabrda that option is currently not production ready and there is a good bit of design and dev work that needs to be done and hasn't been scheduled.  We've already had a support case where someone was trying to use it and getting errors so we'd prefer to remove it.

bkabrda:
@unclebconnor ok, that's fair, let's remove the field for now.

I was going through the terraform docs for removal of attribute [1]. The "right" way to do this seems to be to leave the field there, but add a `Removed` attribute to it, e.g. `Removed: "We identified problems with usage of this field and thus we're marking it as 'removed' until these are fixed"` or something similar. This will make sure that our users who tried to define this field will get a proper error message when running terraform and will know what's going on.

[1] https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal

unclebconnor:
@bkabrda Ok I've made that update.  I'm pretty sure this is the format that is described in the docs but let me know if you have other feedback

unclebconnor:
@bkabrda not sure of process for this repo but I'm assuming you want to do the final merge.  Let me know if there's anything else you need from me here.

bkabrda:
LGTM now, merging. 👍 thanks for your patience :)

Issue #292, opened 2019-08-20
Feature Request: query_table support on dashboard


endzyme:
Currently it doesn't appear I can implement the following dashboard json in the terraform provider. Can someone point me in the right direction if I'm using the `datadog_dashboard` resource incorrectly or if I should be seeking help elsewhere. 

Thanks in advance!

Example Dashboard Config:
```json
{
    "title": "Some Timeboard",
    "description": "",
    "widgets": [
        {
            "id": 1,
            "definition": {
                "type": "query_table",
                "requests": [
                    {
                        "q": "max:kubernetes.memory.usage{$cluster,$daemonset,$deployment} by {kube_container_name,some_cluster}",
                        "aggregator": "max",
                        "limit": 10,
                        "order": "desc",
                        "alias": "Max Mem Usage"
                    },
                    {
                        "q": "max:kubernetes.cpu.usage.total{$cluster,$daemonset,$deployment} by {kube_container_name,some_cluster}",
                        "aggregator": "max",
                        "alias": "Max CPU Usage"
                    }
                ],
                "title": "Max Resources"
            }
        },
        {
            "id": 2,
            "definition": {
                "type": "query_table",
                "requests": [
                    {
                        "q": "avg:kubernetes.memory.usage{$cluster,$daemonset,$deployment} by {kube_container_name,some_cluster}",
                        "aggregator": "avg",
                        "limit": 10,
                        "order": "desc",
                        "alias": "Average Mem Usage"
                    },
                    {
                        "q": "avg:kubernetes.cpu.usage.total{$cluster,$daemonset,$deployment} by {kube_container_name,some_cluster}",
                        "aggregator": "avg",
                        "alias": "Average CPU Usage"
                    }
                ],
                "title": "Average Resources"
            }
        }
    ],
    "template_variables": [
        {
            "name": "cluster",
            "default": "*",
            "prefix": "cluster_name"
        },
        {
            "name": "daemonset",
            "default": "*",
            "prefix": "kube_daemon_set"
        },
        {
            "name": "deployment",
            "default": "*",
            "prefix": "kube_deployment"
        }
    ],
    "layout_type": "ordered",
    "is_read_only": false,
    "notify_list": [],
    "dashboard_id": "blorp"
}
```

### Terraform Version: v0.11.14
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

### Affected Resource(s)
[...]

PR #373, opened 2019-12-20
Add Datadog dashboard query table

nuclearnic:
This is the one outstanding feature for our team to fully implement dashboards. I've looked at a [similar pull request here](https://github.com/terraform-providers/terraform-provider-datadog/pull/355/files) to try and follow standards. I've used the resulting binary already to successfully create query tables for our org.

Also, I see there is an open issue for this feature here: [#292](https://github.com/terraform-providers/terraform-provider-datadog/issues/292)

Comments and feedback welcome


website/docs/r/dashboard.html.markdown
```diff
@@ -715,6 +739,21 @@ Nested `widget` blocks have the following structure:
         - `title_size`: (Optional) The size of the widget's title. Default is 16.
         - `title_align`: (Optional) The alignment of the widget's title. One of "left", "center", or "right".
         - `time`: (Optional) Nested block describing the timeframe to use when displaying the widget. The structure of this block is described [below](dashboard.html#nested-widget-time-blocks).
+  - `query_table_definition`: The definition for a Query Value widget. Exactly one nested block is allowed with the following structure:
```
jbenais: I think it should be `The definition for a Query Table widget` instead

website/docs/r/dashboard.html.markdown
```diff
@@ -184,6 +184,30 @@ resource "datadog_dashboard" "ordered_dashboard" {
...
+				  palette = "white_on_red"
+			  }
+      }
+			title = "Widget Title"
+			time = {
+			  live_span = "1h"
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
nuclearnic: It was indeed! Fixed

nuclearnic:
> The changes look good, but the tests don't pass. The issue seems to be that the added `query_table_definition` is now added at index `9`, but you're testing `widget.8`. I suggest adding the widget at the end of the widgets list, because otherwise you'd have to shift all the `widget.9` and following indexes by one.

Hi @bkabrda  thanks for the response. I've made the change as requested. Definitely an oversight on my part there.

On a sidenote, running `make test` (as stated in the README) doesn't actually test changes when I run it locally here so I was relying on the checks. I might be missing something regarding running the tests.

datadog/resource_datadog_dashboard_test.go
```diff
@@ -721,6 +744,21 @@ func TestAccDatadogDashboard_update(t *testing.T) {
...
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.0.comparator", "<"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.0.value", "2"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.0.palette", "white_on_green"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.1.comparator", ">"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.1.value", "2.2"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.conditional_formats.1.palette", "white_on_red"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.alias", "Lookups"),
```
bkabrda: The check for `Lookups` fails, as no `alias` is set on the widget. Let's just remove this.

datadog/resource_datadog_dashboard_test.go
```diff
@@ -721,6 +744,21 @@ func TestAccDatadogDashboard_update(t *testing.T) {
...
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.aggregator", "sum"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.limit", "10"),
+					resource.TestCheckResourceAttr("datadog_dashboard.ordered_dashboard", "widget.14.query_table_definition.0.request.0.order", "desc"),
```
bkabrda: There's no `order` defined on the widget, so let's remove this line.

nuclearnic:
> Thanks for the fix. There are still two lines to remove in the test file to get the tests to pass - I left comments inline.
> 
> As for the tests, the README also explains that you can run acceptance tests, which use real API (this is one of the acceptance tests). You can use `make testacc` to run all acceptance tests or, to run this specific test that you're working on, `TF_ACC=1 go test ./datadog -v -run TestAccDatadogDashboard_update`. Do note that you have to have your DD API and APP keys exported in your env (`DATADOG_API_KEY` and `DATADOG_APP_KEY`). Also note that some of the tests do destructive operations via the DD API, so I would only ever recommend running these on an org where you don't have any useful data.

@bkabrda thanks for the speedy response!

I've removed those two lines. We've recently moved over to DD so I will have to enquire about running these kinds of tests there before going ahead. Not ideal, but I will do so - especially if there are more contributions to make.

This could've gone a little smoother so thanks for the patience and being so helpful 👍

bkabrda:
@nuclearnic thanks for prompt update. The code LGTM now, so I'll merge this PR.

FTR, we're also working towards having a public CI that would actually run acceptance tests, but it will probably take us some time to get there. In the meantime, for some of the tests it's also possible to just create a free account that you can use to do tests (but obviously this wouldn't work for features that need to be paid for to get the API enabled, such as logs pipelines).

PR #396, opened 2020-01-21
Add new monitor summary widget parameters

chrismdd:
This PR adds two new, optional parameters for the monitor summary (manage status) widget: `summary_type`, and `show_last_triggered`: https://docs.datadoghq.com/graphing/widgets/monitor_summary/#api.

go-datadog-api was updated here: https://github.com/zorkian/go-datadog-api/pull/290

It also updates the acceptance tests for screenboards and dashboards, both of which pass with these changes.


datadog/resource_datadog_dashboard.go
```diff
@@ -2335,6 +2335,10 @@ func getManageStatusDefinitionSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Required: true,
 		},
+		"summary_type": {
+			Type:     schema.TypeString,
+			Optional: true,
```
nmuesch: Can we add description in this schema as well https://github.com/terraform-providers/terraform-provider-datadog/pull/396/files#diff-22ba5b0fbe5d1b3cdfe783aeb61bf51aR561

nmuesch: Should we also set a validation here to ensure its one of the three types - https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#validatefunc or will these change?

chrismdd: Thanks for the review. I added a description and validation. Please let me know if I should move the validation function somewhere or otherwise alter it (I don't use Go much at the moment).

chrismdd:
@jirikuncar @nmuesch thanks for reviewing. I don't have write access to this repo, so could you please merge this PR for me?

PR #403, opened 2020-02-04
Hide deprecated monitor summary widget params

chrismdd:
Remove the count and start monitor summary widget params from the docs and deprecate them.

While these params are still supported (and will continue to be), they don't appear to be used by anyone and we don't know of any use cases for them, so we should remove them from the docs to avoid confusion.

Related: https://github.com/terraform-providers/terraform-provider-datadog/pull/300 (the `sort` param does have use cases so we're adding it to the docs here: https://github.com/DataDog/documentation/pull/6568).

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -2357,13 +2357,15 @@ func getManageStatusDefinitionSchema() map[string]*schema.Schema {
 		},
 		// The count param is deprecated
 		"count": {
-			Type:     schema.TypeInt,
-			Optional: true,
+			Type:       schema.TypeInt,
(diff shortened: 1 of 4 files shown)


datadog/resource_datadog_dashboard.go
```diff
@@ -2355,10 +2355,12 @@ func getManageStatusDefinitionSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Optional: true,
 		},
+		// The count param is deprecated
```
bkabrda: Could you please add the `Deprecated` attribute to these two? https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema

bkabrda:
LGTM now, merging. Thanks for the PR!

PR #404, opened 2020-02-04
Properly set screenboard attributes from client response to not produce non-empty plans

bkabrda:
This fixes https://github.com/terraform-providers/terraform-provider-datadog/issues/105#issuecomment-581938484


bkabrda:
Thanks for the review. Merging.

PR #405, opened 2020-02-04
Gracefully handle downtimes that were canceled manually

### LINKED ISSUE #315: Support cancelled downtime gracefully
When a downtime managed by terraform is cancelled via the web UI, we get the following: `API error 400 Bad Request: {"errors":["cannot update a canceled downtime"]}`

@bkabrda suggests that we switch to the following behavior: when the downtime is canceled, we should consider it deleted in our provider, which would automatically make TF suggest recreating it. so when reading it, we have to look at “canceled” and if that’s true, we have to pretend like it doesn’t exist - that should make TF behave correctly and suggest recreating it. 

This sounds good from our side

bkabrda:
Fixes #315 - if the downtime gets deleted (==canceled) manually in UI, we have to consider it canceled and recreate it.

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -297,11 +297,18 @@ func resourceDatadogDowntimeExists(d *schema.ResourceData, meta interface{}) (b 
 		return false, err
 	}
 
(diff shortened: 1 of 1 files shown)


bkabrda:

PR #406, opened 2020-02-06
Improve monitor example with ignoring changes on silenced

### LINKED ISSUE #389: Clarify deprecation of `silenced` on datadog_monitor
Hi there,

I'm looking for some clarification of the 'silencing by hand' section of the [datadog_monitor](https://www.terraform.io/docs/providers/datadog/r/monitor.html#silencing-by-hand-and-by-downtimes) documentation. It seems that setting `silenced` is [deprecated](https://github.com/terraform-providers/terraform-provider-datadog/pull/221/files), but it may still be populated and merged with current downtimes entered into the DD UI.

Is there a best practice/recommended way to say "don't remove existing downtimes" if they show up in the silenced map, as returned by the DD API? In other words, is there a way to do this, without referring to the deprecated `silenced` attribute:

```
resource "datadog_monitor" "foo" {
  lifecycle {
    ignore_changes = [silenced]
  }
}
```

Thanks in advance.

bkabrda:
Fixes #389

### DIFF
--- a/website/docs/r/monitor.html.markdown
+++ b/website/docs/r/monitor.html.markdown
@@ -37,8 +37,9 @@ resource "datadog_monitor" "foo" {
   timeout_h    = 60
   include_tags = true
 
-  silenced = {
-    "*" = 0
+  # ignore any changes in silenced value; using silenced is deprecated in favor of downtimes
+  lifecycle {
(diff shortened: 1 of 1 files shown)


bkabrda:

Issue #67, opened 2018-05-02
Add support for Slack integration


snemetz:
Working on fully automating our monitoring setup.
It would be very helpful if Terraform could configure Datadog's Slack integration.

My plan is for monitoring to be setup with each service. That would include a Slack channel and that channel added to Datadog's Slack integration.

I'd expect 2 resources to be needed

1.  Datadog Slack integration base configuration
2. Datadog Slack channel

chenrui333:
👍

h3x89:
👍

matt-canty-dragon:
Note: I reviewed go-datadog and it does not appear to support the Monitor Alert Message Options

<img width="346" alt="image" src="https://user-images.githubusercontent.com/56073682/73348784-eb0e0380-4281-11ea-98d9-bc5a8b286d1c.png">

I have found that I prefer them all to be off which is likely to be not default.

chrisjohnson:
This is currently a problem for us, we are hoping to grow adoption and making it easy to full configure datadog with a terraform module tailored to our setup will go a long way toward getting the tooling in front of people so they can really evaluate it.

terowz:
same as others mentioned here. Currently need to use a webhook so there is no way to fully automate this and just use `{{kube_deployment.name}}`, as the webhook needs to be set up in each channel.

skarimo:
Hi all, wanted to follow up here with an update. We have added a slack channels resource: https://github.com/DataDog/terraform-provider-datadog/pull/932

coopergillan:
@snemetz - based on @skarimo's comment and #932, seems like this issue could be considered closed, right?

snemetz:
Probably. I've change companies a couple times since this was opened and currently don't use DataDog

coopergillan:
@skarimo - would you consider this closed?

chrisjohnson:
Based on the original comment, a resource for both the slack channels as well as the base slack integration would be desirable. I'm not actively working in this space but my company does use datadog (managed via terraform) and that still sounds like the desired state. Does the linked PR provide both pieces or only just the configuration of the slack channels themselves?

skarimo:
Hi all, we will go ahead and keep this issue open as the original request was looking for the ability to configure both the base slack integration as well as slack channels. The Datadog terraform provider currently supports configuring slacks channels only.

Thanks!

ohookins:
Seems like the resource misses the "muting" property of the channel configuration.

[...]

PR #411, opened 2020-02-11
Add slack integration #67

bewiwi:
Add slack integration #67 

Allow the possibility to manage slack integration.

Example:

```hcl
# Create a new Datadog - Slack integration
resource "datadog_integration_slack" "foo1" {
	account = "slack_pro_account"
	url = "http:///myslack/url"
   
   channels {
	   channel_name = "#prod-outage"
	   transfer_all_user_comments = false       
   }

   channels {
	   channel_name = "#staging-outage"
	   transfer_all_user_comments = false       
   }
}
```


datadog/resource_datadog_integration_slack.go
```diff
@@ -0,0 +1,220 @@
...
+	return nil
+}
+
+func resourceDatadogIntegrationSlackRead(d *schema.ResourceData, meta interface{}) error {
+	integrationSlackMutex.Lock()
+	defer integrationSlackMutex.Unlock()
+
+	client := meta.(*datadog.Client)
+	account := d.Id()
+	d.SetId("")
```
jirikuncar: why do you set it to `""` if [later](https://github.com/terraform-providers/terraform-provider-datadog/pull/411/files#diff-e35c3d1575be0cb0b644b1b7caaac061R146) you check for its value?

bewiwi: I set it to blank in a first place and re-set it if hook was found. I updated code to be more readable using a boolean var instead.

msuterski:
Thanks for adding this! We're about to move some of the manual maintenance to TF and this will help a lot.

What's the current status of this PR? Are there any plans to merge and release it any time soon?

bkabrda:
Hey 👋 thanks a lot for sending the PR! What I'm concerned about here is a concern mentioned in [1] that we eventually had to fix for Pagerduty integration. In short, for Pagerduty integration there was no way to configure individual service objects (analogous to `channels` in the Slack integration) independently (from multiple TF configs spread across several repos). The path forward implemented in [2] isn't completely straightforward and I'd like to avoid it.

That said, the DD team responsible for this API currently has no plans to implement manipulating individual channels, so I guess we can go forward with the approach in this PR, while understanding the caveat mentioned above and bearing in mind that in future we might need to do the same thing we did for the Pagerduty integration. I'll do an in-depth review and hopefully we would get this merged before the next release.  

[1] https://github.com/terraform-providers/terraform-provider-datadog/issues/75
[2] https://github.com/terraform-providers/terraform-provider-datadog/pull/237

tonglil:
Any update on maybe giving this a review? :)

therve:
The integration team is pushing a proper slack channels management API. This should make the slack resource more manageable through terraform, so that we don't reproduce the issue we had with pagerduty. We'll keep you posted.

ken5scal:
Can't wait for the functionality! Any update since the last time?

JordanP:
Small "up" here. Slack is soooo widely used and sooo needed :) Thanks !

andymac4182:
@therve Thanks for that update. It looks like the new APIs are available. Any chance of this resource coming soon?

tdewitt:
Adding another "status update" comment.  This would be a pretty great addition to our pipelines.  Right now this is one of the manual prerequisites for new services and that means it's easy to miss.

PR #415, opened 2020-02-19
LI-104: Add support for lookup processor

blemale:
Support:
* Lookup processor

### DIFF
--- a/datadog/resource_datadog_logs_custom_pipeline.go
+++ b/datadog/resource_datadog_logs_custom_pipeline.go
@@ -15,6 +15,7 @@ const (
 	tfDateRemapperProcessor      = "date_remapper"
 	tfGeoIPParserProcessor       = "geo_ip_parser"
 	tfGrokParserProcessor        = "grok_parser"
+	tfLookupProcessor            = "lookup_processor"
 	tfMessageRemapperProcessor   = "message_remapper"
 	tfNestedPipelineProcessor    = "pipeline"
 	tfServiceRemapperProcessor   = "service_remapper"
@@ -32,6 +33,7 @@ var tfProcessorTypes = map[string]string{
 	tfDateRemapperProcessor:      datadog.DateRemapperType,
 	tfGeoIPParserProcessor:       datadog.GeoIPParserType,
 	tfGrokParserProcessor:        datadog.GrokParserType,
+	tfLookupProcessor:            datadog.LookupProcessorType,
 	tfMessageRemapperProcessor:   datadog.MessageRemapperType,
 	tfNestedPipelineProcessor:    datadog.NestedPipelineType,
 	tfServiceRemapperProcessor:   datadog.ServiceRemapperType,
@@ -49,6 +51,7 @@ var tfProcessors = map[string]*schema.Schema{
 	tfDateRemapperProcessor:      dateRemapper,
 	tfGeoIPParserProcessor:       geoIPParser,
 	tfGrokParserProcessor:        grokParser,
+	tfLookupProcessor:            lookupProcessor,
 	tfMessageRemapperProcessor:   messageRemapper,
 	tfServiceRemapperProcessor:   serviceRemapper,
 	tfStatusRemapperProcessor:    statusRemmaper,
@@ -65,6 +68,7 @@ var ddProcessorTypes = map[string]string{
 	datadog.DateRemapperType:           tfDateRemapperProcessor,
 	datadog.GeoIPParserType:            tfGeoIPParserProcessor,
 	datadog.GrokParserType:             tfGrokParserProcessor,
+	datadog.LookupProcessorType:        tfLookupProcessor,
 	datadog.MessageRemapperType:        tfMessageRemapperProcessor,
 	datadog.NestedPipelineType:         tfNestedPipelineProcessor,
 	datadog.ServiceRemapperType:        tfServiceRemapperProcessor,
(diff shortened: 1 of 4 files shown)


website/docs/r/logs_custom_pipeline.html.markdown
```diff
@@ -89,6 +89,16 @@ resource "datadog_logs_custom_pipeline" "sample_pipeline" {
             is_enabled = true
         }
     }
+    processor {
+        lookup_processor {
+            source = "service_id"
+            target = "service_name"
+            lookup_table = ["1", "my service"]
```
jirikuncar: Did you want to map `1` to `my service`?
```suggestion
            lookup_table = ["1,my service"]
```

blemale: Indeed, nice catch! It is fixed now.

datadog/resource_datadog_logs_custom_pipeline.go
```diff
@@ -623,6 +661,27 @@ func buildDatadogUserAgentParser(tfProcessor map[string]interface{}) datadog.Use
 	return ddUserAgentParser
 }
 
+func buildDatadogLookupProcessor(tfProcessor map[string]interface{}) datadog.LookupProcessor {
+	ddLookupProcessor := datadog.LookupProcessor{}
```
jirikuncar: are `name` and `is_enabled` missing here?

datadog/resource_datadog_logs_custom_pipeline_test.go
```diff
@@ -97,6 +98,16 @@ resource "datadog_logs_custom_pipeline" "my_pipeline_test" {
 			sources = ["ip1"]
 		}
 	}
+	processor {
+		lookup_processor {
+			name = "lookup processor"
```
jirikuncar: we should start only with required fields

datadog/resource_datadog_logs_custom_pipeline.go
blemale: `name` & `is_enabled` are handled in a generic way here https://github.com/terraform-providers/terraform-provider-datadog/blob/e0e835892b09901ae260e26bfc9d2feccd88c07c/datadog/resource_datadog_logs_custom_pipeline.go#L626-L631

datadog/resource_datadog_logs_custom_pipeline_test.go
blemale: What's the reasoning? Because all the processors around in the test are defined that way.

datadog/resource_datadog_logs_custom_pipeline.go
jirikuncar: Thanks!

datadog/resource_datadog_logs_custom_pipeline_test.go
jirikuncar: we should make sure that optional fields are really optional. you are right that there is no example for this case.

gzussa: Still this is a good point, we should create such test. This makes us realize that the same kind of issue may be elsewhere.

datadog/resource_datadog_logs_custom_pipeline.go
```diff
@@ -623,6 +661,27 @@ func buildDatadogUserAgentParser(tfProcessor map[string]interface{}) datadog.Use
...
+		ddLookupProcessor.SetTarget(tfTarget)
+	}
+	if tfLookupTable, exists := tfProcessor["lookup_table"].([]interface{}); exists && len(tfLookupTable) > 0 {
+		ddLookupTable := make([]string, len(tfLookupTable))
+		for i, tfLookupLine := range tfLookupTable {
+			ddLookupTable[i] = tfLookupLine.(string)
+		}
+		ddLookupProcessor.LookupTable = ddLookupTable
+	}
+	if tfDefaultLookup, exists := tfProcessor["default_lookup"].(string); exists {
```
gzussa: Based on the schema definition,  `default_lookup`, `is_enabled` and the `name` are optional, shouldn't we check they exist before to cast them?
Also, do we need the exists check on required fields?

gzussa: Apparently we are good here @bkabrda will add details for documentation purpose.

bkabrda: I was almost certain that this would fail if the typecasting was there and the value was `nil`, but it doesn't. `tfProcessor["default_lookup"].(string)` correctly works in all possible scenarios (the `default_lookup` key is not present in the map, it's present but nil, it's present and non-nil but of a different type than string - in all these cases, `exists` will be `false`). So everything ok here.

datadog/resource_datadog_logs_custom_pipeline.go
```diff
@@ -186,6 +191,26 @@ var grokParser = &schema.Schema{
...
+			"name":       {Type: schema.TypeString, Optional: true},
+			"is_enabled": {Type: schema.TypeBool, Optional: true},
+			"source":     {Type: schema.TypeString, Required: true},
+			"target":     {Type: schema.TypeString, Required: true},
+			"lookup_table": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"default_lookup": {Type: schema.TypeString, Optional: true},
```
nmuesch: Should this be marked as required? It looks like trying to create a lookup processor without this value doesn't work.

datadog/resource_datadog_logs_custom_pipeline_test.go
blemale: I have added a test with only the mandatory fields.

datadog/resource_datadog_logs_custom_pipeline.go
blemale: It is optional on the API side, what is the error that you get when you try to create a lookup processor without default lookup?

bkabrda: I'm getting this:

```
Error: error updating logs pipeline: (API error 400 Bad Request: {"error":{"code":"InvalidArgument","message":"Invalid Pipeline","details":[{"code":"InvalidArgument","message":"Invalid default lookup: must not be blank"}]}})
```

blemale:
👋 Hi @gzussa! I have finally found some time to cross the finish line for this PR.
So in  78657a3 I have added a test with only the mandatory fields and I have made `default_lookup` effectively optional in terraform. Tell me if it is ok for you?

PR #418, opened 2020-02-20
[MA-902] no data timeframe doc clarification

unclebconnor:
Update some language based on recent discovery/research


website/docs/r/monitor.html.markdown
```diff
@@ -99,9 +99,9 @@ The following arguments are supported:
     For example, if the value is set to 300 (5min), the timeframe is set to last_5m and the time is 7:00,
     the monitor will evaluate data from 6:50 to 6:55. This is useful for AWS CloudWatch and other backfilled
     metrics to ensure the monitor will always have data during evaluation.
-* `no_data_timeframe` (Optional) The number of minutes before a monitor will notify when data stops reporting. Must be at
-    least 2x the monitor timeframe for metric alerts or 2 minutes for service checks. Default: 2x timeframe for
-    metric alerts, 2 minutes for service checks. Defaults to 10 minutes.
+* `no_data_timeframe` (Optional) The number of minutes before a monitor will notify when data stops reporting. Provider defaults to 10 minutes.
+    We recommend at least 2x the monitor timeframe for metric alerts or 2 minutes for service checks.  If omitted, we fall back to
```
nmuesch: Should we remove the `if omitted` sentence?  Since the provider sets a default, I don't think you can get the api defaults values. 

The 10 minute default was originally added in the provider to solve - https://github.com/terraform-providers/terraform-provider-datadog/issues/206 However, if thats not an issue anymore and there is a default imposed by the API, maybe we can remove terraform providing a default value and look into treating this as [computed](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed)?

phillip-dd: Oh, fair point about the provider default @nmuesch. Also related: https://github.com/terraform-providers/terraform-provider-datadog/issues/270.

I don't think computed would work because it can be specified by the user (and most often is?).

unclebconnor: Hmm good point re: omitted.  My personal preference tends toward having an explicit value set so I wonder if we could set the "default" value based on our recommendations?

Also, we do validate to make sure that `no_data_timeframe` is greater than 1:  https://github.com/DataDog/dogweb/blob/e878e61339d8f3ff14bdad22d8f13d8c3eb975b7/kima/lib/validation.py#L736 so I'm not sure how we even accepted values of 0 before.

unclebconnor:
@phillip-dd I just discussed this in slack with @nmuesch.  

When `no_data_timeframe` is omitted,
- the API accepts and leaves the value blank
- we have fallback values (1x eval timeframe for metric monitors, 24h for service checks)

Terraform (or maybe just go), however, seems to respond to that omitted value by unmarshalling it as 0, which is why we added the arbitrary value of 10 as a default.  

Nick suggested that we return a 'default' value for `no_data_timeframe` via the API, which we could use to replace the arbitrary default value of 10.  This could be a good alternative to our previous plan of validating against omitted values, instead just replacing empty values with the ones that we're using anyway.  In this scenario, we don't change any behaviors or break anyone's Terraform setup.  

Thoughts?

phillip-dd:
For now, I'd vote to go with @nmuesch's original suggestion to just remove the `If omitted` so that we get all the docs straightened out. And maybe link to the API docs like we do elsewhere so:
> * `no_data_timeframe` (Optional) The number of minutes before a monitor will notify when data stops reporting. Provider defaults to 10 minutes.
    We recommend at least 2x the monitor timeframe for metric alerts or 2 minutes for service checks. Please see the [API Reference](https://docs.datadoghq.com/api/?lang=bash#common-options) for details.

> we return a 'default' value for no_data_timeframe via the API, which we could use to replace the arbitrary default value of 10.

Not opposed, but not sure I totally follow. This sounds sounds like a separate improvement/refactor - want to throw into a ticket and cc me?

website/docs/r/monitor.html.markdown
unclebconnor: Ok dropped the omitted piece.  Will create a new card for follow up on the rest of our conversation below.

PR #424, opened 2020-02-28
Gzu/users

gzussa:

PR #426, opened 2020-03-03
synthetics: add support for retry (count, interval)

bendrucker:
This adds `options.retry_count` and `options.retry_interval` for configuring retry options for a synthetics test. I'm not sure if/how I should be updating the cassettes as introduced in #423, but the tests should pass once those are re-generated.

This has significant overlap with #424, which also adds support for `renotify_interval`. This PR makes fewer changes to the existing patterns in the library. It uses a flat naming scheme for retry options instead of placing them in their own map.

I'm supportive of moving to `schema.TypeList` for `options` as [recommended by the Terraform docs](https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-a-more-complex-read). However, I think that should wait until a major release and a provider-wide change rather than maintaining both for backwards compatibility.

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -197,6 +197,14 @@ func syntheticsTestOptions() *schema.Schema {
 					Type:     schema.TypeBool,
 					Optional: true,
 				},
+				"retry_count": {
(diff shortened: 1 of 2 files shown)


bendrucker:
Close in favor of #428. State migration is definitely the way to go!

PR #442, opened 2020-03-11
Stabilize TestDatadogDashListImport test by always creating resources in the same order

bkabrda:
Previously, the test would fail in ~50 % of runs, as the `screen` and `time` dashboards were created in random order.


bkabrda:

PR #448, opened 2020-03-16
Clean implementation for consistency

gzussa:


datadog/resource_datadog_service_level_objective.go
```diff
@@ -149,25 +149,71 @@ func ValidateServiceLevelObjectiveTypeString(v interface{}, k string) (ws []stri
...
 
-	slo := datadog.ServiceLevelObjective{
-		Type: datadog.String(d.Get("type").(string)),
-		Name: datadog.String(d.Get("name").(string)),
+	slo := &datadog.ServiceLevelObjective{
+		ID: datadog.String(d.Id()),
+	}
+
+	if attr, ok := d.GetOk("name"); ok {
+		slo.SetName(attr.(string))
```
bkabrda: Minor nit: `name` is required, so we don't have to check for its presence, we can just `slo.SetName(d.Get("name"))`

datadog/resource_datadog_service_level_objective.go
```diff
@@ -149,25 +149,71 @@ func ValidateServiceLevelObjectiveTypeString(v interface{}, k string) (ws []stri
...
 	if attr, ok := d.GetOk("description"); ok {
-		slo.Description = datadog.String(attr.(string))
+		slo.SetDescription(attr.(string))
 	}
 
-	if attr, ok := d.GetOk("tags"); ok {
-		tags := make([]string, 0)
-		for _, s := range attr.(*schema.Set).List() {
-			tags = append(tags, s.(string))
+	if attr, ok := d.GetOk("type"); ok {
```
bkabrda: Same as for `name` - `type` is required, so no need to check for its presence.

PR #450, opened 2020-03-17
Api error mgmt + enhanced provider

gzussa:
- Enhanced Client provider to easily support more clients in the future
- Improve Error management so that we can support distinct client errors in the future as well.

### DIFF
--- a/datadog/resource_datadog_dashboard_list.go
+++ b/datadog/resource_datadog_dashboard_list.go
@@ -1,11 +1,12 @@
 package datadog
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	datadog "github.com/zorkian/go-datadog-api"
+	"github.com/zorkian/go-datadog-api"
 )
 
 func resourceDatadogDashboardList() *schema.Resource {
@@ -53,7 +54,7 @@ func resourceDatadogDashboardListCreate(d *schema.ResourceData, meta interface{}
 
 	dashboardList, err := buildDatadogDashboardList(d)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to parse resource configuration: %s", err.Error())
 	}
 	dashboardList, err = client.CreateDashboardList(dashboardList)
 	if err != nil {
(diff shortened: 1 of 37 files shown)


datadog/resource_datadog_dashboard_list.go
```diff
@@ -49,102 +49,117 @@ func resourceDatadogDashboardList() *schema.Resource {
 }
 
 func resourceDatadogDashboardListCreate(d *schema.ResourceData, meta interface{}) error {
+	providerConf := meta.(*ProviderConfiguration)
+	client := providerConf.CommunityClient
+
 	dashboardList, err := buildDatadogDashboardList(d)
 	if err != nil {
 		fmt.Printf("Error building the dashboard list %s", err.Error())
```
nmuesch: As discussed in person, we should return the error here.

datadog/resource_datadog_integration_pagerduty_service_object.go
```diff
@@ -82,41 +84,46 @@ func resourceDatadogIntegrationPagerdutySORead(d *schema.ResourceData, meta inte
...
+	providerConf := meta.(*ProviderConfiguration)
+	client := providerConf.CommunityClient
 
 	_, err := client.GetIntegrationPDService(d.Id())
 	if err != nil {
 		if strings.Contains(err.Error(), "404 Not Found") {
 			return false, nil
 		}
-		return false, err
+		return false, translateClientError(err, "error getting pager duty integration service")
```
nmuesch: Nit, these should stay PagerDuty

nmuesch: Another nit, ,should we also say error checking exists to keep it consistent, like - https://github.com/terraform-providers/terraform-provider-datadog/pull/450/files#diff-f51916c63534715f41c1d444931ae717R370

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -203,7 +203,8 @@ func syntheticsTestOptions() *schema.Schema {
 }
 
 func resourceDatadogSyntheticsTestCreate(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*datadog.Client)
+	providerConf := meta.(*ProviderConfiguration)
+	client := providerConf.CommunityClient
```
nmuesch: Nit: Should we print the same error messages in the create/update calls this resource makes?

datadog/resource_datadog_timeboard.go
```diff
@@ -824,7 +824,9 @@ func resourceDatadogTimeboardCreate(d *schema.ResourceData, meta interface{}) er
 	if err != nil {
 		return fmt.Errorf("Failed to parse resource configuration: %s", err.Error())
 	}
-	timeboard, err = meta.(*datadog.Client).CreateDashboard(timeboard)
+	providerConf := meta.(*ProviderConfiguration)
+	client := providerConf.CommunityClient
+	timeboard, err = client.CreateDashboard(timeboard)
 	if err != nil {
 		return fmt.Errorf("Failed to create timeboard using Datadog API: %s", err.Error())
```
nmuesch: Should we use the new translateError method for these?

datadog/resource_datadog_synthetics_test_.go
gzussa: Good Catch!!

Issue #439, opened 2020-03-09
[bug] Deleting monitors that are part of an SLO fails


Towerthousand:
Hey!

Declaring a monitor and using it in a declared SLO and then removing the monitor (also removing it from the SLO in the same apply, of course) fails with an API error. There should be a check to remove it from all SLOs first or to remove it from the SLO automatically on deletion.

Happy to open a PR if pointed in the right direction. Thanks!

### Terraform Version
```
Terraform v0.12.23
+ provider.datadog v2.7.0
```

### Affected Resource(s)
- `datadog_monitor`
- `datadog_service_level_objective`

### Terraform Configuration Files
Version 1:
```hcl
datadog_monitor "this" {
  [...]
}
datadog_monitor "that" {
  [...]
}

resource datadog_service_level_objective "overrall_slo" {
  type = "monitor"
  monitor_ids = [datadog_monitor.this.id, datadog_monitor.that.id]
  [...]
}
```

Version 2:
```hcl
datadog_monitor "this" {
  [...]
}

resource datadog_service_level_objective "overrall_slo" {
  type = "monitor"
  monitor_ids = [datadog_monitor.this.id]
  [...]
}
```

### Debug Output
```
Error: API error 401 Unauthorized: {"errors":["monitor [16195606,test monitor 'this'] is referenced in slos: [34dbd856b8c8591dae09f9db0458adb2,test slo]"]}
```

### Expected Behavior
Apply runs without errors

### Actual Behavior
Apply errored with a 401, monitors and SLO untouched.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` with V1
2. `terraform apply` with V2

Towerthousand:
Datadog API explicitly supports the `force` flag for this, but the go client does not support it

Towerthousand:
I opened a PR on the go client: https://github.com/zorkian/go-datadog-api/pull/311

nmuesch:
Hey @Towerthousand Thanks for the report and the pull request to the go client. Would you be interested in working on adding this to the terraform provider?

@bkabrda updated the vendored go client in this provider that includes your change - https://github.com/terraform-providers/terraform-provider-datadog/pull/447/files

Towerthousand:
Hey! adding a PR :)

Towerthousand:
fixed in #535

PR #451, opened 2020-03-17
Use ForceDeleteMonitor instead of DeleteMonitor

Towerthousand:
This implements #439

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -512,7 +512,7 @@ func resourceDatadogMonitorDelete(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	if err = client.DeleteMonitor(i); err != nil {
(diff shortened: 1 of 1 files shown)


platinummonkey:
I think I would prefer this to be somehow configurable. This is there to protect accidental deletes of resources (monitors in this case) that might be referenced in: dashboard widgets and SLOs for example.

nmuesch:
@Towerthousand Thanks for this PR! To @platinummonkey's point, I think it would be helpful to maybe add an element to the schema so that users would explicitly be choosing to force delete their monitors and not have any accidental deletions?

I imagine the workflow in this case being, a user tries to delete a monitor without the force field set, and if its used somewhere they'd get a failure message with the monitor not deleted. Then, if it's really intended to delete, the force flag would be added to the schema with a terraform apply to store it in the state. Then remove the monitor and run another apply.
Though I'm not sure if this field can be read from the state from within the delete call?

Towerthousand:
fixed in #535

PR #455, opened 2020-03-19
Add terraform sdk + cli versions to user agent header

nmuesch:
This PR includes a couple extra pieces of info about the running terraform version in the user agent header:
* The Terraform SDK version (This is vendored into the provider, so this could be extracted indirectly by knowing the provider version, but this explicitly includes it)
* The terraform CLI version used to invoke the provider. 

Inspiration from - https://github.com/terraform-providers/terraform-provider-oci/blob/8ed6736016dfc4ee36cb77c80ea7f519ae2c84f4/oci/provider.go#L958
SDK Meta version from the `meta` package - https://github.com/hashicorp/terraform-plugin-sdk/blob/dbee9ba9e9bca050f5fa5d56d3232eefc64e44be/meta/meta.go

### DIFF
--- a/datadog/provider.go
+++ b/datadog/provider.go
@@ -9,13 +9,16 @@ import (
 	cleanhttp "github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/logging"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/meta"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 	"github.com/terraform-providers/terraform-provider-datadog/version"
 	"github.com/zorkian/go-datadog-api"
 )
 
+var datadogProvider *schema.Provider
+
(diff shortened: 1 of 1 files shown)


datadog/provider.go
```diff
@@ -79,8 +84,8 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	c := cleanhttp.DefaultClient()
 	c.Transport = logging.NewTransport("Datadog", c.Transport)
+	communityClient.ExtraHeader["User-Agent"] = fmt.Sprintf("Datadog/%s/terraform (%s)/terraform (%s)/terraform-cli (%s)", version.ProviderVersion, runtime.Version(), meta.SDKVersionString(), datadogProvider.TerraformVersion)
```
nmuesch:
```suggestion
	communityClient.ExtraHeader["User-Agent"] = fmt.Sprintf("terraform-provider-datadog/%s (go %s; terraform %s; terraform-cli %s)", version.ProviderVersion, runtime.Version(), meta.SDKVersionString(), datadogProvider.TerraformVersion)
```

PR #456, opened 2020-03-20
Add generated Go Client

gzussa:


datadog/provider.go
```diff
@@ -104,5 +134,12 @@ func translateClientError(err error, msg string) error {
 		msg = "an error occurred"
 	}
 
+	if errBody, ok := err.(datadog.GenericOpenAPIError); ok {
+		return fmt.Errorf(msg+" (datadog.GenericOpenAPIError): %s", string(errBody.Body()))
```
zippolyte: Let's still print `err.Error()` in that case too, it at least gives us the status code, and there can be cases when a `datadog.GenericOpenAPIError` can be returned before any request is made, so it won't have a `Body`

datadog/provider.go
```diff
@@ -91,16 +97,40 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+		},
+	)
+	config := datadog.NewConfiguration()
+	if apiURL := d.Get("api_url").(string); apiURL != "" {
+		if strings.Contains(apiURL, "datadoghq.eu") {
+			auth = context.WithValue(auth, datadog.ContextServerVariables, map[string]string{
+				"site": "datadoghq.eu",
+			})
+		}
+	}
```
zippolyte, line 128: I think we need to be smarter here, in case someone is not sending to a datadog URL directly

datadog/provider.go
```diff
@@ -109,5 +139,12 @@ func translateClientError(err error, msg string) error {
 		msg = "an error occurred"
 	}
 
+	if _, ok := err.(datadog.GenericOpenAPIError); ok {
+		return fmt.Errorf(msg+": %s", err.Error())
```
zippolyte, line 143:
```suggestion
		return fmt.Errorf(msg+": %v: %s", err, err.(GenericOpenAPIError).Body())
```

datadog/provider.go
```diff
@@ -109,5 +139,12 @@ func translateClientError(err error, msg string) error {
...
+	}
```
zippolyte, line 144:
```suggestion
	if apiErr, ok := err.(datadog.GenericOpenAPIError); ok {
		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
	}
```

PR #457, opened 2020-03-20
Update client in User Resource

gzussa:
- Based on `gzu/client` branch
- The `isAdmin` field is not support in the new Client.


gzussa:
@zippolyte Do you know if it is ok to ignore the `isAdmin` field?

gzussa:
Won't do because it will introduce a regression otherwise

PR #460, opened 2020-03-25
doc: AWS integration & EU Datadog clarifications

NeverTwice:
Missing information on AWS integration documentation while using the API Endpoint and EU version of Datadog

### DIFF
--- a/website/docs/r/integration_aws.html.markdown
+++ b/website/docs/r/integration_aws.html.markdown
@@ -41,8 +41,7 @@ The following arguments are supported:
   e.x. `env:production,instance-type:c1.*,!region:us-east-1`
 
 * `host_tags` - (Optional) Array of tags (in the form key:value) to add to all hosts and metrics reporting through this integration.
-* `account_specific_namespace_rules` - (Optional) Enables or disables metric collection for specific AWS namespaces for this AWS account only. A list of namespaces can be found at the [available namespace rules API endpoint](https://api.datadoghq.com/api/v1/integration/aws/available_namespace_rules). 
-In order to use this API you'll have to manually set the `api_key` and `application_key` parameters in the URL (ex: `.../aws/available_namespace_rules?api_key=<your_api_key>&application_key=<your_application_key>`) and not forget to use `.eu` instead of `.com` extension if you're working with "EU" Datadog. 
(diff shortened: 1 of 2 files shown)


website/docs/r/integration_aws.html.markdown
```diff
@@ -41,7 +41,8 @@ The following arguments are supported:
   e.x. `env:production,instance-type:c1.*,!region:us-east-1`
 
 * `host_tags` - (Optional) Array of tags (in the form key:value) to add to all hosts and metrics reporting through this integration.
-* `account_specific_namespace_rules` - (Optional) Enables or disables metric collection for specific AWS namespaces for this AWS account only. A list of namespaces can be found at the [available namespace rules API endpoint](https://api.datadoghq.com/api/v1/integration/aws/available_namespace_rules).
+* `account_specific_namespace_rules` - (Optional) Enables or disables metric collection for specific AWS namespaces for this AWS account only. A list of namespaces can be found at the [available namespace rules API endpoint](https://api.datadoghq.com/api/v1/integration/aws/available_namespace_rules). 
+In order to use this API you'll have to manually set the `api_key` and `application_key` parameters in the URL (ex: `.../aws/available_namespace_rules?api_key=<your_api_key>&application_key=<your_application_key>`) and not forget to use `.eu` instead of `.com` extension if you're working with "EU" Datadog.
```
nmuesch: Thanks for calling this out! What do you think of instead changing the link to point to the documentation for this endpoint - https://docs.datadoghq.com/api/?lang=bash#list-namespace-rules That'd keep the description here short while staying in sync with the documentation.

NeverTwice: Great idea, just one minor problem... If you're not using the EN version of https://docs.datadoghq.com/ you get redirected to https://docs.datadoghq.com/api/?lang=bash#api-reference ,as "#list-namespace-rules" reference doesn't exist in the new language. Shall we ignore that or do you have an other solution ?

nmuesch: Hey, I chatted internally and it sounds like the different languages has a bit of a lag but should sync in the future. So in that case I think it should be OK to use this link.

Issue #402, opened 2020-02-03
Support DD_API_KEY, DD_APP_KEY and DD_HOST environment variables


bigwheel:
DATADOG_ prefix environment variables are legacy.
https://github.com/DataDog/dd-trace-py/blob/b4aaa29d/ddtrace/utils/formats.py#L11-L12

It seems like it was official environment variable in the past, but now most datadog documents describe only "DD_" prefix environment variables.
https://docs.datadoghq.com/ja/agent/docker/?tab=standard#global-options

I guess if datadog provider supports DD_API_KEY, DD_APP_KEY and DD_HOST it makes us happy.

fyi:
[Search · DATADOG\_API\_KEY](https://github.com/terraform-providers/terraform-provider-datadog/search?q=DATADOG_API_KEY&unscoped_q=DATADOG_API_KEY)

PR #469, opened 2020-04-14
Add support for env variable with DD prefixes

skarimo:
Closes: #402 

This PR adds support for: DD_API_KEY, DD_APP_KEY and DD_HOST env variables. Retained support for DATADOG_API_KEY, DATADOG_APP_KEY and DATADOG_HOST for backward compatibility


datadog/provider_test.go
```diff
@@ -162,10 +162,10 @@ func testAccPreCheck(t *testing.T) {
 	if isReplaying() {
 		return
 	}
-	if v := os.Getenv("DATADOG_API_KEY"); v == "" {
-		t.Fatal("DATADOG_API_KEY must be set for acceptance tests")
+	if v := os.Getenv("DD_API_KEY"); v == "" {
```
nmuesch, line 165: I believe the CI relies on the secrets being this name. Can we do an `or` and look for the old one as well for now too?

datadog/provider.go
```diff
@@ -23,17 +23,17 @@ func Provider() terraform.ResourceProvider {
 			"api_key": {
 				Type:        schema.TypeString,
 				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("DATADOG_API_KEY", nil),
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{"DD_API_KEY", "DATADOG_API_KEY"}, nil),
```
nmuesch, line 26: Nice, TIL `MultiEnvDefaultFunc` I'm torn on which variable should be first here. According to https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L287, the first available variable in the list that has a value is used.

I'm wondering if the safest path is to have the old value (`DATADOG_`) be the first one, so in the case a user has two set, i.e. one for testing and another for another org, we'd choose the old value. WDYT?

skarimo: Definitely agree with you there, just in case customer has both set previously.

datadog/provider_test.go
skarimo: Broke it down into its own functions. Let me know what you think:

```
func testAccPreCheck(t *testing.T) {
	if isReplaying() {
		return
	}
	if !isAPIKeySet() {
		t.Fatal("DD_API_KEY must be set for acceptance tests")
	}
	if !isAPPKeySet() {
		t.Fatal("DD_APP_KEY must be set for acceptance tests")
	}
}
```

PR #471, opened 2020-04-16
Add resources for API and application keys

bendrucker:
The team at @takescoop is looking to move from a small number of Datadog API keys to provisioning a unique key for each infrastructure service we run, in each environment. I reached out to support and they graciously raised our account limit from 5 keys (the default) to 200, though we currently need maybe 10% of that. All of the places we distribute Datadog credentials to are already managed via Terraform so we're hoping to distribute new credentials by adding something like this:

```tf
resource "datadog_api_key" "agent" {
  name = "Kubernetes (${var.environment})"
}
```

Or to allow Terraform Cloud workspaces to use the Datadog provider:

```tf
resource "datadog_api_key" "terraform" {
  name = "Terraform Cloud"
}

resource "datadog_app_key" "terraform" {
  name = "Terraform Cloud"
}

resource "tfe_variable" "datadog_api_key" {
  key          = "DATADOG_API_KEY"
  value        = datadog_api_key.terraform.key
  sensitive    = true
  category     = "env"
  description  = "API key required by Datadog provider"
  workspace_id = "..."
}

resource "tfe_variable" "datadog_app_key" {
  for_each = tfe_workspace.project

  key          = "DATADOG_APP_KEY"
  value        = datadog_app_key.terraform.hash
  sensitive    = true
  category     = "env"
  description  = "Application key required by Datadog provider"
  workspace_id = each.value.id
}
```

* Acceptance tests are passing and cover creating keys and updating their names
* Cassettes are updated, and I replaced my email with `frog@datadoghq.com`, which was present in some other cassettes
* Docs are included

Notes:

* Application keys return a `hash` attribute but the UI calls the column "key." I chose to maintain parity with the API but if you're interested in using `key` uniformly in Terraform I'm happy to do that too.
* The fact that keys don't have a unique non-secret identifier means that the key/hash value is used as the Terraform resource ID. Setting `Sensitive: true` for key/hash is relatively meaningless because the resource ID is printed during a refresh. 

My wishlist for these APIs:

* Return a unique but non-secret random ID with each API/application key (`id`)
* Accept it in requests to `GET /{api,application}_key/*`, in addition to the secret key/hash for backwards compatibility
* Return key/hash on `POST`, but never again (i.e. GET/PUT)
  * Breaking and requires UI changes, but would leave the keys a lot less exposed

Thanks and let me know if there are any other changes/revisions you'd like! 🔑🐶

### DIFF
--- a/datadog/cassettes/TestAccDatadogAPIKey_Basic.yaml
+++ b/datadog/cassettes/TestAccDatadogAPIKey_Basic.yaml
@@ -1,0 +1,354 @@
+---
+version: 1
+interactions:
+- request:
+    body: '{"name":"foo"}'
+    form: {}
+    headers:
+      Content-Type:
+      - application/json
+      User-Agent:
+      - Datadog/dev/terraform (go1.14)
+    url: https://api.datadoghq.com/api/v1/api_key
+    method: POST
+  response:
+    body: '{"api_key":{"name":"foo","created":"2020-04-16 04:02:05","disabled_by":"","is_active":true,"created_by":"frog@datadoghq.com","disabled":"","key":"84e9a927e6cc2a3b4145b531316311d0"}}'
+    headers:
+      Cache-Control:
+      - no-cache
+      Connection:
+      - keep-alive
+      Content-Length:
+      - "182"
+      Content-Security-Policy:
+      - frame-ancestors 'self'; report-uri https://api.datadoghq.com/csp-report
+      Content-Type:
+      - application/json
+      Date:
+      - Thu, 16 Apr 2020 04:02:05 GMT
+      Pragma:
+      - no-cache
+      Set-Cookie:
+      - DD-PSHARD=141; Max-Age=604800; Path=/; expires=Thu, 23-Apr-2020 04:02:05 GMT;
+        secure; HttpOnly
+      Strict-Transport-Security:
+      - max-age=15724800;
+      X-Content-Type-Options:
+      - nosniff
+      X-Dd-Version:
+      - "35.2396352"
+      X-Frame-Options:
+      - SAMEORIGIN
+    status: 200 OK
+    code: 200
+    duration: ""
+- request:
+    body: ""
+    form: {}
+    headers:
+      User-Agent:
+      - Datadog/dev/terraform (go1.14)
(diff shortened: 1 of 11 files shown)


bkabrda:
Hey, thanks for the PR. We know that this is an important feature that our users are missing, but there are some security concerns - notably using the key itself as the ID doesn't hide it in the state file even when it's marked as sensitive. We've been discussing this with our API team internally and we think that this is something that needs to be fixed before we can add these resources.

One possible solution that we're discussing is adding non-secret identifiers to all keys, which would effectively allow for a safe implementation in the provider. I'll try to keep this PR updated when there's progress.

Thanks.

bendrucker:
Thanks for the response! I'd certainly welcome non-secret identifiers!

Wanted to refine this a bit though:

> notably using the key itself as the ID doesn't hide it in the state file even when it's marked as sensitive

The "sensitive" attribute in Terraform is purely advisory. It is used by `terraform output` and `terraform apply` for output blocks and `terraform state show` for resource schemas. Sensitive data is still persisted to the state file in plain text. The fact that the ID attribute for an API or app key is the key itself means that keys are _also_ printed in logs during Terraform runs.

As long as users also have access to Datadog, they can see every key we issue, whether via the UI or API. For my team, everyone with Terraform access also has Datadog access. We'd therefore be willing to accept that keys are logged in Terraform Cloud for the time being.

In the long run, I would argue that the security impact of a non-secret identifier is negligible, compared to never re-transmitting the secret in an API response. That is the only way to truly reduce the number of people who can access a key.

pietdaniel:
We are deprecating these APIs in favor a more secure version which is not prone to leaking information about keys. The secure version will be added to terraform.

bendrucker:
Very glad to hear! I'm going to close this since:

* There's no path to getting this merged ever
* There's some commitment towards new APIs

Happy to help write the resources for the new API. Let me know if I can help @pietdaniel, happy to give API feedback, maybe beta test on our account, anything you want.

bendrucker:
@gzussa reached out via email to alert me that my keys are leaked in the cassettes. I believe that these keys were destroyed by Terraform during the acceptance test run and were invalidated by the time I committed them. Hard to remember, but I think I was aware that I was committing keys and forgot to mention it in the original description. To be extra safe, I've rotated the keys for that account that I used to run the acceptance tests. Just leaving this note in case someone notices this in the future.

PR #474, opened 2020-04-18
provider: add "validate" option that can disable validation

### LINKED ISSUE #368: Allow to use data source for Datadog IP anonymously (no api key requirement)
I am very glad we got #216 resolved through #298. Thanks again to @dabcoder for this.

It would be good though if we could call [`datadog_ip_ranges`](https://www.terraform.io/docs/providers/datadog/d/ip_ranges.html) without an API key (so anonymously).

This is similar to the [Github Provider](https://www.terraform.io/docs/providers/github/index.html) where there is an `anonymous` flag. 

Is there a chance we could do this for the datadog provider?

The reasons why this is useful
- In some terraform states, we sometimes are defining just the firewall rules that also include datadog IPs. This terraform state is usually separate from our Datadog monitors.
- This makes it very hard to having to populate Datadog API keys in remote states do not actually to actually create/manage Datadog resources extensively.

bendrucker:
This allow uses to disable credential validation by setting `validate = false` in the provider configuration.

### Use Case

I recently wrote some tests for a set of complex modules that abstract details of Datadog resources, e.g. monitors and synthetic tests. While I might E2E test these configurations in a dedicated sandbox account, initially I wanted to write tests that assert some of the contents of:

```sh
terraform plan -out ./plan.out
terraform show -json ./plan.out
```

The resulting JSON document has `planned_resources` and `changed_resources`. There's no existing state so I only care about the `planned_resources`. Currently these tests require valid credentials but they would pass just the same with mock credentials if validation could be disabled.

Closes #368


enieuw:
+1 on this, we have a use case where we conditionally create DD resources. Right now even when no DD resources are created we need to provide the provider with valid credentials so this would help us improve that!

bendrucker:
Fixing the build, forgot to close a `{` 🤦‍♂️

datadog/provider.go
```diff
@@ -35,6 +35,11 @@ func Provider() terraform.ResourceProvider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("DATADOG_HOST", nil),
 			},
+			"validate": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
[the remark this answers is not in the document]
bendrucker: Yes, great idea!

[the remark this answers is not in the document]
bendrucker: That would mean making making `api_key` and `app_key` not required though. With `validate = false` and no other changes you'd still need to set empty strings for the API and app keys.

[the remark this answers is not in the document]
nmuesch: Ahh yeah good point! Maybe we could make api, app, and validate all optional, and just have quick validation in the provider init code that exactly one of these is set?

The expected workflow would then be:
Set the api/app key, and not validate. -> OK
Set validate and not api/app key -> OK
Set validate and api/app key -> Error message during provider initialization
Set neither -> Error message during initialization

Mostly I think it'd be helpful to avoid the scenario of a user not knowing which takes precedence; `validate` or the actual setting of the api/app keys.

bendrucker:
Changes:

* Keys are now optional and validation is performed by the `ConfigureFunc`
* Keys must not be set if validate is false, and must be set if validate = true
* The docs explain this

Notes:

* I opted not to use `ConflictsWith` since that expresses that entire attributes conflict regardless of values. If a user ends up setting `validate = true` and providing valid keys I don't want to treat that as an error.

datadog/provider.go
```diff
@@ -97,17 +114,20 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+			return nil, err
+		} else if !ok {
+			err := errors.New(`Invalid or missing credentials provided to the Datadog Provider. Please confirm your API and APP keys are valid and are for the correct region, see https://www.terraform.io/docs/providers/datadog/ for more information on providing credentials for the Datadog Provider`)
+			log.Printf("[ERROR] Datadog Client validation error: %v", err)
+			return nil, err
+		}
+	} else {
+		log.Println("[INFO] Skipping key validation (validate = false)")
 	}
-	log.Printf("[INFO] Datadog Client successfully validated.")
```
nmuesch, line 110: Lets keep this and move it after line 127, at the end of the validate block, so no matter the code path we get some log validation about this step.

datadog/provider.go
```diff
@@ -97,17 +114,20 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
 
 	// Initialize the official Datadog V1 API client
 	authV1 := context.WithValue(
```
nmuesch, line 133: We should use the new apiKey and appKey variables for the other client initialization as well. (This is a bit outside the scope of your changes, so we can always do this later :))

website/docs/index.html.markdown
```diff
@@ -37,6 +37,7 @@ resource "datadog_timeboard" "default" {
 ## Argument Reference
 
 The following arguments are supported:
-* `api_key` - (Required) Datadog API key. This can also be set via the `DD_API_KEY` environment variable.
-* `app_key` - (Required) Datadog APP key. This can also be set via the `DD_APP_KEY` environment variable.
+* `api_key` - (Optional) Datadog API key. This can also be set via the `DD_API_KEY` environment variable.
```
nmuesch, line 40: This is a nit, but maybe something like:
```suggestion
* `api_key` - (Required unless `validate` is false) Datadog API key. This can also be set via the `DD_API_KEY` environment variable.
```
(For the app key as well) Let me know what you think

website/docs/index.html.markdown
```diff
@@ -37,6 +37,7 @@ resource "datadog_timeboard" "default" {
...
+* `app_key` - (Optional) Datadog APP key. This can also be set via the `DD_APP_KEY` environment variable.
 * `api_url` - (Optional) The API Url. This can be also be set via the `DD_HOST` environment variable. Note that this URL must not end with the `/api/` path. For example, `https://api.datadoghq.com/` is a correct value, while `https://api.datadoghq.com/api/` is not. And if you're working with  "EU" version of Datadog, use `https://api.datadoghq.eu/`.
+* `validate` - (Optional) Enables validation of the provided API and APP keys during provider initialization. Default is true. When false, `api_key` and `app_key` may only be set to empty strings.
```
nmuesch, line 43: 💜

datadog/provider.go
bendrucker: Whoops, yes

bendrucker:
Changes made!

PR #477, opened 2020-04-23
Add Generated client v2 API 

skarimo:
This PR adds Generate go client V2 api to the Terraform provider

### DIFF
--- a/datadog/provider.go
+++ b/datadog/provider.go
@@ -186,20 +186,8 @@ func translateClientError(err error, msg string) error {
 	if apiErr, ok := err.(datadogV1.GenericOpenAPIError); ok {
 		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
 	}
-	if errUrl, ok := err.(*url.Error); ok {
-		return fmt.Errorf(msg+" (url.Error): %s", errUrl)
-	}
-
-	return fmt.Errorf(msg+": %s", err.Error())
-}
-
-func translateClientErrorV2(err error, msg string) error {
-	if msg == "" {
-		msg = "an error occurred"
-	}
-
 	if apiErr, ok := err.(datadogV2.GenericOpenAPIError); ok {
-		fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
+		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
 	}
 	if errUrl, ok := err.(*url.Error); ok {
 		return fmt.Errorf(msg+" (url.Error): %s", errUrl)
--- a/datadog/provider.go
+++ b/datadog/provider.go
@@ -9,7 +9,7 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/DataDog/datadog-api-client-go/api/v1/datadog"
+	datadogV1 "github.com/DataDog/datadog-api-client-go/api/v1/datadog"
 	datadogV2 "github.com/DataDog/datadog-api-client-go/api/v2/datadog"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/logging"
@@ -77,9 +77,9 @@ func Provider() terraform.ResourceProvider {
 //ProviderConfiguration contains the initialized API clients to communicate with the Datadog API
 type ProviderConfiguration struct {
 	CommunityClient *datadogCommunity.Client
-	DatadogClientV1 *datadog.APIClient
+	DatadogClientV1 *datadogV1.APIClient
 	DatadogClientV2 *datadogV2.APIClient
-	Auth            context.Context
+	AuthV1          context.Context
 	AuthV2          context.Context
 }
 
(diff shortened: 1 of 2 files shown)


datadog/provider.go
```diff
@@ -148,3 +177,18 @@ func translateClientError(err error, msg string) error {
...
 }
+
+func translateClientErrorV2(err error, msg string) error {
+	if msg == "" {
+		msg = "an error occurred"
+	}
+
+	if _, ok := err.(datadogV2.GenericOpenAPIError); ok {
+		return fmt.Errorf(msg+": %s", err.Error())
+	}
```
zippolyte, line 188:
```suggestion
	if apiErr, ok := err.(datadogV2.GenericOpenAPIError); ok {
		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
	}
```

datadog/provider_test.go
```diff
@@ -153,11 +168,25 @@ func testProviderConfigure(r *recorder.Recorder) schema.ConfigureFunc {
...
+		configV2 := datadogV2.NewConfiguration()
+		configV2.Debug = true
+		configV2.HTTPClient = c
+		if apiURL := d.Get("api_url").(string); apiURL != "" {
+			if strings.Contains(apiURL, "datadoghq.eu") {
+				auth = context.WithValue(auth, datadogV2.ContextServerVariables, map[string]string{
+					"site": "datadoghq.eu",
+				})
+			}
+		}
```
zippolyte, line 181: I think we need to be smarter here, in case someone is not sending to a datadog URL directly

datadog/provider.go
```diff
@@ -77,7 +78,9 @@ func Provider() terraform.ResourceProvider {
 type ProviderConfiguration struct {
 	CommunityClient *datadogCommunity.Client
 	DatadogClientV1 *datadog.APIClient
+	DatadogClientV2 *datadogV2.APIClient
 	Auth            context.Context
```
jirikuncar, line 82:
```suggestion
	AuthV1          context.Context
```

datadog/provider_test.go
skarimo: Interesting. How would one configure other `site` for the generated go client? From what I can tell, it seems the possible configurations for the `site` variable on the clients are `datadoghq.com` or `datadoghq.eu`

zippolyte: There is another server possible, by setting `ServerIndex` to 2, you can specify an arbitrary URL
See https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/header.yaml#L22-L29

datadog/provider.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/DataDog/datadog-api-client-go/api/v1/datadog"
```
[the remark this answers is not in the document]
skarimo: Updated all corresponding configs imports and appended V1

datadog/provider_test.go
skarimo: Thanks for that link very helpful! Pushed a commit to better handle non-datadog urls. Let me know what you think.

datadog/provider.go
```diff
@@ -118,19 +120,75 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+			return nil, fmt.Errorf(`missing protocol or host : %v`, apiURL)
+		}
+		// Set site to "datadoghq.eu" on ServerIndex{0} if Datadog EU API URL is passed.
+		// If custom api url (not datadoghq.eu or datadoghq.com) is passed, set the api name and protocol on ServerIndex{-1}
+		if parsedApiUrl.Host == "datadoghq.eu" {
+			authV1 = context.WithValue(authV1, datadogV1.ContextServerVariables, map[string]string{
+				"site": "datadoghq.eu",
+			})
+		} else if parsedApiUrl.Host != "datadoghq.com" {
+			authV1 = context.WithValue(authV1, datadogV1.ContextServerIndex, len(configV1.Servers)-1)
```
zippolyte, line 139:
```suggestion
			authV1 = context.WithValue(authV1, datadogV1.ContextServerIndex, 1)
```

datadog/provider.go
```diff
@@ -118,19 +120,75 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+		}
+		if parsedApiUrl.Host == "" || parsedApiUrl.Scheme == "" {
+			return nil, fmt.Errorf(`missing protocol or host : %v`, apiURL)
+		}
+		// Set site to "datadoghq.eu" on ServerIndex{0} if Datadog EU API URL is passed.
+		// If custom api url (not datadoghq.eu or datadoghq.com) is passed, set the api name and protocol on ServerIndex{-1}
+		if parsedApiUrl.Host == "datadoghq.eu" {
+			authV1 = context.WithValue(authV1, datadogV1.ContextServerVariables, map[string]string{
+				"site": "datadoghq.eu",
+			})
```
zippolyte, line 137: I don't think we need this case. The second case covers everything, even Datadog URLs

skarimo: Updated and removed the unnecessary case you mentioned above.

datadog/provider.go
```diff
@@ -139,8 +183,23 @@ func translateClientError(err error, msg string) error {
...
+		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
+	}
+	if errUrl, ok := err.(*url.Error); ok {
+		return fmt.Errorf(msg+" (url.Error): %s", errUrl)
+	}
+
+	return fmt.Errorf(msg+": %s", err.Error())
+}
+
+func translateClientErrorV2(err error, msg string) error {
```
gzussa, line 196: Why not have everything in one method?

PR #478, opened 2020-04-23
Use gotestsum

nmuesch:
Use gotestsum for the test execution https://github.com/gotestyourself/gotestsum to get a nice summary at the end. 

Updates the Makefile to use this wherever `go test` was being executed. Also adds a get-test-deps step so that we can ensure we get gotestsum without modifying the go.mod we ship with the provider.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -76,6 +76,6 @@ update-go-client:
 	go mod tidy
 
 get-test-deps:
-	cd `mktemp -d`;	GO111MODULE=on GOFLAGS='' go get -u gotest.tools/gotestsum@v0.4.1; cd -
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -17,17 +17,17 @@ install: fmtcheck
 uninstall:
 	@rm -vf $(DIR)/terraform-provider-datadog
 
-test: fmtcheck
-	go test -i $(TEST) || exit 1
+test: get-test-deps fmtcheck
+	gotestsum --format testname -- -i $(TEST) || exit 1
```
jirikuncar, line 21: This should only install dependencies.
```suggestion
	go test -i $(TEST) || exit 1
```

nmuesch:
Following up on this through https://github.com/DataDog/terraform-provider-datadog/pull/600

PR #479, opened 2020-04-29
Update Dashboard_list resource to use Datadog Go client

skarimo:
Updates Dashboard_list to use Datadog GO client.


datadog/resource_datadog_dashboard_list.go
```diff
@@ -80,61 +84,51 @@ func resourceDatadogDashboardListCreate(d *schema.ResourceData, meta interface{}
...
-		if err != nil {
-			return fmt.Errorf("failed to parse resource configuration: %s", err.Error())
-		}
-		_, err = client.UpdateDashboardListItemsV2(id, dashboardListV2Items)
-		if err != nil {
-			return translateClientError(err, "error updating dashboard list item")
-		}
-	}
-
+	// Update the dashboard list
```
gzussa, line 108: Why is this different from what is above?

datadog/resource_datadog_dashboard_list_test.go
```diff
@@ -106,21 +107,22 @@ func TestDatadogDashListImport(t *testing.T) {
...
 	for _, r := range s.RootModule().Resources {
 		if !strings.Contains(r.Primary.Attributes["name"], "List") {
 			continue
 		}
 		id, _ := strconv.Atoi(r.Primary.ID)
-		_, errList := client.GetDashboardList(id)
+		_, _, errList := datadogClientV1.DashboardListsApi.GetDashboardList(authV1, int64(id)).Execute()
 		if errList != nil {
-			if strings.Contains(errList.Error(), "not found") {
+			if strings.Contains(errList.Error(), "Not Found") {
```
gzussa, line 125: Why did it change?

datadog/resource_datadog_dashboard_list.go
```diff
@@ -183,30 +179,32 @@ func resourceDatadogDashboardListImport(d *schema.ResourceData, meta interface{}
 	return []*schema.ResourceData{d}, nil
 }
 
-func buildDatadogDashboardList(d *schema.ResourceData) (*datadog.DashboardList, error) {
-	var dashboardList datadog.DashboardList
+func buildDatadogDashboardList(d *schema.ResourceData) (datadogV1.DashboardList, error) {
+	var dashboardList datadogV1.DashboardList
 	dashboardList.SetName(d.Get("name").(string))
-	return &dashboardList, nil
+	return dashboardList, nil
```
gzussa, line 185: Using pointer is memory efficient

datadog/resource_datadog_dashboard_list.go
```diff
@@ -183,30 +179,32 @@ func resourceDatadogDashboardListImport(d *schema.ResourceData, meta interface{}
...
+	}
+	dashboardListV2Items := datadogV2.NewDashboardListItems(dashboardListV2ItemsArr)
+	return *dashboardListV2Items, nil
 }
 
-func buildTerraformDashboardListItemsV2(completeItemListV2 []datadog.DashboardListItemV2) ([]map[string]interface{}, error) {
+func buildTerraformDashboardListItemsV2(completeItemListV2 datadogV2.DashboardListItems) ([]map[string]interface{}, error) {
 	dashItemListV2 := make([]map[string]interface{}, 0, 1)
-	for _, item := range completeItemListV2 {
+	for _, item := range completeItemListV2.Dashboards {
```
gzussa, line 204: Should we use a getter instead here?

datadog/resource_datadog_dashboard_list.go
skarimo: I will sync with Nick on this one to see why it was previously doing the extra step of both deleting the list items and manually adding the items in the config. The api endpoint receives the updated Dash and replaces the previous list. Since the api does this automatically, we were doing extra unnecessary steps. 

But I will follow up here after discussing with Nick

datadog/resource_datadog_dashboard_list_test.go
skarimo: Previously the community client returned the error object as well: `404 Not Found {"errors": "...{dash_id} not found"}`. Where as the go client returns `404 Not Found` alone. Which is why the following never failed when changing from communityclient -> datadog go client: https://github.com/skarimo/terraform-provider-datadog/blob/master/datadog/resource_datadog_dashboard_list.go#L171

datadog/resource_datadog_dashboard_list.go
gzussa: I think we should keep the old logic. However, I am not sure why we do `client.UpdateDashboardListItemsV2(id, dashboardListV2Items)` when in fact we should create them.

PR #485, opened 2020-04-29
Update monitor resource to use Datadog Go client

skarimo:
Updates monitor to use Datadog GO client.

Notes: 
- The tests were updated to remove trailing zeroes during assertions. This is ok since we already have `suppressDataDogFloatIntDiff` method making sure value match as float. Precision is not effected. Only trailing zeroes


gzussa:
Let's hold off on this one for now.

PR #487, opened 2020-04-29
Update pagerduty resource to use Datadog Go client

skarimo:
Updates pagerduty to use Datadog GO client.

Note: Currently the Go client does not have option to update `subdomain` and `api_keys`. Hacky solution is to use the `CreatePagerDutyIntegration` method to both update and create the resource.


gzussa:
Actually this resource is deprecated. Users should use the PD service Object resource instead. As a consequence, we won't migrate this resource. Official SDK client won't support this endpoint anyway.

PR #497, opened 2020-04-30
Migrate AWS logs to use the official Go Client

gzussa:


datadog/resource_datadog_integration_aws_lambda_arn.go
```diff
@@ -82,39 +80,40 @@ func resourceDatadogIntegrationAwsLambdaArnExists(d *schema.ResourceData, meta i
...
 	attachLambdaArnRequest := buildDatadogIntegrationAwsLambdaArnStruct(d)
-	err := client.AttachLambdaARNIntegrationAWS(attachLambdaArnRequest)
-
+	_, _, err := datadogClientV1.AWSLogsIntegrationApi.CreateAWSLambdaARN(authV1).Body(*attachLambdaArnRequest).Execute()
 	if err != nil {
 		return translateClientError(err, "error attaching Lambda ARN to AWS integration account")
 	}
 
-	d.SetId(fmt.Sprintf("%s %s", *attachLambdaArnRequest.AccountID, *attachLambdaArnRequest.LambdaARN))
+	d.SetId(fmt.Sprintf("%s %s", attachLambdaArnRequest.AccountId, attachLambdaArnRequest.LambdaArn))
```
nmuesch, line 92: Should these use the `Get`ters?

datadog/resource_datadog_integration_aws_lambda_arn.go
```diff
@@ -56,9 +53,10 @@ func resourceDatadogIntegrationAwsLambdaArnExists(d *schema.ResourceData, meta i
 	// Exists - This is called to verify a resource still exists. It is called prior to Read,
 	// and lowers the burden of Read to be able to assume the resource exists.
 	providerConf := meta.(*ProviderConfiguration)
-	client := providerConf.CommunityClient
+	datadogClientV1 := providerConf.DatadogClientV1
+	authV1 := providerConf.AuthV1
 
-	logCollections, err := client.GetIntegrationAWSLogCollection()
+	logCollections, _, err := datadogClientV1.AWSLogsIntegrationApi.ListAWSLogsIntegrations(authV1).Execute()
```
nmuesch, line 59: Should we be checking the httpresp and error code (in addition to the `err` object) and failing if not 2xx? (Same question for all the api calls here)

datadog/resource_datadog_integration_aws_lambda_arn.go
skarimo: Addressed here https://github.com/terraform-providers/terraform-provider-datadog/pull/498

datadog/resource_datadog_integration_aws.go
```diff
@@ -110,6 +111,7 @@ func resourceDatadogIntegrationAwsPrepareCreateRequest(d *schema.ResourceData, a
 		for _, s := range attr.([]interface{}) {
 			hostTags = append(hostTags, s.(string))
 		}
+		iaws.SetHostTags(hostTags)
```
nmuesch, line 114: Do you know if this changes the scenario when these fields aren't set in the TF config? I'm wondering if a null value gets sent over an empty list, and if that makes a difference in behavior (since we're now setting this field in the if branch instead of always)

datadog/resource_datadog_integration_aws_lambda_arn.go
skarimo: I am updating the PR @nick here is the pr to address the first comment: https://github.com/terraform-providers/terraform-provider-datadog/pull/498

In regards to the second comment none of our resources actively check for the http_response at the moment. I think checking the http resp would unnecessarily complicate the code at the moment. None 2xx responses are marked as errors. What do you think?

gzussa: If needed we will do in another pr

nmuesch: My main question here is if there is a case where `err` isn't set but httpresp isn't a 2xx. 
I just want to confirm we're continuing to have the same behavior as before this switch.

datadog/cassettes/TestAccDatadogIntegrationAWS.yaml
```diff
@@ -2,17 +2,22 @@
...
     headers:
+      Accept:
+      - application/json
       Content-Type:
       - application/json
+      Dd-Operation-Id:
+      - CreateAWSAccount
       User-Agent:
-      - Datadog/dev/terraform (go1.13.4)
+      - datadog-api-client-go/1.0.0-beta.1+dev.1 (go go1.13; os darwin; arch amd64)
```
jirikuncar, line 16: we need to update the `User-Agent` for `datadog-api-client-go` to `terraform-provider-datadog/{version} datadog-api-client-go/1.0.0-beta.1+dev.1 (go go1.13; os darwin; arch amd64)` format.

gzussa: I will do that in another PR because it will be too many changes for this PR. Good catch and let's update it once we have `beta.3`

zippolyte: Could we instead ignore the user agent in the cassettes ? It's really system and version dependent, so not great for replaying as people won't always be running tests on the same system

jirikuncar: What I meant is that we are recording new cassette but the user-agent is not right.

PR #502, opened 2020-05-01
use schema.ImportStatePassthrough for id-only imports

bendrucker:
This PR makes use of `schema.ImportStatePassthrough` for resources that require only an ID to refresh. 

Terraform will attempt to refresh a resource before an import succeeds and the result is persisted to state. The `schema.StateFunc` only needs to set the ID to a value `Read` can use. If the resource has attributes that are not set by `Read`, it should set these too. This is the case for `resourceDatadogIntegrationAwsImport` with `external_id`. 

Some resources were using `schema.ImportStatePassthrough` while others still had a custom `StateFunc`.

### DIFF
--- a/datadog/cassettes/TestAccDatadogDashboard_import.yaml
+++ b/datadog/cassettes/TestAccDatadogDashboard_import.yaml
@@ -14,18 +14,18 @@ interactions:
       Content-Type:
       - application/json
       User-Agent:
-      - Datadog/dev/terraform (go1.13.4)
+      - Datadog/dev/terraform (go1.14.2)
     url: https://api.datadoghq.com/api/v1/dashboard
     method: POST
   response:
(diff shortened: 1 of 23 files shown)


bendrucker:
Will look into the test failures

bendrucker:
Hmm, was expecting some potential failures from fewer API calls happening and therefore the cassettes would need to be updated. Perhaps that's causing these errors. Will re-record and see if the tests pass live.

bendrucker:
Confirmed, was just a matter of generating cassettes! I regenerated them for all the import-related  tests. Happy to scrub my personal info if that's a concern but it doesn't bother me, it's all public info.

bendrucker:
Because of the conflicts and the lack of response here I'm going to ahead and close this. The changes are still valid. If you are interested in removing the unnecessary code feel it should be enough to just ignore the conflicts and re-generate the affected cassettes.

PR #505, opened 2020-05-04
Do not set empty structures in request to create aws integration

### LINKED ISSUE #504: Error regarding host_tags while using datadog_integration_aws
I'm running custom built binaries for this provider because I needed a feature introduced in PR #474. Since that PR is now merged I built new binaries based on latest master (at commit 0464c2e).

While setting up AWS integration I keep getting an error regarding host tags (which I do not even set).
This is the error from the DD api `{"errors": ["'host_tags' should be array"]}`.

Digging a bit deeper in the provider code and in the APIs being called, it seems that the DD api is quite robust. It accepts requests without host tags, as well as and empty array. It does produce this exact error if a string, for instance, is passed in host tags.

### Terraform Version
```
Terraform v0.12.24
+ provider.datadog v2.7.99-master
```

### Affected Resource(s)
- datadog_integration_aws

```hcl
provider "datadog" {
  api_key = "..."
  app_key = "..."
  api_url = "https://api.datadoghq.eu/"
  version = "v2.7.99-master"   # 👈this is the version I chose for the binaries I built from master (at commit 0464c2e)
}

resource "datadog_integration_aws" "integration" {
  account_id = "123456789101"
  role_name  = "role"
}
```

### Debug Output
https://gist.github.com/miguelaferreira/5658e9a79f0ccda4c2c00ad96d9b060b

### Panic Output
No panic.

### Expected Behavior
That the integration be set up just as it is with version `v2.7.0` of the provider.

### Actual Behavior
There is an error applying the configuration.
```bash
$  tf apply -auto-approve  
datadog_integration_aws.integration: Creating...

Error: error creating AWS integration: 400 Bad Request: {"errors": ["'host_tags' should be array"]}

  on main.tf line 8, in resource "datadog_integration_aws" "integration":
   8: resource "datadog_integration_aws" "integration" {
```

### Steps to Reproduce
1. `terraform apply -auto-approve`

### Important Factoids
I'm running custom build (unreleased) binaries. 
If I set host tags to an empty list `host_tags = []` I still get the same error.
If I set host tags to a list with a single value `host_tags = ["key:value"]`, then I get a similar error regarding filter tags:
```
 {"errors": ["'filter_tags' should be array"]}
```

### References
None.

miguelaferreira:
This PR Fixes #504 by only setting host tags, filter tags and account specific namespace rules on the http request if these values have been set by the user. 

Somehow, setting these structures while they are empty affects the serialisation of the request in such a way that the DataDog API rejects the values as being of the wrong type.

### DIFF
--- a/datadog/resource_datadog_integration_aws.go
+++ b/datadog/resource_datadog_integration_aws.go
@@ -102,6 +102,7 @@ func resourceDatadogIntegrationAwsPrepareCreateRequest(d *schema.ResourceData, a
 		for _, s := range attr.([]interface{}) {
 			filterTags = append(filterTags, s.(string))
 		}
+		iaws.SetFilterTags(filterTags)
 	}
 
 	var hostTags []string
@@ -110,6 +111,7 @@ func resourceDatadogIntegrationAwsPrepareCreateRequest(d *schema.ResourceData, a
 		for _, s := range attr.([]interface{}) {
 			hostTags = append(hostTags, s.(string))
 		}
+		iaws.SetHostTags(hostTags)
 	}
 
 	accountSpecificNamespaceRules := make(map[string]bool)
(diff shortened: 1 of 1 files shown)


miguelaferreira:
Thanks for taking the time @gzussa. Do you know when can we expect a new release?

PR #506, opened 2020-05-06
updating go-datadog-api v2.29.0, fix int issue dash id

sepulworld:
Fixes https://github.com/terraform-providers/terraform-provider-datadog/issues/459


sepulworld:
@gzussa could you cut a new release so we can use the provider with the bug fix? Thanks in advance

PR #511, opened 2020-05-11
Update User Agent

gzussa:


datadog/provider.go
```diff
@@ -111,7 +111,12 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	c := cleanhttp.DefaultClient()
 	c.Transport = logging.NewTransport("Datadog", c.Transport)
-	communityClient.ExtraHeader["User-Agent"] = fmt.Sprintf("terraform-provider-datadog/%s (go %s; terraform %s; terraform-cli %s)", version.ProviderVersion, runtime.Version(), meta.SDKVersionString(), datadogProvider.TerraformVersion)
+	communityClient.ExtraHeader["User-Agent"] = fmt.Sprintf(
```
nmuesch, line 114: Could we use `getUserAgent` and just pass "" as the last argument?

datadog/provider.go
```diff
@@ -228,5 +228,10 @@ func translateClientError(err error, msg string) error {
 }
 
 func getUserAgent(clientUserAgent string) string {
-	return fmt.Sprintf("terraform-provider-datadog/%s (terraform %s) %s", version.ProviderVersion, meta.SDKVersionString(), clientUserAgent)
+	return fmt.Sprintf("terraform-provider-datadog/%s (go %s; terraform %s; terraform-cli %s) %s",
+		version.ProviderVersion,
+		runtime.Version(),
```
nmuesch, line 233: This one may appear twice (as I believe its defined inside the datadogV1 user agent header, but I don't think it'll be an issue

PR #512, opened 2020-05-11
Upgrade Latest Go Client - UserAgent Header adjustment

gzussa:


datadog/provider.go
```diff
@@ -111,7 +110,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	c := cleanhttp.DefaultClient()
 	c.Transport = logging.NewTransport("Datadog", c.Transport)
-	communityClient.ExtraHeader["User-Agent"] = getUserAgent("")
+	communityClient.ExtraHeader["User-Agent"] = getUserAgent("go-datadog-api")
```
jirikuncar, line 113: maybe we could manually add the version since you are pinning it in `GNUmakefile`

GNUmakefile
```diff
@@ -3,7 +3,7 @@ GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=datadog
 DIR=~/.terraform.d/plugins
-ZORKIAN_VERSION=master
+ZORKIAN_VERSION=v2.29.0
```
jirikuncar, line 6:
```suggestion
# Please update `User-Agent` value in `datadog/provider.go` when you modify the value bellow:
ZORKIAN_VERSION=v2.29.0
```

datadog/resource_datadog_monitor.go
```diff
@@ -524,6 +639,7 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 
 	// Similarly, if the silenced attribute is -1, lets unmute those scopes
 	if len(unmutedScopes) != 0 {
+		m.Options.SetSilenced(monitorResp.Options.GetSilenced())
```
nmuesch, line 642: Same, why set this here?

datadog/resource_datadog_monitor.go
```diff
@@ -503,13 +617,14 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 	// they're "drift")
 	unmutedScopes := getUnmutedScopes(d)
 	if newSilenced, ok := d.GetOk("silenced"); ok && !silenced {
+		m.Options.SetSilenced(monitorResp.Options.GetSilenced())
```
nmuesch, line 620: Don't we set this below?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -151,12 +151,110 @@ func ValidateServiceLevelObjectiveTypeString(v interface{}, k string) (ws []stri
...
+				if query, ok := rawQuery.(map[string]interface{}); ok {
+					queries = append(queries, query)
+				}
+			}
+			if len(queries) >= 1 {
+				// only use the first defined query
+				slo.SetQuery(*datadogV1.NewServiceLevelObjectiveQuery(
+					queries[0]["denominator"].(string),
+					queries[0]["numerator"].(string)))
+			}
```
nmuesch, line 195: I may not fully understand, but since query is defined in the schema as only allowing a MaxItems of `1`, couldn't we just setQuery as d.query and avoid this for loop ?

datadog/resource_datadog_service_level_objective.go
```diff
@@ -151,12 +151,110 @@ func ValidateServiceLevelObjectiveTypeString(v interface{}, k string) (ws []stri
...
+
+			if targetDisplayValue, ok := d.GetOk(prefix + "target_display"); ok {
+				if s, ok := targetDisplayValue.(string); ok && strings.TrimSpace(s) != "" {
+					t.SetTargetDisplay(strings.TrimSpace(targetDisplayValue.(string)))
+				}
+			}
+
+			if warningDisplayValue, ok := d.GetOk(prefix + "warning_display"); ok {
+				if s, ok := warningDisplayValue.(string); ok && strings.TrimSpace(s) != "" {
+					t.SetWarningDisplay(strings.TrimSpace(warningDisplayValue.(string)))
```
nmuesch, line 238: Can be later/not as part of this PR, but I wonder if a StateFunc or ValidateFunc on these to trim spaces would help here - terraform.io/docs/extend/schemas/schema-behaviors.html#statefunc

datadog/provider.go
jirikuncar, line 113:
```suggestion
	communityClient.ExtraHeader["User-Agent"] = getUserAgent(fmt.Sprintf("go-datadog-api/2.29.0+ (go %s)", runtime.Version()))
```

datadog/resource_datadog_monitor.go
```diff
@@ -466,8 +466,8 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 	datadogClientV1 := providerConf.DatadogClientV1
 	authV1 := providerConf.AuthV1
 
-	m := buildMonitorStruct(d)
-
+	monitor := buildMonitorStruct(d)
+	m := datadogV1.MonitorUpdateRequest(*monitor)
```
zippolyte, line 470: let's not try to convert type here, and do like for SLO with `buildMonitorStruct` return two types

gzussa: Oups good catch!! Thx

PR #514, opened 2020-05-12
Add default to deprecated `count` field to avoid sending 0

zippolyte:
The count field is set to 0 by terraform if not set, which prevents the widget from getting any data.
Since this field is marked as deprecated, it is not expected to be set, so having a default makes sense.


terrificsoysauce:
I see that the default value is `50`. Does it mean that if the monitor summary has 50+ monitors, it won't display the overage?

bkabrda:
@terrificsoysauce so the widget has two sections - "Counts" and "List" (you can display them both or just one of them). The "Counts" section doesn't get affected by this setting, but the "List" does - it will indeed only display at max 50 monitors (using the default) even if there are more.

PR #517, opened 2020-05-22
Add column revamp properties to dashboard log stream

Prepager:
This PR adds four new properties to the dashboard log stream schema originating from the column revamp: `show_date_column`, `show_message_column`, `message_display`, and `sort`.

From: ~~https://github.com/zorkian/go-datadog-api/pull/321~~ https://github.com/DataDog/datadog-api-client-go/commit/4f430b217fb0d2fb33c887fe101362a9d30d7037


datadog/resource_datadog_dashboard.go
```diff
@@ -2398,6 +2398,37 @@ func getLogStreamDefinitionSchema() map[string]*schema.Schema {
...
+					break
+				default:
+					errs = append(errs, fmt.Errorf(
+						"%q contains an invalid value %q. Valid values are `inline`, `expanded-md`, or `expanded-lg`", key, value))
+				}
+				return
+			},
+		},
+		"sort": {
+			Type:     schema.TypeMap,
```
gzussa, line 2426: Let's convert this to a `TypeList` since using the `Elem` block to define specific keys for the map is currently not possible. See https://www.terraform.io/docs/extend/schemas/schema-types.html#typemap. You can use a TypeList with a maxItem of 1 instead.

nmuesch:
Thanks for the work on this @Prepager!! Could you also add these fields to the docs page here - https://github.com/terraform-providers/terraform-provider-datadog/blob/master/website/docs/r/dashboard.html.markdown under `log_stream_definition` and the example config?

PR #540, opened 2020-06-15
Sort datadog_downtime monitor tags to avoid plan diffs

jurajseffer:
Once an unsorted list of `monitor_tags` is applied for `datadog_downtime`, Terraform shows a permanent diff because Datadog API responds with an ordered list.

This PR builds on https://github.com/terraform-providers/terraform-provider-datadog/pull/214 by changing the type from list to set and applying ordering. I've extended the tests with additional values and replayed cassettes. The fix appears to work using a binary for my use case.


datadog/resource_datadog_downtime.go
```diff
@@ -135,7 +136,10 @@ func resourceDatadogDowntime() *schema.Resource {
 				Description:   "When specified, this downtime will only apply to this monitor",
 			},
 			"monitor_tags": {
-				Type:        schema.TypeList,
```
gzussa, line 138: Could we make this work without changing the type here. I fear it won't be backward compatible otherwise. What do you think?

datadog/resource_datadog_downtime_test.go
```diff
@@ -130,7 +134,7 @@ func TestAccDatadogDowntime_BasicMultiScope(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"datadog_downtime.foo", "message", "Example Datadog downtime message."),
 					resource.TestCheckResourceAttr(
-						"datadog_downtime.foo", "monitor_tags.0", "*"),
+						"datadog_downtime.foo", "monitor_tags.2679715827", "*"),
```
gzussa, line 137: I am not sure i understand these changes here. Why do we have to change assertions?

datadog/resource_datadog_downtime.go
jurajseffer: I only have enough experience with Golan to go by what `datadog_monitor` resource does - https://github.com/terraform-providers/terraform-provider-datadog/commit/6552bcee02a611b4b36e8f7dcdbc1622838bb921#diff-f313670eee4640665f3fee01de4c6ce7R166

It states that set is used because of the ordering problem. After I added all the sorting into this resource it still wasn't working correctly and tests were failing (it was still being reordered by TF). Making it a set fixed that. I assume there is some internal logic terraform that interferes.

datadog/resource_datadog_downtime_test.go
jurajseffer: Since it's not a list anymore, the keys aren't zero indexed and these values are hashes of the values. See https://github.com/terraform-providers/terraform-provider-datadog/commit/6552bcee02a611b4b36e8f7dcdbc1622838bb921#diff-01173e30cd220a18166d845ecb9bcd11R106

datadog/resource_datadog_downtime.go
```diff
@@ -272,9 +276,10 @@ func buildDowntimeStruct(authV1 context.Context, d *schema.ResourceData, client
 	}
 	dt.SetScope(scope)
 	var tags []string
-	for _, mt := range d.Get("monitor_tags").([]interface{}) {
+	for _, mt := range d.Get("monitor_tags").(*schema.Set).List() {
 		tags = append(tags, mt.(string))
 	}
+	sort.Strings(tags)
```
jirikuncar, line 282: do we still need to sort them if we are using `schema.TypeSet` [1] ?

--
[1] https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset

datadog/resource_datadog_downtime.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"sort"
```
jirikuncar, line 8:
```suggestion
```

datadog/resource_datadog_downtime.go
jirikuncar, line 282:
```suggestion
```

datadog/resource_datadog_downtime.go
```diff
@@ -403,7 +408,9 @@ func resourceDatadogDowntimeRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("scope", dt.Scope)
 	// See the comment for monitor_tags in the schema definition above
 	if !reflect.DeepEqual(dt.GetMonitorTags(), []string{"*"}) {
-		d.Set("monitor_tags", dt.GetMonitorTags())
+		tags := dt.GetMonitorTags()
+		sort.Strings(tags)
+		d.Set("monitor_tags", tags)
```
jirikuncar, line 413:
```suggestion
		d.Set("monitor_tags", dt.GetMonitorTags())
```

jirikuncar:
Addresses #577

jurajseffer:
Thanks for cleaning it up @jirikuncar and merging :1st_place_medal:

PR #541, opened 2020-06-16
Allow to pass http client when creating test provider

txominpelu:
This change will allow to mock the tests using gock since we will be able to pass the DefaultHttpClient that is used by default by gock.

### DIFF
--- a/datadog/provider_test.go
+++ b/datadog/provider_test.go
@@ -100,7 +100,7 @@ func removeURLSecrets(u *url.URL) *url.URL {
 	return u
 }
 
-func initAccProvider(t *testing.T) (*schema.Provider, func(t *testing.T)) {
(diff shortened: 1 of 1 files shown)


datadog/provider_test.go
```diff
@@ -245,7 +255,10 @@ func testAccProvider(t *testing.T, accProviders map[string]terraform.ResourcePro
 }
 
 func TestProvider(t *testing.T) {
-	accProvider, cleanup := initAccProvider(t)
+	rec := initRecorder(t)
+	c := cleanhttp.DefaultClient()
+	c.Transport = logging.NewTransport("Datadog", rec)
+	accProvider, cleanup := initAccProvider(t, c, rec)
```
jirikuncar, line 261: Can it stay?
```suggestion
	accProvider, cleanup := initAccProvider(t)
```

Issue #564, opened 2020-06-30
datadog_integration_pagerduty breaks expected terraform contract, allows for creating same resource twice 


geota:
### Terraform Version
```
$ terraform -v
terraform v0.12.26
```

### Affected Resource(s)
Please list the resources as a list, for example:
- datadog_integration_pagerduty

### Terraform Configuration Files
test/module_a/main.tf

```
resource "datadog_integration_pagerduty" "pd" {
  individual_services = true
  subdomain           = "test-geota"
}
```

test/module_b/main.tf
```
resource "datadog_integration_pagerduty" "pd" {
  individual_services = true
  subdomain           = "test-geota"
}
```

### Debug Output

Module a
```
❯ terraform apply

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # datadog_integration_pagerduty.pd will be created
  + resource "datadog_integration_pagerduty" "pd" {
      + id                  = (known after apply)
      + individual_services = true
      + subdomain           = "test-geota"
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

datadog_integration_pagerduty.pd: Creating...
datadog_integration_pagerduty.pd: Still creating... [10s elapsed]
datadog_integration_pagerduty.pd: Creation complete after 11s [id=test-geota]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

Module b
```
❯ terraform apply

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # datadog_integration_pagerduty.pd will be created
  + resource "datadog_integration_pagerduty" "pd" {
      + id                  = (known after apply)
      + individual_services = true
      + subdomain           = "test-geota"
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

datadog_integration_pagerduty.pd: Creating...
datadog_integration_pagerduty.pd: Creation complete after 7s [id=test-geota]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```
### Expected Behavior
[...]

PR #565, opened 2020-07-01
existing PD integration should not be deleted on create #564

skang0601:
Addressed #564

Added a new optional field in the datadog_pagerduty_integration resource the maintain backwards compatibility with the old behavior but allows users to also prevent the resource from deleting the existing PagerDuty integration silently.

This seems like something that the API should handle by returning an appropriate error status code, but the provider should also follow conventions/expectations for Terraform users regarding resource creation/deletion.

### DIFF
--- a/datadog/resource_datadog_integration_pagerduty.go
+++ b/datadog/resource_datadog_integration_pagerduty.go
@@ -1,6 +1,7 @@
 package datadog
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -63,6 +64,12 @@ func resourceDatadogIntegrationPagerduty() *schema.Resource {
 				Optional:  true,
 				Sensitive: true,
 			},
+			"preserve_existing_integration": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Set to true to preserve any existing Pager Duty Integrations if they exist",
+			},
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


skang0601:
Looks like the test changes will require me to re-record the cassettes since the code change will add an additional GET request to the integration endpoint. I"ll do so later with a non-production organization, same for any new test cases

datadog/resource_datadog_integration_pagerduty.go
```diff
@@ -150,7 +161,6 @@ func resourceDatadogIntegrationPagerdutyRead(d *schema.ResourceData, meta interf
 	d.Set("services", services)
 	d.Set("subdomain", pd.GetSubdomain())
 	d.Set("schedules", pd.Schedules)
-
 	return nil
```
jirikuncar, line 164:
```suggestion

	return nil
```

datadog/resource_datadog_integration_pagerduty.go
```diff
@@ -112,6 +119,10 @@ func resourceDatadogIntegrationPagerdutyCreate(d *schema.ResourceData, meta inte
 		return fmt.Errorf("failed to parse resource configuration: %s", err.Error())
 	}
 
+	if exists, _ := resourceDatadogIntegrationPagerdutyExists(d, meta); exists && d.Get("preserve_existing_integration").(bool) {
+		return translateClientError(errors.New("PagerDuty integration already exists"), "Please import into your existing state or remove from your HCL.")
```
jirikuncar, line 123: @skang0601 can you please add new test to cover this branch? Thank you!

therve:
We discussed another approach, thanks for your patch.

Issue #566, opened 2020-07-01
Synthetics JSON assertions is not handle in the terraform provider. (datadog_synthetics_test, operator = `validatesJSONPath`)


chussenot:
Hello,

It seems that the operator `validatesJSONPath` is not available in the provider.
Have you planned to managed it?
Cheers,

```
Error: datadog_synthetics_test.sonarq (import id: 8ep-2ta-ss2): 1 error occurred:
        * import datadog_synthetics_test.sonarq result: 8ep-2ta-ss2: datadog_synthetics_test.sonarq: error getting synthetics test:  is not a valid SyntheticsAssertionOperator: 
```
```JSON
{"status":"live","public_id":"8ep-2ta-ss2","tags":["slo_type:availability","type:synthetic","site:https://sonarqube.domain.tld","owner:sre-delivery","project:sonarqube","service:sonarqube","env:prd"],"locations":["pl:private-ecs-support-s678vnjh"],"message":"sonarqube.domain.tld is down","name":"Test on https://sonarqube.domain.tld","monitor_id":188618,"type":"api","subtype":"http","config":{"variables":[],"request":{"url":"https://sonarqube.domain.tld/api/system/status","method":"GET","timeout":30},"assertions":[{"operator":"is","type":"statusCode","target":200},{"operator":"validatesJSONPath","type":"body","target":{"operator":"contains","targetValue":"UP","jsonPath":"$.status"}}]},"options":{"min_failure_duration":0,"min_location_failed":1,"tick_every":60}}

```

```
$ tf --version
Terraform v0.11.14
+ provider.datadog v2.11.0

Your version of Terraform is out of date! The latest version
is 0.12.28. You can update by downloading from www.terraform.io/downloads.html

$ tree .terraform  
.terraform
└── plugins
    └── linux_amd64
        ├── lock.json
        └── terraform-provider-datadog_v2.11.0_x4
```

### Affected Resource(s)

- datadog_synthetics_test

:warning:  This type is missing on the available Synthetics Assertion Operators, on the datadog go client.
https://github.com/DataDog/datadog-api-client-go/issues/370

Art3mK:
yeah, I'm facing the same issue, using terraform 0.12.28 and datadog provider v2.11

```
[terragrunt] 2020/07/06 11:12:29 Running command: terraform import datadog_synthetics_test.synth["account.example.com health status"] kwr-ema-ar5
[...]

PR #571, opened 2020-07-08
Update go client, use new synthetics target

therve:
This updates the go client to the latest version to be able to use the new
SyntheticsTest model, in particular the validatesJSONPath operator.

Closes ##545, #566


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -330,33 +378,81 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
...
+					}
+					if v, ok := assertionMap["target"]; ok {
+						if isTargetOfTypeInt(assertionTarget.GetType()) {
+							assertionTargetInt, _ := strconv.Atoi(v.(string))
+							assertionTarget.SetTarget(assertionTargetInt)
+						} else if assertionTarget.GetOperator() == datadogV1.SYNTHETICSASSERTIONOPERATOR_VALIDATES {
+							assertionTarget.SetTarget(v.(string))
+						} else {
+							assertionTarget.SetTarget(v.(string))
+						}
```
nmuesch, line 402: Should this just be one `else`?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -330,33 +378,81 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
...
 	config.SetVariables([]datadogV1.SyntheticsBrowserVariable{})
 
+	// Deprecated path
 	if attr, ok := d.GetOk("assertions"); ok && attr != nil {
 		for _, attr := range attr.([]interface{}) {
-			assertion := datadogV1.SyntheticsAssertion{}
 			assertionMap := attr.(map[string]interface{})
 			if v, ok := assertionMap["type"]; ok {
 				assertionType := v.(string)
-				assertion.SetType(datadogV1.SyntheticsAssertionType(assertionType))
```
nmuesch, line 339: Any reason to change this logic if we're deprecating it?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -330,33 +378,81 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
 	config := datadogV1.NewSyntheticsTestConfig([]datadogV1.SyntheticsAssertion{}, *request)
 	config.SetVariables([]datadogV1.SyntheticsBrowserVariable{})
 
+	// Deprecated path
```
nmuesch, line 381:
```suggestion
	// Deprecated path, the assertions field is replaced with assertion
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -466,20 +562,48 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
 	d.Set("request_headers", actualRequest.Headers)
 
 	actualAssertions := syntheticsTest.GetConfig().Assertions
-	var localAssertions []map[string]string
```
nmuesch, line 469: Do we need to keep any of the existing logic for users that still use `assertions` for a bit?

datadog/resource_datadog_synthetics_test_.go
therve: Yeah most likely.

datadog/resource_datadog_synthetics_test_.go
therve: That's because of the api client change I believe.

therve:
Imports don't work properly just yet.

datadog/resource_datadog_synthetics_test_.go
therve: Yeah I think I found what was needed there.

therve:
OK I believe I found a correct path forward. The trick is to keep using the old field if it's referenced in the current state, otherwise use the new field everywhere. AFAIU it just a bit weird if you use import with the old schema.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -330,33 +376,77 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
...
-				} else if assertion.GetOperator() == datadogV1.SYNTHETICSASSERTIONOPERATOR_VALIDATES {
-					assertion.SetTarget(v.(string))
-				} else {
-					assertion.SetTarget(v.(string))
+		}
+	}
+
+	if attr, ok := d.GetOk("assertion"); ok && attr != nil {
+		for _, attr := range attr.([]interface{}) {
+			assertionMap := attr.(map[string]interface{})
```
jirikuncar, line 408: Can you use different variable name for `attr` used in range?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -49,12 +49,60 @@ func resourceDatadogSyntheticsTest() *schema.Resource {
...
+						},
+						"property": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"target": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"targetjsonpath": {
```
nmuesch, line 82: Should we have this conflictsWith `target`? They seem mutually exclusive

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -330,33 +376,77 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
...
-				assertion.SetType(datadogV1.SyntheticsAssertionType(assertionType))
-			}
-			if v, ok := assertionMap["property"]; ok {
-				assertionProperty := v.(string)
-				assertion.SetProperty(assertionProperty)
-			}
-			if v, ok := assertionMap["operator"]; ok {
-				assertionOperator := v.(string)
-				assertion.SetOperator(datadogV1.SyntheticsAssertionOperator(assertionOperator))
+				if v, ok := assertionMap["operator"]; ok {
```
nmuesch, line 385: It looks like this code path is pretty much the same as before, but some is further indented now. From the schema aren't `type` and `operator` siblings? Do the `operator` and `target` checks need to happen inside the `type` check?

datadog/resource_datadog_synthetics_test_test.go
```diff
@@ -49,6 +73,8 @@ func TestAccDatadogSyntheticsSSLTest_importBasic(t *testing.T) {
 				ResourceName:      "datadog_synthetics_test.ssl",
 				ImportState:       true,
 				ImportStateVerify: true,
+				// Assertions will be imported into the new schema by default, but we can ignore them as users need to update the local config in this case
+				ImportStateVerifyIgnore: []string{"assertions", "assertion"},
```
nmuesch, line 77: Is the issue here that on import we will have both `assertions` and `assertion` in the state? Is this going to cause an unempty diff on future `apply`s? 

Actually, re-reading some of the code above, it looks like you check for the config before saving this field, in `updateSyntheticsTestLocalState`, so is this ignore still a requirement?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -466,24 +556,61 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
...
+			if target := assertionTarget.GetTarget(); target != nil {
+				localAssertion["target"] = convertToString(target)
+			}
+			if v, ok := assertionTarget.GetTypeOk(); ok {
+				localAssertion["type"] = string(*v)
+			}
 		}
-		if assertion.HasProperty() {
-			localAssertion["property"] = assertion.GetProperty()
+		if assertion.SyntheticsAssertionJSONPathTarget != nil {
```
nmuesch, line 577: Maybe a nit, but if these are mutually exclusive, would it make sense to make this an elif?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -466,24 +556,61 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
...
+			}
+			if v, ok := assertionTarget.GetTypeOk(); ok {
+				localAssertion["type"] = string(*v)
+			}
 		}
-		if target := assertion.GetTarget(); target != nil {
-			localAssertion["target"] = convertToString(target)
+		localAssertions[i] = localAssertion
+	}
+	// If the config still uses assertions, keep using that in the state to not generate useless diffs
```
nmuesch, line 604: I'd have to test around with this a bit more, but I'm not sure if `d` is always from the config. I think depending on which lifecycle piece we're in, `d` can reflect the object from the state instead.

therve:
> @therve can you bump the version of datadog-api-client-go in separate PR?

It needs a code change so I'd rather merge that one if we're close.

datadog/resource_datadog_synthetics_test_.go
therve: Good idea.

datadog/resource_datadog_synthetics_test_.go
therve: They are now required, so it's indented because we only create the structure when we already have the previous fields.

datadog/resource_datadog_synthetics_test_test.go
therve: I check the config before saving the field, but it only works if I already have state. IIUC, the `ImportStateVerify` compares to a plain import, which doesn't compare to anything locally.

datadog/resource_datadog_synthetics_test_.go
therve: Yeah it mostly comes from the state, but that's what I care about here.

datadog/resource_datadog_synthetics_test_.go
jirikuncar: Can we try to bump the SDK version to https://github.com/hashicorp/terraform-plugin-sdk/releases/tag/v1.15.0

PR #576, opened 2020-07-13
[Synthetics] Add support for browser test steps

romainberger:
Add support for steps for Synthetics browser tests

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -599,8 +599,6 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
 			// todo params
 			localSteps = append(localSteps, localStep)
 		}
-		// fmt.Println("####################################### ADD STEPS")
-		// fmt.Println(len(localSteps))
 		d.Set("steps", localSteps)
 	}
 
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -90,6 +90,7 @@ func resourceDatadogSyntheticsTest() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"step": syntheticsTestStep(),
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -516,6 +577,31 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
 	d.Set("message", syntheticsTest.GetMessage())
 	d.Set("status", syntheticsTest.GetStatus())
 	d.Set("tags", syntheticsTest.Tags)
+
+	if syntheticsTest.GetType() == "browser" {
+		// d.Set("steps", syntheticsTest.GetSteps())
+
```
jirikuncar, line 583:
```suggestion
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -216,6 +217,44 @@ func syntheticsTestOptions() *schema.Schema {
 	}
 }
 
+func syntheticsTestStep() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
```
romainberger, line 222: This uses a `TypeSet` because a `TypeMap` wouldn't fit (the `params` field is a map), but we need to use a `position` field as steps order is important for browser tests.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -433,6 +472,28 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
 		}
 	}
 
+	if attr, ok := d.GetOk("step"); ok && syntheticsTest.GetType() == "browser" {
+		var nbSteps = len(attr.(*schema.Set).List())
+		steps := make([]datadogV1.SyntheticsStep, nbSteps)
+
+		fmt.Println()
```
jirikuncar, line 479:
```suggestion
```

datadog/resource_datadog_synthetics_test_.go
jirikuncar: Can we use `TypeList` to avoid the need for `position` field?

romainberger: Unless I'm mistaken, we can't use a `TypeList` here because one of the field is a `TypeMap` (`params`). So the solution would be to use a `TypeSet`, but with a `position` field since the order is important.

therve: I think you're mistaken. Nothing in `TypeList` prevents you from having a map in it.

romainberger: Sorry you're right I was confused. We can indeed use maps in a list, but the issue is the `params` field is a map, and as far as I know we can't have nested maps. So instead of a list of maps with a map inside, I use sets, so we can have a map inside, but we need a `position` field to make sure order is kept.

romainberger:
Closing to avoid too many conflicts, see https://github.com/DataDog/terraform-provider-datadog/pull/638 for new PR

PR #587, opened 2020-07-21
Fix pagerduty integration tests

therve:
With the removal of the delete tests were broken.

### DIFF
--- a/datadog/resource_datadog_integration_pagerduty_test.go
+++ b/datadog/resource_datadog_integration_pagerduty_test.go
@@ -2,7 +2,6 @@ package datadog
 
 import (
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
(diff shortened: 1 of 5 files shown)


datadog/resource_datadog_integration_pagerduty_test.go
```diff
@@ -149,19 +148,8 @@ func datadogIntegrationPagerdutyExistsHelper(client *datadog.Client) error {
 
 func testAccCheckDatadogIntegrationPagerdutyDestroy(accProvider *schema.Provider) func(*terraform.State) error {
```
nmuesch, line 149: Could we just remove this method and the CheckDestroy that uses them?

jirikuncar: It's related to https://github.com/terraform-providers/terraform-provider-datadog/pull/584/files#diff-f2ab83c1cfbd1f73e2279cdc962210a0L209-L217

I missed that part.

therve: It seems I need to keep it, it fails weirdly without it.

PR #591, opened 2020-07-22
Add file with licenses and copyrights of deps

bkabrda:


bkabrda:
Thanks for the reviews! Merging.

PR #595, opened 2020-07-23
Setup unit-test CI using GH actions, remove Travis configuration

bkabrda:


.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ${{ matrix.platform }}
+    steps:
+    # if we don't do this, `gofmt` will want to rewrite all Go files due to bad line endings,
+    # because Git will convert all line endings to CRLF when cloning on windows
+    - name: Set Git to use Linux-style line endings
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
+    - name: Install Go
+      uses: actions/setup-go@v2-beta
```
jirikuncar, line 24: @bkabrda maybe we need to change the `GOPATH` to a place where the website can be cloned.

.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
...
+        go-version: 1.14
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Debug Go env
+      run: go env
+    - name: Vet
+      run: make vet
+    - name: Website test
+      if: matrix.platform == 'ubuntu-latest'
+      run: make website-test GOPATH=`go env GOPATH`
```
jirikuncar, line 33:
```suggestion
      shell: bash
      run: make -e GOPATH=$(go env GOPATH) website-test 
```

bkabrda: Why? It seems to work fine as it is.

jirikuncar: - `bash` is safer to use as the shell might change
- `-e` is GNUMake option for env overrides (I am not sure why the hack with target `NAME=...` works)

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        go-version: 1.14
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Vet
+      run: make vet
+    - name: Website test
+      if: matrix.platform == 'ubuntu-latest'
+      shell: bash
+      run: make -e GOPATH=$(go env GOPATH) website-test
```
jirikuncar, line 34: nit
```suggestion
      run: make -e GOPATH="$(go env GOPATH)" website-test 
```

PR #596, opened 2020-07-23
Add AZP configuration for CI

bkabrda:


.azure-pipelines/all.yml
```diff
@@ -0,0 +1,25 @@
...
+- job: IntegrationTests
+  pool:
+    vmImage: "Ubuntu-16.04"
+  container:
+    image: golang:latest
+  steps:
+    - script: make testacc
+      displayName: Run integration tests
+      env:
+        CI: "true"
```
nmuesch, line 22: Do we also need RECORD=true or does this handle that?

bkabrda: Using `make testacc` actually does the right thing automatically. It runs tests against the real API without re-recording cassettes.

nmuesch: Ahh ok thanks!

bkabrda:
The test suite seems to be pretty flaky, but working on that is not the aim of this PR. I'm going to merge and work on the flakiness next. Thanks for the review!

PR #597, opened 2020-07-24
Ignore diff to start on downtime resource

### LINKED ISSUE #507: start is marked as optional in docs but it seems to be required? (datadog_downtime)
Terraform v0.12.12
+ provider.datadog v2.7.0

I want to mute a monitor indefinitely so I do this:

```
resource "datadog_downtime" "test" {
  scope = ["*"]
  monitor_id = datadog_monitor.testsdfsdfds.id
  // Mute now and forever
  active = true
}
```

In never comes into state, every time I apply I see this:

```
  ~ resource "datadog_downtime" "test" {
        active     = true
        disabled   = false
        end        = 0
        id         = "123456"
        monitor_id = 212312312
        scope      = [
            "*",
        ]
      - start      = 1588945921 -> null
        timezone   = "UTC"
    }
```

While I can do this, its strange. Requiring ignore of a property as part of normal behaviour I think is poor form, or should I say poor... terra-form HA!
```
  lifecycle {
    ignore_changes = ["start"]
  }
```

I would characterize this as a bug. If active is set with the absence of start, start should just not be checked at all. Maybe Im missing something here

therve:
If the start attributes generates a diff from a value to nothing, it
means that the attribute was never set to being with. Let's ignore the
diff in that case.

Closes #507


datadog/resource_datadog_downtime.go
```diff
@@ -46,7 +46,7 @@ func resourceDatadogDowntime() *schema.Resource {
 				Optional: true,
 				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
 					_, startDatePresent := d.GetOk("start_date")
-					return startDatePresent
+					return startDatePresent || (newVal == "0" && oldVal != "0")
```
jirikuncar, line 49: @therve can you add examples when the diff should (not) be supressed?

jirikuncar:
```suggestion
					// "start_date" is set or was previously set (oldVal != "0") and it is being removed (newVal == "0")
					return startDatePresent || (oldVal != "0" && newVal == "0")
```

therve: I added a comment, let me know if it makes sense.

PR #599, opened 2020-07-27
Retry GET on dashboard create

therve:
Reading dashboard just after create can return a 404 for things to sync
up. Let's retry a bit in that case to make things more reliable.

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -98,8 +98,8 @@ func resourceDatadogDashboardCreate(d *schema.ResourceData, meta interface{}) er
 	d.SetId(*dashboard.Id)
 
 	return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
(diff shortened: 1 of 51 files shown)


datadog/resource_datadog_dashboard.go
```diff
@@ -95,7 +96,16 @@ func resourceDatadogDashboardCreate(d *schema.ResourceData, meta interface{}) er
 		return translateClientError(err, "error creating dashboard")
 	}
 	d.SetId(*dashboard.Id)
-	return resourceDatadogDashboardRead(d, meta)
+
+	return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		if _, _, err := datadogClientV1.DashboardsApi.GetDashboard(authV1, *dashboard.Id).Execute(); err != nil {
+			if strings.Contains(err.Error(), "404 Not Found") {
```
jirikuncar, line 102:
```suggestion
		if _, httpresp, err := datadogClientV1.DashboardsApi.GetDashboard(authV1, *dashboard.Id).Execute(); err != nil {
		        # TODO check that err is APIError
			if httpresp.Status == 404 {
```

PR #606, opened 2020-07-29
Send APM traces

jirikuncar:
Send traces using `dd-trace-go` and DD agent in Azure pipelines.


datadog/provider_test.go
```diff
@@ -134,17 +136,36 @@ func initRecorder(t *testing.T) *recorder.Recorder {
...
+	t.Helper()
+	span, ctx := tracer.StartSpanFromContext(
+		ctx,
+		"test",
+		tracer.SpanType("test"),
+		tracer.ResourceName(t.Name()),
+		tracer.Tag(ext.AnalyticsEvent, true),
+		tracer.Measured(),
+	)
+	// span.SetTag("version", tag)
```
jirikuncar, line 149: @bkabrda 👀  🏷️

bkabrda: I would suggest doing the same thing as we do in https://github.com/DataDog/datadog-api-client-go/blob/master/tests/test_utils.go#L203

go.sum
```diff
@@ -58,6 +58,8 @@ github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.m
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
+github.com/fatih/color v1.9.0 h1:8xPHl4/q1VyqGIPif1F+1V3Y3lSmrq01EabUW3CoW5s=
```
bkabrda, line 61: Please also update `LICENSE-3rdparty.csv` with new dependencies that appeared in `go.sum`.

jirikuncar: I have added dd-trace-go.

PR #611, opened 2020-07-31
Fix DiffSuppressFunc for downtime start

bkabrda:
We want to ignore the diff even if start is set "right now". This fixes a test flake like [this](https://dev.azure.com/datadoghq/terraform-provider-datadog/_build/results?buildId=20841&view=logs&j=abc39910-4c32-5be0-0b39-6d3f303c7218&t=632788a2-3bb3-5d1a-83fe-bfd285143df8&l=573).


bkabrda:
There are more flakes unrelated to this PR that I'll fix in other PRs. This seems to be working fine, so merging.

PR #619, opened 2020-08-03
[Synthetics] Add retry to synthetics options

romainberger:
Add support for fast retry options to synthetics tests.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -253,6 +253,14 @@ func syntheticsTestOptions() *schema.Schema {
...
 				}
 			}
+			// retryRaw, ok := val.(map[string]interface{})["retry"]
+			// if ok {
+			// 	retryInt := int
+			// }
+			// if v < 0 || v > 3 {
+			// 	errs = append(errs, fmt.Errorf("%q must be between 0 and 3 inclusive, got: %d", key, v))
+			// }
+			// return
```
therve, line 263: Probably should remove that? Do you need validation on the number of retries.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -523,6 +535,13 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
 		allowInsecure, _ := strconv.ParseBool(attr.(string))
 		options.SetAllowInsecure(allowInsecure)
 	}
+	if attr, ok := d.GetOk("options.retry"); ok {
+		retryCount, _ := strconv.Atoi(attr.(string))
+		retry := datadogV1.SyntheticsTestOptionsRetry{}
+		retry.SetCount(int64(retryCount))
+		retry.SetInterval(300)
```
therve, line 542: Should we add support for retry interval right now? It's going to be awkward to add it later on if retry is a int instead of a struct.

datadog/resource_datadog_synthetics_test_.go
romainberger: Yes I removed it, It might be annoying to keep this in sync with the api if it changes

datadog/resource_datadog_synthetics_test_.go
romainberger: You're right, I'll change that to a struct 👍

romainberger:
To keep this light I've changed and only added the retry options, I'll make a separate PR for the renotify options.

docs/resources/synthetics.md
```diff
@@ -141,6 +141,8 @@ The following arguments are supported:
   - `min_location_failed` - (Optional) Threshold below which a synthetics test is allowed to fail before sending notifications
   - `accept_self_signed` - (Optional) For type=ssl, true or false
   - `allow_insecure` - (Optional) For type=api, true or false. Allow your HTTP test go on with connection even if there is an error when validating the certificate.
+  - `retry_count` - (Optional) Number of retries needed to consider a location as failed before sending a notification alert.
+  - `retry_interval` - (Optional) Interval between a failed test and the next retry.
```
therve, line 145: Can you put the unit there? I believe it's milliseconds?

PR #629, opened 2020-08-07
Support multi_compute field in log_query

### LINKED ISSUE #628: [Feature Request] Support multi_compute in query_table_definition
It would be good to support multi_compute in widgets that do support it. To my knowledge, at least the query_table widget supports it.
The API returns

```
                    "multi_compute": [
                      {
                        "facet": "@timestamp",
                        "aggregation": "cardinality"
                      },
                      {
                        "aggregation": "count"
                      }
                    ]
```

as a property of the log_query field.

therve:
This adds support for the `multi_compute` field alternative to the
`compute` one.

Closes #628


datadog/resource_datadog_dashboard.go
```diff
@@ -4713,7 +4713,27 @@ func getApmLogNetworkRumSecurityQuerySchema() *schema.Schema {
 				},
 				"compute": {
```
bkabrda, line 4714: Since `compute` and `multi_compute` are mutually exclusive (as per the docs below), should we mark them with `ConflictsWith`? (see it's doc under https://godoc.org/github.com/hashicorp/terraform-plugin-sdk/helper/schema#Schema)

datadog/resource_datadog_dashboard.go
```diff
@@ -4713,7 +4713,27 @@ func getApmLogNetworkRumSecurityQuerySchema() *schema.Schema {
...
 					Type:     schema.TypeMap,
-					Required: true,
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
```
bkabrda, line 4718: This Schema definition seems the same as the one for `multi_compute` AFAICS. Would it make sense to create a function to return it instead of duplicating it?

datadog/resource_datadog_dashboard.go
bkabrda: Hmm, actually after re-reading this I'm not sure. Are they mutually exclusive or can they coexist?

docs/resources/dashboard.md
```diff
@@ -937,19 +937,23 @@ Nested `widget` `time` blocks have the following structure:
...
 
   - `index` - (Required)
-  - `compute` - (Required). Exactly one nested block is required with the following structure:
+  - `compute` - (Optional). One of `compute` and `multi_compute` is required. The map has the following keys:
     - `aggregation` - (Required)
     - `facet` - (Optional)
     - `interval` - (Optional)
-  - `search` - (Optional). One nested block is allowed with the following structure:
-    - `query` - (Optional)
+  - `multi_compute` - (Optional). One of `compute` and `multi_compute` is required. Multiple nested blocks are allow with the following structure:
```
bkabrda, line 947:
```suggestion
  - `multi_compute` - (Optional). One of `compute` and `multi_compute` is required. Multiple nested blocks are allowed with the following structure:
```

datadog/resource_datadog_dashboard.go
therve: AFAIK you can't use `ConflictsWith` instead lists or maps.

datadog/resource_datadog_dashboard.go
therve: We could. The schema for `compute` is actually useless because it's a `TypeMap`, but it doesn't hurt.

datadog/resource_datadog_dashboard.go
bkabrda: Ah, that's right. So we're basically relying on the API to return the proper error message if both are used, right?

therve: Yeah I added a TODO below, but I need to change 5 different functions to propagate errors :/

PR #630, opened 2020-08-10
Prevent nil pointer dereference in dashboard import

juliogreff:
When importing a dashboard with a template variable with no prefix, the
provider panics with the following error message:

```
panic: runtime error: invalid memory address or nil pointer dereference
2020-08-10T13:42:16.345+0200 [DEBUG] plugin.terraform-provider-datadog_v2.12.1: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1bddcf1]
2020-08-10T13:42:16.345+0200 [DEBUG] plugin.terraform-provider-datadog_v2.12.1:
2020-08-10T13:42:16.345+0200 [DEBUG] plugin.terraform-provider-datadog_v2.12.1: goroutine 81 [running]:                                                                                                     2020-08-10T13:42:16.345+0200 [DEBUG] plugin.terraform-provider-datadog_v2.12.1: github.com/terraform-providers/terraform-provider-datadog/datadog.buildTerraformTemplateVariables(0xc0004a1a18, 0x1fa2ab9)
2020-08-10T13:42:16.345+0200 [DEBUG] plugin.terraform-provider-datadog_v2.12.1:         github.com/terraform-providers/terraform-provider-datadog/datadog/resource_datadog_dashboard.go:296 +0x291
```

This commit prevents by first checking if the pointer that we're going
to dereference is nil. If that's the case, we assume that the template
variable doesn't have a name, prefix or default, as appropriate.

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -308,8 +308,8 @@ func buildTerraformTemplateVariables(datadogTemplateVariables *[]datadogV1.Dashb
 		if v, ok := templateVariable.GetNameOk(); ok {
 			terraformTemplateVariable["name"] = *v
 		}
-		if v, ok := templateVariable.GetPrefixOk(); ok {
-			terraformTemplateVariable["prefix"] = *v
+		if v := templateVariable.GetPrefix(); len(v) > 0 {
+			terraformTemplateVariable["prefix"] = v
 		}
 		if v, ok := templateVariable.GetDefaultOk(); ok {
 			terraformTemplateVariable["default"] = *v


datadog/resource_datadog_dashboard.go
```diff
@@ -305,13 +305,13 @@ func buildTerraformTemplateVariables(datadogTemplateVariables *[]datadogV1.Dashb
 	terraformTemplateVariables := make([]map[string]string, len(*datadogTemplateVariables))
 	for i, templateVariable := range *datadogTemplateVariables {
 		terraformTemplateVariable := map[string]string{}
-		if v, ok := templateVariable.GetNameOk(); ok {
+		if v, ok := templateVariable.GetNameOk(); ok && v != nil {
```
therve, line 308: Isn't prefix the only one where that case could happen?

datadog/resource_datadog_dashboard.go
```diff
@@ -305,13 +305,13 @@ func buildTerraformTemplateVariables(datadogTemplateVariables *[]datadogV1.Dashb
...
 			terraformTemplateVariable["name"] = *v
 		}
-		if v, ok := templateVariable.GetPrefixOk(); ok {
+		if v, ok := templateVariable.GetPrefixOk(); ok && v != nil {
```
therve, line 311:
```suggestion
		if v := templateVariable.GetPrefix(); len(v) > 0 {
```

I think that makes more sense in this case, the Ok doesn't really help us. You need to change the next line assignment though.

datadog/resource_datadog_dashboard.go
juliogreff: In theory, yes, I haven't tried torturing the API a little bit more to see if I could make the other cases also trigger. I don't mind having the check just for the prefix, but I preferred having it more consistent.

datadog/resource_datadog_dashboard.go
```diff
@@ -305,14 +305,14 @@ func buildTerraformTemplateVariables(datadogTemplateVariables *[]datadogV1.Dashb
 	terraformTemplateVariables := make([]map[string]string, len(*datadogTemplateVariables))
 	for i, templateVariable := range *datadogTemplateVariables {
 		terraformTemplateVariable := map[string]string{}
-		if v, ok := templateVariable.GetNameOk(); ok {
-			terraformTemplateVariable["name"] = *v
+		if v := templateVariable.GetName(); len(v) > 0 {
```
therve, line 308: Ah sorry, my suggestion only works on prefix, because it's marked nullable. Other fields don't have the same property.

juliogreff: No worries :) I reverted the other two as they were before then, given that it's unlikely they'll map to a nil pointer anyway.

7oas7er:
Thanks for fixing this. When will it be released in a new version of the provider?

PR #636, opened 2020-08-14
[Synthetics] Add retry and monitor options to options_list

### LINKED ISSUE #376: Renotify_interval doesn't exist in datadog_synthetics_test
Hello.  There is no available option to enable re-notification if the monitor has not been resolved.

### Affected Resource(s)
- datadog_synthetics_test

### Expected Behavior
As in datadog_monitor resource:

> renotify_interval (Optional) The number of minutes after the last notification before a monitor will re-notify on the current status. It will only re-notify if it's not resolved.

romainberger:
Add `retry` and `monitor_options` to `options_list` for Synthetics.

I had to update the go client to be able to use the new monitor options model.

Closes #376


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -335,6 +335,37 @@ func syntheticsTestOptionsList() *schema.Schema {
 					Type:     schema.TypeInt,
 					Optional: true,
 				},
+				"monitor_options": {
+					Type:     schema.TypeMap,
```
therve, line 339: It's probably better to not use TypeMap and use the same trick with TypeList and MaxItems 1. Otherwise your schema isn't really useful, and you need to do all the type casting below.

docs/resources/synthetics.md
```diff
@@ -134,7 +143,7 @@ The following arguments are supported:
     - `targetvalue` - (Required) Expected matching value
     - `jsonpath` - (Required) The JSON path to assert
   - `property` - (Optional) if assertion type is "header", this is a the header name
-- `options` - (Required)
+- `options` - (Required) **Deprecated**
```
therve, line 146: I think you need to add the deprecation now.

romainberger: Ah right I forgot to add it in the schema

docs/resources/synthetics.md
```diff
@@ -150,6 +159,11 @@ The following arguments are supported:
   - `min_location_failed` - (Optional) Threshold below which a synthetics test is allowed to fail before sending notifications. Default is 1.
   - `accept_self_signed` - (Optional) For type=ssl, true or false
   - `allow_insecure` - (Optional) For type=api, true or false. Allow your HTTP test go on with connection even if there is an error when validating the certificate.
+  - `retry` - (Optional)
+    - `count` - (Optional) Number of retries needed to consider a location as failed before sending a notification alert.
+    - `interval` - (Optional) Interval between a failed test and the next retry in milliseconds.
+  - `monitor_options` - (Optional)
+    - `renotification_interval` - (Optional) Specify a renotification frequency.
```
justinsousa, line 166: obviously this is long closed but it looks like this is in minutes. I can submit a PR to add that to the docs to help clarify the unit on this number for folks

PR #639, opened 2020-08-18
Validate monitor in pre-flight

### LINKED ISSUE #414: (feature request) Validate monitor during plan with datadog api
This is a feature request same as https://github.com/terraform-providers/terraform-provider-datadog/issues/66 (closed) - datadog now provides `/validate` endpoint which I believe can be used by terraform-provider-datadog.

### Terraform Version
0.12.19

### Affected Resource(s)
- datadog_monitor

Note query field is invalid (no closing `}`). Other fields are not relevant with issue.

```hcl
resource "datadog_monitor" "my_monitor" {
  name    = "example"
  type    = "query alert"
  message = <<EOF
example
EOF
  query   = "sum(last_15m):sum:my_metrics.foo.bar{env:production.as_count() < 15"

  thresholds = {
    critical = 15.0
  }

  timeout_h           = 0
  no_data_timeframe   = 0
  renotify_interval   = 0
  notify_no_data      = false
  require_full_window = false
  notify_audit        = false
  include_tags        = true

  lifecycle {
    ignore_changes = [silenced]
  }

  tags = [
    "example"
  ]
}
```

### Expected Behavior
`terraform plan` fails because this is invalid monitor definition

### Actual Behavior
`terraform plan` succeeds, however when you run `terraform apply` you will get error

### Steps to Reproduce
1. copy above definition
1. `terraform plan` -> no error
1. `terraform apply` -> error

### References
https://github.com/terraform-providers/terraform-provider-datadog/issues/66 was closed due to philosophy that we shouldn't duplicate validation logic on provider side. However datadog provides /validate endpoint that could be used https://docs.datadoghq.com/api/?lang=bash#validate-a-monitor

therve:
Using CustomizeDiff, we can call the validation API to provide more validation for monitor creation. This allows returning an error during the `plan` phase, instead of waiting for `apply`.

Closes #414


datadog/resource_datadog_monitor.go
```diff
@@ -342,6 +349,22 @@ func resourceDatadogMonitorExists(d *schema.ResourceData, meta interface{}) (b b
 	return true, nil
 }
 
+func resourceDatadogMonitorCustomizeDiff(diff *schema.ResourceDiff, meta interface{}) error {
+	if _, ok := diff.GetOk("query"); !ok {
```
nmuesch, line 353: Does this mean we only validate the monitor if the query changed? 
Will we have access to all the other non changed fields like thresholds?
(Can we debug log something here about not validating)

therve: No it means we only validate if we have a value, in particular when the query depends on another resource that's not resolved.

nmuesch: Thanks. My main concern was that this `diff` object wouldn't contain the entire monitor object and only include what changed. But we discussed offline and that shouldn't be the case.

datadog/resource_datadog_monitor.go
```diff
@@ -342,6 +349,22 @@ func resourceDatadogMonitorExists(d *schema.ResourceData, meta interface{}) (b b
...
+		// If query depends on previous resources, we can't validate
```
nmuesch, line 354: Suggestion on this log line based on our offline chat that clarifies things for me :) 
```suggestion
		// If "query" depends on other resources, we can't validate as the variables may not be interpolated yet. 
```

datadog/resource_datadog_monitor.go
```diff
@@ -195,11 +202,15 @@ func resourceDatadogMonitor() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"validate": {
```
phillip-dd, line 205: What happens if the only change is to flip this from default true to false? If there aren't any changes to send to the API, I'm not sure if the endpoint accepts `PUT /monitor/<id> {}`, would be good to verify.

therve: You mean if the monitor object is already created, and we change the flag? I'll test it.

datadog/resource_datadog_monitor.go
```diff
@@ -342,6 +353,31 @@ func resourceDatadogMonitorExists(d *schema.ResourceData, meta interface{}) (b b
...
+		// Explicitly skip validation
+		return nil
+	}
+	m, _ := buildMonitorStruct(diff)
+
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV1 := providerConf.DatadogClientV1
+	authV1 := providerConf.AuthV1
+	if _, _, err := datadogClientV1.MonitorsApi.ValidateMonitor(authV1).Body(*m).Execute(); err != nil {
+		return translateClientError(err, "error validating monitor")
```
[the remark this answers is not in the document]
therve: I think that's what `translateClientError` is supposed to do.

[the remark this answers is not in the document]
phillip-dd: sounds good!

datadog/resource_datadog_monitor.go
therve: Good catch, removed it from the diff explicitly.

zippolyte (approved):
LGTM

ktham:
Awesome stuff. Just wondering @therve , are you planning to cut a release soon so we can leverage plan-time validation? Looking forward trying this out

PR #654, opened 2020-09-09
Add a dashboard attribute to add to a dashboard list

### LINKED ISSUE #379: Feature request: Add to a dashboard_list from a dashboard resource
Hi there 👋 !

We've thought it would be nice to have an additional optional attribute in the `dashboard` resource to pass a `dashboard_list`'s id.
Right now the DD `dashboard_list` resource need the list of dashboard ids, not the other way around: the dashboard don't have a parameter to specify a dashboard_list id.
Given our current terraform codebase layout, we don't have a way of making it nice and scalable without a lot of tweaking.

We've notice that the DD API support adding and removing dashboard items from a dashboard_list [1] and the API client used under the hood for the DD terraform provider already implement that endpoint [2]. 
Given the facts above, we're confident that this functionality can be added to the dashboard resource as an optional field.

Do you have already something similar in mind or in the roadmap, or any design concerns against such feature ? 

Have a nice day !

[1] https://docs.datadoghq.com/api/?lang=bash#add-items-to-a-dashboard-list
[2] https://github.com/zorkian/go-datadog-api/blob/master/dashboard_list_items_v2.go#L42-L52

therve:
This adds a new attribute to the dashboard resource to link the added dashboard to dashboard lists. It also handles removals. For now it needs to take IDs, but we'll add a data source in the future.

Closes #379


datadog/resource_datadog_dashboard.go
```diff
@@ -122,9 +149,47 @@ func resourceDatadogDashboardUpdate(d *schema.ResourceData, meta interface{}) er
...
+
+	updateDashboarLists(d, providerConf, *dashboard.Id)
+
 	return resourceDatadogDashboardRead(d, meta)
 }
 
+func updateDashboarLists(d *schema.ResourceData, providerConf *ProviderConfiguration, dashboardId string) {
+	dashTypeString := "custom_screenboard"
+	if d.Get("layout_type").(string) == "free" {
+		dashTypeString = "custom_timeboard"
```
zippolyte, line 161: shouldn't it be the opposite ? `custom_screenboard` for `free` and `custom_timeboard` otherwise ?

therve: It's very possible. I think the backend corrects anyway, but I'm always confused by terminology.

zippolyte (approved):

PR #657, opened 2020-09-14
Add a datasource for dashboard lists

### LINKED ISSUE #651: [Feature Request] expose a dashboard_list data source
### Feature request

It could be a good thing to create a data source around : `datadog_dashboard_list`. 
I have a parent project that is responsible to manage the dashboard_list. Also, multiple terraform workspace are refering to different sub part of this parent project. it would be great to store all dashboards of those sub parts in the same dashboard list.
In order to be able to automate component dashboard registration, we thought that the better way to handle it is that each component have to register itself to the dashboard_list

### Current workaround

For now the only way I found to register each dashboard in the same list is to import the `datadog_dashboard_list` resource in each sub projects. Note that because each of those project uses it's own terraform workspace, if I have to modify the dashboard_list, I have to modify it in all of these workspaces in order to keep a consistency between each terraform states.
If this consistency is not assured, a terraform apply in a subproject will potentially remove a dashboard for another component from the dashboard_list.

therve:
This adds a new datasource for referencing dashboard lists by name.

Closes #651


zippolyte (approved):
LGTM

docs/data-sources/dashboard_list.md
```diff
@@ -0,0 +1,24 @@
...
+
+```
+data "datadog_dashboard_list" "test" {
+  name = "My super list"
+}
+```
+
+## Argument Reference
+
+- `name`: (Required) A monitor name to limit the search.
```
zippolyte, line 19:
```suggestion
- `name`: (Required) A dashboard list name to limit the search.
```

zippolyte (changes requested):

zippolyte (approved):

PR #663, opened 2020-09-16
Update to go client beta 8

therve:


GNUmakefile
```diff
@@ -3,7 +3,7 @@ GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 PKG_NAME=datadog
 DIR=~/.terraform.d/plugins
 ZORKIAN_VERSION=master
-API_CLIENT_VERSION=master
```
bkabrda, line 6: We don't generally update this to contain the new release tag. The assumption is that when `make update-go-client` is run, you want to *update*, so fixing `API_CLIENT_VERSION` to the tag is not really useful.

therve: Yeah that was unwanted.

bkabrda (approved):

Issue #667, opened 2020-09-17
Handle dependency between monitors, SLOs and dashboards


Tony-Proum:
### Terraform Version

Terraform v0.12.29

### Affected Resource(s)

- datadog_dashboard
- datadog_service_level_objective
- datadog_monitor
(and maybe some other but didn't know enough to provide a full list of this resources)

### Terraform Configuration Files

```hcl
resource "datadog_monitor" "availability" {
  name               = "Availability"
  type               = "metric alert"
  message            = "Monitor triggered."
  query = "<any request>"
  lifecycle {
    ignore_changes = [silenced]
  }
}

resource "datadog_service_level_objective" "availability" {
  name        = "Availability"
  type        = "monitor"
  description = "Availability"
  monitor_ids = [
    datadog_monitor.availability.id
  ]
  thresholds {
    timeframe = "7d"
    target    = 99.95
    warning   = 99.96
  }
}

resource "datadog_dashboard" "service_level_objectives" {
  layout_type = "ordered"
  title       = "Service Level Objectives"
  widget {
    service_level_objective_definition {
      title             = "Availability SLO"
      view_type         = "detail"
      slo_id            = datadog_service_level_objective.availability.id
      show_error_budget = true
      view_mode         = "overall"
      time_windows      = ["7d", "30d", "90d"]
    }
  }
}
```

### Actual Behavior

Modifying the monitor type is not possible in this configuration without using some workaround: 
When modifying the type of the monitor, terraform needs to destroy the resource and to recreate it (due to an API restriction ? The documentation don't mention that but I didn't tried).
The plan shows that the operation is needed, and it make sense as illustrated here: 

```
  # datadog_monitor.availability must be replaced
-/+ resource "datadog_monitor" "availability" {
      ~ evaluation_delay    = 0 -> (known after apply)
      ~ id                  = "269981" -> (known after apply)
... TRUNCATED ...
      - timeout_h           = 0 -> null
      ~ type                = "query alert" -> "log alert" # forces replacement
    }

  # datadog_service_level_objective.availability will be updated in-place
  ~ resource "datadog_service_level_objective" "availability" {
... TRUNCATED ...
      ~ monitor_ids = [
          - 269981,
        ] -> (known after apply)
... TRUNCATED ...
}

```
But at runtime, terraform output displays throws this type of error : 
```
[...]

PR #668, opened 2020-09-17
Add dependencies between monitor, SLO & dashboards

Tony-Proum:
Don't know if this could handle of the complexity of this issue : https://github.com/DataDog/terraform-provider-datadog/issues/667 
But maybe this simple fix could do the trick


jirikuncar:
@Tony-Proum it looks good. Would you mind adding a test scenario?

Tony-Proum:
Oh, ok, I thought that the solution may be more complicated, but as it looks good, I'll add a test for this feature

Tony-Proum:
@jirikuncar I think that those test are able to demonstrate what we are doing here. But I'm not sure that they follow the practices in this repository ( naming / way of writing and so forth ). Feel free to tell me if something is wrong.

Tony-Proum:
Oh, sorry, I just realized that I forgot to assert that the dashboard were forced to be recreated in the last test I have commit, I'll fix that

Tony-Proum:
@jirikuncar It works on my machine :laughing: 
I removed the fail fast flag on the github action part, in order to see if tests succeed only on my machine, but seems like their works on linux and not on other types of OS. unfortunately I does not have access to a mac or windows build environment. And even if so, I'm not a golang expert and don't know if I'll be able to locate the problem

Tony-Proum:
I managed to fix the test for MacOs and Windows case by removing the destroyCheck of this test. Even if it was running properly on linux, it seems to have some kind of race condition on mac or windows that's make the test fail for those OS.

Do you think that we could omit this check for this particular test ?

.github/workflows/test.yml
```diff
@@ -27,6 +27,7 @@ jobs:
     env:
       GOFLAGS: "-mod=vendor"
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
Tony-Proum: Done

datadog/resource_datadog_dashboard.go
```diff
@@ -3802,6 +3802,7 @@ func getServiceLevelObjectiveDefinitionSchema() map[string]*schema.Schema {
 		"slo_id": {
 			Type:     schema.TypeString,
 			Required: true,
+			ForceNew: true,
```
nmuesch, line 3805: Will this cause the whole dashboard to be deleted + recreated if the slo changes? I'm not sure if that'll be expected for users, since it'd mean any dashboard lists or favorited dashboards will need to be updated with the new dashboard id.

nmuesch (commented):
Taking a look at the related issue I'm wondering if there's a way to use the newer force_delete flag on the monitor resource - https://github.com/DataDog/terraform-provider-datadog/pull/535/files to satisfy this use case.

I believe that would avoid having to delete+create any dependent dashboards/SLOs and let those update in place with the new monitor id.

Tony-Proum:
> Taking a look at the related issue I'm wondering if there's a way to use the newer force_delete flag on the monitor resource - https://github.com/DataDog/terraform-provider-datadog/pull/535/files to satisfy this use case.
> 
> I believe that would avoid having to delete+create any dependent dashboards/SLOs and let those update in place with the new monitor id.

Oh seems to be a way better solution ! (I didn't known that feature exists through the API)

datadog/resource_datadog_dashboard.go
Tony-Proum: I don't know if it's an issue any more: the dasboard_list can now be handle by the dashboard itself (see: https://github.com/DataDog/terraform-provider-datadog/pull/654/files) so in theory it should works just fine. But I' m agree, delete + recreate could be a little ... overkill :smile:

nmuesch:
Hey @Tony-Proum I apologize for my delay here. 

Please let me know if this isn't the case, but I believe monitors is the only resource that would fail to delete if its referenced in an SLO? So we'd only need the force_delete option there? 

Originally we decided against enabling forceDelete to be on by default, since the option is designed as a safegaurd (see - https://github.com/DataDog/terraform-provider-datadog/pull/451#issuecomment-600653140) from inadvertently deleting monitors that are referenced inside an SLO definition. Though I do think the experience could be a bit better. Maybe we could call this out in the error thats logged and raised to the user and explicitly note the force_delete option is available?

Tony-Proum:
If I remember correctly, when doing some tests, the same validation exists between SLO widget and Dasboards. Yes I think that we can remove this validation on some resources using the force_delete flag. 
But a thing keep struggle me: with this solution, we will have some dangling dashboards with no widget or SLO with no metrics associated to, and don't know which is better: having irrelevant dashboards/SLOs or having to recreate all the things when a dependency have to change (As a developer I have a preference for the second case: as it seems similar to what's happen when we build a package but well, this is not a package :sweat_smile: )

datadog/resource_datadog_dashboard.go
```diff
@@ -3809,6 +3809,7 @@ func getServiceLevelObjectiveDefinitionSchema() map[string]*schema.Schema {
...
```
jirikuncar, line 3812:
```suggestion
```

datadog/resource_datadog_dashboard_dependencies_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"strings"
+	"testing"
+)
+
+func datadogDashboardDepConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_monitor" "foo" {
+  name = "Name for monitor foo"
+  type = "query alert"
+  message = "some message Notify: @hipchat-channel"
```
jirikuncar, line 17:
```suggestion
  message = "some message"
```

datadog/resource_datadog_dashboard_dependencies_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}`, uniq)
+}
+
+func datadogDashboardDepChangedConfig(uniq string) string {
+	return fmt.Sprintf(
+		`resource "datadog_monitor" "foo" {
+  name = "Name for monitor foo"
+  type = "metric alert"
+  message = "some message Notify: @hipchat-channel"
```
jirikuncar, line 63:
```suggestion
  message = "some message"
```

datadog/resource_datadog_dashboard_dependencies_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"strings"
+	"testing"
+)
+
+func datadogDashboardDepConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_monitor" "foo" {
+  name = "Name for monitor foo"
```
jirikuncar, line 15: Please include `uniq` in the monitor name:
```suggestion
  name = "%s monitor"
```

datadog/resource_datadog_dashboard_dependencies_test.go
```diff
@@ -0,0 +1,166 @@
...
+  type = "query alert"
+  message = "some message Notify: @hipchat-channel"
+  query = "avg(last_1h):anomalies(avg:system.cpu.system{name:cassandra}, 'basic', 3, direction='above', alert_window='last_5m', interval=20, count_default_zero='true') >= 1"
+}
+
+resource "datadog_service_level_objective" "foo" {
+  name = "Name for slo foo"
```
jirikuncar, line 22:
```suggestion
  name = "%s slo"
```

datadog/resource_datadog_dashboard_dependencies_test.go
```diff
@@ -0,0 +1,166 @@
...
+			view_type = "detail"
+		}
+	}
+}`, uniq)
+}
+
+func datadogDashboardDepChangedConfig(uniq string) string {
+	return fmt.Sprintf(
+		`resource "datadog_monitor" "foo" {
+  name = "Name for monitor foo"
```
jirikuncar, line 61:
```suggestion
  name = "%s monitor"
```

Tony-Proum:
@nmuesch 
I tried to use the force_delete and it seems to do the job in a certain way. I got some SLOs with no monitor when something went wrong during deleting and recreating a monitor (successful delete and erroneous creation) As you said, this is a good safeguard and I'm not sure that it should be used in a ¨normal¨ use case. 
Also, I'm agree, the error message could help to find this solution.

I also tried, to change the SLO on a monitor and for this part, I think that the issue were a PEBKAC as I were able to change it without issue.

If you are agree with this, I'll remove the `ForceNew` flag from dashboard resource as suggested by @jirikuncar, I also have to delete the associated integration test that remains. And would like to pursue with this PR: I think that it allows a better developer experience on this specific situation and a is a more straight forward solution for a normal use case. I keep in mind that the `force_delete` flag exists but I'm keep struggled by the workflow I had encounter to finally have it working. It's not really simple as we have to know that the `force_delete` flag exists, so even if the solution is in the error, the user will have to rollback is changes and then applying the `force_delete` flag on all impacted monitor and then apply a change on the impacted monitor. 

IMHO Instead of simply delete the SLO, it seems cumbersome.

datadog/resource_datadog_service_level_objective_test.go
```diff
@@ -101,6 +101,58 @@ resource "datadog_service_level_objective" "foo" {
 }`, uniq)
 }
 
+func testAccCheckDatadogServiceLevelObjectiveConfigMonitor(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_monitor" "foo" {
+  name = "Name for monitor foo"
```
jirikuncar, line 107: please make sure that _all_ resource names contain `%s` with `uniq` so we can identify these resources in test org

Tony-Proum: Oh, thanks, I missed this point

Tony-Proum:
I were not able to test this anymore as my trial accounts life ends during our forth and back about this PR, do you know if it exists some developer accounts on datadog ? (did'nt found this information in the documentation)

Tony-Proum:
@nmuesch & @jirikuncar  just happens to me this morning while trying to modify an SLO: 
```Error: error deleting service level objective: json: cannot unmarshal string into Go struct field SLODeleteResponse.errors of type map[string]string: {"errors":"slo <slo_id> is referenced in dashboard_id: <dashboard_id>","data":null}
```

So I finally think that the link exists both between dashboard and slo, and between slo and monitors.

therve:
We've added the `force_delete` flag to SLO resources, I think this is the recommended solution for now. Thanks for your patience around this issue.

Tony-Proum:
@therve Maybe should this flag be added to dashboards as well ?

PR #673, opened 2020-09-28
[Synthetics] Add support for DNS tests

romainberger:
Add support for DNS tests for Synthetics.


therve (changes requested):
Code looks good, can you update the documentation? Thanks.

PR #675, opened 2020-09-29
[Synthetics] Add support for global variables

romainberger:
Add support for synthetics' global variables.

It also updates the go client since it needed a new endpoint.


datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,143 @@
...
+
+func resourceDatadogSyntheticsGlobalVariableRead(d *schema.ResourceData, meta interface{}) error {
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV1 := providerConf.DatadogClientV1
+	authV1 := providerConf.AuthV1
+
+	syntheticsGlobalVariable, _, err := datadogClientV1.SyntheticsApi.GetGlobalVariable(authV1, d.Id()).Execute()
+
+	if err != nil {
+		if strings.Contains(err.Error(), "404 Not Found") {
```
therve, line 69: You should be able to use the HTTP response instead and check the status code. I know we used that pattern elsewhere, but it's super awkward.

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,143 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Required: true,
```
therve, line 26: This is marked as optional in the doc, which seems to make sense to me.

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,143 @@
...
+		Read:   resourceDatadogSyntheticsGlobalVariableRead,
+		Update: resourceDatadogSyntheticsGlobalVariableUpdate,
+		Delete: resourceDatadogSyntheticsGlobalVariableDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
```
therve, line 22: Maybe we want to check it's not an empty string?

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,143 @@
...
+			},
+			"tags": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"value": {
+				Type:     schema.TypeString,
+				Required: true,
```
therve, line 35: Do you support empty strings here as well?

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,143 @@
...
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"tags": {
+				Type:     schema.TypeList,
+				Required: true,
```
therve, line 30: Is it really required? I would assume an empty list works.

therve (changes requested):
Looking good, some minor comments about the schema.

datadog/resource_datadog_synthetics_global_variable.go
romainberger: Yes it works with an empty list, but it needs to be set (like for synthetics tests). I suppose we could default to an empty list if nothing is specified in the config.

datadog/resource_datadog_synthetics_global_variable.go
romainberger: For the value yes, it is supported.

therve (approved):

PR #676, opened 2020-09-29
[APM/Dashboards] add apm_stats_query

anbnyc:
We are supporting a new query type in dashboards for APM data sources. This PR would allow users to import and create dashboards with this query using Terraform.

See https://github.com/DataDog/datadog-api-spec/pull/579 for API spec.


datadog/resource_datadog_dashboard.go
```diff
@@ -5086,6 +5093,111 @@ func buildTerraformProcessQuery(datadogQuery datadogV1.ProcessQueryDefinition) m
...
+	if v, ok := terraformQuery["name"].(string); ok && len(v) != 0 {
+		datadogQuery.SetName(v)
+	}
+	if v, ok := terraformQuery["env"].(string); ok && len(v) != 0 {
+		datadogQuery.SetEnv(v)
+	}
+	if v, ok := terraformQuery["primary_tag"].(string); ok && len(v) != 0 {
+		datadogQuery.SetPrimaryTag(v)
+	}
+	if v, ok := terraformQuery["row_type"].(datadogV1.ApmStatsQueryRowType); ok && len(v) != 0 {
```
anbnyc, line 5152: I was getting an error for `terraformQuery["row_type"].(string)` and I think this is the solution, but please let me know if there's a more correct way

therve: I think it might more make sense to do `datadogQuery.SetRowType(datadogV1.ApmStatsQueryRowType(v))`.

therve (changes requested):
A small comment, please add a test, and change the documentation. Thanks!

therve (approved):

PR #681, opened 2020-10-05
Update vendor

vinckama:
Vendor update to support new endpoints and attributes


jirikuncar (approved):

therve (changes requested):
Looks like the CI failure is legit.

vinckama:
That's a backward incompatibility, I'm going to fix that quickly on `api-spec`& each client

therve:
Opened #687

PR #680, opened 2020-10-01
[APM] change columns from string[] to object[]

anbnyc:
Implements changes from https://github.com/DataDog/datadog-api-spec/pull/599

PR #684, opened 2020-10-06
Update vendor

sighrobot:
In support of #685


therve:
I think we'll need #680 in to make it work.

PR #685, opened 2020-10-06
[VIZZ-953] Add timeseries dual y-axis to provider

sighrobot:
Adds support for dual y-axis for timeseries widgets

Related:
https://github.com/DataDog/datadog-api-spec/pull/593
https://github.com/DataDog/dogweb/pull/47456


nmuesch (dismissed):
Thanks, this looks good to me. 

One minor nit, it may be worthwhile to set one of the on_right_yaxis to "false" so we're testing both conditions.

therve (approved):

PR #686, opened 2020-10-06
Update table widget definition and request

sighrobot:
Adds new properties to the table widget definition and request


datadog/resource_datadog_dashboard.go
```diff
@@ -3297,6 +3297,10 @@ func getQueryTableDefinitionSchema() map[string]*schema.Schema {
 				Schema: getWidgetTimeSchema(),
 			},
 		},
+		"has_search_bar": {
+			Type:     schema.TypeString,
```
nmuesch, line 3301: Should this be a bool?

datadog/resource_datadog_dashboard.go
```diff
@@ -3317,6 +3321,9 @@ func buildDatadogQueryTableDefinition(terraformDefinition map[string]interface{}
 	if v, ok := terraformDefinition["time"].(map[string]interface{}); ok && len(v) > 0 {
 		datadogDefinition.SetTime(*buildDatadogWidgetTime(v))
 	}
+	if v, ok := terraformDefinition["has_search_bar"].(datadogV1.TableWidgetHasSearchBar); ok && len(v) != 0 {
```
nmuesch, line 3324: (If this winds up being a bool, we can probably remove the length check)

nmuesch (commented):
Quick question on the type

datadog/resource_datadog_dashboard.go
sighrobot: No, this is a string enum (`"always"`, `"never"`, or `"auto"`) – is there a way to specify an enum?

nmuesch: Ahh ok thanks. You could do something like set a ValidateFunc with stringInSlice - https://github.com/hashicorp/terraform/blob/master/helper/validation/validation.go#L32 but I think its OK to just document the allowed fields too.

therve (changes requested):
That looks good, can you tweak the documentation to mention the new field? Thanks.

therve (approved):

PR #688, opened 2020-10-07
Update vendor

vinckama:
Update vendor to remove `target_format` default value in logs  `attribute remapper`


therve:
I think superced by #687

PR #689, opened 2020-10-07
Update `ApmStatsQuery` column interface parsing

anbnyc:
`ApmStatsQuery` column object needs to be built/parsed property by property, instead of casting as `ApmStatsQueryColumnType`.


datadog/resource_datadog_dashboard.go
```diff
@@ -5181,6 +5181,26 @@ func getApmStatsQuerySchema() *schema.Schema {
...
+
+	datadogColumn := datadogV1.NewApmStatsQueryColumnTypeWithDefaults()
+
+	if value, ok := terraformColumn["name"].(string); ok && len(value) != 0 {
+		datadogColumn.SetName(value)
+	}
+	if value, ok := terraformColumn["alias"].(string); ok && len(value) != 0 {
+		datadogColumn.SetAlias(value)
+	}
+	if value, ok := terraformColumn["cell_display_mode"].(datadogV1.TableWidgetCellDisplayMode); ok && len(value) != 0 {
```
nmuesch, line 5194: You may have to build these as well?

nmuesch (commented):
Overall looks good to me! Can you add to the dashboard tests to include a widget that includes this apm stats column?

datadog/resource_datadog_dashboard.go
anbnyc: @nmuesch `TableWidgetCellDisplayMode` and `WidgetSort` are enums, so I don't think there's anything to build --would it be better to set them as `.(string)` here?

anbnyc: (Update: I tried `.(string)` and it throws a type error)

therve: You need to cast to string, and then call `datadogV1.TableWidgetCellDisplayMode` on the value.

datadog/resource_datadog_dashboard.go
```diff
@@ -5181,6 +5181,26 @@ func getApmStatsQuerySchema() *schema.Schema {
...
+		datadogColumn.SetCellDisplayMode(value)
+	}
+	if value, ok := terraformColumn["order"].(datadogV1.WidgetSort); ok && len(value) != 0 {
```
therve, line 5197: Same here.

therve (changes requested):
I think we have an issue with the spec, it shouldn't set the default value of `cell_display_mode`.

therve (approved):

PR #694, opened 2020-10-08
[Logs] add the new endpoint to reorder the logs archives

vinckama:
Related to that [PR](https://github.com/DataDog/datadog-api-spec/pull/528) in api-spec


datadog/resource_datadog_logs_archive_order.go
```diff
@@ -0,0 +1,91 @@
...
+				Type:     schema.TypeList,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func resourceDatadogLogsArchiveOrderCreate(d *schema.ResourceData, meta interface{}) error {
+	return resourceDatadogLogsArchiveOrderUpdate(d, meta)
```
therve, line 32: Are you sure we want to do that? Not the same thing as the integration pipeline, where we just support imports?

datadog/resource_datadog_logs_archive_order.go
```diff
@@ -0,0 +1,91 @@
...
+		Create: resourceDatadogLogsArchiveOrderCreate,
+		Update: resourceDatadogLogsArchiveOrderUpdate,
+		Read:   resourceDatadogLogsArchiveOrderRead,
+		Delete: resourceDatadogLogsArchiveOrderDelete,
+		Exists: resourceDatadogLogsArchiveOrderExists,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {Type: schema.TypeString, Required: true},
```
therve, line 21: Do we really need to have this for the user? It looks like it's an internal field.

datadog/resource_datadog_logs_archive_order.go
```diff
@@ -0,0 +1,91 @@
...
+			if getErr != nil {
+				return translateClientError(err, "error getting logs archive order")
+			}
+			return fmt.Errorf("cannot map archives to existing ones\n existing archives: %s\n archive to be updated: %s",
+				ddArchiveOrder.Data.Attributes.ArchiveIds,
+				ddList)
+		}
+		return translateClientError(err, "error updating logs archive order")
+	}
+	d.SetId(tfId)
```
therve, line 79: If we expose the name, I wonder what happens here when it changes.

therve (changes requested):
Thanks, looks like a good start. Those kind of resources are always a tough fit, asked some questions. You also need to create the tests using your test data.

datadog/resource_datadog_logs_archive_order.go
vinckama: It's the same behavior of endpoint to reorder pipeline

vinckama:
I juste reused the pipeline reorder to create that the endpoint. I will also update, in another PR, the tests of pipeline reordering

datadog/resource_datadog_logs_archive_order.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// The deletion of archive order is not supported from config API.
+// This function simply delete the archive order resource from terraform state.
+func resourceDatadogLogsArchiveOrderDelete(_ *schema.ResourceData, _ interface{}) error {
+	return nil
+}
+
+func resourceDatadogLogsArchiveOrderExists(_ *schema.ResourceData, _ interface{}) (bool, error) {
+	return true, nil
```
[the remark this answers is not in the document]
vinckama: yes, it fact the order always exists and it can be empty

datadog/resource_datadog_logs_archive_order.go
vinckama: the name is useless to me I will remove it

docs/resources/logs_archive_order.md
```diff
@@ -0,0 +1,42 @@
...
+# datadog_logs_archive_order Resource
+
+Provides a Datadog [Logs Archive API](https://docs.datadoghq.com/api/v2/logs-archives/) resource, which is used to manage Datadog log archives order.
+
+## Example Usage
+
+```hcl
+resource "datadog_logs_archive_order" "sample_archive_order" {
+    name = "sample_archive_order"
+    depends_on = [
```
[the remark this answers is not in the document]
vinckama: okay nice 👌

datadog/resource_datadog_logs_archive_order.go
```diff
@@ -0,0 +1,91 @@
...
+				ddList)
+		}
+		return translateClientError(err, "error updating logs archive order")
+	}
+	d.SetId(tfId)
+	return resourceDatadogLogsArchiveOrderRead(d, meta)
+}
+
+// The deletion of archive order is not supported from config API.
+// This function simply delete the archive order resource from terraform state.
```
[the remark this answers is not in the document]
vinckama: I assume that the `id` is deleted automatically by terraform

vinckama:
PR updates
----

 - remove `name` attribute
 - set Id to `archiveOrderID` after each change (create/update of the ressource)
 - change `archive_ids` to optional & computed
 - support of create action, behavior depend if the user specifies values in `archive_ids` attribute
 - update the doc

datadog/cassettes/TestAccDatadogLogsArchiveOrder_basic.freeze
```diff
@@ -0,0 +1 @@
+2020-10-13T15:00:35.986278+02:00
```
luchenyuxx, line 1: missing new line character

datadog/cassettes/TestAccDatadogLogsArchiveOrder_empty.freeze
```diff
@@ -0,0 +1 @@
+2020-10-13T15:00:07.182132+02:00
```
luchenyuxx, line 1: missing new line character

datadog/resource_datadog_logs_archive_order_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"testing"
+)
+
+func ArchiveOrderConfig() string {
+	return fmt.Sprintf(`
+resource "datadog_logs_archive_order" "archives" {
+	archive_ids = []
```
therve, line 16: Can you create a test with real logs archive?

datadog/resource_datadog_logs_archive_order_test.go
```diff
@@ -0,0 +1,95 @@
...
+}`)
+}
+
+func TestAccDatadogLogsArchiveOrder_basic(t *testing.T) {
+	rec := initRecorder(t)
+	accProviders, _, cleanup := testAccProviders(t, rec)
+	defer cleanup(t)
+	accProvider := testAccProvider(t, accProviders)
+
+	resource.ParallelTest(t, resource.TestCase{
```
therve, line 32: Given that there is a unique resource behind it I don't think we want to use parallel tests.

datadog/cassettes/TestAccDatadogLogsArchiveOrder_basic.freeze
therve: This is a generated file.

therve (commented):
Looks good, some comments about testing if possible.

datadog/resource_datadog_logs_archive_order_test.go
vinckama: I added a test to see if the archive_ids order from the state match with the real one.
If I would add archive order with real archive, I would mean get the archive from a real client and import the id before the test start, which seems not possible with the Test helper

therve (approved):

PR #696, opened 2020-10-08
Add custom_links support to dashboard widget definitions

### LINKED ISSUE #557: [Feature Request] Support custom_links in timeseries_definition
It would be good to support Custom Links in widgets that do support it. To my knowledge, at least the timeseries widget supports it.
The API returns 
```
"custom_links":[{"link":"https://example.com","label":"example"}]
```
as a property of the `definition` field.

sighrobot:
Resolves #557


datadog/resource_datadog_dashboard_timeseries_test.go
```diff
@@ -6,253 +6,197 @@ import (
...
+//            "legend_size":"2",
+//            "type":"timeseries",
+//            "events":[
+//               {
+//                  "q":"sources:test tags:1",
+//                  "tags_execution":"and"
+//               }
+//            ]
+//         },
+//         "id":203606066001657
```
nmuesch, line 196: Since we redact most ids/urls/handles, we should probably redact these widget ids

datadog/resource_datadog_dashboard.go
```diff
@@ -4669,6 +4783,45 @@ func buildTerraformWidgetConditionalFormat(datadogWidgetConditionalFormat *[]dat
...
+
+func getWidgetCustomLinkSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"label": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"link": {
+			Type:     schema.TypeString,
+			Optional: true,
```
nmuesch, line 4796: Is this not a required field? My assumption would be this would be reversed :)

sighrobot: thanks for this catch – [both are required](https://github.com/DataDog/dogweb/blob/prod/dogweb/model/dashboard_widget/widget/lib/common_schemas.py#L81)!

nmuesch (approved):
Looks good to me! Thanks. 
@DataDog/dashboards if someone could take a pass here as well 🙇

nmuesch:
Took a look at the AZP failure, looks like the errors are an unrelated 409 on an AWS test, and a known issue about dashboard deletion. This is good to merge once dashboards approves 🙂

datadog/resource_datadog_dashboard.go
```diff
@@ -4669,6 +4783,45 @@ func buildTerraformWidgetConditionalFormat(datadogWidgetConditionalFormat *[]dat
...
+			Label: terraformCustomLink["label"].(string),
+			Link:  terraformCustomLink["link"].(string),
+		}
+		datadogWidgetCustomLinks[i] = datadogWidgetCustomLink
+	}
+	return &datadogWidgetCustomLinks
+}
+func buildTerraformWidgetCustomLinks(datadogWidgetCustomLinks *[]datadogV1.WidgetCustomLink) *[]map[string]string {
+	terraformWidgetCustomLinks := make([]map[string]string, len(*datadogWidgetCustomLinks))
+	for i, datadogWidget := range *datadogWidgetCustomLinks {
```
jbenais, line 4814: nit: I would rename `datadogWidget` to datadogWidgetCustomLink` since you're iterating over `custom_links` here

jbenais:
Looks good to me for Dashboards! 👏
I would just recommend to update the Terraform documentation with the new `custom_link` field.

therve (approved):

PR #698, opened 2020-10-09
Add command for versionless building

nmuesch:
Use the workaround provided in https://github.com/hashicorp/terraform/issues/ 25906 to allow the ability to build a local version of the provider that will work with terraform 0.13. 

Currently, there is no way to build a versionless provider and use that when doing local testing for development purposes. The `init` will always pull the latest release. This allows us to build and store a version on the directory tf 0.13 is expecting, and utilize a local filesystem mirror to pull that build plugin.


GNUmakefile
```diff
@@ -5,11 +5,25 @@ DIR=~/.terraform.d/plugins
 ZORKIAN_VERSION=master
 API_CLIENT_VERSION=master
 
+# Local variables for installing the plugin to a local
+# plugin mirrior, used for manual build/testing with terraform 0.13
```
zippolyte, line 9:
```suggestion
# plugin mirror, used for manual build/testing with terraform 0.13
```

zippolyte (dismissed):
small nit, feel free to merge after

nmuesch:
Thanks for the review, applied your suggestion 🙇 
Merging

PR #703, opened 2020-10-12
Document dual y-axis attributes.

therve:
See #685.


docs/resources/dashboard.md
```diff
@@ -860,6 +860,7 @@ Nested `widget` blocks have the following structure:
       - `rum_query`: (Optional) The rum query to use in the widget. The structure of this block is described [below](dashboard.html#nested-apm_query-log_query-network_query-and-rum_query-blocks).
       - `process_query`: (Optional) The process query to use in the widget. The structure of this block is described [below](dashboard.html#nested-process_query-blocks).
       - `display_type`: (Optional) Type of display to use for the request. Available values are: `area`, `bars`, or `line`.
+      - `on_right_axis`: (Optional) Boolean indicating whether the request will use the right or left Y-Axis.
```
nmuesch, line 863:
```suggestion
      - `on_right_yaxis`: (Optional) Boolean indicating whether the request will use the right or left Y-Axis.
```

docs/resources/dashboard.md
```diff
@@ -878,6 +879,7 @@ Nested `widget` blocks have the following structure:
       - `q`: (Required) The event query to use in the widget.
       - `tags_execution`: (Optional) The execution method for multi-value filters.
     - `yaxis`: (Optional) Nested block describing the Y-Axis Controls. The structure of this block is described [below](dashboard.html#nested-widget-axis-blocks)
+    - `right_yaxis`: (Optional) Nested block describing the right Y-Axis Controls. See the `on_right_axis` property for which request will use this axis. The structure of this block is described [below](dashboard.html#nested-widget-axis-blocks)
```
nmuesch, line 882: I'm not sure I follow this:
> See the `on_right_axis` property for which request will use this axis.

Is the goal to mention that in order for this to take effect you need `on_right_yaxis` field set? Maybe something like:

> Set `on_right_yaxis` to True to use this configuration

nmuesch (changes requested):
Thanks! From #685 it looks like the option is `on_right_yaxis`

docs/resources/dashboard.md
therve: No the fields are on 2 different objects. You have 2 axis, some will use the left, some will use the right, depending on `on_right_yaxis`.

sighrobot (approved):
This looks good, thanks!

docs/resources/dashboard.md
nmuesch: Ahh ok thanks

PR #708, opened 2020-10-14
Fix logs archive order tests

therve:
Do a single assert so that it works in our test account.


datadog/resource_datadog_logs_archive_order_test.go
```diff
@@ -69,7 +68,6 @@ func TestAccDatadogLogsArchiveOrder_empty(t *testing.T) {
 					resource.TestCheckResourceAttrSet(
 						"datadog_logs_archive_order.archives", "archive_ids.#"),
 					testAccCheckArchiveOrderResourceMatch(accProvider, "datadog_logs_archive_order.archives", "archive_ids.0"),
-					testAccCheckArchiveOrderResourceMatch(accProvider, "datadog_logs_archive_order.archives", "archive_ids.1"),
```
nmuesch, line 72: Would this work if we were to start testing against a fresh account? (If not, is there a way to "setup" the account to be prepared for the test?)

vinckama: I think the fix is just to avoid the error [here](https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_logs_archive_order_test.go#L119) (replace it by println)

vinckama: This depends on the customer's configuration. To me it is not possible to setup the right configuration.

 it is not possible to create another one, since we have predefined an archive limit, and we can reach the limit.

vinckama (approved):
LGTM !

PR #711, opened 2020-10-19
[Synthetics] Add support to dnsServer and client certificate options

romainberger:
This PR updates the go-client to add support to `dnsServer` and client certificate for synthetics tests.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -231,6 +242,34 @@ func syntheticsTestRequest() *schema.Schema {
 					Optional: true,
 					Default:  60,
 				},
+				"dnsServer": {
```
therve, line 245: In this case we can easily hide the API case, let's rename it to `dns_server`.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -231,6 +242,34 @@ func syntheticsTestRequest() *schema.Schema {
...
+					Type:     schema.TypeString,
+					Required: true,
+				},
+				"filename": {
+					Type:     schema.TypeString,
+					Required: true,
+				},
+				"updated_at": {
+					Type:     schema.TypeString,
+					Required: true,
```
therve, line 271: Are those last 2 required? Not sure why the API case about filename, or what updated_at means.

therve (changes requested):
Looks good, some minor comments

PierreGuyot (dismissed):

datadog/resource_datadog_synthetics_test_.go
romainberger: Yes that's two fields that make sense when using the UI but not really with terraform, I'll make the `updatedAt` option automatically fill in the backend, and see what we can do for `filename`.

romainberger: I've removed the `updatedAt` option and we decided to automatically fill the `filename` (required in the backend) so it makes more sense.

therve (approved):

PR #715, opened 2020-10-22
Add include_tags in the log external archives

vinckama:
PR to add `include_tags` in the log external archives

related to this [PR](https://github.com/DataDog/datadog-api-spec/pull/607)  in api-spec and this [Jira Ticket](https://datadoghq.atlassian.net/jira/software/projects/LB/boards/386?selectedIssue=LB-252)


datadog/resource_datadog_logs_archive.go
```diff
@@ -65,6 +65,11 @@ func resourceDatadogLogsArchive() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"include_tags": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
gabrielreid, line 71: The default in the UI is `true`, so I'm wondering about the choice to have this set to `false` by default here.

This is the first time I've looked at terraform-provider code, so I'm definitely not fully familiar with the general concepts, but what I would _guess_ that we want is:
* existing terraformed archives won't be affected if this isn't specified (i.e. they're setting will remain unchanged)
* newly-created archives will have this enabled by default

I'm not sure if the above is in line with how everything works within the terraform provider, or the introduction of other flags like this, but it's what I would expect as a user who doesn't know anything about this stuff.

docs/resources/logs_archive.md
```diff
@@ -46,6 +46,7 @@ The following arguments are supported:
   - `client_id`: (Required) Your client id.
   - `storage_account`: (Required) The associated storage account.
 - `rehydration_tags`: (Optional) An array of tags to add to rehydrated logs from an archive.
+- `include_tags`: (Optional, default=false) To store the tags in the archive, set the value "true". If it is set to "false", the tags will be deleted when the logs are sent to the archive. By default.
```
gabrielreid, line 49: Small suggestions on wording:

```suggestion
- `include_tags`: (Optional, default=false) To store the tags in the archive, set the value "true". If it is set to "false", the tags will be dropped when the logs are sent to the archive.
```

datadog/resource_datadog_logs_archive.go
vinckama: _I discussed a lot with @pcarioufr at the end of last week of that._

In fact, there is a divergence between UI and API.We made this choice because in the other case, we would need to create a completely new scheme for the PUT resource.
Today, the schemas for POST and PUT are the same in the archives. This means that if we want to have default tags in the archives on POST, it would be the same for PUT, it would change the behavior of the old archives. 
Another possibility is to create a new schema for the PUT resource, but this would create a backward incompatibility for users. 
We conclude that we will not enable default tags on resources created with the API. We note that if users complain about this behavior, there would still be time to update the default value and create this new schema. 

However, if you think this is not clear to you, or if you are the first user to complain about this divergence, I can update the behavior of the default value... let me know 😃

gabrielreid (approved):
Looks good to me 👍

therve (approved):

therve (approved):

PR #719, opened 2020-10-23
[MA-1516] add monitor priority

unclebconnor:


unclebconnor:
hmmm.  weird merge.  starting with a fresh one.

PR #725, opened 2020-10-28
Fix indentation in documentation

### LINKED ISSUE #723: Argument hierarchy not rendered correctly on website
On the [dashboard reference](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/dashboard) for the datadog provider, the nesting that the description references is not there:

> A widget should have exactly one of the following nested blocks describing the widget definition

It looks like this: 

Screen Shot 2020-10-27 at 12 10 13 PM

 And we can see it's meant to look differently according to the [markdown](https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/resources/dashboard.md) file. 

The rendering of nested lists might be a terraform problem but this could also be fixed with headers 

### Affected Resource(s)

- dashboard

therve:
Hashicorp rendering of markdown requires some specific indentation for
deep lists: reindent those lists to properly display them in the doc
website.

Verified with https://registry.terraform.io/tools/doc-preview

Closes #723


zippolyte (dismissed):

zippolyte:
Will need to fix the recommit check then. It doesn't seem to be happy with the new indentation

thieman:
This seems to be changing a lot more than just documentation, was that intentional?

therve:
> This seems to be changing a lot more than just documentation, was that intentional?

Yeah we use prettier over yaml files as well, but it shouldn't touch any code.

nmuesch (approved):

PR #726, opened 2020-10-28
Handle 0 in monitor new_host_delay

therve:
Terraform doesn't return 0 values on fields, but it's appropropriate for
new_host_delay on the monitor resource. Let's make sure we send it in
this case.


datadog/resource_datadog_monitor.go
```diff
@@ -274,9 +274,8 @@ func buildMonitorStruct(d BuiltResource) (*datadogV1.Monitor, *datadogV1.Monitor
 	if attr, ok := d.GetOk("notify_no_data"); ok {
 		o.SetNotifyNoData(attr.(bool))
 	}
-	if attr, ok := d.GetOk("new_host_delay"); ok {
-		o.SetNewHostDelay(int64(attr.(int)))
-	}
+	// Don't check with GetOk, doesn't work with 0
+	o.SetNewHostDelay(int64(d.Get("new_host_delay").(int)))
```
zippolyte, line 278: What if new_host_delay isn't actually set ? This will error out no ?

therve: There is a default so it should always be set I think.

zippolyte: Ah ok, do we have a test case for it to make sure  ?

therve: Yeah tests like TestAccDatadogMonitor_SilencedRemove ought to trigger this.

zippolyte (approved):

PR #728, opened 2020-10-28
replace 4d63.com/tz with time/tzdata

bendrucker:
This uses the native `time/tzdata` package added in Go 1.15 to embed time zone data. It replaces #560 and removes the `4d63.com/tz` package added there.


therve (approved):
Thanks a lot!

PR #739, opened 2020-11-02
Document monitor priority

therve:
Missed in #729

PR #741, opened 2020-11-02
Revert "Document monitor priority"

therve:
Reverts DataDog/terraform-provider-datadog#739 

We don't want to make that public public yet.


phillip-dd:
Thanks!

PR #750, opened 2020-11-12
[Synthetics] Set client certificate content as sensitive

romainberger:
Set the client certificate content in synthetics config to prevent leaking content in CLI.


therve (approved):

jdrouet (approved):
LGTM

PR #752, opened 2020-11-16
Use v2 API in User resource

therve:
This makes the User resource use the V2 API instead of V1 by default. Instances created by the V1 API still uses it for update and delete, but creation happens on V2 only.


datadog/resource_datadog_user.go
```diff
@@ -90,75 +121,186 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
...
+		d.Set("disabled", userAttributes.GetDisabled())
+		d.Set("handle", userAttributes.GetHandle())
+	} else {
+		client := providerConf.CommunityClient
+		u, err := client.GetUser(d.Id())
+		if err != nil {
+			return err
+		}
+
+		d.Set("email", u.GetEmail())
```
zippolyte, line 242: `disabled` should still be set in v1 too no ?

therve: Yeah removed it by mistake.

datadog/resource_datadog_user.go
```diff
@@ -90,75 +121,187 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
 	return &u
 }
 
+func buildDatadogUserV2Struct(d *schema.ResourceData, providerConf *ProviderConfiguration) (*datadogV2.UserCreateRequest, error) {
```
zippolyte, line 124: doesn't look like it can return any error, let's remove it

zippolyte (changes requested):
One nit, LGTM otherwise

zippolyte (approved):

zippolyte (dismissed):
Need a quick update to the documentation too

datadog/resource_datadog_user.go
```diff
@@ -66,13 +84,26 @@ func resourceDatadogUserExists(d *schema.ResourceData, meta interface{}) (b bool
 	// Exists - This is called to verify a resource still exists. It is called prior to Read,
 	// and lowers the burden of Read to be able to assume the resource exists.
 	providerConf := meta.(*ProviderConfiguration)
-	client := providerConf.CommunityClient
 
-	if _, err := client.GetUser(d.Id()); err != nil {
-		if strings.Contains(err.Error(), "404 Not Found") {
-			return false, nil
+	if isV2User(d.Id()) {
```
pietdaniel, line 88: does this ID need to be the same as the V1 API. currently the v2 returns the uuid as the id and v1 returns the primary key as the id.

therve: Not sure I completely understand the question, but we store whichever ID the API returns, and make decision based on the format like here.

datadog/resource_datadog_user.go
```diff
@@ -54,6 +67,11 @@ func resourceDatadogUser() *schema.Resource {
 				Optional:   true,
 				Deprecated: "This parameter was removed from the API and has no effect",
 			},
+			"roles": {
+				Type:     schema.TypeList,
```
zippolyte, line 71: For preventing multiple values and ignoring order
```suggestion
				Type:     schema.TypeSet,
```

datadog/resource_datadog_user.go
```diff
@@ -90,75 +121,184 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
...
+			return err
+		}
+
+		d.Set("disabled", u.GetDisabled())
+		d.Set("email", u.GetEmail())
+		d.Set("handle", u.GetHandle())
+		d.Set("name", u.GetName())
+		d.Set("verified", u.GetVerified())
+		d.Set("access_role", u.GetAccessRole())
+		d.Set("is_admin", u.GetIsAdmin())
```
jirikuncar, line 245: shall we use the email value and fetch v2 user?

therve: That'd be nice if you describe more what that means. What would that achieve?

pietdaniel: FWIW, the preferable mechanisms for fetching single users in the V2 API is the UUID. The email is dynamic and the handle is a field we'd like to move away from as the canonical identifier in favor of UUIDs as it is unique.

zippolyte (dismissed):

pietdaniel (dismissed):

datadog/resource_datadog_user_test.go
```diff
@@ -137,3 +319,34 @@ func datadogUserExistsHelper(s *terraform.State, client *datadog.Client) error {
...
+			return fmt.Errorf("received an error retrieving user %s", err)
+		}
+
+		userData := userResponse.GetData()
+		userAttributes := userData.GetAttributes()
+		// Datadog only disables user on DELETE
+		if userAttributes.GetDisabled() {
+			continue
+		}
+		return fmt.Errorf("user still exists")
```
nmuesch, line 341:
```suggestion
		return fmt.Errorf("user still enabled")
```

datadog/resource_datadog_user.go
```diff
@@ -90,75 +118,244 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
 	return &u
 }
 
+func buildDatadogUserV2Struct(d *schema.ResourceData, providerConf *ProviderConfiguration) *datadogV2.UserCreateRequest {
+	userAttributes := datadogV2.NewUserCreateAttributesWithDefaults()
+	userAttributes.SetEmail(d.Get("email").(string))
+	userAttributes.SetName(d.Get("name").(string))
```
nmuesch, line 124: If this is optional now, will if fail if not set? Should we GetOk instead

datadog/resource_datadog_user.go
```diff
@@ -90,75 +118,244 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
...
+	userRequest := datadogV2.NewUserCreateRequestWithDefaults()
+	userRequest.SetData(*userCreate)
+
+	return userRequest
+}
+
+func buildDatadogUserV2UpdateStruct(d *schema.ResourceData, userId string) *datadogV2.UserUpdateRequest {
+	userAttributes := datadogV2.NewUserUpdateAttributesWithDefaults()
+	userAttributes.SetEmail(d.Get("email").(string))
+	userAttributes.SetName(d.Get("name").(string))
```
nmuesch, line 153: Same getOk question here

datadog/resource_datadog_user.go
```diff
@@ -90,75 +118,244 @@ func buildDatadogUserStruct(d *schema.ResourceData) *datadog.User {
...
+		if httpResponse, err := datadogClientV2.UsersApi.DisableUser(authV2, d.Id()).Execute(); err != nil {
+			if httpResponse != nil && httpResponse.StatusCode == 404 {
+				return nil
+			}
+			return translateClientError(err, "error disabling user")
+		}
+	} else {
+		client := providerConf.CommunityClient
+
+		// Datadog does not actually delete users, but instead marks them as disabled.
```
nmuesch, line 350: Does v2 not have the same 400 behavior if try to disable an already disabled user?

datadog/resource_datadog_user.go
```diff
@@ -30,30 +39,36 @@ func resourceDatadogUser() *schema.Resource {
...
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: "This parameter is deprecated and will be removed from the next Major version",
 			},
 			"is_admin": {
 				Type:       schema.TypeBool,
 				Computed:   true,
 				Optional:   true,
-				Deprecated: "This parameter will be replaced by `access_role` and will be removed from the next Major version",
+				Deprecated: "This parameter is replaced by `roles` and will be removed from the next Major version",
```
nmuesch, line 50: From my read of the code, these deprecated fields will be ignored in the v2 create code path, but the user will get a deprecation warning during plan/apply. 

WDYT of checking if these are set during the create call and error-ing out if so. It means that we'd lose some backwards compatibility, but would avoid users missing expected permissions.

nmuesch (commented):
Nice! I left a few minor inline comments, but this approach looks good to me!!

Do we have recommended steps users can take to convert their existing v1 user resources to v2?

datadog/resource_datadog_user.go
therve: I don't think it will fail, but it will set it to the empty string instead of null. Using GetOk.

datadog/resource_datadog_user.go
therve: No we get a 404 in this case.

datadog/resource_datadog_user.go
therve: Yeah I think it's worth raising an error for those for security reasons.

therve: After looking into it, I'll leave like that. I had to keep the default value to access_roles so it messes up with checking the value.

nmuesch: Hmm, yeah the default throws an issue here. 👍 Thanks for looking into it

nmuesch (approved):
This looks good to me 💯 (The CI failure looks to be on an unrelated resource)
If possible, can we add a note about how to migrate from an existing user config to use the newer v2 methods?

pietdaniel (approved):

nmuesch (approved):

PR #753, opened 2020-11-17
Add roles resource and permissions datasource

zippolyte:


datadog/resource_datadog_role.go
```diff
@@ -0,0 +1,287 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the role.",
+			},
+			"permission": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: "List of permission IDs to give to this role.",
```
jirikuncar, line 35: is it list of permission IDs or objects with id and name?

zippolyte: It is a list of objects with IDs and name (name being computed so user only gives the IDs). I'm not using the name in the provider code directly anymore (needed it when I was exploring the suppressDiff approach), but I figured I'd leave it like this as I thought it might be useful for users to reference maybe.
I'm fine going back to a pure list of ID strings only, if you think it makes more sense

jirikuncar: I would only update the description to match the new schema.

zippolyte:
```suggestion
				Description: "List of objects containing the permission ID the name of the permissions assigned to this role.
```

.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
                   git config --global core.autocrlf false
                   git config --global core.eol lf
             - name: Install Go
-              uses: actions/setup-go@v2-beta
+              uses: actions/setup-go@v2.1.3
```
jirikuncar, line 41: just to get security updates
```suggestion
              uses: actions/setup-go@v2
```

datadog/data_source_datadog_permissions.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return translateClientError(err, "error listing permissions")
+	}
+	perms := res.GetData()
+	permsNameToID := make(map[string]string, len(perms))
+	for _, perm := range perms {
+		// Don't list restricted permissions, as they cannot be granted/revoked to/from a role
+		if !perm.Attributes.GetRestricted() {
+			permsNameToID[perm.Attributes.GetName()] = perm.GetId()
+		}
```
jirikuncar, line 39:
```suggestion
		// Don't list restricted permissions, as they cannot be granted/revoked to/from a role
		if perm.Attributes.GetRestricted() {
			continue
		}
		permsNameToID[perm.Attributes.GetName()] = perm.GetId()
```

jirikuncar: it's a personal preference so feel free to ignore it

jirikuncar: the happy path in main loop, error handling under if statements

datadog/resource_datadog_role.go
```diff
@@ -0,0 +1,287 @@
...
+		Exists: resourceDatadogRoleExists,
+		Create: resourceDatadogRoleCreate,
+		Read:   resourceDatadogRoleRead,
+		Update: resourceDatadogRoleUpdate,
+		Delete: resourceDatadogRoleDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceDatadogRoleImport,
+		},
+		CustomizeDiff: customdiff.ValidateValue("permission", validatePermissionsUnrestricted),
+		Schema: map[string]*schema.Schema{
```
pietdaniel, line 26: should this include the role UUID?

pietdaniel (approved):
Looks good, mostly questions, my terraform provider knowledge is still rather fresh so take the approval with a grain of salt

datadog/resource_datadog_role.go
zippolyte: ID's are specific in terraform. They are not in the schema, but they are set in the state (look for `SetId()`) and user can reference it in other resources with `resource_name.id`

datadog/cassettes/TestAccDatadogPermissionsDatasource.yaml
```diff
@@ -0,0 +1,451 @@
...
+      X-Datadog-Parent-Id:
+      - "4776416262837454408"
+      X-Datadog-Sampling-Priority:
+      - "1"
+      X-Datadog-Trace-Id:
+      - "8052966640934693994"
+    url: https://api.datadoghq.com/api/v2/permissions
+    method: GET
+  response:
+    body: '{"data":[{"type":"permissions","id":"984a2bd4-d3b4-11e8-a1ff-a7f660d43029","attributes":{"name":"admin","display_name":"Privileged
```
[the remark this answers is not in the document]
zippolyte: I don't test for the exact set of returned permissions, so this well still pass as new permissions are added

datadog/provider.go
```diff
@@ -100,6 +103,12 @@ type ProviderConfiguration struct {
 	DatadogClientV2 *datadogV2.APIClient
 	AuthV1          context.Context
 	AuthV2          context.Context
+
+	now func() time.Time
```
jirikuncar, line 107: Why not directly?
```suggestion
	Now func() time.Time
```

zippolyte: I didn't even think of that haha. Guess it felt more "go" the other way, but yeah, I guess this works too

datadog/resource_datadog_role_test.go
```diff
@@ -0,0 +1,187 @@
...
+				Config: testAccCheckDatadogRoleConfigNoPerm(rolename + "noperm"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDatadogRoleExists(accProvider, "datadog_role.foo"),
+					resource.TestCheckResourceAttr("datadog_role.foo", "permission.#", "0"),
+				),
+			},
+		},
+	})
+}
+func TestAccDatadogRole_RestrictedPerm(t *testing.T) {
```
nmuesch, line 70: Super small nit, but s/restricted/invalid since this permission doesn't exist
```suggestion
func TestAccDatadogRole_InvalidPerm(t *testing.T) {
```

datadog/resource_datadog_monitor.go
```diff
@@ -509,7 +508,7 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
 
 	// Ignore any timestamps in the past that aren't -1 or 0
 	for k, v := range configSilenced {
-		if v.(int) < int(time.Now().Unix()) && v.(int) != 0 && v.(int) != -1 {
+		if v.(int) < int(providerConf.Now().Unix()) && v.(int) != 0 && v.(int) != -1 {
```
nmuesch, line 511: Not sure I fully understand what this change addresses?

datadog/resource_datadog_role.go
```diff
@@ -0,0 +1,288 @@
...
+
+func getRolePermissionSchema() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "ID of the permission to assign.",
+			},
+			"name": {
```
nmuesch, line 99: Is this mostly here for the ability to reference?

datadog/resource_datadog_role.go
```diff
@@ -0,0 +1,288 @@
...
+				Computed:    true,
+				Description: "Number of users that have this role.",
+			},
+		},
+	}
+}
+
+func getValidPermissions(client *datadog.APIClient, auth context.Context) (map[string]string, error) {
+	// Get a list of all permissions, to ignore restricted perms
+	if validPermissions == nil {
```
nmuesch, line 49: Nice 💯

nmuesch (approved):
This looks great! Left a few inline notes/comments.

nmuesch (changes requested):
Sorry, I meant to "request changes" to discuss my above notes 🤦

datadog/data_source_datadog_permissions.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceDatadogPermissions() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDatadogPermissionsRead,
+
+		Schema: map[string]*schema.Schema{
+			// Computed values
+			"permissions": {
```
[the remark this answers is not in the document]
zippolyte: Honestly, I don't know. And I do need this mapping so that it's practical for user to reference in other resources. A list wouldn't allow that

datadog/data_source_datadog_permissions_test.go
```diff
@@ -0,0 +1,28 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: accProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: `data "datadog_permissions" "foo" {}`,
+				// Check at least one permission exists
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.datadog_permissions.foo", "id", "datadog-permissions"),
+					resource.TestCheckResourceAttrSet("data.datadog_permissions.foo", "permissions.admin"),
+					resource.TestCheckNoResourceAttr("data.datadog_permissions.foo", "permissions.dashboards_read"),
```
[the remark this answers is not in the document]
zippolyte: yup

datadog/resource_datadog_monitor.go
zippolyte: problem in replaying tests, since we do a match on the body now. So we need to be able to set the frozen clock in the test, this allows us to do that

datadog/resource_datadog_role.go
zippolyte: Yeah. I'm not sure about this one though. At first I had a simple string list of IDs, then I had to add it for my first try at implanting `diffSupressFunc` that i didn't end up doing, but kept this as is 🤷

datadog/data_source_datadog_permissions.go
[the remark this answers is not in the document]
nmuesch: Yeah you're right, we don't want users to have to care about the order of the list...

I just want to make sure will be able to add new fields in the future without compatibility issues if needed. (But maybe we won't need it) Another possibility may be this, though I'm unsure if that means you'd need to know the hash
```permissions: [
    {
        "name": {
            "id": <id>,
            "future_fields": ....,
         }
    }
]```

datadog/resource_datadog_monitor.go
nmuesch: Ahh ok, so no functional user difference, just for tests 👍

datadog/data_source_datadog_permissions.go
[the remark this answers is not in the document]
zippolyte: So in a map I can only have primitive types though
 ```Elem represents the element type. For a TypeMap, it must be a *Schema with a Type that is one of the primitives: TypeString, TypeBool, TypeInt, or TypeFloat```

nmuesch (approved):
Thanks, This looks good to me 👍

PR #754, opened 2020-11-18
Reflect synthetics type in resource name example

filiptepper:
I know it's minor and kinda nit-picky, but I have just noticed this minor inconsistency in the docs.


romainberger (approved):
Good catch!

PR #763, opened 2020-11-24
[SA-535]Add support for security monitoring rules

### LINKED ISSUE #509: [Feature Request] Support security detection rules
Given the new [security detection rules](https://www.datadoghq.com/blog/announcing-security-monitoring/), I'd very much like to be able to configure this through the Terraform provider as well.

greentruff:
Add support for Security Monitoring public API.
See: https://docs.datadoghq.com/api/v2/security-monitoring/

Closes #509


datadog/data_source_datadog_security_monitoring_rules.go
```diff
@@ -0,0 +1,246 @@
...
+			"default_only_filter": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Limit the search to default rules",
+			},
+			"user_only_filter": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Limit the search to user rules",
+			},
```
greentruff, line 42: Two different attributes are used for boolean values to work around `GetOk` only checking if there was truthy value.

datadog/data_source_datadog_security_monitoring_rules.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	d.SetId(computeSecMonDataSourceRulesId(nameFilter, defaultFilter, tagFilter))
+	d.Set("rules", rules)
+	d.Set("rule_ids", ruleIds)
+
+	return nil
+}
+
+func computeSecMonDataSourceRulesId(nameFilter *string, defaultFilter *bool, tagFilter map[string]bool) string {
```
greentruff, line 133: This is a costly way to get a hash, I am not sure if this matters

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,558 @@
...
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "The target field to aggregate over when using the sum or max aggregations.",
+					},
+					"name": {
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "Name of the query.",
+					},
+					"query": {
```
greentruff, line 144: `q` may be closer to the existing terraform API. I mostly used the same field names as the public API.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,558 @@
...
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Whether the rule is enabled.",
+		},
+
+		"disabled": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Whether the rule is disabled.",
+		},
```
btoueg, line 68: Why do we need two fields?

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,558 @@
...
+		payloadCases[idx] = *structRuleCase
+	}
+	payload.Cases = payloadCases
+
+	_, enabled := d.GetOk("enabled")
+	_, disabled := d.GetOk("disabled")
+
+	if enabled && disabled {
+		return payload, errors.New("rule can not be both enabled and disabled")
+	}
```
btoueg, line 231: Can it be both "not enabled" and "not disabled"?

datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,278 @@
...
+			"enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Enable the rule.",
+			},
+
+			"disabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Disable the rule.",
```
therve, line 59: Same as the filter, would need some more details.

datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,278 @@
...
+import (
+	"errors"
+	"fmt"
+	datadogV2 "github.com/DataDog/datadog-api-client-go/api/v2/datadog"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceDatadogSecurityMonitoringDefaultRule() *schema.Resource {
+	return &schema.Resource{
+		Exists: resourceDatadogSecurityMonitoringDefaultRuleExists,
```
therve, line 12: You can remove exist method, this is not useful in the end.

datadog/resource_datadog_security_monitoring_rule.go
greentruff: https://github.com/DataDog/terraform-provider-datadog/pull/763#discussion_r530557706
It is a limitation in the way to check for optional fields in terraform `d.GetOk()` will only return `ok = true` if a value is present and truthy. Having only an enabled field means that rules would need to be disabled by default. This would be consistent with [logs_custom_pipelines](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/logs_custom_pipeline) but it seems strange to me to have to specify `enabled: true` for each rule resource.

docs/resources/security_monitoring_rule.md.md
```diff
@@ -0,0 +1,83 @@
+---
```
therve, line 1: There is an extra ".md" in the file.

datadog/resource_datadog_security_monitoring_rule.go
greentruff: In this case, neither value has been defined in the conf and the value defaults to `true`

docs/resources/security_monitoring_rule.md.md
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    query {
+        name = "warnings"
+        query = "status:warning"
+        aggregation = "count"
+        group_by_fields = ["host"]
+    }
+
+ 	case {
```
therve, line 35: Indentation is wrong.

therve (changes requested):
Looks great! Some minor question/remarks, plus need clarifications around boolean usage. Need to check why CI is red too.

datadog/resource_datadog_security_monitoring_rule.go
therve: Could we set the default value of enabled to true instead?

greentruff: Ah! I was trying to implement default in code instead of using the default in the resource definition. I'll do that!

Bit-Doctor (commented):
I've made a first pass at custom security rules. 
I'll check default rule resource and data source in a bit.

So far looks good with a few nits.

docs/resources/security_monitoring_default_rule.md
```diff
@@ -0,0 +1,51 @@
...
+    default_only_filter = true
+    tags_filter = ["security:attack"]
+}
+
+# Create a resource for each default rule id and enable the rule
+resource "datadog_security_monitoring_default_rule" "attack_rule" {
+    for_each = toset(data.datadog_security_monitoring_rules.rules.rule_ids)
+    rule_id = each.key
+    enabled = true
+}
```
btoueg, line 50: IIUC this is the motivation behind `datadog_security_rules` datasource. It is a peculiar use of datasources AFAICT. For instance, how does it behave if you create datasources that overlaps? Say you have `attack_default_rules` and `specific_framework` rules. A rule could belong to both, and would result in multiple `datadog_security_monitoring_default_rule`. My guess is that in this case, last write wins. but with lots of subtleties. Are we comfortable to recommend such usage?

I would have recommended dealing with default rules with a prior import.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,539 @@
...
+	return map[string]*schema.Schema{
+		"case": {
+			Type:        schema.TypeList,
+			Required:    true,
+			Description: "Cases for generating signals.",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"name": {
+						Type:        schema.TypeString,
+						Optional:    true,
```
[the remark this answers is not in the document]
btoueg: There's a trade-off between client-side validation and server-side validation. I think it easier and more maintenable to delegate most of it to the server, but we have to make sure the error message is descriptive and visible.

datadog/data_source_datadog_security_monitoring_rules.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if v, ok := d.GetOk("name_filter"); ok {
+		filter := v.(string)
+		nameFilter = &filter
+	}
+
+	_, filterDefault := d.GetOk("default_only_filter")
+	_, filterUser := d.GetOk("user_only_filter")
+	if filterDefault && filterUser {
+		return errors.New("error: cannot filter both default and user rules")
```
Bit-Doctor, line 78: If you keep both attributes think you might be able to handle that with `ConflictsWith: []string{"user_only_filter"}` in `default_only_filter` schema definition.

docs/resources/security_monitoring_default_rule.md
Bit-Doctor: IIUC your edge case is something like this ?
```terraform
resource "datadog_security_monitoring_default_rule" "rule1" {
    rule_id = "SAMEID"
    enabled = true
}

resource "datadog_security_monitoring_default_rule" "rule2" {
    rule_id = "SAMEID"
    enabled = false
}
```

I'm actually unsure what whould Terraform do here, It will probably conflict on the resource ID depending on how you build it.

greentruff: I am asking @NicolasLevaique what he thinks about this as well

datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,261 @@
...
+	// create only updates an existing rule
+	err := resourceDatadogSecurityMonitoringDefaultRuleUpdate(d, meta)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("rule_id").(string))
+
+	return nil
+}
```
Bit-Doctor, line 74: IIRC The way "fake" resource like this one are handled in official provider is to not define an `Update` method and have `ForceNew` attribute on all the schema elements.

Not sure about the benefits of doing this 🤷 Might help with the conflict @btoueg talked.

datadog/data_source_datadog_security_monitoring_rules.go
greentruff: Updated with a default value

datadog/data_source_datadog_security_monitoring_rules_test.go
```diff
@@ -0,0 +1,106 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    accProviders,
+		CheckDestroy: testAccCheckDatadogSecurityMonitoringRuleDestroy(accProvider),
+		Steps: []resource.TestStep{
+			{
+				// Create a rule to make sure we have at least one non-default rule
+				Config: testAccCheckDatadogSecurityMonitoringCreatedConfig(ruleName),
+			},
+			{
+				// Ideally we would like to check the size and content of the lists, but the terraform
```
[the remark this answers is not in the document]
greentruff: I went over it again and figured out how to test with size in the end.

datadog/resource_datadog_security_monitoring_default_rule.go
greentruff: Removed

datadog/resource_datadog_security_monitoring_rule_test.go
```diff
@@ -0,0 +1,453 @@
...
+
+	case {
+        name = "high case"
+        status = "high"
+        condition = "first > 3 || second > 10"
+        notifications = ["@user"]
+    }
+
+    case {
+		name = "warning case"
```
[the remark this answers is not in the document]
greentruff: tabs vs spaces 😞

datadog/resource_datadog_security_monitoring_default_rule.go
Bit-Doctor: I think you missed this `enabled/disabled` actually 🤔

greentruff: @Bit-Doctor fixed

docs/resources/security_monitoring_default_rule.md
Bit-Doctor: Also I just thought of a resource in the AWS provider that has a similar behavior, with no real `create/delete`.
So at least this behavior exist.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_vpc
https://github.com/hashicorp/terraform-provider-aws/blob/master/aws/resource_aws_default_vpc.go

Might be worth checking if they handle this edge case with two resource with different params.

Bit-Doctor: Also maybe a bit late but there's `setintersection`, `setunion` and `setsubtract` functions that could have helped user with that issue.

greentruff:
@therve Can you take another look? After discussing with @NicolasLevaique, I changed the default rule so that it needs to be imported like other resources. I kept the datasource though since it can be useful and I don't know how to test the default rule without it. 
There is still the question of the `default_only_filter`/`user_only_filter` in the datasource: in this case there are three valid states, no filter or filter by user/default rules. Merging the attributes would mean that we would have to always return either default or user rules. Do you think it is ok to leave both? The alternative is to change the format to something like 
```
filter {
"is_default_rule": true,
}
```
but this would not be consistent with other datasources in the provider.

datadog/data_source_datadog_security_monitoring_rules.go
therve: You're right, because you use `nil` as a possible value, so there are effectively 3 states, not just 2.

datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,212 @@
...
+
+	ruleUpdate, shouldUpdate, err := buildSecMonDefaultRuleUpdatePayload(response, d)
+
+	if err != nil {
+		return err
+	}
+
+	if shouldUpdate {
+		_, _, err := datadogClientV2.SecurityMonitoringApi.UpdateSecurityMonitoringRule(authV2, ruleId).Body(*ruleUpdate).Execute()
+		if err != nil {
```
therve, line 126: You should inline that if with the previous command.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,553 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+
+	datadogV2 "github.com/DataDog/datadog-api-client-go/api/v2/datadog"
+)
+
+func resourceDatadogSecurityMonitoringRule() *schema.Resource {
+	return &schema.Resource{
+		Exists: resourceDatadogSecurityMonitoringRuleExists,
```
therve, line 11: You should remove the exists, and check for 404 in read instead (and set the id to "" when you get a 404)

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,553 @@
...
+}
+
+func resourceDatadogSecurityMonitoringRuleDelete(d *schema.ResourceData, meta interface{}) error {
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV2 := providerConf.DatadogClientV2
+	authV2 := providerConf.AuthV2
+
+	_, err := datadogClientV2.SecurityMonitoringApi.DeleteSecurityMonitoringRule(authV2, d.Id()).Execute()
+
+	if err != nil {
```
therve, line 548: You can inline that condition too.

docs/resources/security_monitoring_default_rule.md
```diff
@@ -0,0 +1,44 @@
...
+
+Default rules need to be imported using their ID before applying.
+
+```hcl
+resource "datadog_security_monitoring_default_rule" "adefaultrule" {
+}
+```
+
+```
+terraform import datadog_security_monitoring_default_rule.adefaultrule
```
therve, line 43: You're missing an ID here.

therve (changes requested):
A few minor comments, otherwise almost there.

datadog/resource_datadog_security_monitoring_rule.go
greentruff: Done in the last commit, can you double check how I did it though? I didn't find any other resources which omit `exists`.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,539 @@
...
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV2 := providerConf.DatadogClientV2
+	authV2 := providerConf.AuthV2
+
+	id := d.Id()
+	ruleResponse, httpResponse, err := datadogClientV2.SecurityMonitoringApi.GetSecurityMonitoringRule(authV2, id).Execute()
+	if err != nil {
+		if httpResponse != nil && httpResponse.StatusCode == 404 {
+			d.SetId("")
+			return errors.New("security monitoring rule does not exist")
```
therve, line 313: Almost that, just return nil in that case instead of an error. Thanks.

therve (changes requested):

therve (approved):

PR #770, opened 2020-12-03
Add example for initializing provider with tf 0.13+

nmuesch:
Add a note about how to declare the provider to use the Terraform Registry when using terraform 0.13 and greater.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,6 +12,37 @@ Use the navigation to the left to read about the available resources.
 
 ## Example Usage
 
+Terraform 0.13+ uses the Terraform Registry:
+```hcl
+terraform {
+  required_providers {
+    datadog = {
+      source  = "DataDog/datadog"
+      # version = "~> 1.0"
+    }
+  }
+}
+
+# Configure the Datadog provider
+provider "datadog" {
+  api_key = "${var.datadog_api_key}"
+  app_key = "${var.datadog_app_key}"
+}
+
+# Create a new monitor
+resource "datadog_monitor" "default" {
+  # ...
+}
+
+# Create a new timeboard
+resource "datadog_timeboard" "default" {
+  # ...
+}
+
+```
+
+Terraform 0.12- can be specified as:
+
 ```hcl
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -12,6 +12,37 @@ Use the navigation to the left to read about the available resources.
 
 ## Example Usage
 
+Terraform 0.13+ uses the Terraform Registry:
+```hcl
+terraform {
+  required_providers {
+    datadog = {
+      source  = "DataDog/datadog"
+      # version = "~> 1.0"
```
therve, line 21: Maybe remove that line? I'm afraid people are just going to uncomment it, and I don't really want to list a version here.

docs/index.md
```diff
@@ -12,6 +12,37 @@ Use the navigation to the left to read about the available resources.
...
+    }
+  }
+}
+
+# Configure the Datadog provider
+provider "datadog" {
+  api_key = "${var.datadog_api_key}"
+  app_key = "${var.datadog_app_key}"
```
therve, line 29: Can you take the chance to use the new syntax without `"${}"` ?

docs/index.md
```diff
@@ -12,6 +12,37 @@ Use the navigation to the left to read about the available resources.
...
+}
+
+# Create a new monitor
+resource "datadog_monitor" "default" {
+  # ...
+}
+
+# Create a new timeboard
+resource "datadog_timeboard" "default" {
```
therve, line 38: dashboard please! If you can change it below too that'd be great.

therve (changes requested):

docs/index.md
nmuesch: Fair point, I originally left to show an example of where you'd put the version in this setup, but we never showed it the old way either, so removing.

docs/index.md
nmuesch: I tried with terraform 0.14, the variables themselves are defined the same so I didn't include it here, but it works to reference them without all the `${}`

therve (approved):

PR #771, opened 2020-12-03
Add force_delete to SLO resource

nmuesch:
Add `force_delete` to the SLO resource schema in the same way we have it for the monitor resource. 

This already exists in the API layer, so this brings TF up to date.


phillip-dd (approved):
lgtm, thanks!

therve (approved):

PR #774, opened 2020-12-07
Add ForceNew to `layout_type` dashboard attribute

### LINKED ISSUE #772: Error when updating Dashboard from `ordered` layout type to `free` layout type.
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

```
v0.12.29 on Terraform Enterprise.
```

### Affected Resource(s)

-   datadog_dashaboard.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```
Reach c-marin.salinas@hulu.com for some additional info.
### Debug Output

Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output

If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Actual Behavior

I found something on dd provider, if I want to update a dashboard from ordered  to free  layout, I got the following error:

```
Error: error updating dashboard: 400 Bad Request: {"errors": ["'layout_type' is not editable for existing dashboards. This dashboard uses ordered layout."]}
```

### Expected Behavior

The dashboard will update in place or I'm wondering if `layout_type`  should be changed as ForceNew attribute to avoid this error?
https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_dashboard.go#L52

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Since we are writing Terraform modules for all our organization we used to have an ordered layout in v1.2.4 of our Datadog module so then we upgraded the module to use free layout and the users who jump to this new version are getting this error.

A workaround of this is to run `terraform taint` to force the dashboard recreation but since we are using Terraform Enterprise the users don't have permissions to perform that manual actions.

zippolyte:
Add ForceNew to `layout_type` dashboard attribute so that users can change the layout type of a dashboard.
Note that changing the type only is not sufficient, a `layout` property need to be added/removed to/from all the widgets in that dashboard depending on the change. Also a few widgets might be available in one type and not the other, which is another thing to lookout for when changing the layout type

fixes #772


matt-miller-ddog:
We don't support changing a dashboard's layout type; our API will reject these payloads even if our Terraform provider doesn't.

zippolyte:
That's the thing, it won't try to update the dashboard, it will just delete the old one, and create a new one from scratch

datadog/resource_datadog_dashboard_test.go
```diff
@@ -452,6 +452,56 @@ resource "datadog_dashboard" "ordered_dashboard" {
...
+func datadogSimpleOrderedDashboardConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_dashboard" "simple_ordered_dashboard" {
+	title         = "%s"
+	description   = "Created using the Datadog provider in Terraform"
+	layout_type   = "ordered"
+	is_read_only  = true
+	widget {
+		alert_graph_definition {
+			alert_id = "895605"
```
nmuesch, line 464: Should this be a random value like `123456` ?

datadog/resource_datadog_dashboard_test.go
```diff
@@ -634,6 +684,130 @@ resource "datadog_dashboard" "free_dashboard" {
...
+func datadogSimpleFreeDashboardConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_dashboard" "simple_free_dashboard" {
+	title         = "%s"
+	description   = "Created using the Datadog provider in Terraform"
+	layout_type   = "free"
+	is_read_only  = true
+	widget {
+		alert_graph_definition {
+			alert_id = "895605"
```
nmuesch, line 696: Same

nmuesch (approved):
LGTM, minor question/note about the alert id, but other than that this seems 💯

datadog/resource_datadog_dashboard_test.go
```diff
@@ -634,6 +684,130 @@ resource "datadog_dashboard" "free_dashboard" {
 }`, uniq)
 }
 
+func datadogSimpleFreeDashboardConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_dashboard" "simple_free_dashboard" {
```
therve, line 689: You need to keep the same resource name in your update, otherwise terraform ends doing a delete/create and you don't test forceNew

therve (changes requested):

datadog/resource_datadog_dashboard_test.go
zippolyte: isn't 895605 random ?

datadog/resource_datadog_dashboard_test.go
zippolyte: 🤦

datadog/resource_datadog_dashboard_test.go
nmuesch: Sorry, I meant fake. Wasn't sure if this was a real id or not 🙂

zippolyte: well, it's not exactly sensitive anyway. Plus I just c/p it from the other dashboard defs

zippolyte: pretty sure 123456 is also a real id ;)

nmuesch: hahaha, fair enough :)

therve (approved):

nmuesch (approved):

matt-miller-ddog:
I get the idea behind ForceNew now, but I'm still worried about approving. The issue is with widgets' `layout` blocks. Traditionally they're screenboard-only, and our API will force users to add/remove them when changing layout types. But the Multi-Size Layout project uses `layout` in timeboard widgets as well, using the same structure but for different purposes and with different expected values. That means that for orgs with MSL enabled, the API will allow users to switch from `free` to `ordered` (or the other way, once MSL goes GA and makes its way into TF) and produce a corrupt / bad-looking board.

(But if we _don't_ do this PR, TF users may end up renaming their resources with new types to accomplish the same thing in a hackier way, which leaves us in the same situation.)

matt-miller-ddog:
Clarification: if Terraform is able to detect and prevent using `layout` on TBs + not using `layout` on SBs and prevent the API calls appropriately, this should be fine; if we're counting on the API to do it then this will become an issue once we start rolling out MSL.

therve:
> Clarification: if Terraform is able to detect and prevent using `layout` on TBs + not using `layout` on SBs and prevent the API calls appropriately, this should be fine; if we're counting on the API to do it then this will become an issue once we start rolling out MSL.

I suppose terraform could do that, but I'd rather not add more logic if we can avoid it. I'm not exactly sure how it's related to the issue at hand?

IIUC, right now you can either specify `ordered`, and no layout, or `free`, and layout everything. If you try to change the type, the API fails. In the future, with MSL, you can specify layout in `ordered`? And thus the API will allow changing the type? If that's the case, I agree we probably shouldn't merge this.

matt-miller-ddog (approved):
> IIUC, right now you can either specify ordered, and no layout, or free, and layout everything. If you try to change the type, the API fails. In the future, with MSL, you can specify layout in ordered? And thus the API will allow changing the type? If that's the case, I agree we probably shouldn't merge this.

The API still won't allow `Dashboard.update` to change board types even with MSL (meaning that we'd still need `ForceNew` to do it); but other than that your statement is accurate, and TB payloads will have `layout` in them to manage MSL layout. As it stands now, the API won't be able to tell the difference between a SB `layout` vs an MSL TB `layout` and will allow users to change their resources' layout types without also forcing them to fix their `layout` values.

I'd classify this as "a future dashboards backend bug that will apply to TF users once MSL is rolled out" (and not merging this reduces its severity but also interferes with this layout_type edit bugfix) so I'll see if dashboards-backend is willing to address this.

At any rate, this future MSL bug will be present for TF users with MSL enabled regardless of whether this PR is merged or not, so I'm willing to approve this PR to fix the existing bug report. I'll raise awareness of this situation so that hopefully we can dodge the future MSL layout bug before it happens.

PR #778, opened 2020-12-08
Automatically upgrade users when `roles` is set

### LINKED ISSUE #777: [v2.18.0] Existing Admin user converted to Standard
### Terraform Version

Terraform v0.13.5
+ provider registry.terraform.io/datadog/datadog v2.18.0
+ provider registry.terraform.io/hashicorp/aws v3.20.0

### Affected Resource(s)

-   datadog_user
-   datadog_role (data source)

Before the change:
```hcl
resource "datadog_user" "user" {
  count       = 1 - max(local.disabled ? 1 : 0, local.client ? 1 : 0)
  email       = "${var.username}@contoso.com"
  handle      = "${var.username}@contoso.com"
  name        = "${title(element(split(".", var.username), 0))} ${title(element(split(".", var.username), 1))}"
  disabled    = local.disabled
  access_role = local.is-sre || local.datadog-only-admin ? "adm" : "st"
  lifecycle {
    ignore_changes = [name]
  }
}
```
After the change:
```hcl
data "datadog_role" "standard" {
  filter = "Datadog Standard Role"
}

data "datadog_role" "admin" {
  filter = "Datadog Admin Role"
}

resource "datadog_user" "user" {
  count       = 1 - max(local.disabled ? 1 : 0, local.client ? 1 : 0)
  email       = "${var.username}@contoso.com"
  name        = "${title(element(split(".", var.username), 0))} ${title(element(split(".", var.username), 1))}"
  disabled    = local.disabled
  roles       = local.is-sre || local.datadog-only-admin ? [data.datadog_role.admin.id] : [data.datadog_role.standard.id]
  lifecycle {
    ignore_changes = [name]
  }
}
```

### Expected Behavior

After refactoring the code to remove the deprecated attributes and using the new `roles` list, an admin user would keep its admin role after the apply

### Actual Behavior

1. (OK) A user with the standard role kept its role 
1. (NOK) The admin user was downgraded to the Standard role, although the role id showed in the plan was clearly different from the users assigned to the Standard role.
1. (NOK) After the apply the handle attribute still appears in the plan and the deprecated `access_role` attribute is being changed from "adm" to "st" : 
```
  # module.user.datadog_user.user[0] will be updated in-place
  ~ resource "datadog_user" "user" {
      ~ access_role = "adm" -> "st"
        disabled    = false
        email       = "first.last@contoso.com"
      - handle      = "first.last@contoso.com" -> null
        id          = "first.last@contoso.com"
        is_admin    = true
        name        = "First Last"
        roles       = [
            "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXX5f",
        ]
        verified    = true
    }

```

What is the right way to convert existing users to the new provider version? I see this was asked in the [PR](https://github.com/DataDog/terraform-provider-datadog/pull/752#pullrequestreview-543294009) but couldn't find an answer. Should I keep access_role along with `roles` until the next provider version? How does that affect existing users? Differentiating existing and new users isn't really feasible in a module approach.

Thanks.

therve:
If roles field is set we need to make sure that we're using v2,
otherwise we fall back on the standard access role.

Closes #777

### DIFF
--- a/datadog/resource_datadog_user.go
+++ b/datadog/resource_datadog_user.go
@@ -258,6 +258,24 @@ func resourceDatadogUserRead(d *schema.ResourceData, meta interface{}) error {
 func resourceDatadogUserUpdate(d *schema.ResourceData, meta interface{}) error {
 	providerConf := meta.(*ProviderConfiguration)
 
+	if !isV2User(d.Id()) && (d.Get("roles").(*schema.Set)).Len() > 0 {
+		datadogClientV2 := providerConf.DatadogClientV2
+		authV2 := providerConf.AuthV2
+		email := d.Get("email").(string)
+		log.Printf("[INFO] Migrating existing Datadog user %s", email)
+		// Find user ID by listing user and filtering by email
+		listResponse, _, err := datadogClientV2.UsersApi.ListUsers(authV2).Filter(email).Execute()
+		if err != nil {
+			return translateClientError(err, "error searching user")
+		}
+		responseData := listResponse.GetData()
+		if len(responseData) != 1 {
+			return fmt.Errorf("could not find single user with email %s", email)
+		}
+		userId := responseData[0].GetId()
+		d.SetId(userId)
+	}
+
 	if isV2User(d.Id()) {
 		datadogClientV2 := providerConf.DatadogClientV2
 		authV2 := providerConf.AuthV2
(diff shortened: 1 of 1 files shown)


zippolyte (approved):
LGTM
Did a manual test and it seems to work OK

pietdaniel (dismissed):

zippolyte (approved):

fishfacemcgee:
@therve Does upgrading to the v2 client make the provider ignore the `handle` property? Something else mentioned in #777 and that I'm seeing on my end is that removing the deprecated `handle` attribute causes a constant dirty state on the user resource.

Issue #783, opened 2020-12-11
Error when trying to use "matches" operator with "statusCode"


zachkknowbe4:
Even though the DataDog API fully supports sending,
```
"assertions": [
            {
                "operator": "matches",
                "type": "statusCode",
                "target": "20[04]"
            }
        ]
```

The provider is changing the JSON structure to,
```
"assertions": [
            {
                "operator": "matches",
                "type": "statusCode",
                "target": 0
            }
        ]
```
Which obviously causes the API to throw an error because it wasn't expecting a `0`. Not sure why the API docs and this provider have not been updated to show that this combination is supported.

As you can see below, the web interface also supports setting status code to a regex,
![image](https://user-images.githubusercontent.com/42388953/101949316-68ecac00-3bc1-11eb-8c91-ef9ca93bbd1d.png)

### Terraform Version
Terraform v0.13.5
+ provider registry.terraform.io/datadog/datadog v2.18.1
+ provider registry.terraform.io/hashicorp/aws v3.21.0
+ provider registry.terraform.io/hashicorp/null v3.0.0

### Affected Resource(s)

-   datadog_synthetics_test

### Terraform Configuration Files

```hcl
resource "datadog_synthetics_test" "test_api" {
  for_each = var.internal_apps

  type = "api"
  request = {
    method = "GET"
    url    = "https://${each.value}"
  }
  tags = ["env:${terraform.workspace}"]
  assertion {
      type     = "statusCode"
      operator = "matches"
      target   = "20[04]"
  } 
...
```

### Debug Output
```
2020/12/11 14:37:00 [WARN] Provider "registry.terraform.io/datadog/datadog" produced an unexpected new value for datadog_synthetics_test.test_api["gitlab_readiness"], but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .assertion[0].target: was cty.StringVal("20[40]"), but now cty.StringVal("0")
```
---
```
Error: error updating synthetics test: 400 Bad Request: {"errors": ["0.type: 0 is not of type 'string'\n\nFailed validating 'type' in schema[0]:\n    {'maxLength': 51200, 'type': 'string'}\n\nOn instance:\n    0"]}
```

### Expected Behavior

It should have created the synthetic monitor using the regex.

### Actual Behavior

It converted the regex to 0 and made the API request fail to DataDog.

### Steps to Reproduce

1. Create a `datadog_synthetics_test` resource.
[...]

PR #784, opened 2020-12-14
Fix error when setting synthetics status code assertion with regex

zippolyte:
Fixes https://github.com/DataDog/terraform-provider-datadog/issues/783

The target type is int only for operators `is` and `isNot` for the type `statusCode`, and string for operators `matches` and `doesNotMatch`


romainberger (approved):
👍  LGTM for Synthetics

therve (approved):

PR #794, opened 2020-12-17
Validate enum values

### LINKED ISSUE #787: Widget with invalid palette name fails with incomprehensible error and stays broken until dashboard is deleted
When a dashboard is created with a widget using an invalid palette name, the dashboard creates with the widget and fails terraform with a horrible json output that isn't helpful in any way to identify what failed.  Even worse once the problem is identified and fixed re-applying does not fix the problem it will continue to keep failing.  The only way once the problem is identified and fixed to resolve is to delete the dashboard and re-apply.

### Terraform Version

```
Terraform v0.13.5
+ provider registry.terraform.io/datadog/datadog v2.18.1
+ provider registry.terraform.io/hashicorp/aws v3.21.0
+ provider registry.terraform.io/hashicorp/local v2.0.0
+ provider registry.terraform.io/hashicorp/null v3.0.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
```

### Affected Resource(s)

-   datadog_dashboard

```hcl
resource "datadog_dashboard" "lb" {
  title = "LB"
  layout_type = "ordered"
  is_read_only = true

  template_variable {
    name = "scope"
    prefix = "use-case"
    default = "funstuff"
  }

  template_variable {
    name = "name"
    prefix = "name"
    default = "default"
  }

  widget {
    query_value_definition {
      title = "Response time (avg)"
      request {
        q = "avg:aws.applicationelb.target_response_time.average{$scope,$name} * 1000"
        aggregator = "avg"
        conditional_formats {
          comparator = ">"
          value = 500
          palette = "white_on_red"
        }
        conditional_formats {
          comparator = ">"
          value = 400
          palette = "white_on_yello" # BAD palette name
        }
        conditional_formats {
          comparator = "<="
          value = 400
          palette = "white_on_green"
        }
      }
      autoscale = false
      custom_unit = "ms"
      precision = 0
    }
  }
}
```

### Output

```
Error: error checking dashboard exists: Data failed to match schemas in oneOf(WidgetDefinition): {"notify_list":[],"description":null,"author_name": ... horrible incomprehensible json object here}
```

### Expected Behavior

Expect that if a resource value is not valid that a clear concise error explaining the problem is displayed.

In this case probably "Error: unknown color palette name: white_on_yello" with context of which resource/widget line number.

### Actual Behavior

Dashboard was created with invalid widget.  All subsequent attempts to apply will fail, even after the palette name is fixed.  The dashboard has to be removed (or maybe manually fixed) and then the resource can apply changes w/o errors.

### Steps to Reproduce

1. create a dashboard like above with invalid palette name
2. `terraform apply`
3. fix the color name
4. `terraform apply` this will still fail
5. delete the dashboard
6. `terraform apply`this will function properly

zippolyte:
Add a validate function to all enum values for clear error messages and fail early to avoid sending bad payloads to API
Fixes #787


therve (dismissed):

romainberger (dismissed):
LGTM for Synthetics 👍

datadog/resource_datadog_dashboard.go
```diff
@@ -5622,25 +5653,234 @@ func buildTerraformHostmapRequestStyle(datadogStyle datadogV1.HostMapWidgetDefin
 }
 
 // Schema validation
+func validateWidgetPalette(val interface{}, key string) (warns []string, errs []error) {
+	_, err := datadogV1.NewWidgetPaletteFromValue(val.(string))
+	if err != nil {
```
therve, line 5658: With a little bit of elbow grease, I feel that we could define a wrapper function that you can directly assign to `ValidateFunc`. Seems non trivial. In the mean time, please inline the err check like `if _, err := ..; err != nil`

therve (changes requested):
Looks good, just a small but annoying comment.

datadog/validators.go
```diff
@@ -18,3 +23,17 @@ func validateAggregatorMethod(v interface{}, k string) (ws []string, errors []er
...
 }
+
+// validateEnumValue returns a validate func for an enum value. It takes the constructor with validation for the enum as an argument.
+// Such a constructor is for instance `datadogV1.NewWidgetLineWidthFromValue`
+func validateEnumValue(newEnumFunc interface{}) schema.SchemaValidateFunc {
+	return func(val interface{}, key string) (warns []string, errs []error) {
+		arg := reflect.ValueOf(val)
+		outs := reflect.ValueOf(newEnumFunc).Call([]reflect.Value{arg})
+		err := outs[1].Interface()
+		if err != nil {
```
therve, line 34: Please inline the condition withe previous line.

therve (approved):
That's awesome, one minor nit.

nmuesch (approved):

PR #798, opened 2020-12-21
Start using tfplugindocs

nmuesch:
Start using the [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) project to generate the documentation pages

**NOTE** Anything under the `docs` folder is now generated using tplugindocs, and anything else was handwritten

To do that, this PR:
* Bumps the sdk to the next minor to resolve hitting this issue - https://github.com/hashicorp/terraform-plugin-test/issues/42
* Created a new `examples/data-sources` folder to include the hard-coded examples we show. These can't currently be generated
* Created a new `examples/resources` folder to include hard-coded import.sh and resource.tf files. These are used to show the example config and how to import, and currently can't be generated
* ~Add a `Description` field to each schema element so that it can be generated into the docs~ Descriptions were followed up on in #803 
* Adds a `templates` folder so that we can still have our special `guides` section where we hardcode different monitor examples
* Adds a new `tools` folder as recommended by the project to define the tfplugindocs as a tool dependency
* Added `docs` to the prettier ignore file. There are a few open issues about how prettier should handle the number of spaces after a list item, and this is conflicting with the single space approach `tfplugindocs` uses. So since the docs are autogenerated, we don't need to prettier them. 

FOLLOWUP:
* Done in a followup PR since we'll want that regardless of this PR

## Additional Notes
There isn't a way to preview the entire site currently, but there is an open feature request, and a way to preview each individual markdown file - https://registry.terraform.io/tools/doc-preview


nmuesch:
Closing this for now. 
Everything in the original PR description was done via other PRs (updating schema descriptions, individually committing generated tfplugindocs changes, etc)

Screenboard + timeboard are deprecated resources, so their docs are not currently autogenerated

Generating docs using tfplugindocs is also now a section in the DEVELOPMENT.md file. We'll likely want this behind CI, but for now am closing this PR

PR #801, opened 2020-12-25
Allow dependabot to check go packages

jlosito:
This change allows dependabot to check go dependencies which this project uses and submit PRs with version bumps in order to keep dependencies up-to-date.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates


jlosito:
Closing due to inactivity.

PR #803, opened 2020-12-29
Add descriptions to schema

nmuesch:
Add descriptions to missing schema fields.
We currently had Descriptions on some fields but not others, so this gets us further towards having descriptions everywhere. 
This is also a pre-cursor if we want to use tfplugindocs to generate documentation from schemas.


nmuesch:
@zippolyte Really appreciate the thorough review, and 💜 for the inline suggestions. Github seemed to struggle a bit to show them all, but I think I've accepted all comments

zippolyte (approved):

wackywendell (approved):
LGTM on behalf of @DataDog/metrics-query

pietdaniel (approved):

nmuesch:
Thanks for the reviews!!! Merging as all descriptions were taken from the existing documentation pages for these resources

PR #804, opened 2020-12-30
[monitor] Fix updating priorities.

phillip-dd:
And adds a test to cover updating priorities.

cc @DataDog/monitor-app

### DIFF
--- a/datadog/cassettes/TestAccDatadogMonitor_Basic_float_int.yaml
+++ b/datadog/cassettes/TestAccDatadogMonitor_Basic_float_int.yaml
@@ -799,7 +799,7 @@ interactions:
     duration: ""
 - request:
     body: |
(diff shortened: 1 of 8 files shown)


zhengshizhao (approved):
awesome! thanks for fixing it!

zippolyte (approved):

phillip-dd:
Looks like silenced tests are failing:
```
=== FAIL: datadog TestAccDatadogMonitor_SilencedRemove (2.54s)
    testing.go:684: Step 1 error: Check failed: Check 4/4 error: datadog_monitor.foo: Attribute 'silenced.*' expected "-1", got "0"

=== FAIL: datadog TestAccDatadogMonitor_MuteUnmuteSpecificScopes
    testing.go:684: Step 1 error: After applying this step, the plan was not empty:
```

We fixed the issue around accepting `0`s in the silenced - did some other behavior break here?

zippolyte:
I think it's not completely back to the behavior it had previously in the backend. IIUC, the backend also returned 0 in the response payload, but now returns `null` for the values where we put 0.

zippolyte:
Regardless, I think we can merge this as it's not related to the failure

PR #807, opened 2021-01-04
[Synthetics] Add support for config variables

romainberger:
Add support for `config_variable` for Synthetics tests.

To make things clearer, the `variable` has been renamed to `browser_variable` (keeping support for `variable` to avoid breaking changes).


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -444,6 +419,35 @@ func syntheticsTestOptionsList() *schema.Schema {
...
+					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[A-Z][A-Z0-9_]+[A-Z0-9]$`), "must be all uppercase with underscores"),
+				},
+				"pattern": {
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+				"type": {
+					Type:         schema.TypeString,
+					Required:     true,
+					ValidateFunc: validation.StringInSlice(syntheticsConfigVariableTypes, false),
```
therve, line 444:
```suggestion
					ValidateFunc: validateEnumValue(datadogV1.NewSyntheticsConfigVariableTypeFromValue),
```

therve (changes requested):
One minor change, and one question: couldn't we keep using `variable`, and use the right type depending on the test? The structures look very similar, it would make things slightly cleaner I think.

romainberger:
> couldn't we keep using `variable`, and use the right type depending on the test?

No, but I guess this means the documentation is not clear enough 😬 . 

- `config_variable` are used in tests configuration (for example in the test url), they can be used both for the API and browser tests.
- `browser_variable` are used in a browser test step. 

So for API tests you can only have `config_variable`, but for browser tests you can have both.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -444,6 +419,35 @@ func syntheticsTestOptionsList() *schema.Schema {
...
 }
 
+func syntheticsConfigVariable() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"example": {
+					Type:     schema.TypeString,
```
nmuesch, line 429: Can you add a description to each of these fields (in each schema)? We have an ongoing effort to ensure all fields have descriptions so we can work on autogenerating the terraform documentation - #798

therve (approved):

PR #827, opened 2021-01-09
Fix order of synthetic check's location by adding sorting

Bleacks:
Items in "locations" list wasn't fixed, so it would produce unecessary changes on each "terraform apply"

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	_nethttp "net/http"
 	"regexp"
+	"sort"
 	"strconv"
 
 	datadogV1 "github.com/DataDog/datadog-api-client-go/api/v1/datadog"
@@ -1011,6 +1012,7 @@ func buildSyntheticsTestStruct(d *schema.ResourceData) *datadogV1.SyntheticsTest
 		for _, s := range attr.([]interface{}) {
 			locations = append(locations, s.(string))
 		}
+		sort.Strings(locations)
 		syntheticsTest.SetLocations(locations)
 	}
 


therve:
Thanks for your patch. `locations` is a list, so it's ordered by default, I'm surprised you would need to sort it. Can you detail the problem you encounter?

Bleacks:
I thought it might solve the problem, but in fact I realized afterwards that my test case might be flawed.. The problem is still there on my side sadly

I encounter an issue when using some synthetic checks with several test locations, changes are listed by terraform on each apply just based on the order of items in said list. Here is a example of plan output that detail this behavior:

       ~ locations       = [
          + "aws:ap-south-1",
            "aws:ca-central-1",
          + "aws:eu-central-1",
            "aws:eu-west-1",
          - "aws:eu-west-3",
            "aws:eu-west-2",
          - "aws:ap-south-1",
          - "aws:eu-central-1",
          + "aws:eu-west-3",
        ]

Those are dummy locations, but as you can see, no changes have been applied to those element whatsoever, and terraform is still reporting a change. Which is inconvenient for us as we rely on those a lot, and it can produce a lot of false positive changes.

This is my very time creating a PR, I'd hope this is the good approach, I'd be glad to add required informations about this if need be

therve:
OK that makes more sense. The issue probably comes from the server side then. We should go the opposite direction, and instead of sorting the list turn it into a `TypeSet` then. See #540 for an example of a similar problem.

Bleacks:
This seems to require a bit more tinkering than just adding a line as I thought. I was able to update field's type, but there are some errors, as I am not a Go developer whatsoever, I'll have to take a deeper look on Go and how to make a terraform provider and after try to update this one.

therve:
Handled in #864

PR #833, opened 2021-01-12
Provide alternative to TypeMap complex fields for monitor resource and datasource

zippolyte:
Add alternative of `TypeList` with `MaxItems: 1`, and deprecate the map field


datadog/data_source_datadog_monitor.go
```diff
@@ -94,8 +95,43 @@ func dataSourceDatadogMonitor() *schema.Resource {
...
 			},
+			"monitor_thresholds": {
+				Description: "Alert thresholds of the monitor.",
+				Type:        schema.TypeList,
+				Computed:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"ok": {
+							Type:     schema.TypeString,
```
zippolyte, line 106: Forced to use `TypeString` to be able to differentiate between set to 0 and unset in config....

therve (approved):

PR #835, opened 2021-01-13
[Synthetics] Store SHA 256 hash of certificates in state instead of the actual cert

romainberger:
Do not store the content of client certificate in the state file for privacy.

Note: When this is released, it will trigger a non empty plan for every user with certificates in their config


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1106,15 +1106,11 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
...
-			localCertificate["cert"][0]["content"] = configCertificateContent.(string)
-		}
-		if configKeyContent, ok := d.GetOk("request_client_certificate.0.key.0.content"); ok {
-			localCertificate["key"][0]["content"] = configKeyContent.(string)
-		}
+		// make sure the content of the certificate is not set in the state file
+		// the field is write-only (not returned by the API) so it will generate
+		// some useless diff when planning
+		localCertificate["cert"][0]["content"] = ""
+		localCertificate["key"][0]["content"] = ""
```
zippolyte, line 1113: Maybe instead of storing nothing, we can store some sort of hash of the cert ? That would allow us to store it in the state as the hash itself would be unusable, and allow the diff to stay functional. WDYT ?
I don't think we can accept having a perpetual non empty plan

zippolyte (changes requested):

datadog/resource_datadog_synthetics_test_.go
romainberger: That would definitely be a better solution, I'll try to do that 👍

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -273,6 +274,9 @@ func syntheticsTestRequestClientCertificateItem() *schema.Schema {
 					Type:        schema.TypeString,
 					Required:    true,
 					Sensitive:   true,
+					StateFunc: func(val interface{}) string {
+						return convertToSha256(val.(string))
+					},
```
zippolyte, line 279: Won't this make terraform hash the value twice ? once in the read function when you use `getCertificateStateValue`, and once by terraform automatically when you set the value in the state ?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1106,14 +1110,14 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
...
 
-		// the content of the certificate and the key are write-only
-		// so we need to get them from the config since they will
-		// not be in the api response
+		// the content of client certificate is write-only so it will not be returned by the API.
+		// To avoid useless diff but also prevent storing the value in clear in the state
+		// we store a hash of the value.
 		if configCertificateContent, ok := d.GetOk("request_client_certificate.0.cert.0.content"); ok {
-			localCertificate["cert"][0]["content"] = configCertificateContent.(string)
+			localCertificate["cert"][0]["content"] = getCertificateStateValue(configCertificateContent.(string))
```
zippolyte, line 1117: I think `GetOk` takes the value from the config, not the state, so it should never be hashed already. Unless I'm missing something, there should be no need to do this, and instead rely on the `StateFunc` you added above to let terraform store in the state no ?

datadog/resource_datadog_synthetics_test_.go
zippolyte: Well actually, from my testing, it seems this doesn't do anything at all, let's just remove it and manually set the hashed value directly.

datadog/resource_datadog_synthetics_test_.go
romainberger: I thought so as well, but when I print the value to debug, we get a hashed value (even after commenting the `StateFunc`). I assumed it is when terraform reads the config to compare the state and the config.

datadog/resource_datadog_synthetics_test_.go
romainberger: No this function never receives a hashed value.

zippolyte:
Well ignore all I said... terraform does too many weird things, what you have seems to be working fine, so let's go for it.

datadog/resource_datadog_synthetics_test_.go
zippolyte: Yeah looks like depending on when you call it, it picks from state or from config...

zippolyte (approved):

PR #821, opened 2021-01-07
Generate index and user docs from tfplugindocs

nmuesch:
Based on PR #798 start generating the user + index documentation files
I noticed a couple of small typos, and fixed those here as well

PR #839, opened 2021-01-14
Add tfplugindocs scaffolding and generate monitor docs

nmuesch:
This PR continues #821 and #798 by adding the needed templates and example files to generate Users, Monitors, and the index documentation pages. This also autogenerates those pages based on those files + the Schema definitions using `tfplugindocs`


examples/resources/datadog_monitor/resource.tf
```diff
@@ -0,0 +1,30 @@
+# Create a new Datadog monitor
+resource "datadog_monitor" "foo" {
+  name               = "Name for monitor foo"
+  type               = "metric alert"
+  message            = "Monitor triggered. Notify: @hipchat-channel"
+  escalation_message = "Escalation message @pagerduty"
+
+  query = "avg(last_1h):avg:aws.ec2.cpu{environment:foo,host:foo} by {host} > 4"
+
+  thresholds = {
```
zippolyte, line 10:
```suggestion
  monitor_thresholds {
```

zippolyte (changes requested):

examples/resources/datadog_monitor/resource.tf
nmuesch: Good catch

docs/data-sources/monitor.md
```diff
@@ -1,49 +1,75 @@
...
 }
 ```
 
-## Argument Reference
-
--   `name_filter`: (Optional) A monitor name to limit the search.
--   `tags_filter`: (Optional) A list of tags to limit the search. This filters on the monitor scope.
--   `monitor_tags_filter`: (Optional) A list of monitor tags to limit the search. This filters on the tags set on the monitor itself.
-
-~> **NOTE:** If more or less than a single match is returned by the search, Terraform will fail. Ensure that your search is specific enough to return a single monitor.
```
nmuesch, line 24: One downside to the autogeneration is that we lose these kinds of notes. If we wanted to, I can add a template that copies the base template but had this note hard-coded in. Though I don't know if its strictly necessary here.

docs/resources/monitor.md
```diff
@@ -41,176 +44,73 @@ resource "datadog_monitor" "foo" {
...
-You can compose monitors of all types in order to define more specific alert conditions (see the [doc](https://docs.datadoghq.com/monitors/monitor_types/composite/)). You just need to reuse the ID of your `datadog_monitor` resources. You can also compose any monitor with a `datadog_synthetics_test` by passing the computed `monitor_id` attribute in the query.
-
-```hcl
-resource "datadog_monitor" "bar" {
-  name = "Composite Monitor"
-  type = "composite"
-  message = "This is a message"
-
-  query = "${datadog_monitor.foo.id} || ${datadog_synthetics_test.foo.monitor_id}"
-}
```
nmuesch, line 215: Probably worth keeping this and adding into the hardcoded monitors guide page instead

docs/resources/monitor.md
```diff
@@ -41,176 +44,73 @@ resource "datadog_monitor" "foo" {
...
--   `monitor_threshold_windows` (Optional) A list of one element containing `recovery_window` and `trigger_window` values, e.g. `last_15m` . Can only be used for, and are required for, anomaly monitors.
-    -   `recovery_window` describes how long an anomalous metric must be normal before the alert recovers.
-    -   `trigger_window` describes how long a metric must be anomalous before an alert triggers.
--   `threshold_windows` (Deprecated, Optional) A dictionary defining the threshold windows for an anomaly monitor. Use `monitor_threshold_windows` instead.
--   `validate` (Optional) If set to false, skip the validation call done during `plan` .
--   `priority` (Optional) Integer from 1 (high) to 5 (low) indicating alert severity.
--   `silenced` (Optional) Each scope will be muted until the given POSIX timestamp or forever if the value is 0. Use `-1` if you want to unmute the scope. **Deprecated** The `silenced` parameter is being deprecated in favor of the downtime resource. This will be removed in the next major version of the Terraform Provider.
```
nmuesch, line 147: Since this is deprecated, I think its OK we're not generating this section anymore ^

nmuesch:
I looked over the generation and one thing I realized we probably don't want to lose is the `Composite Monitor` example. So I updated this PR to also include the temlpate and examples pieces for the `Monitor Guide` page. PR is a little bigger but does all the monitor pieces together (resource + data-source + guide)

docs/data-sources/monitor.md
zippolyte: We can always add it in the description of the filter arguments

zippolyte (approved):

PR #849, opened 2021-01-20
[Synthetics] Add new browser_step field for browser tests

romainberger:
This PR is a preliminary work for the next PR (https://github.com/DataDog/terraform-provider-datadog/pull/854), I've divided the work in two PR for easier review.

- mark as deprecated but still support `step` field
- add new field `browser_step` to replace `step`, to make it more explicit (we'll have `api_step` in the future), and details all parameters for the steps. This is needed for the next PR.

Note: The documentation is _willingly_ vague, because documenting what parameter a step type accepts would be hard to document clearly. I've documented the description of each field, but for now the goal was to be able to do what https://github.com/DataDog/terraform-provider-datadog/pull/854 does, not to document precisely everything.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -494,10 +507,130 @@ func syntheticsTestStep() *schema.Schema {
...
+					Description: "ID of the tab to play the subtest.",
+					Type:        schema.TypeString,
+					Optional:    true,
+				},
+				"request": {
+					Description: "Request for an API step.",
+					Type:        schema.TypeString,
+					Optional:    true,
+				},
+				"subtest_public_id": {
```
PierreGuyot, line 590: For confirmation, is it expected we don't have a `subtest` field here?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1396,3 +1588,153 @@ func getCertificateStateValue(content string) string {
...
+		return convertToString(value)
+
+	case "variable":
+		return []interface{}{value}
+	}
+
+	return value
+}
+
+func convertStepParamsKey(key string) string {
```
PierreGuyot, line 1712: It seems a bit weird to have a conversion method which handle both sides, would it make sense to divide between two methods (something like `toSnakeCase`/`toCamelCase`, or `toState`/`toConfig`)?

PierreGuyot (approved):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1396,3 +1588,153 @@ func getCertificateStateValue(content string) string {
...
+		return []string{"value"}
+
+	case datadogV1.SYNTHETICSSTEPTYPE_GO_TO_URL:
+		return []string{"value"}
+
+	case datadogV1.SYNTHETICSSTEPTYPE_HOVER:
+		return []string{"element"}
+
+	case datadogV1.SYNTHETICSSTEPTYPE_PLAY_SUB_TEST:
+		return []string{"playing_tab_id", "subtest_public_id"}
```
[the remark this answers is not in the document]
romainberger: Yes this is the list of parameters needed for the step, which are `playingTabId` and `subtestPublicId`.

datadog/resource_datadog_synthetics_test_.go
romainberger: Same as above, the parameters are exact matches (except for casing) of what is needed in the backend.

datadog/resource_datadog_synthetics_test_.go
romainberger: This was to avoid having to create the `toSnakeCase` / `toCamelCase` methods but I can do that 👍

therve (approved):

PR #851, opened 2021-01-21
Fixed error in doc where Int defaults to `false`.

venthur:
timeout_h is of type Int and defaults to `false` according to the docs.
There are no defaults defined in the code, so i removed the comment
about the default.


therve (approved):

therve:
Looks great, thank you!

PR #854, opened 2021-01-21
[Synthetics] Add parameter to prevent useless diffs for MML

romainberger:
Based on https://github.com/DataDog/terraform-provider-datadog/pull/849.

The `element` parameter of a browser test step (`browser_step`) is updated automatically server-side, which makes the terraform config and state not up-to-date. As a consequence there are some unwanted diffs, and if a config is re-applied it can break the test by overwriting the `element` parameter with an outdated value.
To fix this, the `element` parameter diff is ignored (which is the wanted behavior most of the time), and if the user actually wants to update the parameter, the new `force_element_update` field will force the update.


PierreGuyot (approved):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1510,6 +1534,10 @@ func updateSyntheticsTestLocalState(d *schema.ResourceData, syntheticsTest *data
 				}
 
 				localStep["params"] = []interface{}{localParams}
+
+				if forceElementUpdate, ok := d.GetOk("browser_step." + convertToString(stepIndex) + ".force_element_update"); ok {
```
zippolyte, line 1538:
```suggestion
				if forceElementUpdate, ok := d.GetOk(fmt.Sprintf("browser_step.%d.force_element_update", stepIndex)); ok {
```

therve (approved):

PR #855, opened 2021-01-21
Support empty strings for custom pipeline filter query

skarimo:
For the custom pipelines resource, an empty string can be set for the `filter.query` attribute and returned via the API. This ensures we handle empty queries. 

Example:
```
{"id":"","type":"pipeline","name":"Testing","is_enabled":true,"is_read_only":false,"filter":{"query":""},"processors":[]}
```

Needed for: https://github.com/GoogleCloudPlatform/terraformer/pull/752


datadog/resource_datadog_logs_custom_pipeline.go
```diff
@@ -1078,9 +1077,11 @@ func buildDatadogDateRemapperProcessor(tfProcessor map[string]interface{}) *data
 
 func buildDatadogFilter(tfFilter map[string]interface{}) datadogV1.LogsFilter {
 	ddFilter := datadogV1.LogsFilter{}
+	var query string
 	if tfQuery, exists := tfFilter["query"].(string); exists {
-		ddFilter.SetQuery(tfQuery)
+		query = tfQuery
 	}
+	ddFilter.SetQuery(query)
```
zippolyte, line 1084: Curious, why do we need this ? `query` is set as required, so it should always exist, regardless if it is empty string or not

therve: I think this is something Slavek discovered in the codegen project: if you do `d.Get("filter.X.query")` you'll get an empty string, but when you do `d.Get("filter")` it can remove some empty values.

therve (approved):

PR #859, opened 2021-01-22
Update timeout on gotestsum to be 20 minutes

nmuesch:
Bump timeout to be 20 minutes instead of 15. 
As new resources and endpoints are being added, the tests seem to be starting to hit the 15 minute timeout


therve (approved):

nmuesch:
Merging as CI failures are unrelated and seem to have been happening for a bit. Will investigate those separately

PR #864, opened 2021-01-25
Turn synthetics test locations into a set

therve:
The items in the location list need to be unique, and server can
re-order them, so using a set is the best course of actions.

### DIFF
--- a/datadog/resource_datadog_synthetics_test_test.go
+++ b/datadog/resource_datadog_synthetics_test_test.go
@@ -417,7 +417,7 @@ func createSyntheticsAPITestStep(accProvider *schema.Provider, clock clockwork.F
 				"datadog_synthetics_test.foo", "assertions.3.target", "terraform"),
 			resource.TestCheckResourceAttr(
 				"datadog_synthetics_test.foo", "locations.#", "1"),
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_synthetics_test_test.go
```diff
@@ -417,8 +417,8 @@ func createSyntheticsAPITestStep(accProvider *schema.Provider, clock clockwork.F
 				"datadog_synthetics_test.foo", "assertions.3.target", "terraform"),
 			resource.TestCheckResourceAttr(
 				"datadog_synthetics_test.foo", "locations.#", "1"),
-			resource.TestCheckResourceAttr(
-				"datadog_synthetics_test.foo", "locations.0", "aws:eu-central-1"),
+			CheckResourceAttr(
```
zippolyte, line 420: What's this change ?

jirikuncar:
```suggestion
			resource.TestCheckResourceAttr(
```

therve: Used for debugging, good catch.

zippolyte (approved):

PR #870, opened 2021-01-26
Deprecate TypeMap field in synthetics

zippolyte:
deprecate typeMap `request` in favor of typeList `request_definition`


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -46,7 +46,22 @@ func resourceDatadogSyntheticsTest() *schema.Resource {
...
+			"request": {
+				Description:   "The synthetics test request. Required if `type = \"api\"` and `subtype = \"http\"`.",
+				Deprecated:    "Define `request_definition` list with one element instead.",
+				ConflictsWith: []string{"request_definition"},
+				Type:          schema.TypeMap,
+				Optional:      true,
+				Elem:          syntheticsTestRequest(),
+			},
+			"request_definition": {
+				Description:   "The synthetics test request. Required if `type = \"api\"` and `subtype = \"http\"`.",
```
romainberger, line 58: This part may be legacy but `Required if `type = \"api\"` and `subtype = \"http\"`.` is not true, it should be just `Required if `type = \"api\"``

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -830,29 +838,50 @@ func isTargetOfTypeInt(assertionType datadogV1.SyntheticsAssertionType, assertio
...
 	}
-	if attr, ok := d.GetOk("request.body"); ok {
+	if attr, ok := k.GetOkWith("body"); ok {
 		request.SetBody(attr.(string))
 	}
-	if attr, ok := d.GetOk("request.timeout"); ok {
-		timeoutInt, _ := strconv.Atoi(attr.(string))
+	if attr, ok := k.GetOkWith("timeout"); ok {
+		var timeoutInt int
+		// first try to convert to int if we're getting from the new TypeList field
```
nmuesch, line 860: Sanity checking; is this because everything in a TypeMap is a string, regardless of what the Type is set as?

datadog/resource_datadog_synthetics_test_test.go
```diff
@@ -30,33 +30,6 @@ func TestAccDatadogSyntheticsAPITest_importBasic(t *testing.T) {
 				ResourceName:      "datadog_synthetics_test.foo",
 				ImportState:       true,
 				ImportStateVerify: true,
-				// Assertions will be imported into the new schema by default, but we can ignore them as users need to update the local config in this case
-				ImportStateVerifyIgnore: []string{"assertions", "assertion", "options", "options_list"},
```
nmuesch, line 34: Why remove this?

datadog/resource_datadog_synthetics_test_test.go
```diff
@@ -30,33 +30,6 @@ func TestAccDatadogSyntheticsAPITest_importBasic(t *testing.T) {
...
-			},
-		},
-	})
-}
-
-func TestAccDatadogSyntheticsAPITest_importBasicNewAssertionsOptions(t *testing.T) {
```
nmuesch, line 40: Are we covering this import elsewhere?

nmuesch (commented):
Code looks good to me. Couple minor questions on the tests

datadog/resource_datadog_synthetics_test_.go
zippolyte: yes

datadog/resource_datadog_synthetics_test_test.go
zippolyte: We want verification on these fields

datadog/resource_datadog_synthetics_test_test.go
zippolyte: Deprecated field cannot be imported, as the implementation of read prefers the new fields when not set in config (which is the case for import)
I modified the main test config to not have any deprecated fields in it, so the other import tests now does the proper verifications (without the importIgnore)
This one is then rendered useless, as it's just a duplication

nmuesch (approved):

PR #871, opened 2021-01-27
fix TestAccDatadogIntegrationAWSLogCollection test

skarimo:
This PR fixes the `TestAccDatadogIntegrationAWSLogCollection` by ensuring we are using a unique `account_id` and making sure lambda `datadog_integration_aws_lambda_arn` resource is created for the AWS account prior to attempting to add AWS log collection resource.

If no lambda arn is set up for the account we recieve the following object:
```
{"status":"error","errors":[{"message":"AWS account 029491012312 has no lambda config to update","code":"no_such_config"}]}
```


datadog/resource_datadog_integration_aws_log_collection_test.go
```diff
@@ -19,17 +19,22 @@ resource "datadog_integration_aws" "account" {
   role_name                        = "testacc-datadog-integration-role"
 }
 
+resource "datadog_integration_aws_lambda_arn" "lambda" {
+  account_id = "%s"
```
nmuesch, line 23: nit: could we use the variable from datadog_integration_aws.account.account_id instead of passing another `uniq`?
Same for the one below. I think in that case we'd only have to pass one or two `uniq` here

skarimo: Good call! Thanks

nmuesch (approved):

datadog/resource_datadog_integration_aws_log_collection_test.go
```diff
@@ -19,17 +19,22 @@ resource "datadog_integration_aws" "account" {
...
   services = ["lambda"]
-  depends_on = [datadog_integration_aws.account]
+  depends_on = [datadog_integration_aws_lambda_arn.lambda]
 }`, uniq, uniq)
 }
 
 func TestAccDatadogIntegrationAWSLogCollection(t *testing.T) {
 	accProviders, clock, cleanup := testAccProviders(t, initRecorder(t))
 	accountID := uniqueAWSAccountID(clock, t)
-	accountID = "1234567890"
```
zippolyte, line 32: lol 🤦 
good catch

zippolyte (approved):

Issue #878, opened 2021-01-30
Child Organizations


andymac4182:
Hi,

We manage a multi-org account where we have an organization for each of our customers. 

It would be great if we could automate the creation of these organizations via Terraform.

https://docs.datadoghq.com/api/latest/organizations/

PR #879, opened 2021-01-30
Initial work on organization resource

andymac4182:
Based on the request #878 

This is the start of support for an datadog_organisation resource. 

Be great to get some feedback. I know I still need to add tests.


therve:
Hi,

Thanks for your patch! That looks a good start. I think it should indeed be called `datadog_child_organization`, that matches the resource name a bit better. I also think you should store the api key and application key, as we don't provide those in any other way later on.

andymac4182:
Thanks for the feedback @therve Will look at updating these to include those values.

andymac4182:
@therve Can you please let me know which properties from ApplicationKey (https://github.com/DataDog/datadog-api-client-go/blob/6f5dd76317c8818530d95ab1f64c9f3482271b70/api/v1/datadog/docs/ApplicationKey.md) and ApiKey (https://github.com/DataDog/datadog-api-client-go/blob/6f5dd76317c8818530d95ab1f64c9f3482271b70/api/v1/datadog/docs/ApiKey.md)
 I should be storing?

therve:
@andymac4182 As documented at https://docs.datadoghq.com/api/latest/organizations/#create-a-child-organization I believe you need the api_key.key and app_key.hash.

andymac4182:
Thanks. Sorry I missed that in the docs.

andymac4182:
@therve There isn't a remove/delete api endpoint. I am getting an error from the validation where delete is required. `resource datadog_child_organization: Delete must be implemented` any preference to handling this?

therve:
Yes I think there is no choice, just set the ID to '' in delete, without any API call.

andymac4182:
The tests run locally. Not sure if I need to do something to make them run in the CI environment. Currently getting `Error: error getting organization: 400 Bad Request: {"errors": ["Not allowed to access this organization"]}` Sorry if I didn't get this right this is my first end to end TF resource.

datadog/resource_datadog_child_organization.go
```diff
@@ -0,0 +1,112 @@
+package datadog
+
+import (
+	datadogV1 "github.com/DataDog/datadog-api-client-go/api/v1/datadog"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+func resourceDatadogChildOrganisation() *schema.Resource {
```
fanny-jiang, line 9: Small nit--can you please harmonize the spelling of this function name with the file name?
```suggestion
func resourceDatadogChildOrganisation() *schema.Resource {
```
```suggestion
func resourceDatadogChildOrganization() *schema.Resource {
```

datadog/resource_datadog_child_organization.go
```diff
@@ -0,0 +1,112 @@
...
+	return &schema.Resource{
+		Description: "Provides a Datadog Organization resource. This can be used to create and manage Datadog organizations.",
+		Create:      resourceDatadogChildOrganizationCreate,
+		Read:        resourceDatadogChildOrganizationRead,
+		Update:      resourceDatadogChildOrganizationUpdate,
+		Delete:      resourceDatadogChildOrganizationDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceDatadogOrganizationImport,
```
fanny-jiang, line 17: Update this function name to be consistent with the other function names.

```suggestion
			State: resourceDatadogChildOrganizationImport,
```

datadog/resource_datadog_child_organization.go
```diff
@@ -0,0 +1,112 @@
...
+			"name": {
+				Description:  "The name of the organization",
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 32),
+			},
+			"api_key": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "",
```
fanny-jiang, line 30: Each schema should have a description. Feel free to reference the model descriptions in the API docs: https://docs.datadoghq.com/api/latest/organizations/#create-a-child-organization

datadog/resource_datadog_child_organization.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+			"application_key": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "",
+			},
+		},
```
fanny-jiang, line 37: I think you might also need to specify the `public_id` parameter here in order to read organizations: https://docs.datadoghq.com/api/latest/organizations/#get-organization-information

fanny-jiang:
Hey @andymac4182 

> The tests run locally. Not sure if I need to do something to make them run in the CI environment. Currently getting `Error: error getting organization: 400 Bad Request: {"errors": ["Not allowed to access this organization"]}` Sorry if I didn't get this right this is my first end to end TF resource.

I got this error too when I tested your resource and also when I manually made a request to the API endpoint. I'm checking with our team to get some more information about that and will follow up with you here. 

Besides that, your PR looks good and thanks for working on this! There are just some small naming inconsistencies in the functions that can be fixed. Also, instead of returning `resourceDatadogChildOrganizationRead()` in your create, read, and update functions, can you add a helper function to call `d.Set()` for the `name` and `public_id`? We're moving away from the pattern of returning the read function so that the provider can return values more consistently--sometimes there's some latency returning data from the backend when making a GET request immediately after another request. See this recent PR as an example: https://github.com/DataDog/terraform-provider-datadog/pull/909

andymac4182:
Thanks for the information. I will get this updated. 

Thanks for checking with the team I hope it is something simple I missed.

nmuesch:
@andymac4182 On a side note, I just merged a PR that changes the structure of the repo a bit. 
Would you be able to rebase against latest master? It mostly moves cassettes and tests into a new `tests` package under `datadog`, and moves some common shared utils and validators into a new `internal` package structure.

Let me know if you need any help on that!

fanny-jiang:
Hey @andymac4182, I have some more info from the team about the `"Not allowed to access this organization"` error. You were doing everything correctly and it's not a config issue. It looks like this error is being returned from the Get Organization Information endpoint (`/v1/org/{public_id}`), which currently doesn't support returning child orgs when using a parent org user's api/app key credentials. It currently returns information about the user's current org. 

I looked into some potential workarounds, but since the read org endpoint needs the child org's api/app key, this wouldn't be something that can be implemented within the read function of the resource. 

Unfortunately this means that adding this resource would be blocked until an endpoint is added that supports getting child orgs since the resource wouldn't be able to confirm that the child org has been added without a read endpoint. Adding support for this feature/endpoint is currently a feature request with the team that owns the Organizations endpoint. 

In the meantime, I did find that if you handle the `400` status code and allow the read function to continue or omit an API call in the read function, the resource does apply successfully to create an org. It's not ideal without a read, but that's something to consider if you'd still like to implement the create org on a local build of the datadog provider. I hope this helps!

korenyoni:
@fanny-jiang has the situation on `/v1/org/{public_id}` for child organizations changed at all since your last comment? https://github.com/DataDog/terraform-provider-datadog/pull/879#issuecomment-782195211

PR #880, opened 2021-01-30
Use custom SDK version

therve:
This gives us a big performance boost, at least in test, until we manage to migrate to v2.


nmuesch (approved):
LGTM, thanks

Issue #568, opened 2020-07-06
Add Support for Filtering AWS Namespaces


brandonstevens:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add support for filtering AWS namespaces (e.g. Lambda) using tags. It's currently only possible to limit resource collection using tags for Lambdas in the UI or through an undocumented API /api/v1/integration/aws/filtering. For organizations with a large number of Datadog accounts, limited resource collection by requiring an opt-in tag makes it a lot easier to manage Datadog costs. Making changes across a large number of Datadog accounts is tedious. Filters should be configurable with Terraform.

### New or Affected Resource(s)

* `datadog_integration_aws_filters`
or
* `filters` block inside `datadog_integration_aws` resource

### Potential Terraform Configuration

```hcl
resource "datadog_integration_aws" "sandbox" {
    account_id = "1234567890"
    role_name = "DatadogAWSIntegrationRole"
    filter_tags = ["key:value"]
    host_tags = ["key:value", "key2:value2"]
    account_specific_namespace_rules = {
        auto_scaling = false
        opsworks = false
    }
    excluded_regions = ["us-east-1", "us-west-2"]
    filters = {
        lambda = ["key:value"]
    }
}
```

### References

```
$ curl -sX GET https://api.datadoghq.com/api/v1/integration/aws/filtering?account_id=XXXXXXXXXXXX -H "Content-Type: application/json" -H "DD-API-KEY: XXXXXXXXXXXX" -H "DD-APPLICATION-KEY: XXXXXXXXXXXX"

{
  "filters": [
    {
      "tag_filter_str": null,
      "namespace": "elb"
    },
    {
      "tag_filter_str": null,
      "namespace": "application_elb"
    },
    {
      "tag_filter_str": null,
      "namespace": "sqs"
    },
    {
      "tag_filter_str": null,
      "namespace": "rds"
    },
    {
      "tag_filter_str": null,
      "namespace": "custom"
    },
    {
      "tag_filter_str": null,
      "namespace": "network_elb"
    },
    {
      "tag_filter_str": "monitoring:true",
[...]

PR #881, opened 2021-02-01
Add support for aws tag filter resource

skarimo:
Closes: https://github.com/DataDog/terraform-provider-datadog/issues/568

Requires a datadog-go-client version bump after spec changes have been merged


datadog/resource_datadog_integration_aws_tag_filter.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, ns := range resp.GetFilters() {
+		if ns.GetNamespace() == namespace {
+			d.Set("account_id", accountID)
+			d.Set("namespace", ns.GetNamespace())
+			d.Set("tag_filter_str", ns.GetTagFilterStr())
+			return nil
+		}
+	}
+
+	d.SetId("")
```
fanny-jiang, line 112: Would be nice to include a comment here indicating that we're resetting the ID and clearing the local state and that we haven't found the filter.

therve (approved):

PR #883, opened 2021-02-02
[monitor] Add restricted roles.

phillip-dd:
Adds restricted roles to monitors.

Note: Not generally available yet, so documentation is commented out. Rollout plan is being finalized.

cc @NBParis @psionn


datadog/resource_datadog_monitor.go
```diff
@@ -268,9 +268,32 @@ func resourceDatadogMonitor() *schema.Resource {
...
 			},
 			"locked": {
-				Description: "A boolean indicating whether changes to to this monitor should be restricted to the creator or admins. Defaults to `false`.",
-				Type:        schema.TypeBool,
-				Optional:    true,
+				Description:   "A boolean indicating whether changes to to this monitor should be restricted to the creator or admins. Defaults to `false`.",
+				Type:          schema.TypeBool,
+				Optional:      true,
+				ConflictsWith: []string{"restricted_roles"},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
therve, line 275: Why do we need those diff suppress if it can't be set by terraform? Is it returned by the API somehow?

phillip-dd: Ya, there's some logic that if restricted roles are set, we also return `locked` as true (regardless of its value in the request). This is in theory supposed to help with migration across tools.

therve (approved):

PR #891, opened 2021-02-03
Emphasize deprecated fields in generated docs

zippolyte:
Use `SchemaDescriptionBuilder` as mentioned [here](https://github.com/hashicorp/terraform-plugin-docs/issues/10#issuecomment-767682837) to prepend the deprecated string to the description of objects

Same thing could be used for default to avoid having to define it manually in the description.


nmuesch (approved):
LGTM, thanks!

nmuesch (approved):

PR #899, opened 2021-02-05
Add retry helper to tests and use it in SLO destroy helper

nmuesch:
Add a retry method (took heavy inspiration from what we have in datadog-api-client-go test utilities) 
However the contract is slightly different (commented in code)
Also moved the destroyHelper method to be right next to the code using it. 

I'd like to get feedback on the pattern before starting to apply in all the resources tests


datadog/test_utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"time"
+)
+
+// Retry calls the call function for count times every interval while it returns true
+// Call function should return (false, nil) to indicate a success
+// (true, err) to indicate there was an error but we should retry
+// and (false, err) to indicate there was a fatal error
+func Retry(interval time.Duration, count int, call func() (bool, error)) error {
+	for i := 0; i < count; i++ {
+		shouldRetry, err := call()
```
therve, line 14: WDYT of using a special error type like `RetriableError? I think we should retry in all other cases anyway, and it would simplify signature changes.

datadog/test_utils.go
```diff
@@ -0,0 +1,26 @@
...
+		if !shouldRetry && err == nil {
+			return nil
+		} else if shouldRetry {
+			fmt.Errorf("Failed with retry-able error: %s", err)
```
therve, line 18: This doesn't do anything.

datadog/test_utils.go
```diff
@@ -0,0 +1,26 @@
...
+			time.Sleep(interval)
+		} else {
+			fmt.Errorf("Failed with an error and shouldn't retry: %s", err)
```
therve, line 21: Same.

datadog/resource_datadog_service_level_objective_test.go
```diff
@@ -228,18 +228,33 @@ func TestAccDatadogServiceLevelObjective_InvalidMonitor(t *testing.T) {
...
-	for _, r := range s.RootModule().Resources {
-		if r.Primary.ID != "" {
-			if _, _, err := datadogClientV1.ServiceLevelObjectivesApi.GetSLO(authV1, r.Primary.ID).Execute(); err != nil {
-				if strings.Contains(strings.ToLower(err.Error()), "not found") {
-					continue
+	Retry(2, 5, func() (bool, error) {
+		for _, r := range s.RootModule().Resources {
+			if r.Primary.ID != "" {
+				if _, _, err := datadogClientV1.ServiceLevelObjectivesApi.GetSLO(authV1, r.Primary.ID).Execute(); err != nil {
+					if strings.Contains(strings.ToLower(err.Error()), "not found") {
```
therve, line 248: If you can take the change to check the HTTP response for 404 instead that'd be cool.

therve (changes requested):
Looks like the right path to me! Some questions/suggestions.

datadog/test_utils.go
nmuesch: You're right, I keep forgetting I need `log.` instead here. Will make the change

datadog/test_utils.go
nmuesch: Yeah I think that makes sense. Will look into implementing that

datadog/test_utils.go
```diff
@@ -0,0 +1,48 @@
+package datadog
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"time"
+)
+
+// Retry calls the call function for count times every interval while it returns true
```
nmuesch, line 10:
```suggestion
// Retry calls the call function for count times every interval while it returns nil
```

datadog/test_utils.go
```diff
@@ -0,0 +1,48 @@
...
+// Call function should return either a RetryableError or FatalError
```
nmuesch, line 11:
```suggestion
// Call function should return either: nil, RetryableError or FatalError
```

datadog/test_utils.go
```diff
@@ -0,0 +1,48 @@
...
+// RetryableError means we'll retry up to the count
```
nmuesch, line 12:
```suggestion
// nil indicates a success condition was met
// RetryableError means we'll retry up to the count
```

datadog/test_utils.go
```diff
@@ -0,0 +1,49 @@
...
+type RetryableError struct {
+	prob string
+}
+
+func (e *RetryableError) Error() string {
+	return fmt.Sprintf("[INFO] Failed with retry-able error: %s", e.prob)
+}
+
+// FatalError should be considered final and not trigger any retries
+type FatalError struct {
```
nmuesch, line 43: Having an explicit fatal error type may be a bit of an overkill here, but I think it keeps the interface of this utility pretty explicit, WDYT?

datadog/test_utils.go
```diff
@@ -0,0 +1,49 @@
...
+			return nil
+		} else if errors.Is(err, retryErrorType) {
+			log.Printf(err.Error())
+			time.Sleep(interval)
+		} else if errors.Is(err, fatalErrorType) {
+			log.Printf(err.Error())
+			return err
+		}
+	}
+	return &FatalError{prob: fmt.Sprintf("Retry error: failed to satisfy the condition after %d times\n", count)}
```
zippolyte, line 30:
```suggestion
	return &FatalError{prob: fmt.Sprintf("failed to satisfy the condition after %d times\n", count)}
```

zippolyte (changes requested):

datadog/test_utils.go
```diff
@@ -0,0 +1,49 @@
...
+			return nil
+		} else if errors.Is(err, retryErrorType) {
+			log.Printf(err.Error())
+			time.Sleep(interval)
+		} else if errors.Is(err, fatalErrorType) {
+			log.Printf(err.Error())
+			return err
+		}
+	}
+	return &FatalError{prob: fmt.Sprintf("failed to satisfy the condition after %d times\n", count)}
```
jirikuncar, line 30:
```suggestion
	return &FatalError{prob: fmt.Sprintf("failed to satisfy the condition after %d times", count)}
```

zippolyte (approved):

jirikuncar (approved):

nmuesch:
Thanks all, CI is green. Going to merge and see where in the test suite we can use this!

PR #901, opened 2021-02-08

zippolyte:

Also fixes some linting warnings, and reads error returned by `d.Set`


datadog/resource_datadog_monitor.go
```diff
@@ -559,41 +559,8 @@ func resourceDatadogMonitorCreate(d *schema.ResourceData, meta interface{}) erro
 	return resourceDatadogMonitorRead(d, meta)
 }
 
-func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error {
+func updateMonitorState(d *schema.ResourceData, meta interface{}, m *datadogV1.Monitor) error {
 	providerConf := meta.(*ProviderConfiguration)
```
nmuesch, line 563: is providerConf still used in this method?

datadog/resource_datadog_monitor.go
```diff
@@ -632,45 +599,88 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
 	sort.Strings(tags)
 
 	log.Printf("[DEBUG] monitor: %+v", m)
-	d.Set("name", m.GetName())
-	d.Set("message", m.GetMessage())
-	d.Set("query", m.GetQuery())
-	d.Set("type", m.GetType())
-	d.Set("priority", m.GetPriority())
-	d.Set("restricted_roles", m.GetRestrictedRoles())
+	if err := d.Set("name", m.GetName()); err != nil {
```
nmuesch, line 602: Does this error if the value being set is empty, or is it just if an internal TF SDK piece fails?

datadog/resource_datadog_monitor.go
```diff
@@ -632,45 +599,88 @@ func resourceDatadogMonitorRead(d *schema.ResourceData, meta interface{}) error
...
+	if err := d.Set("escalation_message", m.Options.GetEscalationMessage()); err != nil {
+		return err
+	}
+	if err := d.Set("include_tags", m.Options.GetIncludeTags()); err != nil {
+		return err
+	}
+	if err := d.Set("tags", tags); err != nil {
+		return err
+	}
+	// TODO Is this one of those options that we neeed to check?
```
nmuesch, line 672: Do we know what this TODO is for?

datadog/resource_datadog_monitor.go
```diff
@@ -771,12 +822,13 @@ func resourceDatadogMonitorUpdate(d *schema.ResourceData, meta interface{}) erro
 				delete(silencedList, scope)
 			}
 		}
-		if _, _, err = datadogClientV1.MonitorsApi.UpdateMonitor(authV1, i).Body(*m).Execute(); err != nil {
+		monitorResp, _, err = datadogClientV1.MonitorsApi.UpdateMonitor(authV1, i).Body(*m).Execute()
```
nmuesch, line 825: Perhaps you have already (not sure if we have an explicit test for this) but can you confirm we get the full monitor back here when we unmute scopes?

nmuesch (commented):
Overall LGTM, left a few inline questions. Thanks!

datadog/resource_datadog_monitor.go
zippolyte: It errors if an internal SDK piece fails, like wrong type or something

datadog/resource_datadog_monitor.go
zippolyte: absolutely not

datadog/resource_datadog_monitor.go
zippolyte: It's still the same update call, so I'd say yes, and monitors team said I could rely on output of PUT request

zippolyte: (also yeah we have a specific unmute test)

datadog/resource_datadog_monitor.go
zippolyte: yes, to get the time near the end

nmuesch (approved):

PR #902, opened 2021-02-08
Move tests to its own package

nmuesch:
This PR aims to move the tests for this project into its own `test` sub folder/package. The hope is to make the structure of this repo a bit easier to parse through. 

While most of the PR is simply copying `*_test.go` files into a new `tests` folder, there are a couple of functional changes:
* Since the tests now reside in a different package than `datadog`, any code used from the main project needs to be imported
* Due to ^, this means there were a few helper methods we defined on resources that required being `Exported`. (These are helpers like: `translateClientError ` -> `TranslateClientError`, `GetUserAgent`, etc)
* The cassettes folder now lives under `tests` as well. 
* All filenames under `testFiles2EndpointTags` are now prefixed with `tests/`


pietdaniel (dismissed):
approved for aaa code owner

DrkSephy (approved):
LGTM for Dashboards

zippolyte (dismissed):

nmuesch:
Apologies for everyone being notified on this one. The goal is to try and clean up the structure of the repo a bit since we're getting

zippolyte (dismissed):

nmuesch:
Few changes since the last commits/reviews (in addition to rebasing latest master)

Most notably: 
* Moved utils + validations to their own respective packages in `internal`
  * Moved the tests for these alongside their code in the `internal` package. The `make test` command still picks this up since its specifying the packages as `./...` wildcard
* Added a "Deprecated" comment to the utilities section of the `screenbaord` resource. It looks like that was previously used to place any and all utils.

datadog/resource_datadog_integration_aws.go
```diff
@@ -6,13 +6,15 @@ import (
...
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
+
 	datadogV1 "github.com/DataDog/datadog-api-client-go/api/v1/datadog"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
 var integrationAwsMutex = sync.Mutex{}
 
-func accountAndRoleFromID(id string) (string, string, error) {
+func AccountAndRoleFromID(id string) (string, string, error) {
```
zippolyte, line 17: shouldn't this be moved to utils to follow the pattern ?

datadog/resource_datadog_role.go
```diff
@@ -90,32 +90,14 @@ func validatePermissionsUnrestricted(value interface{}, meta interface{}) error
 	return nil
 }
 
-func getRolePermissionSchema() *schema.Resource {
```
zippolyte, line 93: This I'd keep here, it's for the schema it's interesting to have in the same file

datadog/tests/key_test.go
```diff
@@ -1,8 +1,10 @@
-package datadog
```
zippolyte, line 1: this and it's main file `key.go` should be moved to internal package

zippolyte (changes requested):

datadog/resource_datadog_integration_aws.go
nmuesch: Good catch, I thought I moved this but I only moved the `AccountAndNamespaceFromID` function. 
Moving now

datadog/resource_datadog_role.go
nmuesch: Yeah thats fair. The schema is used in tests, so it'd have to be exported, but I agree it should live here.
Thinking more, the Schema is pretty much the external API of a provider, so 👍 to exporting.

zippolyte (approved):

PR #903, opened 2021-02-08
Add a DEVELOPMENT file

nmuesch:
Add a DEVELOPMENT.md file to the repository and remove a large portion of the README. 
The goal is to more readily document the steps we take when working with and developing on the provider.


zippolyte (dismissed):

DEVELOPMENT.md
```diff
@@ -0,0 +1,71 @@
+# Development
+
+Terraform provides helpful [Extending Terraform][1] documentation for best practices around writing provider code. This document provides some guidelines for working with this project.
+
+## Prerequisites:
+
+-   [Terraform][2] 0.10.x
```
therve, line 7: We should probably advise 0.12 right away.

DEVELOPMENT.md
```diff
@@ -0,0 +1,71 @@
...
+-   [Go][3] 1.15 (to build the provider plugin)
+-   A clone of this repository and the [\$GOPATH environment variable][7] set
+-   [tfplugindocs][8]
```
therve, line 10: Do we have instructions on how to install it? Not sure I managed to do it right last time.

DEVELOPMENT.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Building the provider
+
+The Datadog Provider can be built to use the binary as a terraform plugin. This is most useful when attempting to build off a feature branch and manually run the `terraform plan|apply` commands on a real HCL configuration.
+
+This provider can be built by running `make build`, or just `make`. This will place the binary in `$GOPATH/bin`
+
+Running `terraform init` in a folder containing `.tf` files will let terraform look for any needed providers. The default behavior is to pull the latest released binary, but you can tell the terraform CLI to look for your locally built binary instead.
+
+The steps for this approach can differ depending on the version of Terraform being used. More information can be found on the official [Terraform documentation][4].
```
therve, line 24: Do you think it's worth adding the bits that work for 0.14?

DEVELOPMENT.md
```diff
@@ -0,0 +1,71 @@
...
+In order to update the underlying API Clients that are used by this provider to interact with the Datadog API, run:
+
+```sh
+API_CLIENT_VERSION=vx.y.z ZORKIAN_VERSION=vx.y.z make update-go-client
+```
+
+where:
+
+-   `API_CLIENT_VERSION` is the version or commit ref of the https://github.com/DataDog/datadog-api-client-go client.
+-   `ZORKIAN_VERSION` is the version or commit ref of the https://github.com/zorkian/go-datadog-api client.
```
therve, line 62: I'll add a note like "if you run it just after a release, this will automatically pick up the latest tag without needing to specify the version".

DEVELOPMENT.md
nmuesch: Also added a note about `tfenv`

DEVELOPMENT.md
nmuesch: Installation should be a combination of a standard `go get/install` and having GOPATH/bin in your path? I was hoping to not go too detailed on how to install the dependencies and instead rely on their documentation

DEVELOPMENT.md
nmuesch: I haven't done the dive on 0.14 and how it plays with installing as a plugin yet. I personally use 0.12 when I do these steps. Given that this workflow has changed between 0.12, 0.13, and 0.14, I'm not sure its worthwhile to add here yet. Given that, my hope was, for now, to rely on the official TF docs

Though this PR is also just a start. If we find that people want specific steps for a certain version (or if the 0.14 approach is stable/easy, we can definitely add it in

nmuesch: Spent some time yesterday going through the new `dev_overrides` pieces that 0.14 introduced. I included a section on that. I don't think we should let this section get too verbose though if the process changes again. Let me know what you think! (And thanks for the offline help getting the issues I was hitting sorted out)

fanny-jiang (approved):

therve (approved):

PR #909, opened 2021-02-10
Improve consistency by using response from PUT requests directly to save state

zippolyte:
Avoid consistency issues, due to replication delays, caused by calling the GET endpoint right after PUT.
This leaves the Create method as it is, as there is already a retry on 404 for this problem, and we need this for updating dashboard lists when used.


matt-miller-ddog (dismissed):

datadog/resource_datadog_dashboard.go
```diff
@@ -193,7 +198,7 @@ func updateDashboardLists(d *schema.ResourceData, providerConf *ProviderConfigur
 	}
 }
 
-func loadDatadogDashboard(d *schema.ResourceData, dashboard datadogV1.Dashboard) error {
+func loadDatadogDashboard(d *schema.ResourceData, dashboard *datadogV1.Dashboard) error {
```
nmuesch, line 201: Small nit, can we name this how you're doing the newer ones; i.e. monitor is going to have - `updateMonitorState`

nmuesch (approved):
LGTM, approving but left one nit

nmuesch (approved):

PR #915, opened 2021-02-11

zippolyte:

### DIFF
--- a/datadog/resource_datadog_logs_index_order.go
+++ b/datadog/resource_datadog_logs_index_order.go
@@ -51,24 +51,30 @@ func resourceDatadogLogsIndexOrderUpdate(d *schema.ResourceData, meta interface{
 	datadogClientV1 := providerConf.DatadogClientV1
 	authV1 := providerConf.AuthV1
 
-	if _, _, err := datadogClientV1.LogsIndexesApi.UpdateLogsIndexOrder(authV1).Body(ddIndexList).Execute(); err != nil {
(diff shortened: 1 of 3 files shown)


txominpelu (approved):

datadog/resource_datadog_logs_index_order.go
```diff
@@ -51,24 +51,30 @@ func resourceDatadogLogsIndexOrderUpdate(d *schema.ResourceData, meta interface{
...
+	return nil
 }
 
 func resourceDatadogLogsIndexOrderRead(d *schema.ResourceData, meta interface{}) error {
 	providerConf := meta.(*ProviderConfiguration)
-	client := providerConf.CommunityClient
-	ddIndexList, err := client.GetLogsIndexList()
+	client := providerConf.DatadogClientV1
+	auth := providerConf.AuthV1
+	ddIndexList, _, err := client.LogsIndexesApi.GetLogsIndexOrder(auth).Execute()
```
nmuesch, line 73: is this the same endpoint, just different named method? (Asking because I'm surprised cassettes didn't need re-recording)

nmuesch (approved):
LGTM, just left one inline question, but if CI is green 💯

datadog/resource_datadog_logs_index_order.go
zippolyte: yeah it was using community client before, I just switched to ours
Also no cassettes cause no tests for this, I tested manually

nmuesch: Ahh ok cool! 
Can you mention the client change in the Pr title so its called out in the changelog? Otherwise LGTM

PR #917, opened 2021-02-11

zippolyte:

Also fixes some linting warnings.


txominpelu (approved):
LGTM !

therve (approved):

PR #922, opened 2021-02-12

zippolyte:

### DIFF
--- a/datadog/resource_datadog_metric_metadata.go
+++ b/datadog/resource_datadog_metric_metadata.go
@@ -73,14 +73,37 @@ func resourceDatadogMetricMetadataCreate(d *schema.ResourceData, meta interface{
 	authV1 := providerConf.AuthV1
 
 	id, m := buildMetricMetadataStruct(d)
-	_, _, err := datadogClientV1.MetricsApi.UpdateMetricMetadata(authV1, id).Body(*m).Execute()
+	createdMetadata, _, err := datadogClientV1.MetricsApi.UpdateMetricMetadata(authV1, id).Body(*m).Execute()
 	if err != nil {
 		return translateClientError(err, "error creating metric metadata")
 	}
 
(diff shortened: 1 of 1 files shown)


nmuesch (approved):

datadog/resource_datadog_metric_metadata.go
```diff
@@ -73,14 +73,37 @@ func resourceDatadogMetricMetadataCreate(d *schema.ResourceData, meta interface{
 	authV1 := providerConf.AuthV1
 
 	id, m := buildMetricMetadataStruct(d)
-	_, _, err := datadogClientV1.MetricsApi.UpdateMetricMetadata(authV1, id).Body(*m).Execute()
+	createdMetadata, _, err := datadogClientV1.MetricsApi.UpdateMetricMetadata(authV1, id).Body(*m).Execute()
```
jmfurlott, line 76: Not sure if this is related to this PR, but from testing out the responses from the PUT/GET endpoints, I found a bug where if a user sends a key that does not exist on the metadata table, the user's metadata gets completely wiped.  See the curls below:
```
➜  dogweb git:(prod) ✗ curl 'https://dd.datad0g.com/api/v1/metrics/TimedBillingSparkJob.job_runtime' \
 ... \
{"description":"hello world","short_name":null,"integration":null,"statsd_interval":null,"per_unit":null,"type":"gauge","unit":null}%                                                                                                                          ➜  dogweb git:(prod) ✗ curl 'https://dd.datad0g.com/api/v1/metrics/TimedBillingSparkJob.job_runtime' \
  -X 'PUT' \
  ... \
  --data '{ "foo": "bar", "_authentication_token": "841a3d66eaa21c6abedb4e520a5aee4ea7f8f3f7"}' \
{"description":null,"short_name":null,"integration":null,"statsd_interval":null,"per_unit":null,"type":"gauge","unit":null}%   
```

PR #924, opened 2021-02-12
Cleanup makefile, add testall, add rerun-failed

nmuesch:
This PR aims to cleanup the Makefile and the way tests are run:

* Adds rerun-failed to gotestsum so we can retry individual errors
  * The default here is sane, the test will be retried twice, and if more than 10 tests are failing, tests won't be retried at all by gotestsum since there's likely a larger issue at play
  * gotestsum requires you pass in the packages to it, so I also moved the `go list` to be part of the invocation to `gotestsum` instead of `go test` flags. 
* Add a new `make testall` command that will run both acceptance and unit tests
  * Stop `make test` from running the acceptance tests in record mode
* Set the default `RECORD` to `false` in the makefile and allow it to run `make testacc` or `make testall` with the flag to control if we playback or record during testing
* Remove the special make commands for installing the provider as a plugin for terraform 0.13; we explicitly document how to do this using terraform 0.14 in the new DEVELOPMENT.md file. 
* Update the DEV doc for people to use `testall` instead of `testacc`

### DIFF
--- a/.azure-pipelines/all.yml
+++ b/.azure-pipelines/all.yml
@@ -46,9 +46,10 @@ jobs:
       services:
           datadog-agent: datadog-agent
       steps:
-          - script: make testacc
-            displayName: Run integration tests
+          - script: make testall
+            displayName: Run all tests
             env:
+                RECORD: "none"
                 CI: "true"
                 DD_AGENT_HOST: datadog-agent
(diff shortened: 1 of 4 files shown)


fanny-jiang (approved):
LGTM. The changes are pretty clear and simplifies testing commands

jirikuncar (approved):

PR #925, opened 2021-02-15

zippolyte:


datadog/resource_datadog_role.go
```diff
@@ -119,102 +120,151 @@ func resourceDatadogRoleCreate(d *schema.ResourceData, meta interface{}) error {
...
-		if permName, ok := permsIDToName[permID]; ok {
-			permR := map[string]string{
-				"id":   permID,
-				"name": permName,
-			}
-			perms = append(perms, permR)
+	switch rolePerms := rolePermsI.(type) {
+	case []datadog.RelationshipToPermissionData:
+		for _, permI := range rolePerms {
+			perms = appendPerm(perms, &permI, permsIDToName)
```
therve, line 164: I think you can just pass `permI.GetId()`, this will simplify the code in `appendPerm`.

therve (changes requested):
One small suggestion, otherwise looks great.

datadog/resource_datadog_role.go
zippolyte: True, thanks for the suggestion

therve (approved):

therve (approved):

PR #929, opened 2021-02-16
Update 3rd Party License file and add check to CI

nmuesch:
* Updates the 3rd party csv file to note all the deps we have in go.sum
* Adds a new script `license-check.sh`, essentially a copy from the datadog-api-client-go with a change from `echo` -> `printf`
  * Run the script via `make license-check` and add this to the github action CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,8 @@ jobs:
               uses: actions/checkout@v2
             - name: Vet
               run: make vet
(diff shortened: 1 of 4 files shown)


jeremy-lq (changes requested):
These should be updated to SPDX identifiers

jeremy-lq (approved):

zippolyte (approved):

PR #931, opened 2021-02-17
Add SLO datasources

### LINKED ISSUE #911: Please add datadog_service_level_objective data source to provider
```
We use the DD provider to manage some, but not all dashboards in DataDog.  One common pattern we use is to build out a product specific dashboard and then consolidate some basic stats about that product with other products in a team or group overview dashboard.  Today, that is done in a central workspace that many teams modify to add their dashboards, exporting the widget information that is then consumed by modules that build the central dashboards.  This leads to a very tightly coupled design that results in a workspace that becomes larger and larger, and users have begun to bump into each other.  

Splitting the workspace up is possible, but accessing remote state for those workspaces ends up loading a very large amount of data in the central workspace memory due to an inability to pull "part" of a state.

If the dashboard data source supported exporting additional data about the dashboards, especially the widget information (id, text), that would allow for more decoupled designs, where we search for dashboards, and pull the widgetids out to build the centralized dashboards.
```
Editing this to preserve the original statement, but see conversation below, this initial description of the issue was wrong.

adrien-f:
Closes #911

Signed-off-by: Adrien FILLON <adrien.fillon@manomano.com>


adrien-f:
Hi 👋 

A few notes:

* Quite inspired by what has been done for the monitor Data source
* There's a bit of duplication between the single DS and the multiple DS, let me know if you'd like a refactor on this
* I came across a bit of a refactor here (you moved the tests in its own package/folder), let me know if there was anything else missing
* Exposed fields are basic as discussed in the original issue

Let me know what you think!

datadog/data_source_datadog_service_level_objectives.go
```diff
@@ -0,0 +1,109 @@
...
+			"id":   slo.GetId(),
+			"name": slo.GetName(),
+			"type": slo.GetType(),
+		})
+	}
+	if err := d.Set("slos", slos); err != nil {
+		return err
+	}
+	d.SetId("datadog-service-level-objectives")
+	//d.SetId(fmt.Sprintf("%d", hashcode.String(req.String())))
```
nmuesch, line 106: Looks like this may have been leftover?

nmuesch (commented):
Hey @adrien-f Thanks for the PR, and bearing with us during the recent repo restructure, we're trying to organize things a bit. 🙂 

We're currently running `tfplugindocs` with no flags, but only committing relevant changes as there's a couple final pieces we're working out there. 

I haven't gone fully over the tests yet, but the core code seems good to me. I would like to discuss the single+multiple datasource approach though. It would be nice if we could find a way to get this into a single data-source since the object really is the same its just a difference of how many gets returned.

Perhaps we could have a flag in the config like `multiple: true` (with any name 😉) and differentiate the logic in the READ that way?

docs/data-sources/service_level_objective.md
```diff
@@ -0,0 +1,30 @@
...
+subcategory: ""
+description: |-
+  Use this data source to retrieve information about an existing SLO for use in other resources.
+---
+
+# datadog_service_level_objective (Data Source)
+
+Use this data source to retrieve information about an existing SLO for use in other resources.
+
+
```
nmuesch, line 13: To get an HCL config generated here, can you add an example to the examples/data-source folder?

adrien-f:
Hi 👋 Sorry for the delay!

I've rebased and added examples.

Regarding the issue of having one more two datasources, I really have no opinion on this 😅 Looking at the providers in the registry like AWS or Azure I see they often have two datasources:

* https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/shared_image_versions
* https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/shared_image_version
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ec2_instance_type_offering
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ec2_instance_type_offerings

I'm not sure there are any clear guidelines on this, I suppose there's a reason? Maybe you could get some pointers from Hashicorp here?

In any case, thanks again for this 🙂 Let's me know,

nmuesch:
Hey @adrien-f Thanks for checking around and providing those links. Given that this pattern looks to be the standard, I think it's OK to proceed with two data sources. 

I plan on taking another pass through the PR, but could you update the branch with `master`? We've added some more CI for linting and docs and it would be good to confirm this is succeeding on those 🙂

datadog/data_source_datadog_service_level_objectives.go
```diff
@@ -0,0 +1,108 @@
...
+		slos = append(slos, map[string]interface{}{
+			"id":   slo.GetId(),
+			"name": slo.GetName(),
+			"type": slo.GetType(),
+		})
+	}
+	if err := d.Set("slos", slos); err != nil {
+		return err
+	}
+	d.SetId("datadog-service-level-objectives")
```
nmuesch, line 105: I think since the data-source allows filtering, and isn't a "global"/"unique" data-source that returns all SLOs, we'll need to give it a unique ID. Maybe something similar to - https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/data_source_datadog_security_monitoring_rules.go#L140

datadog/tests/data_source_datadog_service_level_objective_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func TestAccDatadogServiceLevelObjectiveDatasource(t *testing.T) {
+	ctx, accProviders := testAccProviders(context.Background(), t)
+	uniq := strings.ReplaceAll(uniqueEntityName(ctx, t), "-", "_")
+	accProvider := testAccProvider(t, accProviders)
+
+	resource.ParallelTest(t, resource.TestCase{
```
nmuesch, line 18: Sorry, you caught us in a time of some refactors/improvements to the project 🙂 
Could you update the beginning of these tests to match - https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/tests/data_source_datadog_dashboard_test.go#L13-L18

datadog/tests/data_source_datadog_service_level_objectives_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func TestAccDatadogServiceLevelObjectivesDatasource(t *testing.T) {
+	ctx, accProviders := testAccProviders(context.Background(), t)
+	uniq := strings.ReplaceAll(uniqueEntityName(ctx, t), "-", "_")
+	accProvider := testAccProvider(t, accProviders)
+
+	resource.ParallelTest(t, resource.TestCase{
```
nmuesch, line 18: Same here on the test updates 🙇

datadog/tests/data_source_datadog_service_level_objectives_test.go
```diff
@@ -0,0 +1,108 @@
...
+	})
+}
+
+func checkServiceLevelObjectivesDatasourceAttrs(accProvider *schema.Provider, uniq string) resource.TestCheckFunc {
+	return resource.ComposeTestCheckFunc(
+		resource.TestCheckResourceAttrSet("data.datadog_service_level_objectives.foo", "slos.0.id"),
+		resource.TestCheckResourceAttr("data.datadog_service_level_objectives.foo", "slos.0.name", uniq),
+		resource.TestCheckResourceAttr("data.datadog_service_level_objectives.foo", "slos.0.type", "metric"),
+	)
+}
```
nmuesch, line 61: I think it'd be helpful here to assert at least 2 SLOs are returned. Would you be able to add a new config for this?

datadog/data_source_datadog_service_level_objectives.go
```diff
@@ -0,0 +1,108 @@
...
+
+			// Computed values
+			"slos": {
+				Description: "List of SLOs",
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Description: "Id of Datadog service level objective",
```
nmuesch, line 46:
```suggestion
							Description: "ID of the Datadog service level objective",
```

datadog/data_source_datadog_service_level_objectives.go
```diff
@@ -0,0 +1,108 @@
...
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"name": {
+							Description: "Name of Datadog service level objective",
```
nmuesch, line 51:
```suggestion
							Description: "Name of the Datadog service level objective",
```

nmuesch (commented):
Overall these new data-sources look good! Thanks for working on this. 
I left some inline notes, but they're mostly docs + test comments.

adrien-f:
Heya, apologies for the delay in my latest update. 

I've updated the tests to include extra SLOs, let me know if you're satisfied with this. It also now computes a custom ID, I had to mess with pointers and I've tried to keep it concise enough and I've rebased everything.

Cheers,

zippolyte (approved):
Hey @adrien-f, sorry for the delay in taking another look at this.
This looks good to me, I've taken the liberty to rebase on master, as we've updated the go client which contained breaking changes. Hope that's fine.

adrien-f:
Awesome! Thanks a lot 🚀 !

apollocreed:
Yeah thx a lot much to both of you much appreciated

PR #932, opened 2021-02-17
Add support for slack channel resource

skarimo:


datadog/resource_datadog_integration_slack_channel.go
```diff
@@ -0,0 +1,200 @@
...
+	tfChannelDisplay := map[string]interface{}{}
+	tfChannelDisplay["message"] = ddChannelDisplay.GetMessage()
+	tfChannelDisplay["notified"] = ddChannelDisplay.GetNotified()
+	tfChannelDisplay["snapshot"] = ddChannelDisplay.GetSnapshot()
+	tfChannelDisplay["tags"] = ddChannelDisplay.GetTags()
+
+	return tfChannelDisplay
+}
+
+func teamNameAndChannelNameFromID(id string) (string, string, error) {
```
nmuesch, line 194: Would be good to add a unit test for this and maybe move to the internal/utils.go and internal/utils_test.go files?

datadog/tests/resource_datadog_integration_slack_channel_test.go
```diff
@@ -0,0 +1,177 @@
...
+	accProvider := testAccProvider(t, accProviders)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    accProviders,
+		CheckDestroy: testAccCheckDatadogIntegrationSlackChannelDestroy(accProvider),
+		Steps: []resource.TestStep{
+			{
+				// Workaround to ensure we create the slack integration before running the tests.
+				Config: emptyLogsArchiveConfig(),
```
nmuesch, line 30: nit, can we use a datasource here instead. Feels generally safer, though logs_archive_order may be fine.

nmuesch: Could we move this method call to the PreCheck instead actually?

nmuesch (commented):
Left a first pass, looking good though. 
Also please add the relevant doc pieces to the `examples` folder and run `tfplugindocs` to generate documentation

datadog/tests/resource_datadog_integration_slack_channel_test.go
skarimo: I believe the provider is initialized in the steps which is performed after the PreCheck so we might not have access to the datadog provider. Will double check

skarimo: Good idea on the data source but unfortunately we can't use `PreCheck` in this case. It seems PreCheck happens before the provider is resolved.

jirikuncar: You should be able to pass the `accProvider` to the `PreCheck` function via closure.

skarimo: Gave the above a shot and ran into: 
```
panic: interface conversion: interface {} is nil, not *datadog.ProviderConfiguration [recovered]
```
Although we initialize `accProvider`, it seems that the actual provider are loaded in during the test steps only. The accProvider seems to be just a map that is used to check what providers will be used in the tests, e.g. `datadog`.

Looking at the code PreCheck itself seems to be run before the providers are resolved: https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/resource/testing.go#L542-L544

Based on what I can tell, within Config step is the first place where the provider gets loaded in

zippolyte: Yeah precheck is a no go, I tried it too a while ago and couldn't haha.
Do you have to use an actual resource though ? What happens if you provide an empty string ?

skarimo: Tested it out and empty string will not work. Looks like terraform sdk expects strictly a resource in the config to load in the provider

zippolyte (approved):

Issue #884, opened 2021-02-02
Support get list of ipv4 by location in datadog_ip_ranges data source


kuanhn:
### Description

Because in the Synthetics test, sometimes we just want to use certain locations, we want to be able to improve the source so that we can get the IP in the `prefixes_ipv4_by_location` field by location.

https://docs.datadoghq.com/api/latest/ip-ranges/

### New or Affected Resource(s)

* datadog_ip_ranges

### Potential Terraform Configuration

```tf
data "datadog_ip_ranges" "test" {}

resource "aws_waf_ipset" "datadog_ip" {
  name = "datadog_ip"

  dynamic "ip_set_descriptors" {
    for_each = data.datadog_ip_ranges.test.synthetics_ipv4_by_location.ap-northeast-1

    content {
      type  = "IPV4"
      value = ip_set_descriptors.value
    }
  }
}
```

egorchabala:
Might be very convenient - we have a network firewall rule that allows connection from Datadog synthetic locations. The current process to get Datadog IPs is implying parsing JSON and using it inside Terraform code. The requested data resource will make the configuration simple and straightforward.

fanny-jiang:
Hey @kuanhn and @egorchabala, I'm working on a [PR](https://github.com/DataDog/terraform-provider-datadog/pull/934) to add support for this feature (see above). The API returns a map of locations and list of IPs for each location, however due to limitations in the Terraform SDK, creating a map of lists isn't supported, so the above configuration wouldn't work. In my PR, the state would be set with a map of locations and IPs as a concatenated string, delimited by `,`. It's not pretty, but you would be able to do a split on the string to get the IPs by location. All other potential schemas and data structures would make the configuration very cumbersome. Let me know your thoughts.

therve:
Closes with #934

kuanhn:
@fanny-jiang 
It's great. That will help us a lot. Thank you.

PR #934, opened 2021-02-17
add support for reading ipv4/6 prefixes by location

fanny-jiang:
Addresses GH issue [#884](https://github.com/DataDog/terraform-provider-datadog/issues/884). 

The implementation is a bit hacky--since the Terraform SDK currently doesn't support `TypeList` elements in `TypeMap` schema attributes ([see here](https://discuss.hashicorp.com/t/resource-schema-typelist-as-element-for-typemap/4809)). As a workaround, this implementation returns a map with each location as a key and a concatenated string of the IPs as the corresponding value.

```
     "synthetics_ipv4_by_location": {
         "aws:ap-northeast-1": "13.114.211.96/32,52.192.175.207/32,13.115.46.213/32",
        [...]
      },
``` 

The drawback of this implementation is that it wouldn't be possible to iterate over the string value for each individual IP prefix, without performing a split on the string. 

I also considered manipulating the returned map from the Read function into a list of lists structure, but that would make iterating and extracting values even more difficult and cumbersome.


pietdaniel:
not sure why AAA is pinged of reviewers here, let me know how I can lend if ✋ if needed though

datadog/data_source_datadog_ip_ranges.go
```diff
@@ -150,5 +163,28 @@ func dataSourceDatadogIPRangesRead(d *schema.ResourceData, meta interface{}) err
...
+		ipv4PrefixesByLocationMap[key] = strings.Join(value, ",")
+	}
+
+	for key, value := range ipv6PrefixesByLocation {
+		ipv6PrefixesByLocationMap[key] = strings.Join(value, ",")
+	}
+
+	err = d.Set("synthetics_ipv4_by_location", ipv4PrefixesByLocationMap)
+	if err != nil {
+		log.Printf("Error setting value: %s", err)
```
nmuesch, line 182: Lets `return err` here instead so it bubbles up

datadog/data_source_datadog_ip_ranges.go
```diff
@@ -150,5 +163,28 @@ func dataSourceDatadogIPRangesRead(d *schema.ResourceData, meta interface{}) err
...
+	}
+	err = d.Set("synthetics_ipv6_by_location", ipv6PrefixesByLocationMap)
+	if err != nil {
+		log.Printf("Error setting value: %s", err)
```
nmuesch, line 186: Same here

datadog/data_source_datadog_ip_ranges.go
```diff
@@ -48,6 +50,11 @@ func dataSourceDatadogIpRanges() *schema.Resource {
 				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"synthetics_ipv4_by_location": {
+				Description: "List of IPv4 prefixes by location.",
```
nmuesch, line 54: So that it winds up in documentation, do you think it'd be useful to mention how the list can be split here? (Not sure how much wording is needed to show that)

nmuesch (commented):
PR Looks good overall. Left a few minor comments

Can you also regenerate the docs with the latest schema - https://github.com/DataDog/terraform-provider-datadog/blob/master/DEVELOPMENT.md#generating-documentation (I recently merged a PR so you can run `make tools` to get tfplugindocs locally. 

@pietdaniel it looks like aaa is listed as codeowners of this resource - https://github.com/DataDog/terraform-provider-datadog/blob/master/.github/CODEOWNERS#L18-L19 Should this be updated to another team?

datadog/data_source_datadog_ip_ranges.go
fanny-jiang: Yeah I think that'll be useful, thanks for the suggestion!

nmuesch (approved):
Thanks, this LGTM! 

Can you just rebase with master? I regenerated all the docs to be up to date and it looks like that caused a conflict (sorry about that)

sdeprez (approved):

jtappa:
👍🏻  from Docs

PR #937, opened 2021-02-19
Fix test spans duration

therve:
There is an unnecessary span at least, and we don't measure them properly in parallel tests. This works around it by using custom test functions.

### DIFF
--- a/datadog/tests/provider_test.go
+++ b/datadog/tests/provider_test.go
@@ -349,7 +349,7 @@ func testSpan(ctx context.Context, t *testing.T) (context.Context, func()) {
 	if err != nil {
 		t.Fatal(err.Error())
 	}
-	return ddtesting.StartSpanWithFinish(ctx, t, ddtesting.WithSkipFrames(5), ddtesting.WithSpanOptions(
+	return ddtesting.StartSpanWithFinish(ctx, t, ddtesting.WithSkipFrames(3), ddtesting.WithSpanOptions(
 		// We need to make the tag be something that is then searchable in monitors
 		// https://docs.datadoghq.com/tracing/guide/metrics_namespace/#errors
 		// "version" is really the only one we can use here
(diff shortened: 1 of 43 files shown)


zippolyte (dismissed):

datadog/tests/provider_test.go
```diff
@@ -598,3 +599,15 @@ func CheckResourceAttr(name, key, value string) resource.TestCheckFunc {
...
+func parallelTest(ctx context.Context, t *testing.T, c resource.TestCase) {
+	t.Helper()
+	t.Parallel()
+	test(ctx, t, c)
+}
+
+func test(ctx context.Context, t *testing.T, c resource.TestCase) {
+	_, finish := testSpan(ctx, t)
+	t.Cleanup(finish)
+	resource.Test(t, c)
```
jirikuncar, line 612:
```suggestion
func test(ctx context.Context, t *testing.T, c resource.TestCase) context.Context {
	ctx, finish := testSpan(ctx, t)
	t.Cleanup(finish)
	resource.Test(t, c)
	return ctx
```

datadog/tests/provider_test.go
```diff
@@ -598,3 +599,15 @@ func CheckResourceAttr(name, key, value string) resource.TestCheckFunc {
 		return nil
 	}
 }
+
+func parallelTest(ctx context.Context, t *testing.T, c resource.TestCase) {
+	t.Helper()
+	t.Parallel()
+	test(ctx, t, c)
```
jirikuncar, line 606:
```suggestion
func parallelTest(ctx context.Context, t *testing.T, c resource.TestCase) context.Context {
	t.Helper()
	t.Parallel()
	return test(ctx, t, c)
```

datadog/tests/provider_test.go
```diff
@@ -515,11 +516,11 @@ func TestProvider_impl(t *testing.T) {
...
 }
 
-func testAccPreCheck(t *testing.T) {
+func testAccPreCheck(ctx context.Context, t *testing.T) {
 	if isReplaying() {
 		return
 	}
-	_, finish := testSpan(context.Background(), t)
+	_, finish := testSpan(ctx, t)
 	t.Cleanup(finish)
```
jirikuncar: it doesn't seem like the right place to create test span as the context can not be passed to clients.

jirikuncar (approved):

PR #939, opened 2021-02-22
Add note about APM distribution metrics to terraform monitors

andrewardito:
Add note about apm distribution metrics.


docs/resources/monitor.md
```diff
@@ -44,6 +44,8 @@ resource "datadog_monitor" "foo" {
 }
 ```
 
+**Note:** APM Trace Metrics are now available as distributions and your existing monitors have been migrated.  To learn more, or to see examples of how to update your terraform definitions to utilize the new distribution metrics, see the [migration guide](https://docs.datadoghq.com/tracing/guide/ddsketch_trace_metrics/)
```
nmuesch, line 47: The `docs` folder is autogenerated based on both the `examples` folder, and the `Description` field in each resource. 

We'd either have to add this as a note to a description, or create a new template for the monitor doc that hardcodes this in. The template would be `templates/resources/<resource name>.md[.tmpl]`. More info on that here - https://github.com/hashicorp/terraform-plugin-docs#templates

Though it'd be nicer if we could include this as a description of one of the elements if thats possible 🙂

omris1, line 47: Some light suggestions: "APM latency data is now available as Distribution Metrics. Existing monitors have been migrated automatically but all terraformed monitors can still use the existing metrics. To learn more, or to see examples of how to update your terraform definitions to utilize the new distribution metrics, see the [detailed doc]"

pkalmakis: Can we say something like

> Terraformed monitors can still utilize the existing metrics, but we strongly recommend updating monitor definitions to query the new metrics. To learn more...

pkalmakis: It sounds like a new template would be the best bet here. We didn't actually change anything in the terraform elements themselves.

nmuesch: I was thinking about this more. I'm thinking it may be a better idea to just append to the resource description so it appears at the top of the page:
Adding to the description here - https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_monitor.go#L31
Something like: `\n\n<Your note>`

If that doesn't work out/isn't how we want it to display, then I can help make a new template. I'd just like "new template" to be our last option since it means we'd be maintaining that to make sure it was up to date with the doc autogeneration tool changes.

andrewardito:
image

docs/resources/monitor.md
andrewardito: ^ tried this (thanks for the help again Nick! 

Added a comment with the preview https://github.com/DataDog/terraform-provider-datadog/pull/939#issuecomment-783552845

docs/resources/monitor.md
```diff
@@ -44,69 +48,70 @@ resource "datadog_monitor" "foo" {
 }
 ```
 
 ## Schema
 
 ### Required
 
-- **message** (String, Required) A message to include with notifications for this monitor.
+- **message** (String) A message to include with notifications for this monitor.
```
nmuesch, line 56: Everything below this is cosmetic from an upgrade of the `tfplugindoc` tool 👍

nmuesch (approved):

zippolyte (approved):

Issue #926, opened 2021-02-15
Provider can't apply exported dashboard, refuses template_variable_presets without template_variable


mgarstecki:
### Terraform Version

```
$ terraform-0.14.6 -v
Terraform v0.14.6
+ provider registry.terraform.io/datadog/datadog v2.21.0
```

### Affected Resource(s)
- datadog_dashboard

### Terraform Configuration Files

Simplified to best showcase the issue.

```hcl
resource "datadog_dashboard" "redacted_id" {
  is_read_only = "false"
  layout_type  = "ordered"
  title = "Sample dashboard"

  template_variable {
    default = "prod"
    name    = "env"
    prefix  = "env"
  }

  template_variable {
    default = "eu"
    name    = "continent"
    prefix  = "continent"
  }

  template_variable_preset {
    name = "prod-eu"
  }

  template_variable_preset {
    name = "prod-us"

    template_variable {
      name  = "continent"
      value = "us"
    }

    template_variable {
      name  = "env"
      value = "prod"
    }
  }

  widget {
    timeseries_definition {
      request {
        display_type   = "line"
        on_right_yaxis = "false"
        q              = "max:kubernetes_state.deployment.replicas_desired{$env,$continent,kube_deployment:badoom3} by {kube_deployment}, max:kubernetes_state.deployment.replicas_unavailable{$env,$continent,kube_deployment:badoom3} by {kube_deployment}.rollup(max)"
      }
    }
  }
}

```

### Debug Output

https://gist.github.com/mgarstecki/1a20c39f5fc6e5f3dcb76c29e42b6f39

### Panic Output

None

### Expected Behavior

The plan finishes without errors.

### Actual Behavior

The plan gives this error:

```
Error: "template_variable_preset.0.template_variable": required field is not set

  on dashboard.tf line 1, in resource "datadog_dashboard" "redacted_id":
   1: resource "datadog_dashboard" "redacted_id" {
```

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:

1. `terraform plan` or `terraform apply`

### Important Factoids

This TF configuration was generated with [Terraformer](https://github.com/GoogleCloudPlatform/terraformer), but a manual export of the dashboard to JSON through the Datadog UI also returns an empty block, so this seems correct behavior from Terraformer. 

`template_variable_preset` blocks without `template_variable` subblocks look legit to the Datadog API, and can be imported through the Datadog UI, so this looks like to me like the TF provider is too restrictive.

[...]

PR #944, opened 2021-02-22
Fix template_variable_presets to support optional template_variables

fanny-jiang:
Addresses issue [#926](https://github.com/DataDog/terraform-provider-datadog/issues/926) where the provider would not properly apply `template_variable_presets` without specified `template_variables`. This fix aligns better with the API spec: [Dashboards](https://docs.datadoghq.com/api/latest/dashboards/#create-a-new-dashboard).

The fix involves making the `template_variable_presets` schema and its `template_variable` schema attributes optional. Also removes the conditional checking whether the `template_variable_presets.template_variable` list is empty--this allows the resource to set an empty list.


therve (approved):
Looks great!

PR #956, opened 2021-02-26
Fix datadog security monitoring default rule

### LINKED ISSUE #955: Updating datadog_security_monitoring_default_rule cases notifications doesn't work
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

Terraform v0.14.7

### Affected Resource(s)

-   datadog_security_monitoring_default_rule

```hcl
resource "datadog_security_monitoring_default_rule" "aws_cmk_deleted" {
  enabled = true

  case {
    status = "high"
    notifications = ["alexej@foobar.com"]
  }
}
```

### Expected Behavior

After running `terraform import` to import the security monitoring default rule into the state, we expect being able running `terraform apply ` to update the rule notifications recipients.

### Actual Behavior

After running `terraform import` to import the security monitoring default rule into the state, running `terraform apply ` to update the rule notifications recipients does successfully execute, but the rule case notifications recipients update is not reflected datadog side.

### Steps to Reproduce

1. Create a simple terraform configuration file as the one in the example above.
2. Run `terraform import` on the associated datadog security monitoring default rule
3. Run `terraform apply` to update the datadog security monitoring default rule to reflect the changes in the terraform configuration file
4. Verify on datadog console, that the rule case notifications recipients update hasn't been performed

### Important Factoids

While debugging the issue, we attempted running the acceptance tests on the `datadog_security_monitoring_default_rule`:

```
RECORD="none" DD_API_KEY="foo" DD_APP_KEY="bar" TESTARGS="-run TestAccDatadogSecurityMonitoringDefaultRule_Basic" make testacc
```

No matter what we change on the test suite, the acceptance tests seem to be always succeeding, which is suspicious.

`datadog_security_monitoring_default_rule` is a special resource, as it *must* be imported before attempting any allowed manipulation on it. We've noticed, that the acceptance suite test steps are all defining `ImportState: true`, so we're wondering if the `Check` sections actually have any effects, given that they assume to be executed on terraform applied `Config`, from the docs:

> Check is called after the Config is applied. Use this step to make your own API calls to check the status of things, and to inspect the format of the ResourceState itself.

> If an error is returned, the test will fail. In this case, a destroy plan will still be attempted.

rymir:
We're attempting sorting some unexpected behaviours we're experiencing when using the datadog `datadog_security_monitoring_default_rule` resource.

As described over the linked #955 issue, updating the default rule cases notifications doesn't seem to be functional.

An additional fix included in this pull request is related to the terraform state being persisted when attempting manipulating a default rule case that isn't available datadog side. (https://github.com/DataDog/terraform-provider-datadog/commit/8f854d245a3168152087a1db112ddbf20cc6c090)

When starting working on a fix for this, we tried executing the acceptance tests for the resource, but we've been caught by surprise in noticing they're always succeeding, no matter what values we use in tests `Check` steps. More details about this in the linked issue #955 (`Important Factoids` section).

We'd like to contribute fixing the resource acceptance tests, but need some guidance and context on the current test suite  implementation.

Closes #955


greentruff:
Thanks for the report @rymir! I'll take a look to understand why the tests are not failing

greentruff:
I looked through the implementation of `terraform-plugin-sdk` which we rely on for tests and it seems like it isn't possible to check for modifications on an imported resource. I have opened [an issue](https://github.com/hashicorp/terraform-plugin-sdk/issues/717) on that project. In the mean time, the best we can do is to only keep the first import test and remove all the ones with `Check`. 
I have updated the tests and and the cassettes in [this PR](https://github.com/rymir/terraform-provider-datadog/pull/1) if you are not familiar with the test suite.

rymir:
@greentruff rebased and pushed

greentruff (approved):
LGTM, can you check and merge @therve ?

therve (approved):

PR #960, opened 2021-03-02
add datadog_metric_tag_configuration resource

efraese:
This PR adds support for a new `datadog_metric_tag_configuration` resource. This resource utilizes the V2 Metric tag configuration API to configure tags on metrics ([public docs are here](https://docs.datadoghq.com/api/latest/metrics/)). 

Verified the following cases and used the `GET /api/v2/{metric_name}/tags` to verify attributes:
- Create a new distribution, count, gauge tag configuration ✅ 
- Update each type of tag configuration (change tags, toggle percentiles) ✅ 
- Ensure we cannot toggle percentiles for non-distribution metrics ✅ 
- Import each type of tag configuration ✅ 
- attempt to import a tag configuration that does not exist (will error) ✅ 
- modifying metric name or metric type results in a new resource ✅ 
- delete resource ✅


chuck:
> * modifying metric name or metric type results in a new resource ✅

Does this act like a rename where the old resource is removed?

efraese:
> > * modifying metric name or metric type results in a new resource ✅
> 
> Does this act like a rename where the old resource is removed?

Yes, this would first delete the existing resource by calling the delete endpoint and then create the new resource using post.

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+				return utils.TranslateClientError(err, "error validating diff")
+			}
+			if includePercentilesOk && *metricTypeValidated != datadogV2.METRICTAGCONFIGURATIONMETRICTYPES_DISTRIBUTION {
+				return fmt.Errorf("cannot use include_percentiles with a metric_type of %s, must use metric_type of 'distribution'", metricType)
+			}
+
+			return nil
+		},
+		Importer: &schema.ResourceImporter{
+			State: resourceDatadogMetricTagConfigurationImport,
```
therve, line 44: Any reason to not use `ImportStatePassthrough` here?

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"metric_name": {
+				Description: "The metric name for this resource.",
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				ValidateFunc: func(val interface{}, k string) (warns []string, errs []error) {
```
therve, line 53: You should be able to rewrite this using helpers `StringIsValidRegExp` and `StringLenBetween`.

therve (changes requested):
Some minor comments, otherwise looks great. Only required change is in testing.

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -0,0 +1,311 @@
...
+	resultMetricTagConfigurationData, err := buildDatadogMetricTagConfiguration(d)
+	if err != nil {
+		return utils.TranslateClientError(err, "error building MetricTagConfiguration object")
+	}
+	ddObject := datadogV2.NewMetricTagConfigurationCreateRequestWithDefaults()
+	ddObject.SetData(*resultMetricTagConfigurationData)
+	metricName := d.Get("metric_name").(string)
+
+	// check if the tag configuration already exists, if so return an error
+	_, httpresp, err := datadogClient.MetricsApi.ListTagConfigurationByName(auth, metricName).Execute()
```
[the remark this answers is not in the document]
efraese: yes, the create endpoint will return `409` in this case. I'll remove this check here and just let the create call fail.

datadog/resource_datadog_metric_tag_configuration.go
efraese: tested this and `ImportStatePassthrough` works properly!

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -0,0 +1,299 @@
...
+		Schema: map[string]*schema.Schema{
+			"metric_name": {
+				Description: "The metric name for this resource.",
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				ValidateFunc: func(val interface{}, k string) (warns []string, errs []error) {
+					v := val.(string)
+					regExpWarns, regExpErrs := validation.StringMatch(regexp.MustCompile(`^[A-Za-z][A-Za-z0-9\\.\\_]*$`), "metric name must be valid")(v, k)
+					lenWarns, lenErrs := validation.StringLenBetween(1, 200)(v, k)
```
therve, line 57: I think you can use https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/validation#All and remove most of the code in there.

therve (approved):

PR #962, opened 2021-03-02
Explicitly check monitor_id for nil value

skarimo:
This PR checks if monitor_id is nil before setting it. Previous approach was causing value of 0 to be stored in state. See terraformer issue here: https://github.com/GoogleCloudPlatform/terraformer/issues/818

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -374,10 +374,9 @@ func updateDowntimeState(d *schema.ResourceData, dt *datadogV1.Downtime) error {
 	if err := d.Set("message", dt.GetMessage()); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


datadog/resource_datadog_downtime.go
```diff
@@ -374,10 +374,9 @@ func updateDowntimeState(d *schema.ResourceData, dt *datadogV1.Downtime) error {
 	if err := d.Set("message", dt.GetMessage()); err != nil {
 		return err
 	}
-	if err := d.Set("monitor_id", dt.GetMonitorId()); err != nil {
-		return err
+	if v, ok := dt.GetMonitorIdOk(); ok && v != nil {
+		d.Set("monitor_id", v)
```
nmuesch, line 378: I think we still want to check and report if an error gets returned from `d.Set`

nmuesch (approved):

PR #964, opened 2021-03-03
(docs) Update links to datadog documentation

jtappa:
These links were pointing to an older page and were a bit confusing. We can't link to the exact spot in the table that's being referenced, but this should at least get readers a little closer to the info they need.


jtappa:
As a side note, I'm assuming that the automation will take care of updating them in the markdown file that appears to be generated, so I only changed/updated them in the Go file. Cheers!

sdeprez (approved):

nmuesch (approved):
Hey @jtappa Thanks for the PR. I'll get this merged and do the documentation autogeneration in a followup 🙇

PR #970, opened 2021-03-04
Fix lint reporting

therve:
We need to make it fail.

### DIFF
--- a/datadog/tests/resource_datadog_integration_slack_channel_test.go
+++ b/datadog/tests/resource_datadog_integration_slack_channel_test.go
@@ -33,7 +33,7 @@ func TestAccDatadogIntegrationSlackChannel_Basic(t *testing.T) {
 				),
 			},
 			{
-				Config: testAccCheckDatadogIntegrationSlackChannelConfig_Create(uniqueChannelAccountName),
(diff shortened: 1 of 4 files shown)


datadog/tests/resource_datadog_integration_slack_channel_test.go
```diff
@@ -83,7 +83,7 @@ func testAccCheckDatadogIntegrationSlackChannelConfig_Create(uniq string) string
    `, uniq)
 }
 
-func testAccCheckDatadogIntegrationSlackChannelConfig_Update(uniq string) string {
+func testAccCheckDatadogIntegrationSlackChannelConfigUpdate(uniq string) string {
```
nmuesch, line 86: Do cassettes need recording if these change?

therve: No it's private helper.

zippolyte (approved):

datadog/tests/resource_datadog_integration_slack_channel_test.go
nmuesch: Ahh right, I saw testAcc and assumed 🙂 
LGTM then 👍

nmuesch (approved):

PR #1002, opened 2021-03-23
[Synthetics] Various fixes

romainberger:
Various fixes for synthetics:

- mark `tick_every` as required
- use a default value for variables `example` field.


fanny-jiang (changes requested):
Let's see if fixing the duplicate `Default` in the `syntheticsBrowserStepParams` schema will make the tests pass. Otherwise, the rest looks good!

fanny-jiang:
Looks good. I think now you just need to run `make fmt` to reformat the code.

PR #1005, opened 2021-03-24
[Synthetics] Use new models for api tests

romainberger:
Use new endpoint for synthetics api tests so we can use the correct object and implement multistep tests (incoming PR).


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -782,19 +782,39 @@ func resourceDatadogSyntheticsTestCreate(d *schema.ResourceData, meta interface{
...
 
-	// If the Create callback returns with or without an error without an ID set using SetId,
-	// the resource is assumed to not be created, and no state is saved.
-	d.SetId(createdSyntheticsTest.GetPublicId())
+	if testType == datadogV1.SYNTHETICSTESTDETAILSTYPE_API {
+		syntheticsTest := buildSyntheticsAPITestStruct(d)
+		createdSyntheticsTest, _, err := datadogClientV1.SyntheticsApi.CreateSyntheticsAPITest(authV1).Body(*syntheticsTest).Execute()
+		if err != nil {
+			// Note that Id won't be set, so no state will be saved.
+			return utils.TranslateClientError(err, "error creating synthetics test")
```
zippolyte, line 792:
```suggestion
			return utils.TranslateClientError(err, "error creating synthetics API test")
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -782,19 +782,39 @@ func resourceDatadogSyntheticsTestCreate(d *schema.ResourceData, meta interface{
...
+		d.SetId(createdSyntheticsTest.GetPublicId())
+
+		// Return the read function to ensure the state is reflected in the terraform.state file
+		return resourceDatadogSyntheticsTestRead(d, meta)
+	} else if testType == datadogV1.SYNTHETICSTESTDETAILSTYPE_BROWSER {
+		syntheticsTest := buildSyntheticsBrowserTestStruct(d)
+		createdSyntheticsTest, _, err := datadogClientV1.SyntheticsApi.CreateSyntheticsBrowserTest(authV1).Body(*syntheticsTest).Execute()
+		if err != nil {
+			// Note that Id won't be set, so no state will be saved.
+			return utils.TranslateClientError(err, "error creating synthetics test")
```
zippolyte, line 806:
```suggestion
			return utils.TranslateClientError(err, "error creating synthetics browser test")
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -782,19 +782,39 @@ func resourceDatadogSyntheticsTestCreate(d *schema.ResourceData, meta interface{
...
+
+		// If the Create callback returns with or without an error without an ID set using SetId,
+		// the resource is assumed to not be created, and no state is saved.
+		d.SetId(createdSyntheticsTest.GetPublicId())
+
+		// Return the read function to ensure the state is reflected in the terraform.state file
+		return resourceDatadogSyntheticsTestRead(d, meta)
+	}
+
+	return fmt.Errorf("Unrecognized synthetics test type %v", testType)
```
zippolyte, line 817:
```suggestion
	return fmt.Errorf("unrecognized synthetics test type %v", testType)
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -803,18 +823,17 @@ func resourceDatadogSyntheticsTestRead(d *schema.ResourceData, meta interface{})
...
-		syntheticsTest, httpresp, err = datadogClientV1.SyntheticsApi.GetTest(authV1, d.Id()).Execute()
+	// get the generic test to detect if it's an api or browser test
+	syntheticsTest, httpresp, err = datadogClientV1.SyntheticsApi.GetTest(authV1, d.Id()).Execute()
 
-		// re-fetch test if it was actually a browser but we didn't have the info earlier
-		if syntheticsTest.GetType() == "browser" {
-			syntheticsTest, _, err = datadogClientV1.SyntheticsApi.GetBrowserTest(authV1, d.Id()).Execute()
-		}
+	if syntheticsTest.GetType() == datadogV1.SYNTHETICSTESTDETAILSTYPE_BROWSER {
+		syntheticsTest, _, err = datadogClientV1.SyntheticsApi.GetBrowserTest(authV1, d.Id()).Execute()
```
zippolyte, line 834: Why do we get a `SyntheticsTestDetails` here and not a `SyntheticsBrowserTest` ??

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -826,18 +845,34 @@ func resourceDatadogSyntheticsTestRead(d *schema.ResourceData, meta interface{})
...
-		// If the Update callback returns with or without an error, the full state is saved.
-		return utils.TranslateClientError(err, "error updating synthetics test")
+	testType := getSyntheticsTestType(d)
+
+	if testType == datadogV1.SYNTHETICSTESTDETAILSTYPE_API {
+		syntheticsTest := buildSyntheticsAPITestStruct(d)
+
+		if _, _, err := datadogClientV1.SyntheticsApi.UpdateAPITest(authV1, d.Id()).Body(*syntheticsTest).Execute(); err != nil {
+			// If the Update callback returns with or without an error, the full state is saved.
+			return utils.TranslateClientError(err, "error updating synthetics test")
```
zippolyte, line 867:
```suggestion
			return utils.TranslateClientError(err, "error updating synthetics API test")
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -826,18 +845,34 @@ func resourceDatadogSyntheticsTestRead(d *schema.ResourceData, meta interface{})
...
+		}
+	} else if testType == datadogV1.SYNTHETICSTESTDETAILSTYPE_BROWSER {
+		syntheticsTest := buildSyntheticsBrowserTestStruct(d)
+
+		if _, _, err := datadogClientV1.SyntheticsApi.UpdateBrowserTest(authV1, d.Id()).Body(*syntheticsTest).Execute(); err != nil {
+			// If the Update callback returns with or without an error, the full state is saved.
+			return utils.TranslateClientError(err, "error updating synthetics test")
```
zippolyte, line 874:
```suggestion
			return utils.TranslateClientError(err, "error updating synthetics browser test")
```

zippolyte (changes requested):
A few nits and one question

datadog/resource_datadog_synthetics_test_.go
romainberger: The `GetBrowserTest` operation was added before the API/browser objects were split. So it's still the "legacy" object returned by the endpoint. We could change it for a `SyntheticsBrowserTest` but it would create breaking changes.

zippolyte: The breaking changes would not be in terraform right ? Only in the go client ?
If that's so, I think we should do it, while we still can (go client beta) to avoid carrying this legacy weirdness in the client

romainberger: Yes we can do that, the provider would have to be fixed after the go client is updated but I can do it 👍

romainberger: Made the PR to update: https://github.com/DataDog/datadog-api-spec/pull/911
I'll update this PR after the go client is updated.

zippolyte (dismissed):

zippolyte (approved):

therve (approved):

PR #1015, opened 2021-03-31
Remove read from monitor creation

therve:
Remove the read to improve consistency and update the state directly
instead.


zippolyte (approved):
Oops I missed this one 🙈

PR #1020, opened 2021-04-07
[Synthetics] Small fixes for synthetics

romainberger:
Various fixes for synthetics:

- mark `tick_every` as required
- use a default value for variables `example` field.


sdeprez (approved):

zippolyte (dismissed):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -423,7 +423,7 @@ func syntheticsTestOptionsList() *schema.Schema {
 				"tick_every": {
 					Description:  "How often the test should run (in seconds). Current possible values are `900`, `1800`, `3600`, `21600`, `43200`, `86400`, `604800` plus `60` for API tests or `300` for browser tests.",
 					Type:         schema.TypeInt,
-					Optional:     true,
+					Required:     true,
```
zippolyte, line 426: Isn't there a default value here ? instead of marking as required ?

zippolyte: And does it fail currently if you don't specify it ?

romainberger: It does fail as it is required in the API so it will fail if it's not provided. I'm not sure we can use a default value because in the UI browser and API tests have different default values.

zippolyte: Sounds good, just wanted to make sure we wouldn't break current workflows by doing this

zippolyte (approved):

PR #1055, opened 2021-05-04
Update monitor `critical` threshold documentation

theotherjason:
Under the schema for `monitor_thresholds`, `critical` is listed as the "recovery threshold" rather than "threshold".


therve:
Thanks for the patch! This is generated so we'll need to change the source.

therve (approved):

PR #1061, opened 2021-05-11
Use code formatting in description for attribute remapper

murukesh-mohanan-paidy:
From "log attribute or tag", it's not clear what the permitted values of `source_type` and `target_type` are, exactly (`log attribute` and `tag`? `log-attribute` and `tag`? `attribute` and `tag`? ...). The examples all use `tag`, so the correct way to specify attribute isn't clear. The API docs use code formatting for `attribute` and `tag` and make it clear (see option 6 in [the "create a pipeline" section][1] of the API docs):

image

Doing the same here would increase clarity.

[1]: https://docs.datadoghq.com/api/latest/logs-pipelines/#create-a-pipeline


zippolyte (changes requested):
Docs are generated from the schemas definition, please change it in there too https://github.com/DataDog/terraform-provider-datadog/blob/e456e740de4c0c7c062e77f7a82eab5ec7e18ab7/datadog/resource_datadog_logs_custom_pipeline.go#L131-L133

zippolyte (approved):

zippolyte:
Hmm can you rebase on the latest master ? That should make the CI pass.

PR #1062, opened 2021-05-11
[graph-editor] Change Dashboard WidgetCustomLink properties

qincchen:
** Motivation
Terraform change to support new context menu link override functionality.  

API spec change: https://github.com/DataDog/datadog-api-spec/pull/971/


datadog/resource_datadog_dashboard.go
```diff
@@ -6180,35 +6180,63 @@ func getWidgetCustomLinkSchema() map[string]*schema.Schema {
...
+		},
+		"is_hidden": {
+			Description: "The flag for toggling context menu link visibility",
+			Type:        schema.TypeBool,
+			Optional:    true,
+		},
+		"override_label": {
+			Description: "The label id that refers to a context menu link item",
+			Type:        schema.TypeString,
+			Optional:    true,
```
qincchen, line 6198: Could not use `ConflictsWith` property here due to `custom_link` being in a list. 

See  https://github.com/hashicorp/terraform-plugin-sdk/issues/71

datadog/resource_datadog_dashboard.go
```diff
@@ -6180,35 +6180,64 @@ func getWidgetCustomLinkSchema() map[string]*schema.Schema {
...
-			Required:    true,
+			Optional:    true,
 		},
 		"link": {
 			Description: "The URL of the custom link.",
 			Type:        schema.TypeString,
-			Required:    true,
+			Optional:    true,
+		},
+		"is_hidden": {
```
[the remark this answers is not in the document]
zippolyte: @skarimo what causes the perpetual diff ?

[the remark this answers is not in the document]
skarimo: Chatted outside of the Github issue but to close the loop here, the perpetual diff is caused when both `label` and `label_override` is set.

We can keep the behavior as is but let's document that both fields should not be set at the same time.

datadog/resource_datadog_dashboard.go
```diff
@@ -5953,35 +5953,64 @@ func getWidgetCustomLinkSchema() map[string]*schema.Schema {
...
+			Description: "The flag for toggling context menu link visibility",
```
skarimo, line 5964:
```suggestion
			Description: "The flag for toggling context menu link visibility.",
```

datadog/resource_datadog_dashboard.go
```diff
@@ -5953,35 +5953,64 @@ func getWidgetCustomLinkSchema() map[string]*schema.Schema {
...
+			Type:        schema.TypeBool,
+			Optional:    true,
+		},
+		"override_label": {
+			Description: "The label id that refers to a context menu link item",
```
skarimo, line 5969:
```suggestion
			Description: "The label id that refers to a context menu link item.",
```

skarimo: Could we also document that `label` should not be set if `override_label` is set to true

skarimo (approved):

PR #1064, opened 2021-05-12
Update Datadog api go client

skarimo:
Update datadog api go client to use latest release.


phillip-dd:
👋 looks like this is failing due to https://github.com/DataDog/datadog-api-spec/pull/959 which inlined the properties of the SLO instead of the previous $ref. I think we should revert that PR for the time being and @DataDog/slo-app can revisit.

zippolyte (approved):

PR #1092, opened 2021-06-01
Properly handle recurring downtimes definitions

armcburney:
## Overview

This PR implements a workaround for handling recurring downtimes with the Datadog Terraform provider. _Recurring_ downtimes differ from regular 'one-off' downtimes in that subsequent recurrences are scheduled as new downtimes from the previous parent definition. Conceptually, we can think of this as a "linked list" of downtimes where each subsequent downtime is scheduled from the previous scheduled recurrence downtime definition. All fields are copied over from the previous parent with the exception of certain fields like `start` and `end`, which are calculated off of the `recurrence` attribute.

The Datadog Terraform provider keeps reference to this original parent downtime ID. Since subsequent recurrence downtimes are scheduled as new downtimes (with a new ID), updates in the UI/API to the existing 'child' downtime corresponding to the original recurrence would not previously be recognized when comparing the downtime's state with what we store in Terraform. Moreover, after downtimes expire, we delete them from our database after a certain period of time. This behavior was recently changed so that we _don't_ delete downtimes after they expire _if they are the first downtime in the recurrence chain_ (i.e., the original parent downtime).

Instead, we now return that downtime with a new `active_child` field in our `GET /api/v1/downtime` API - which we use to compare state with Terraform. This way updates from the UI/API will be propagated back to Terraform. Additionally, when making updates through Terraform, we call the `PUT /api/v1/downtime/{downtime_id}` endpoint with the `active_child` definition on the `active_child`'s ID, so that changes from Terraform will be made to the current active recurrence downtime.

## Caveats

**UPDATE:** We don't check the `start`/`end` boundaries for changes if the recurring downtime is a child to prevent superfluous diffs every time the downtime is rescheduled. The con to this approach is that if the `start`/`end` values are changed in the UI on a child recurring downtime, the diff will not be picked up by Terraform. We plan to iterate on this solution to address this shortcoming, but feel the benefits of the child/parent references being handled by Terraform are worth merging in.

~Since new downtimes are scheduled each time a recurrence is rescheduled, fields like `start` and `end` will perpetually differ after the first schedule when running `terraform plan`/`terraform apply`.~

```
$ terraform plan
datadog_downtime.recurring_downtime: Refreshing state... [id=1337]

  ~ update in-place

  # datadog_downtime.recurring_downtime will be updated in-place
  ~ resource "datadog_downtime" "recurring_downtime" {
      ~ end             = 1621016100 -> 1620929700
        id              = "1337"
      ~ message         = "test this out edit ui" -> "test this out"
      ~ start           = 1621015800 -> 1620929400

        # (1 unchanged block hidden)
    }

```

~I recommend we update the documentation around recurring downtimes to document this problem and recommended an approach to mitigating this (i.e., using `ignore_changes` to ignore the `start` and `end` updates).~

```
resource "datadog_downtime" "recurring_downtime" {
  scope        = ["*"]
  start        = 1620929400
  end          = 1620929700
  timezone     = "EST"
  monitor_tags = ["github:armcburney"]
  message      = "test this out"

  recurrence {
    type   = "days"
    period = 1
  }

  lifecycle {
    ignore_changes = [
      start,
      end
    ]
  }
}
```

## Fixes
- https://github.com/DataDog/terraform-provider-datadog/issues/109
- https://github.com/DataDog/terraform-provider-datadog/issues/393


datadog/resource_datadog_downtime.go
```diff
@@ -20,6 +21,80 @@ import (
...
+	GetActiveChildOk() (*datadogV1.DowntimeChild, bool)
+}
+
+// downtimeChild wraps the `datadogV1.DowntimeChild` struct via embedding to implement `downtimeOrDowntimeChild`
+// interface missing the `GetActiveChild`, `GetActiveChildOk` methods.
+type downtimeChild struct {
+	child *datadogV1.DowntimeChild
+}
+
+func (d *downtimeChild) GetId() int64 {
```
armcburney, line 46: I'm implementing an interface for the open-api-spec generated code, but the CI is failing on `golint` due to preferring `GetID` over `GetId` (breaks the open-api-spec interface). Not sure the best way to ignore this in CI as I'm unfamiliar with Golang.

armcburney: `golint` was deprecated by Google and `NOLINT` was not implemented as a feature before doing so. 

- https://github.com/golang/lint/issues/114
- https://github.com/golang/go/issues/38968

phillip-dd:
thanks @armcburney! This makes sense to me, except for this part:

>  fields like start and end will perpetually differ after the first schedule 

From a customer perspective I don't think this is what we want -  the provider should not be showing a diff when this is working as expected. As well, if customers use `ignore_changes`, then they won't be able to see any true diffs if the recurrence is updated in the UI.

Some other options:
* ignore start/end explicitly in the provider
* always compare start/end on the parent downtime (this would only catch changes if the original parent was changed)
* compare duration: e.g. end-start
* something else?

docs/resources/downtime.md
```diff
@@ -50,6 +50,7 @@ resource "datadog_downtime" "foo" {
 
 ### Optional
 
+- **active_child_id** (Number) The id corresponding to the downtime object definition of the active child for the original parent recurring downtime. This field will only exist on recurring downtimes.
```
phillip-dd, line 53: if we include this, I think it should be actually read only.

datadog/resource_datadog_downtime.go
```diff
@@ -197,6 +280,12 @@ func getDowntimeBoundaryTimestamp(d *schema.ResourceData, dateAttr, tsAttr strin
 // * `configTs` - desired value (from TF configuration) of the boundary
 // * `updating` - `true` if this call is from Update method of the downtime resource, `false` if from Create
 func downtimeBoundaryNeedsApply(d *schema.ResourceData, tsFrom string, apiTs, configTs int64, updating bool) (apply bool) {
+	// We don't want to apply the start/end boundaries for recurring child downtimes, as doing so will lead to 400s for:
```
phillip-dd, line 283: typo/grammar, i think this sentence isn't quite right

datadog/resource_datadog_downtime.go
```diff
@@ -412,8 +507,27 @@ func updateDowntimeState(d *schema.ResourceData, dt *datadogV1.Downtime) diag.Di
...
+		if err := d.Set("end", dt.GetEnd()); err != nil {
+			return diag.FromErr(err)
+		}
+		if attr, ok := dt.GetActiveChildOk(); ok {
+			if err := d.Set("active_child_id", attr.GetId()); err != nil {
+				return diag.FromErr(err)
+			}
+		}
+	case *downtimeChild:
+		if err := d.Set("active_child_id", dt.GetId()); err != nil {
```
phillip-dd, line 526: 🤔 hm, would this also lead to a terraform diff on each recurrence?

phillip-dd (commented):
👍 couple of questions, but this looks good to me. Definitely an improvement!

datadog/resource_datadog_downtime.go
phillip-dd: talked about this - since it's a computed property this should be okay

GNUmakefile
```diff
@@ -20,11 +20,11 @@ uninstall:
 	@rm -vf $(DIR)/terraform-provider-datadog
 
 # Run unit tests; these tests don't interact with the API and don't support/need RECORD
-test: get-test-deps fmtcheck lint
+test: get-test-deps fmtcheck
 	gotestsum --hide-summary skipped --format testname --debug --packages $(TEST) -- $(TESTARGS) -timeout=30s
 
 # Run acceptance tests (this runs integration CRUD tests through the terraform test framework)
-testacc: get-test-deps fmtcheck lint
+testacc: get-test-deps fmtcheck
```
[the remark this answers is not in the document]
therve: Yes we'll come back to fix it later.

docs/resources/downtime.md
phillip-dd: Bumping on this, is this part auto generated or a copy/paste?

phillip-dd (approved):
lgtm - we may need to update the PR title for the change log

docs/resources/downtime.md
skarimo: The doc is auto generated. But to avoid this, we should remove the line below since the attribute is read only: https://github.com/DataDog/terraform-provider-datadog/blob/f06954f01639447aaeab733c6488555282a96f34/datadog/resource_datadog_downtime.go#L253

therve (approved):

PR #1099, opened 2021-06-07
Re-introduce monitor type diff suppression for query/metric alerts

### LINKED ISSUE #1100: datadog_monitor with complex query forces replacement every apply
### Terraform Version

```
 terraform -v
Terraform v0.15.5
on darwin_amd64
+ provider registry.terraform.io/datadog/datadog v3.0.0
```

### Affected Resource(s)

- datadog_monitor

```hcl
terraform {
  backend "local" {}
  required_providers {
    datadog = {
      source  = "Datadog/datadog"
      version = "~> 3.0.0"
    }
  }
}

resource "datadog_monitor" "example" {
  type    = "metric alert"
  name    = "Forces replacement every apply"
  query   = "sum(last_1d):sum:foo.count{env:prod,slo_p95:missed}.as_count() / ( ( 1 - 0.95 ) * ( sum:foo.count{env:prod,slo_p95:missed}.as_count() + sum:foo.count{env:prod,slo_p95:met}.as_count() ) ) > 4"
  message = "Example"
}
```

### Expected Behavior

Resource should apply idempotently and subsequent plans without changes to config should not force resource replacement.

### Actual Behavior

Each plan/apply determines the monitor type needs to be changed from `query alert` to `metric alert` which will force replacement.

```
tf plan 
datadog_monitor.example: Refreshing state... [id=37214355]

-/+ destroy and then create replacement

  # datadog_monitor.example must be replaced
-/+ resource "datadog_monitor" "example" {
      ~ evaluation_delay    = 0 -> (known after apply)
      ~ id                  = "37214355" -> (known after apply)
      - locked              = false -> null
        name                = "Forces replacement every apply"
      - no_data_timeframe   = 0 -> null
      - notify_audit        = false -> null
      - priority            = 0 -> null
      - renotify_interval   = 0 -> null
      - restricted_roles    = [] -> null
      - tags                = [] -> null
      - timeout_h           = 0 -> null
      ~ type                = "query alert" -> "metric alert" # forces replacement
    }

Plan: 1 to add, 0 to change, 1 to destroy.

------------------------------------------------------------------------

Note: You didn't specify an "-out" parameter to save this plan, so Terraform
can't guarantee that exactly these actions will be performed if
"terraform apply" is subsequently run.
```

### Steps to Reproduce

1. `terraform apply`
2. `terraform apply`

### Important Factoids

~~This is requiring manual intervention each apply in our case because the monitor that is showing up as requiring replacement is also a member of a composite monitor. The composite monitor relationship prevents the datadog_monitor from being deleted and requires manual intervention each cycle.~~

```
Error: error deleting monitor: 400 Bad Request: {"errors": ["monitor [37214355,Forces replacement every apply] is referenced in composite monitors: [37050226,Some composite monitor]"]}
```

~~I will file a separate issue about the inability to recreate a monitor that is referenced by a composite monitor, but thought it worth mentioning that it can exacerbate the negative impact of this `query alert -> metric alert` issue.~~

It looks like the issues around recreating monitors that are part of composite monitors are identified in https://github.com/DataDog/terraform-provider-datadog/issues/667 and adding `force_delete` will allow our applies to proceed.

### References

 - https://github.com/DataDog/terraform-provider-datadog/issues/3
Without investigating deeply, this seems very similar or even a revival of the above issue where sufficiently complex queries result in a metric monitor being classified as a query monitor.

skarimo:
There are still cases where the api will return `query alert` for `metric alert` types. This PR introduces the previous removed `DiffSuppressFunc`.

Fixes #1100

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -75,6 +75,15 @@ func resourceDatadogMonitor() *schema.Resource {
 				Required:         true,
 				ForceNew:         true,
 				ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewMonitorTypeFromValue),
+				// Datadog API quirk, see https://github.com/hashicorp/terraform/issues/13784
+				DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+					if (oldVal == "query alert" && newVal == "metric alert") ||
+						(oldVal == "metric alert" && newVal == "query alert") {
+						log.Printf("[DEBUG] Monitor '%s' got a '%s' response for an expected '%s' type. Suppressing change.", d.Get("name"), newVal, oldVal)
+						return true
+					}
+					return newVal == oldVal
+				},
 			},
 			"priority": {
 				Description: "Integer from 1 (high) to 5 (low) indicating alert severity.",


zippolyte (approved):

zippolyte:
Fixes #1100

Issue #1078, opened 2021-05-25
Add dashboard_lists property to datadog_dashboard_json


mBouamama:
Let's add support for the field in the same fashion as datadog_dashboard.

therve:
Hi,

The `datadog_dashboard` resource supports this via the `dashboard_lists` attribute, but not the `datadog_dashboard_json` one.

mBouamama:
hi, thank for your replying
Do you have any solution ofr my problem? or i need to wait an enhancement?
Regards

therve:
If you want to use the `datadog_dashboard_json`, you can use the `datadog_dashboard_list` resource to defined everything in terraform.

mBouamama:
Hi, thank for your replying
Ok it's alternative but if i want to add my dashboard(`datadog_dashboard_json`) to a existing list, it's impossible as the moment?
it's possible to request an enhancement for this feature?If yes, how ?
Regards

therve:
We'll track this here.

mBouamama:
Thanks

PR #1102, opened 2021-06-08
Add dashboard list support

skarimo:
This pr adds dashboard lists support for dashboard JSON resource.

Closes: https://github.com/DataDog/terraform-provider-datadog/issues/1078


datadog/resource_datadog_dashboard_json.go
```diff
@@ -112,6 +143,12 @@ func resourceDatadogDashboardJSONCreate(ctx context.Context, d *schema.ResourceD
 	}
 	d.SetId(id.(string))
 
+	layoutType, ok := respMap["layout_type"]
+	if !ok {
+		return diag.FromErr(errors.New("error retrieving layout_type from response"))
+	}
+	updateDashboardJSONLists(d, providerConf, id.(string), layoutType.(string))
```
zippolyte, line 150: Hmm this is tricky I think in case of create. We might need to add some retries here, make sure the dashboard is fully replicated in the backend, otherwise I believe the call to add to a list will fail.

skarimo: You are right, thanks!

datadog/resource_datadog_dashboard_json.go
```diff
@@ -176,3 +240,38 @@ func updateDashboardJSONState(d *schema.ResourceData, dashboard map[string]inter
 	}
 	return nil
 }
+
+func updateDashboardJSONLists(d *schema.ResourceData, providerConf *ProviderConfiguration, dashboardID, layoutType string) {
```
zippolyte, line 244: Actually, can't you reuse directly the same function as in the Dashboard resource ? Seems very similar apart from that `layoutType` argument

zippolyte (changes requested):

zippolyte (approved):

PR #1106, opened 2021-06-11
[Synthetics] Add support for global variables in config variables

romainberger:
Fix Synthetics test config variables by adding support for `global` variables.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -772,6 +772,11 @@ func syntheticsConfigVariable() *schema.Schema {
 					Required:         true,
```
zippolyte, line 772: Can you quickly add the new `global` allowed enum value in the description for type while we're here ?

zippolyte (approved):

PR #1109, opened 2021-06-14
Dashboard RBAC roles

matt-miller-ddog:
Adds support for per-dashboard RBAC (`restricted_roles`) configuration to `resource_datadog_dashboard`.

This PR is currently blocked until 1) the Go client has been updated with support for `restricted_roles`, 2) the Terraform provider has been updated to use a newer Go client which includes this support.

Note that this feature is not yet GA, and for now will only work for customers in the beta.


datadog/resource_datadog_dashboard.go
```diff
@@ -75,17 +75,24 @@ func resourceDatadogDashboard() *schema.Resource {
 				Optional:    true,
 				Description: "The description of the dashboard.",
 			},
+			"url": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The URL of the dashboard.",
+			},
 			"is_read_only": {
```
skarimo, line 84: Lets mark that this field conflicts with `restricted_roles`.

datadog/resource_datadog_dashboard.go
```diff
@@ -315,6 +328,10 @@ func buildDatadogDashboard(d *schema.ResourceData) (*datadogV1.Dashboard, error)
 	if v, ok := d.GetOk("is_read_only"); ok {
 		dashboard.SetIsReadOnly(v.(bool))
 	}
+	if v, ok := d.GetOk("restricted_roles"); ok && !dashboard.GetIsReadOnly() {
```
skarimo, line 331: We probably don't need `!dashboard.GetIsReadOnly()` once the fields are set to conflict with one another in the schema.

matt-miller-ddog: This one is necessary, for 1 specific case: when we're trying to set `is_read_only: true`. The issue is that if the API receives both `is_read_only` + `restricted_roles` on a request, the `restricted_roles` will take priority -- this is an issue in the case where a user has `is_read_only = true` in their config.

datadog/resource_datadog_dashboard.go
matt-miller-ddog: `restricted_roles` is marked as conflicting with `is_read_only`; do you mean "mark both fields as conflicting with each other", or "mark `is_read_only` instead of `restricted_roles`"?

skarimo: Sorry yeah I meant marking both fields as conflicting with one another

DrkSephy (dismissed):

DrkSephy (approved):

skarimo (approved):

PR #1116, opened 2021-06-21
[SMD-491] - Introduce the detections methods and the new value rules options

adussarps:
This PR extend the current datadog provider for security rules.

It add support for the detection method "new_value"; as for the detection method fields.

It will be followed by other PRs making the provider up to date with the current state of the API.


datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -108,6 +116,30 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
 						Required:         true,
 						Description:      "A signal will “close” regardless of the query being matched once the time exceeds the maximum duration. This time is calculated from the first seen timestamp.",
 					},
+
+					"new_value_options": {
```
greentruff, line 120: `new_value` would make more sense imo since this is already inside of `options`

greentruff (commented):
Can you also update the doc here? https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/resources/security_monitoring_rule.md

datadog/resource_datadog_security_monitoring_rule.go
adussarps: hmm; agreed; but it's the same "string" in the json api; so like this we have a direct mapping; I think it will help avoiding confusions.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -240,13 +272,15 @@ func buildCreatePayloadCases(d *schema.ResourceData) []datadogV2.SecurityMonitor
...
-		tfOptions = make(map[string]interface{})
-	} else {
-		tfOptions = tfOptionsList[0].(map[string]interface{})
+	tfOptions := extractMapFromInterface(tfOptionsList)
+
+	if v, ok := tfOptions["detection_method"]; ok {
+		detectionMethod := datadogV2.SecurityMonitoringRuleDetectionMethod(v.(string))
+		if detectionMethod.IsValid() {
+			payloadOptions.DetectionMethod = &detectionMethod
+		}
```
zippolyte, line 283: No need for this check, it's already ensured with the ValidateFunc in the schema

zippolyte:
```suggestion
		payloadOptions.DetectionMethod = &detectionMethod
```

PR #1118, opened 2021-06-21
Fix `retention_days` and `daily_limit` attributes 

jaredledvina:
From https://docs.datadoghq.com/api/latest/logs-indexes/#create-an-index it seems like `daily_limit` is optional which would align with the UI as well.


zippolyte (dismissed):

capucinemartin (dismissed):
LGTM

skarimo (approved):

PR #1121, opened 2021-06-23
Updates for monitors' restricted roles

gkharkwal:
1. Allow unsetting `restricted_roles` on a monitor resource.
2. Add `restricted_roles` to the monitor data source.

https://datadoghq.atlassian.net/browse/MA-2382


datadog/resource_datadog_monitor.go
```diff
@@ -556,6 +556,11 @@ func updateMonitorState(d *schema.ResourceData, meta interface{}, m *datadogV1.M
 	if err := d.Set("locked", m.Options.GetLocked()); err != nil {
 		return diag.FromErr(err)
 	}
+	// This helper function is defined in `resource_datadog_dashboard`
+	restrictedRoles := buildTerraformRestrictedRoles(m.RestrictedRoles)
```
phillip-dd, line 560: Should we move this to a shared location? Maybe internal/utils/builders or something? would defer to @DataDog/integrations-tools-and-libraries

phillip-dd (approved):
lgtm - thanks!

datadog/resource_datadog_monitor.go
gkharkwal: Yea, I wasn't sure where to put it but I agree with you.

Let's see what the team says.

jirikuncar: Since it is still in `datadog` package it doesn't really matter in which file it is defined. I agree that eventually we should move it to a separate file or closer to roles resource.

jirikuncar (approved):

zippolyte (approved):

PR #1124, opened 2021-06-24
Datasource datadog user

tleveque69:
Introducing a datasource to retrieve datadog user using their email


datadog/data_source_datadog_user.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+func dataSourceDatadogUserRead(d *schema.ResourceData, meta interface{}) error {
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV2 := providerConf.DatadogClientV2
+	authV2 := providerConf.AuthV2
+	pageNumber := int64(0)
+	filter := d.Get("filter").(string)
+	founded := false
```
zippolyte, line 43: grammar nit:
```suggestion
	found := false
```

datadog/data_source_datadog_user.go
```diff
@@ -0,0 +1,79 @@
...
+				if err := d.Set("id", user.GetId()); err != nil {
+					return err
+				}
+				break
+			}
+		}
+		pageNumber++
+	}
+	if !founded {
+		return fmt.Errorf("didn't found any user mathing this email")
```
zippolyte, line 76:
```suggestion
		return fmt.Errorf("didn't find any user matching filter string  \"%s\"", )
```

datadog/data_source_datadog_user.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func dataSourceDatadogUser() *schema.Resource {
+	return &schema.Resource{
+		Description: "Use this data source to retrieve information about an existing user to use it in an other resources.",
+		Read:        dataSourceDatadogUserRead,
+
+		Schema: map[string]*schema.Schema{
+			"filter": {
+				Description: "email",
```
zippolyte, line 18: Filter all users by the given string.

datadog/internal/utils/utils.go
```diff
@@ -13,15 +13,14 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/meta"
-
 	"github.com/terraform-providers/terraform-provider-datadog/version"
 )
 
 // DatadogProvider holds a reference to the provider
 var DatadogProvider *schema.Provider
 
 // TranslateClientError turns an error into a message
-func TranslateClientError(err error, apiURL *url.URL, msg string) error {
```
zippolyte, line 24: let's not change that

datadog/data_source_datadog_user.go
```diff
@@ -0,0 +1,79 @@
...
+					return err
+				}
+				if err := d.Set("id", user.GetId()); err != nil {
+					return err
+				}
+				break
+			}
+		}
+		pageNumber++
+	}
```
zippolyte, line 74: Let's not go through all the pages. Instead if there is more than one result (i.e. `total_filtered_count > 1`, return an error stating that there's more than one user matching the filter, and that it should be refined.
Similar to what we do in other datasources, like role http://github.com/datadog/terraform-provider-datadog/blob/master/datadog/data_source_datadog_role.go#L67-L71

zippolyte (changes requested):
You'll also need to rebase on top of master. There have been changes in the go client version, and the API for making requests changed a bit.
The terraform plugin SDK v2 is also used, and the signature of the read function is a little different, it return diagnostics, and takes a context as first argument too

tleveque69:
I haven't been able to record the tests because of the following error eventhough i did what was mention (multiple times).
I used a trial version of DD on which i am admin  

`The keys you've set potentially belong to a production environment. Tests do all sorts of create/update/delete calls to the organisation, so only run them against a sandbox environment. If you know what you are doing, set the DD_TEST_ORG environment variable to the public ID of your organization. See https://docs.datadoghq.com/api/latest/organizations/#list-your-managed-organizations to get it.`

datadog/tests/data_source_datadog_user_test.go
```diff
@@ -0,0 +1,63 @@
...
+			},
+		},
+	})
+}
+
+func testAccDatasourceUserConfig(uniq string) string {
+	return fmt.Sprintf(`
+	data "datadog_user" "test" {
+	  filter = "%s"
+	}`, uniq)
```
zippolyte, line 54: you will need to add the config for the resource here, and add a `depends_on` in the datasource piece when you switch to a unique username

zippolyte (changes requested):
Left some more comments.
If you're having issues with recording, I can run it and push to your branch when you've addressed the comments

tleveque69:
I'm sorry i left so much unused code in this PR thank you so much for you review :pray: .

I tried to correct the test using your advices however can't really test them because of the previous error if u have a workaround or can record them i'll be glad :+1:

zippolyte (approved):

PR #1126, opened 2021-06-24
Fix logs_index update method

skarimo:
This PR ensures we properly update `daily_limit` and `retention_days` fields on Update request. 

Resource update + import has been manually tested. We cannot implement a test feature at the moment as Logs index creation is not supported.

### DIFF
--- a/datadog/resource_datadog_logs_index.go
+++ b/datadog/resource_datadog_logs_index.go
@@ -2,6 +2,7 @@ package datadog
 
 import (
 	"context"
+	"log"
 	"strings"
 
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_logs_index.go
```diff
@@ -111,6 +127,9 @@ func updateLogsIndexState(d *schema.ResourceData, index *datadogV1.LogsIndex) di
 	if err := d.Set("name", index.GetName()); err != nil {
 		return diag.FromErr(err)
 	}
+	if err := d.Set("disable_daily_limit", !index.HasDailyLimit()); err != nil {
```
zippolyte, line 130: Why use `HasDailyLimit()` here, and not `GetDisableDailyLimit()` ?

skarimo: Because the `disable_daily_limit` is a field that can be set only on the update request object and is never returned.  See here: https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/logs_indexes.yaml#L287

zippolyte (approved):

txominpelu (approved):

PR #1134, opened 2021-06-30
Save `has_extended_title` and `enabled` in state

zippolyte:
Was missing from https://github.com/DataDog/terraform-provider-datadog/pull/1130, not sure how the create and update tests were passing correctly if we didn't save the value in the state 😕

### DIFF
--- a/datadog/tests/resource_datadog_security_monitoring_rule_test.go
+++ b/datadog/tests/resource_datadog_security_monitoring_rule_test.go
@@ -105,7 +105,7 @@ func TestAccDatadogSecurityMonitoringRule_Import(t *testing.T) {
 				ResourceName:            tfSecurityRuleName,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"enabled","has_extended_title"},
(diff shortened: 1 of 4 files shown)


therve (approved):

datadog/tests/resource_datadog_security_monitoring_rule_test.go
```diff
@@ -102,11 +102,10 @@ func TestAccDatadogSecurityMonitoringRule_Import(t *testing.T) {
 				Config: testAccCheckDatadogSecurityMonitoringCreatedRequiredConfig(ruleName),
 			},
 			{
-				ResourceName:            tfSecurityRuleName,
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"enabled","has_extended_title"},
```
adussarps, line 108: this should never had been added isn't?

adussarps (approved):

datadog/tests/resource_datadog_security_monitoring_rule_test.go
zippolyte: No it shouldn't have

PR #1140, opened 2021-07-12
[Dashboards] Fix `is_column_break` issues

DavidRobertAnsart:
## Changes

- This PR fix 2 issues:

  1. **`is_column_break` is optional and should not be set for screenboards:**
The fix is to not add the `is_column_break` attribute when the value is `false`. It seems that even if the optional value is not provided, Terraform set its by default to `false`.

  2. **Fix an infinite apply plan with the `is_column_break` attribute:**
When adding `is_column_break` to `true`, the internal TF state dont save it, leading to an infinite plan. This PR save the attribute value in the TF state.

- Add tests for the Multi-size layout `reflow_type` property and the `is_column_break` attribute.

## View the changes

I updated lots of cassettes because the `is_column_break` is not in the dashboard payload by default now.

The first file `datadog/resource_datadog_dashboard.go` and the last file `datadog/tests/resource_datadog_dashboard_test.go` are the only one I updated. To view them easily, collapse all files by holding down the alt key and clicking on the inverted caret icon in any file header. Then open only the first and the last file to review them.


matt-miller-ddog (approved):

skarimo:
Could you add a test case for `is_column_break: true`. Thanks!

DavidRobertAnsart:
> Could you add a test case for `is_column_break: true`. Thanks!

Done !

jirikuncar (approved):

PR #1141, opened 2021-07-12
Add security monitoring filters resource

adussarps:
This PR adds terraform support for security filters.

After this PR we can create/update/delete and imports security filters using terraform.

It also adds the corresponding doc so it should be reviewed by "documentation" too.


docs/resources/security_monitoring_filter.md
```diff
@@ -0,0 +1,61 @@
+---
+page_title: "datadog_security_monitoring_filter Resource - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Provides a Datadog Security Monitoring Filter API resource. This can be used to create and manage Datadog security monitoring filters
```
greentruff: nit: Maybe add an explanation of what filters are or link to the documentation?

docs/resources/security_monitoring_filter.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+  
+}
+```
+
+## Schema
+
+### Required
+
+- **is_enabled** (String, Required) The name of the security filter.
```
greentruff: Wrong name

greentruff (approved):
LGTM beside a couple of things in the docs

docs/resources/security_monitoring_filter.md
adussarps: > LGTM beside a couple of things in the docs


Yes, maybe @DataDog/documentation would have some ideas for the main description of the ressource?

ruthnaebeck: Can someone add the docs team to the repo so they can be added as a reviewer?

datadog/resource_datadog_security_monitoring_filters.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func resourceDatadogSecurityMonitoringFilterDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV2 := providerConf.DatadogClientV2
+	authV2 := providerConf.AuthV2
+
+	filterId := d.Id()
+
+	_, diagnostics, failed := fetchFilterFromApi(datadogClientV2, authV2, filterId)
```
therve, line 164: Same here.

datadog/tests/resource_datadog_security_monitoring_filter_test.go
```diff
@@ -0,0 +1,159 @@
...
+func testAccCheckDatadogSecurityMonitorFilterUpdatedCheck(accProvider func() (*schema.Provider, error), filterName string) resource.TestCheckFunc {
+	return resource.ComposeTestCheckFunc(
+		testAccCheckDatadogSecurityMonitoringFilterExists(accProvider),
+		resource.TestCheckResourceAttr(
+			tfSecurityFilterName, "name", filterName),
+		resource.TestCheckResourceAttr(
+			tfSecurityFilterName, "query", "new query"),
+		resource.TestCheckResourceAttr(
+			tfSecurityFilterName, "is_enabled", "false"),
+		resource.TestCheckResourceAttr(
```
therve, line 109: You should add a check for the length, in `exclusion_filter.#`.

therve (changes requested):
Some minor comments, looks mostly good otherwise.

ruthnaebeck (dismissed):
👍  for docs

datadog/resource_datadog_security_monitoring_filters.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func resourceDatadogSecurityMonitoringFilterUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	providerConf := meta.(*ProviderConfiguration)
+	datadogClientV2 := providerConf.DatadogClientV2
+	authV2 := providerConf.AuthV2
+
+	filterId := d.Id()
+
+	response, diagnostics, failed := fetchFilterFromApi(datadogClientV2, authV2, filterId)
```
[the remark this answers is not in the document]
adussarps: the purpose of that call is to get the latest version of the filter; we mostly need to fetch the value of "version" field; to set to to the latest one; otherwise we can't issue a proper update.

datadog/resource_datadog_security_monitoring_filters.go
adussarps: here, it's indeed useless, I'll remove it.

datadog/resource_datadog_security_monitoring_filters.go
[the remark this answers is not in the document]
therve: It seems to me that there should be a `version` computed attribute then, that you update in read. Then you could pass it directly again.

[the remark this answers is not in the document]
adussarps: Will read be called before an update ?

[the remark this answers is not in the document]
adussarps: We did the same for the rule ressource actually.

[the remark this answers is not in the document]
therve: It's done for the default rule, not the normal one for some reason. I don't think it's correct either. But yes read is called during planning before any operation.

datadog/tests/resource_datadog_security_monitoring_filter_test.go
adussarps: Thanks for all the comment; the code looks much cleaner now. I'll update the rule ressource too with this in mind.

therve:
Tests are failing, you may need to re-record?

therve (approved):

PR #1148, opened 2021-07-21
Expand the example to demonstrate how to use the dashboard list data

jyee:
The documentation states that this data source can be used to register dashboards to an existing dashboard list, but isn't clear on exactly how to do that. No examples exist on the dashboard resource, so this is helpful to show exactly how to use the data source.


docs/data-sources/dashboard_list.md
```diff
@@ -16,6 +16,23 @@ Use this data source to retrieve information about an existing dashboard list, f
...
+  is_read_only = true
+  widget {
+    alert_graph_definition {
+      alert_id  = "1234"
+      viz_type  = "timeseries"
+      title     = "Widget Title"
+      live_span = "1h"
+    }
+  }
+}
```
skarimo, line 35:
```suggestion
```

Thanks for the contribution. The datadog terraform provider docs are auto generated. Could you move the example to https://github.com/DataDog/terraform-provider-datadog/blob/master/examples/data-sources/datadog_dashboard_list/data-source.tf and run `make docs` to auto-generate the docs.

PR #1151, opened 2021-07-22
New Datasource: datadog_synthetics_global_variable

jaredledvina:
This PR creates a new datasource for the provider, `datadog_synthetics_global_variable` 

Synthetics global variables may be created and managed via Terraform in a workspace that is not the same as the one configuring a `datadog_synthetics_test` that uses them. Currently, this would require the user to hardcode the global variable ID via the `datadog_synthetics_test`'s `config_variable` parameter. By created a native datasource provider, users can instead create a new datasource definition like: 
```hcl
data "datadog_synthetics_global_variable" "foo" {
  name = "foo"
}
```

The `data.datadog_synthetics_global_variable` object will have an `id` and `tags` attributes associated with it for further use.

This datasource is also helpful in cases where a global variable may have sensitive values, like API keys, where users may not want to commit those into source control and opt to manually create them via the Web UI instead.


zippolyte (dismissed):
LGTM. Do we want to also export the value when it's not a sensitive var ?

jaredledvina:
> Do we want to also export the value when it's not a sensitive var ?

@zippolyte - I'm not aware of a use-case where folks would want the value of a global variable. As far as I'm aware, these only get used directly as a `config_variable` via `datadog_synthetics_test` which does that for the user. I can work on adding it if that'd be wanted of course.

zippolyte (approved):
No if there's no use case, let's not add it. Looks good

datadog/data_source_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,67 @@
...
+		if globalVariable.Name == searchedName {
+			matchedGlobalVariables = append(matchedGlobalVariables, globalVariable)
+		}
+	}
+
+	if len(matchedGlobalVariables) == 0 {
+		return diag.Errorf("Couldn't find synthetics global variable named %s", searchedName)
+	} else if len(matchedGlobalVariables) > 1 {
+		return diag.Errorf("Found multiple synthetics global variables named %s", searchedName)
+	}
```
sdeprez, line 61: I think it's fine to have that protection here, but just for reference this cannot happen at the moment because you cannot create two global variables with same name.

sdeprez (approved):
LGTM for Synthetics

PR #1176, opened 2021-08-18
Add new_group_delay monitor option

chrismdd:
This PR adds a new monitor option, `new_group_delay` to the monitor data source and resource.

`new_group_delay` is a replacement for `new_host_delay` that works for all multi-alert monitors, not only for those grouped by host. When both options are specified in a monitor definition, `new_group_delay` overrides `new_host_delay`.

`new_host_delay` currently defaults to 300 in this provider. Removing this default would be a breaking change and is therefore not possible without a major version update.

The ideal behavior would be for `new_group_delay` and `new_host_delay` to conflict with each other, and for the provider to not set a default value for `new_host_delay` when `new_group_delay` is specified in the config of a monitor resource. Unfortunately, due to limitations of the terraform SDK, it is not possible (as far as I can tell) to reliably distinguish between `new_group_delay` being set to zero in the config and not being set at all. Specifically, I found that it wasn't possible to unset this field in `updateMonitorState` because `ResourceData` stores fields in multiple levels that are merged when reading the field, but we can only unset the field at one level.

Instead, we opted to keep `new_group_delay` and `new_host_delay` independent. This has the downside that disabling delay for monitors grouped by by host requires setting `new_host_delay` to zero, despite the field now being deprecated. This requirement is clearly documented.

Fixes https://github.com/DataDog/terraform-provider-datadog/issues/1183


docs/resources/monitor.md
```diff
@@ -67,7 +67,12 @@ For example, if the value is set to `300` (5min), the `timeframe` is set to `las
 - **locked** (Boolean) A boolean indicating whether changes to to this monitor should be restricted to the creator or admins. Defaults to `false`.
 - **monitor_threshold_windows** (Block List, Max: 1) A mapping containing `recovery_window` and `trigger_window` values, e.g. `last_15m` . Can only be used for, and are required for, anomaly monitors. (see [below for nested schema](#nestedblock--monitor_threshold_windows))
 - **monitor_thresholds** (Block List, Max: 1) Alert thresholds of the monitor. (see [below for nested schema](#nestedblock--monitor_thresholds))
-- **new_host_delay** (Number) Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non negative integer. Defaults to `300`.
+- **new_group_delay** (Number) Time (in seconds) to skip evaluations for new groups.
+
+new_group_delay overrides new_host_delay if it is set to a nonzero value.
```
alai97, line 72:
```suggestion
`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.
```

docs/resources/monitor.md
```diff
@@ -67,7 +67,12 @@ For example, if the value is set to `300` (5min), the `timeframe` is set to `las
...
+
+To disable group delay for monitors grouped by host, new_host_delay must be set to zero due to the default value of 300 for that field (new_group_delay defaults to zero, so setting it to zero is not required).
```
alai97, line 74:
```suggestion
To disable group delay for monitors grouped by host, `new_host_delay` must be set to zero due to the default value of 300 for that field (`new_group_delay` defaults to zero, so setting it to zero is not required).
```

docs/resources/monitor.md
```diff
@@ -67,7 +67,12 @@ For example, if the value is set to `300` (5min), the `timeframe` is set to `las
...
+- **new_host_delay** (Number, Deprecated) Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non negative integer. Defaults to `300` (this default will be removed in a major version release and new_host_delay will be removed entirely in a subsequent major version release). **Deprecated.** Prefer using new_group_delay (except when setting new_host_delay to zero)
```
alai97, line 75:
```suggestion
- **new_host_delay** (Number, Deprecated) Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to `300` (this default will be removed in a major version release and new_host_delay will be removed entirely in a subsequent major version release). **Deprecated.** Prefer using `new_group_delay` (except when setting `new_host_delay` to zero).
```

alai97 (changes requested):
Minor wording nit!

zippolyte (approved):

zippolyte (approved):

jasperbogers-ig:
This change does not appear to work as a replacement for new_host_delay; we see an error `{"errors": ["The new_group_delay option can only be used for multi-alert monitors"]}`, breaking our build.

PR #1178, opened 2021-08-19
[Synthetics] Add support for restricted roles for global variables

romainberger:
Add support for restricted roles for Synthetics global variables. This PR also updates the go clients to have access to the necessary methods.


PierreGuyot (dismissed):

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -48,7 +48,7 @@ func resourceDatadogSyntheticsGlobalVariable() *schema.Resource {
 				Sensitive:   true,
 			},
 			"secure": {
-				Description: "Sets the variable as secure. Defaults to `false`.",
+				Description: "If set to true, the value of the global variable will be hidden. Defaults to `false`.",
```
zippolyte, line 51: Going to need to run `make docs`
Also I think 
```suggestion
				Description: "If set to true, the value of the global variable is hidden. Defaults to `false`.",
```

datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -99,6 +99,12 @@ func resourceDatadogSyntheticsGlobalVariable() *schema.Resource {
 					},
 				},
 			},
+			"restricted_roles": {
+				Description: "A list of role identifiers to associate with the Synthetics global variable",
```
zippolyte, line 103:
```suggestion
				Description: "A list of role identifiers to associate with the Synthetics global variable.",
```

zippolyte (changes requested):
Can you rebase on master ? There's been some fixes to the integration tests pipeline which was broken.

alai97 (dismissed):
Looks good for docs!

datadog/resource_datadog_synthetics_global_variable.go
alai97: +1 on that, we try to avoid future tense as often as we can!

datadog/resource_datadog_synthetics_global_variable.go
alai97: Danke!

zippolyte (approved):

zippolyte (approved):

PR #1199, opened 2021-09-15
[dashboards] Implement support for APM Dependency Stats query in Query Table

DrkSephy:
This pull request implements support for the new APM Dependency Stats query data source which is a part of formulas and functions in the Query Table Widget.


jtappa (approved):
👍🏻  from docs team

therve (approved):

PR #1205, opened 2021-09-17
Add Webhook resource

skarimo:
Adds support for `datadog_webhook` resource


datadog/resource_datadog_webhook.go
```diff
@@ -0,0 +1,204 @@
...
+				Optional:    true,
+				Computed:    true,
+			},
+			"custom_headers": {
+				Description: "The headers attached to the webhook.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"encode_as": {
+				Description:      "Encoding type. Can be given either `json` or `form`.",
```
zippolyte, line 50:
```suggestion
				Description:      "Encoding type.",
```
Allowed enum values automatically described when generating docs thanks to the validateFunc

examples/resources/datadog_webhook/resource.tf
```diff
@@ -0,0 +1,9 @@
+# Create a new Datadog webhook
```
zippolyte, line 1: Let's also add an import example like other resources

zippolyte (changes requested):

zippolyte (approved):

PR #1215, opened 2021-09-27
Update docs for datadog_synthetics_test: device_ids is required

alexjurkiewicz:
If you don't specify this property, you'll see the following error:

> 400 Bad Request: {"errors": ["properties.options.required: 'device_ids' is a required property...


zippolyte (commented):
That's true only for browser tests, but this resource also manages API tests.
We can't merge this.

alexjurkiewicz:
OK, I updated the docs to show this as required for `type = browser` only. WDYT?

PR #1216, opened 2021-09-27
datadog_synthetics_test: device_ids is required for browser type only

alexjurkiewicz:
update of #1215

If you don't specify this property, you'll see the following error:

> 400 Bad Request: {"errors": ["properties.options.required: 'device_ids' is a required property...

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -52,7 +52,7 @@ func resourceDatadogSyntheticsTest() *schema.Resource {
 				ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewSyntheticsTestDetailsSubTypeFromValue),
 			},
 			"request_definition": {
-				Description: "The synthetics test request. Required if `type = \"api\"`.",
+				Description: "Required if `type = \"api\"`. The synthetics test request.",
 				Type:        schema.TypeList,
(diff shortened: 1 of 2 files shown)


zippolyte (commented):
Unfortunately, that won't work either.
The docs are generated automatically based on the schema defined in https://github.com/DataDog/terraform-provider-datadog/blob/4e7641455dd44cd847575171c23bdf3116b7dd72/datadog/resource_datadog_synthetics_test_.go#L25

So it's not possible to discriminate attributes unfortunately. The best we can do is add in the description itself that it is required.

zippolyte (changes requested):
I suggest adding some wording here to emphasise that it is required for browser tests 
https://github.com/DataDog/terraform-provider-datadog/blob/4e7641455dd44cd847575171c23bdf3116b7dd72/datadog/resource_datadog_synthetics_test_.go#L69

kayayarai (approved):

alexjurkiewicz:
Updated as suggested -- thanks for pointing out how the docs are autogenerated 👍 .

zippolyte (commented):
Looks good, thanks !
Can you just run `make docs` to generate the new changes.
I can do it otherwise and push on your branch if your having issues running it

zippolyte (approved):

cswatt (approved):

Issue #1193, opened 2021-09-02
"inconsistent final plan" when using timestamp() with datadog_downtime


midgetspy:
### Terraform Version

1.0.5

### Affected Resource(s)

datadog_downtime

### Terraform Configuration Files

```hcl
resource "datadog_downtime" "foo" {
  scope        = ["*"]
  monitor_tags = ["my:tag"]

  start_date = timestamp()
  end_date   = timeadd(timestamp(), "2m")
}
```

```hcl
terraform {
  required_providers {
    datadog = {
      source  = "DataDog/datadog"
      version = "~> 3.3.0"
    }
  }

  required_version = "= 1.0.5"
}
```

### Debug Output

https://gist.github.com/midgetspy/97870db44525ecfc563ceb7e1cf4f96d

### Expected Behavior

The downtime should be updated with new start and end time.

### Actual Behavior

Terraform fails to apply:

```
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for datadog_downtime.foo to include new values
│ learned so far during apply, provider
│ "registry.terraform.io/datadog/datadog" produced an invalid new value for
│ .end: was null, but now cty.NumberIntVal(1.630612565e+09).
│
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

### Steps to Reproduce

1. `terraform apply`
2. `terraform apply`

The first step creates it, the second step fails to update it with the above error.

midgetspy:
Could be the same issue as #974

marceloboeira:
@midgetspy doesn't this basically create the alarm with the "right now" time stamp and then if you plan 5 min later it will request to edit again?

midgetspy:
> 
> 
> @midgetspy doesn't this basically create the alarm with the "right now" time stamp and then if you plan 5 min later it will request to edit again?

@marceloboeira It's creating a downtime, not an alarm, but yes that's exactly what we want. Every apply should create a short downtime which starts immediately.

sgametrio:
@zippolyte I encounter the same very issue when setting `start_date` and `end_date`. Is it possible to re-open this?

PR #1223, opened 2021-10-01
Set only one of timestamp or date format for start and end to avoid inconsistent plans

zippolyte:
Fixes https://github.com/DataDog/terraform-provider-datadog/issues/1193


skarimo (approved):
Nice!

midgetspy:
@skarimo @zippolyte Do you know which version will contain this fix and when it will be released? Thanks!

PR #1224, opened 2021-10-01
Set dashboard resource's widget attribute to Optional

skarimo:
The resource submit an empty list by default so this pr sets the widget attribute to optional: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_dashboard.go#L867

This is to avoid having to provide an empty `widget` block when constructing a dashboard without widget. E.g.

```
resource "datadog_dashboard" "tfer--dashboard_uxi-002D-bcq-002D-4p7" {
  is_read_only = "false"
  layout_type  = "ordered"
  reflow_type  = "fixed"
  title        = "Test dashboard"
  widget {}
}
```

### DIFF
--- a/docs/resources/dashboard.md
+++ b/docs/resources/dashboard.md
@@ -686,7 +686,6 @@ resource "datadog_dashboard" "free_dashboard" {
 
 - **layout_type** (String) The layout type of the dashboard. Valid values are `ordered`, `free`.
 - **title** (String) The title of the dashboard.
-- **widget** (Block List, Min: 1) The list of widgets to display on the dashboard. (see [below for nested schema](#nestedblock--widget))
 
 ### Optional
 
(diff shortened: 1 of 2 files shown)


docs/resources/dashboard.md
```diff
@@ -699,12 +698,45 @@ resource "datadog_dashboard" "free_dashboard" {
...
 - **url** (String) The URL of the dashboard.
+- **widget** (Block List) The list of widgets to display on the dashboard. (see [below for nested schema](#nestedblock--widget))
 
 ### Read-Only
 
 - **dashboard_lists_removed** (Set of Number) A list of dashboard lists this dashboard should be removed from. Internal only.
 - **id** (String) The ID of this resource.
 
+<a id="nestedblock--template_variable"></a>
+### Nested Schema for `template_variable`
```
sarina-dd, line 709:
```suggestion
### Nested schema for `template_variable`
```

H2 and below sentence-cased for docs

docs/resources/dashboard.md
```diff
@@ -699,12 +698,45 @@ resource "datadog_dashboard" "free_dashboard" {
 - **template_variable** (Block List) The list of template variables for this dashboard. (see [below for nested schema](#nestedblock--template_variable))
 - **template_variable_preset** (Block List) The list of selectable template variable presets for this dashboard. (see [below for nested schema](#nestedblock--template_variable_preset))
 - **url** (String) The URL of the dashboard.
+- **widget** (Block List) The list of widgets to display on the dashboard. (see [below for nested schema](#nestedblock--widget))
 
 ### Read-Only
```
sarina-dd, line 703:
```suggestion
### Read-only
```

docs/resources/dashboard.md
skarimo: Thanks for the review. These sections of the docs are auto-generated so unfortunately we will not be able to change these without making changes to the underlying library used to generate them.

docs/resources/dashboard.md
skarimo: Same as above

therve (approved):

PR #1244, opened 2021-10-20
Add datasource datadog_logs_indexes_order

jlamande:
The purpose of this new data source is to provide a way to get the current order of logs indexes.
It may be useful in cases where other logs indexes might be created outside of the current Terraform state.

In my case (multi-tenant environment) we are deploying resources and specific configurations per tenant. We need this data source to be able to manage some order in the logs indexes.

Community Note
Please vote on this pull request by adding a 👍 reaction to the original pull request comment to help the community and maintainers prioritize this request
Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request


aghandri:
I faced this same use case, and I ended up doing it manually which is pretty annoying. This datasource is needed

PR #1271, opened 2021-11-11
[docs] [r/logs_index_order] Remove depends_on, use modern refs without string interpolation

tobypinder:
The depends on should not be necessary as this establishes an explicit relationship in the DAG that is already present as an implicit relationship in the `.id` calls for listed resources. This therefore seems like a simpler, less error prone example would not use `depends_on` which is generally discouraged unless needed for cases where there are no implicit relationships.

Also, updated the syntax in the example for TF 0.12+ as string interpolation is no longer required

----

Not sure what this means for the codegen part of how these documents are produced, but I did want to surface these changes.

### DIFF
--- a/docs/resources/logs_index_order.md
+++ b/docs/resources/logs_index_order.md
@@ -15,11 +15,8 @@ Provides a Datadog Logs Index API resource. This can be used to manage the order
 ```terraform
 resource "datadog_logs_index_order" "sample_index_order" {
   name = "sample_index_order"
-  depends_on = [
-    "datadog_logs_index.sample_index"
-  ]
   indexes = [
-    "${datadog_logs_index.sample_index.id}"
+    datadog_logs_index.sample_index.id
   ]
 }
 ```


ruthnaebeck (approved):
👍  for docs

zippolyte:
The source for this example is in https://github.com/DataDog/terraform-provider-datadog/blob/master/examples/resources/datadog_logs_index_order/resource.tf, could you make that change there as well ?
Thanks

tobypinder:
> The source for this example is in https://github.com/DataDog/terraform-provider-datadog/blob/master/examples/resources/datadog_logs_index_order/resource.tf, could you make that change there as well ? Thanks

Thanks for taking a look and pointing out how the docs are generated. It looks like this pattern exists in a couple of places, I'll try and find time to put a more comprehensive PR together that's a bit wider in scope and addresses the same issue(s) across multiple resource types.

zippolyte:
Awesome 👍 
Yeah once you make the changes in the example files, you should be able to regenerate everything with `make docs`

skarimo:
Closing this in favor of: https://github.com/DataDog/terraform-provider-datadog/pull/1348

PR #1276, opened 2021-11-15
Bump go client and make necessary fixes

zippolyte:
New required attribute broke constructor in synthetics assertions, and add nullable to two fields in slo correction


therve (dismissed):

apigirl (dismissed):
docs look good

therve (approved):

PR #1278, opened 2021-11-17
fix: panic on resource name change

lhelman:
On resource name change the old is still there, and the new is empty, so there
the `new.(string)` is empty, so `newAttrMap` is `interface{}`
And terraform panics on that line.
This defensive programming action fixes the panic, but I think (guess?) that the code should never
get to this point on this case.

### DIFF
--- a/datadog/resource_datadog_monitor_json.go
+++ b/datadog/resource_datadog_monitor_json.go
@@ -55,7 +55,13 @@ func resourceDatadogMonitorJSON() *schema.Resource {
 			if !ok {
 				return true
 			}
-			return oldType != newAttrMap["type"].(string)
+			
+			newType, ok := newAttrMap["type"].(string)
+			if !ok {
+                                return true
+                        }
+			
+			return oldType != newType
 		}),
 		Schema: map[string]*schema.Schema{
 			"monitor": {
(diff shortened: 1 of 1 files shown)


therve:
Thanks for your patch. Can you describe exactly how to reproduce the problem? I tried just renaming the resource in the terraform config file and it worked without a panic.

lhelman:
sadly I removed the logs after applying with my patch.
But adding some more defensive programming should never
be wrong, right?
The code got there, and it panicked. With the fix, I could plan/apply
the whole deal.
I had some resource name changes, and also some issue with dependencies (as in SLO monitors dependencies between the monitor resource and the service_level_objective resource).
Not sure what triggered it.
I'm sure a taint of the proper resources will trigger it again.
But I need to advance with other stuff.

therve (approved):

therve:
> But adding some more defensive programming should never
> be wrong, right?

Right, the point is to be sure we solved the issue you faced. Can you quickly run `make fmt` so that we can merge this? Thanks.

lhelman:
Reformatted with `make fmt`

therve (approved):

lhelman:
Late, but I've hit my issue again running v3.6.0, and I can confirm that this PR fixed it.
I can attach logs or what not if needed.

PR #1281, opened 2021-11-19
[examples] Fix invalid monitor timeout_h example

phillip-dd:
`timeout_h` values greater than 24 hours are not supported. This example has unfortunately led to many invalid monitors.

### DIFF
--- a/examples/resources/datadog_monitor/resource.tf
+++ b/examples/resources/datadog_monitor/resource.tf
@@ -18,7 +18,6 @@ resource "datadog_monitor" "foo" {
   renotify_interval = 60
 
   notify_audit = false
-  timeout_h    = 60
   include_tags = true
 
   tags = ["foo:bar", "baz"]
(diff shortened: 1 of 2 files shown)


examples/resources/datadog_monitor/resource.tf
```diff
@@ -18,7 +18,7 @@ resource "datadog_monitor" "foo" {
   renotify_interval = 60
 
   notify_audit = false
-  timeout_h    = 60
+  timeout_h    = 18
```
mathispesch, line 21: Do we actually want to have this in the example? There's a bunch of fields that aren't populated here already and I don't know if the autoresolve timeout is a behaviour that people want to have by default.

mathispesch (approved):

examples/resources/datadog_monitor/resource.tf
phillip-dd: fair enough, i've removed it entirely

therve (approved):

PR #1293, opened 2021-12-02
`new_group_delay` and `new_host_delay` docs

### LINKED ISSUE #1292: Clarify `new_group_delay` and `new_host_delay` in docs
### Terraform Version

terraform 1.0.11

### Affected Resource(s)

-   datadog_monitor

```hcl
resource "datadog_monitor" "this" {
  # new_host_delay = 300
  new_group_delay = 0
}
```

### Expected Behavior

Knowing how to deprecate `new_host_delay` and when and how to use `new_group_delay`

### Actual Behavior

This is pretty confusing.

In the terraform provider docs it says the following

> * `new_group_delay` (Number) Time (in seconds) to skip evaluations for new groups.

> `new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.

So `new_group_delay: 300` would override `new_host_delay` and `new_group_delay: 0` would do nothing. Makes sense.

> To disable group delay for monitors grouped by host, `new_host_delay` must be set to zero due to the default value of 300 for that field (new_group_delay defaults to zero, so setting it to zero is not required).

So `new_host_delay` should be set and should be set to `0` to override the default of `300`? Does that mean `new_host_delay` should still be used even tho it's deprecated ? Or is this line an old portion of the documentation that should be removed ?

> * `new_host_delay` (Number, Deprecated) Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to 300 (this default will be removed in a major version release and new_host_delay will be removed entirely in a subsequent major version release). Deprecated. Prefer using new_group_delay (except when setting new_host_delay to zero).

soooo this deprecated part.

> Deprecated. Prefer using new_group_delay (except when setting new_host_delay to zero).

This means that if we set `new_host_delay` to zero, we don't prefer using `new_group_delay` ???

Then in this comment https://github.com/DataDog/terraform-provider-datadog/issues/1221#issuecomment-931006926 by @zippolyte seemed.

> Hi,
> 
> The `new_host_delay` option could be set on simple monitors, but it never had any effect in that case. It would only have an effect for a multi alert monitor scoped on hosts. You should just remove that `new_group_delay` option in your configuration, as it's not useful to have here.

Is the correct solution ?

* remove `new_host_delay` and use `new_group_delay` for multi alert monitors
* remove `new_host_delay` and remove `new_group_delay` for simple monitors

If so, can we update the docs to clarify this?

### References

-   https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/monitor
-   https://github.com/DataDog/terraform-provider-datadog/issues/1221

sarina-dd:
Updates the `new_group_delay` and `new_host_delay` docs to resolve #1292

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -180,7 +180,7 @@ func resourceDatadogMonitor() *schema.Resource {
 			"new_host_delay": {
 				// Removing the default requires removing the default in the API as well (possibly only for
 				// terraform user agents)
-				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and `new_host_delay` to zero for monitors grouped by host.",
+				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and set `new_host_delay` to zero for monitors grouped by host.",
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     300,
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -173,18 +173,18 @@ func resourceDatadogMonitor() *schema.Resource {
 			// refers to this terraform resource. In the API, setting new_group_delay
 			// to any value, including zero, causes it to override new_host_delay.
 			"new_group_delay": {
-				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.\n\nTo disable group delay for monitors grouped by host, `new_host_delay` must be set to zero due to the default value of `300` for that field (`new_group_delay` defaults to zero, so setting it to zero is not required).",
+				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.",
 				Type:        schema.TypeInt,
 				Optional:    true,
 			},
 			"new_host_delay": {
 				// Removing the default requires removing the default in the API as well (possibly only for
 				// terraform user agents)
-				Description: "Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to `300` (this default will be removed in a major version release and `new_host_delay` will be removed entirely in a subsequent major version release).",
+				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and `new_host_delay` to zero for monitors grouped by host.",
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     300,
-				Deprecated:  "Prefer using new_group_delay (except when setting `new_host_delay` to zero).",
+				Deprecated:  "Use `new_group_delay` (except when setting `new_host_delay` to zero).",
 			},
 			"evaluation_delay": {
 				Description: "(Only applies to metric alert) Time (in seconds) to delay evaluation, as a non-negative integer.\n\nFor example, if the value is set to `300` (5min), the `timeframe` is set to `last_5m` and the time is 7:00, the monitor will evaluate data from 6:50 to 6:55. This is useful for AWS CloudWatch and other backfilled metrics to ensure the monitor will always have data during evaluation.",
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -173,7 +173,7 @@ func resourceDatadogMonitor() *schema.Resource {
 			// refers to this terraform resource. In the API, setting new_group_delay
 			// to any value, including zero, causes it to override new_host_delay.
 			"new_group_delay": {
-				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.",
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_monitor.go
```diff
@@ -173,18 +173,18 @@ func resourceDatadogMonitor() *schema.Resource {
...
+				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.",
 				Type:        schema.TypeInt,
 				Optional:    true,
 			},
 			"new_host_delay": {
 				// Removing the default requires removing the default in the API as well (possibly only for
 				// terraform user agents)
-				Description: "Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to `300` (this default will be removed in a major version release and `new_host_delay` will be removed entirely in a subsequent major version release).",
+				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and `new_host_delay` to zero for monitors grouped by host.",
```
phillip-dd, line 183: I think we're missing the word "set" here:

> The only case when this should be used is to override the default and `new_host_delay` to zero

-> 

> The only case when this should be used is to override the default and set `new_host_delay` to zero

datadog/resource_datadog_monitor.go
```diff
@@ -173,18 +173,18 @@ func resourceDatadogMonitor() *schema.Resource {
 			// refers to this terraform resource. In the API, setting new_group_delay
 			// to any value, including zero, causes it to override new_host_delay.
 			"new_group_delay": {
-				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.\n\nTo disable group delay for monitors grouped by host, `new_host_delay` must be set to zero due to the default value of `300` for that field (`new_group_delay` defaults to zero, so setting it to zero is not required).",
+				Description: "Time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.",
```
ruthnaebeck, line 176:
```suggestion
				Description: "The time (in seconds) to skip evaluations for new groups.\n\n`new_group_delay` overrides `new_host_delay` if it is set to a nonzero value.",
```

datadog/resource_datadog_monitor.go
ruthnaebeck:
```suggestion
				Description: "**Deprecated**. See `new_group_delay`. The time (in seconds) to allow a host to boot and applications to start before evaluating the monitor results. It should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and set `new_host_delay` to zero for monitors grouped by host.",
```

datadog/resource_datadog_monitor.go
```diff
@@ -173,18 +173,18 @@ func resourceDatadogMonitor() *schema.Resource {
...
 				// Removing the default requires removing the default in the API as well (possibly only for
 				// terraform user agents)
-				Description: "Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to `300` (this default will be removed in a major version release and `new_host_delay` will be removed entirely in a subsequent major version release).",
+				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and `new_host_delay` to zero for monitors grouped by host.",
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     300,
-				Deprecated:  "Prefer using new_group_delay (except when setting `new_host_delay` to zero).",
+				Deprecated:  "Use `new_group_delay` (except when setting `new_host_delay` to zero).",
```
ruthnaebeck, line 187:
```suggestion
				Deprecated:  "Use `new_group_delay` except when setting `new_host_delay` to zero.",
```

datadog/resource_datadog_monitor.go
sarina-dd: Good catch, will update. Thanks!

docs/resources/monitor.md
```diff
@@ -69,9 +69,7 @@ For example, if the value is set to `300` (5min), the `timeframe` is set to `las
...
-- **new_host_delay** (Number, Deprecated) Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. Defaults to `300` (this default will be removed in a major version release and `new_host_delay` will be removed entirely in a subsequent major version release). **Deprecated.** Prefer using new_group_delay (except when setting `new_host_delay` to zero).
+- **new_host_delay** (Number, Deprecated) **Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and `new_host_delay` to zero for monitors grouped by host. **Deprecated.** Use `new_group_delay` (except when setting `new_host_delay` to zero).
```
ruthnaebeck, line 72:
```suggestion
- **new_host_delay** (Number, Deprecated) **Deprecated**. See `new_group_delay`. The time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and set `new_host_delay` to zero for monitors grouped by host. **Deprecated.** Use `new_group_delay` except when setting `new_host_delay` to zero.
```

ruthnaebeck (changes requested):

datadog/resource_datadog_monitor.go
sarina-dd, line 183:
```suggestion
				Description: "**Deprecated**. See `new_group_delay`. Time (in seconds) to allow a host to boot and applications to fully start before starting the evaluation of monitor results. Should be a non-negative integer. This value is ignored for simple monitors and monitors not grouped by host. Defaults to `300`. The only case when this should be used is to override the default and set `new_host_delay` to zero for monitors grouped by host.",
```

ruthnaebeck (dismissed):
👍 for docs

bateller (approved):

therve (approved):

Issue #1290, opened 2021-11-30
Handle perpetual diff on notify list order changes


lhelman:
### Terraform Version

terraform -v
Terraform v1.0.10
on darwin_amd64
+ provider registry.terraform.io/datadog/datadog v3.6.0
+ provider registry.terraform.io/hashicorp/aws v3.32.0
+ provider registry.terraform.io/pagerduty/pagerduty v1.10.0

### Affected Resource(s)

-  datadog_dashboard_json

### Terraform Configuration Files

```hcl
resource "datadog_dashboard_json" "something" {
  dashboard = <<EOF
  ...
  notify_list: [
    "one.mail@example.com",
    "yet.another@example.com",
    "another.mail@example.com"
  ],
   ...
EOF
```

Once you have that, sometimes it shows the perpetual diff, even after 
successful applies without doing anything, or if it doesn't show the diff,
just change the orderm and it will produce the perpetual diff.

### Expected Behavior

No changes on plan after a successful apply is made.
That, even after changing the order of the list.

### Actual Behavior

A diff is shown after successful apply.

Even more:
I am running that snippets in two different datadog orgs, and
one produces some changes every time and the other doesn't 
produce any change.
If I reorganize the list, I get the inverse behavior.
I tried tainting the resource so terraforms destroys/creates, but
that doesn't produce any change on the results.

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:

1. `terraform apply`
2. `terraform plan`
3. If that doesn't show any changes, reorder the list and
4. `terraform apply`
5. `terraform plan`

therve:
Hi,

I can reproduce the order issue. Note that the handles in the list need to be in the org you're targeting, so if you put unknown users you will get a perpetual diff unrelated to order.

PR #1295, opened 2021-12-02
Handle `notify_list` diffs for dashboard resource

skarimo:
Closes [1290](https://github.com/DataDog/terraform-provider-datadog/issues/1290)


therve (approved):
Nice! Can you handle the issue in the dashboard resource as well? I think there just changing to TypeSet worked for me. You can do it in a follow up PR.

PR #1318, opened 2022-01-04
Bump datadog-api-client-go and deprecate `agent_rule` field

jirikuncar:


andpiccione (commented):
Thanks for removing this 🙇 LGTM, but shouldn't we also modify tests&cassettes and update documentation accordingly?

andpiccione (approved):
👍 Can you also re-run `make docs`? I think markdown files should be updated following the field deprecation

therve (approved):

PR #1330, opened 2022-01-19
Enforce labels on PRs

therve:
We need a changelog and a resource label to generate proper changelog
entries.

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -1,0 +1,27 @@
+name: "Ensure labels"
+on: # yamllint disable-line rule:truthy
+    pull_request:
+        types:
+            - labeled
+            - unlabeled
+            - opened
+            - synchronize
+            - reopened
+            - ready_for_review
+
+jobs:
+    changelog:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Check changelog label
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog.yaml
```diff
@@ -0,0 +1,27 @@
...
+    changelog:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Check changelog label
+              if: github.event.pull_request.draft == false && false == contains(join(github.event.pull_request.labels.*.name, ','), 'changelog/')
+              run: |-
+                  echo "::error::Add 'changelog/*' label";
+                  exit 1;
+            - name: Check resource label
+              if: github.event.pull_request.draft == false && false == contains(join(github.event.pull_request.labels.*.name, ','), 'changelog/no-changelog') && false == contains(join(github.event.pull_request.labels.*.name, ','), 'resource/')
```
jirikuncar, line 22: we can skip this check also for `changelog/note`

jirikuncar: alternatively we could use PR title prefix `[{resource_name}] ...`

therve: Yeah... It's a bit harder to verify it's in a list fo known names.

PR #1338, opened 2022-01-27
Add terraform support for Cloud Workload Security Agent Rules

andpiccione:
This PR adds support for creating/updating/deleting/fetching Cloud Workload Security Agent rules. See related [API reference](https://docs.datadoghq.com/api/latest/cloud-workload-security/) for documentation.


andpiccione:
FYI, I created a new label `resource/datadog_cloud_workload_security_agent_rul` (missing `e` at the end because character limit for labels is 50), I hope it's ok.

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+			Description: "Whether the Agent rule is enabled.",
+		},
+
+		"expression": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "The SECL expression of the Agent rule.",
+		},
+
+		"name": {
```
skarimo, line 53: It seems you cannot update the `name` attribute. Can you set `ForceNew: true` in this case so that the resource is recreated if this is changed. Otherwise, i believe there would be a perpetual diff if a user changes the name.

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+
+	agentRuleId := d.Id()
+
+	agentRuleUpdate := buildCwsAgentRuleUpdatePayload(d)
+
+	if _, httpResponse, err := datadogClientV2.CloudWorkloadSecurityApi.UpdateCloudWorkloadSecurityAgentRule(authV2, agentRuleId, *agentRuleUpdate); err != nil {
+		return utils.TranslateClientErrorDiag(err, httpResponse, "error updating cloud workload security agent rule")
+	}
+
+	return nil
```
skarimo, line 118:
```suggestion
	if agentRuleResponse, httpResponse, err := datadogClientV2.CloudWorkloadSecurityApi.UpdateCloudWorkloadSecurityAgentRule(authV2, agentRuleId, *agentRuleUpdate); err != nil {
		return utils.TranslateClientErrorDiag(err, httpResponse, "error updating cloud workload security agent rule")
	}

	return updateResourceDataAgentRuleFromResponse(d, agentRuleResponse)
```

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+	if httpResponse, err := datadogClientV2.CloudWorkloadSecurityApi.DeleteCloudWorkloadSecurityAgentRule(authV2, agentRuleId); err != nil {
+		return utils.TranslateClientErrorDiag(err, httpResponse, "error deleting cloud workload security agent rule")
+	}
+
+	return nil
+}
+
+func updateResourceDataAgentRuleFromResponse(d *schema.ResourceData, agentRuleResponse datadogV2.CloudWorkloadSecurityAgentRuleResponse) {
+	data := agentRuleResponse.GetData()
+	d.SetId(data.GetId())
```
skarimo, line 137:
```suggestion
```
Typically terraform resources set the ID in the create method.

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+	agentRuleId := d.Id()
+
+	if httpResponse, err := datadogClientV2.CloudWorkloadSecurityApi.DeleteCloudWorkloadSecurityAgentRule(authV2, agentRuleId); err != nil {
+		return utils.TranslateClientErrorDiag(err, httpResponse, "error deleting cloud workload security agent rule")
+	}
+
+	return nil
+}
+
+func updateResourceDataAgentRuleFromResponse(d *schema.ResourceData, agentRuleResponse datadogV2.CloudWorkloadSecurityAgentRuleResponse) {
```
skarimo, line 135:
```suggestion
func updateCloudWorkloadSecurityAgentRuleState(d *schema.ResourceData, agentRuleResponse datadogV2.CloudWorkloadSecurityAgentRuleResponse) {
```

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return &payload
+}
+
+func buildCwsAgentRuleCreatePayload(d *schema.ResourceData) *datadogV2.CloudWorkloadSecurityAgentRuleCreateRequest {
+	payload := datadogV2.CloudWorkloadSecurityAgentRuleCreateRequest{}
+	payload.Data.Type = agentRuleType
+
+	payload.Data.Attributes.SetDescription(d.Get("description").(string))
+	payload.Data.Attributes.SetEnabled(d.Get("enabled").(bool))
```
skarimo, line 163: If these fields are optional, we should check to see if they are set first:
```
	if attr, ok := d.GetOk("description"); ok {
		payload.Data.Attributes.SetDescription(attr.(string))
	}
```

datadog/tests/data_source_datadog_cloud_workload_security_agent_rules_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/terraform-providers/terraform-provider-datadog/datadog"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+const tfAgentRulesSource = "data.datadog_cloud_workload_security_agent_rules.acceptance_test"
+
+func TestAccDatadogCloudWorkloadSecurityAgentRulesDatasource(t *testing.T) {
```
skarimo, line 18: Could you update this test so that a new CloudWorkloadAgentRule is first created and then read using the datasource. Here is an example: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/tests/data_source_datadog_roles_test.go#L31-L54

This will ensure that data source is working properly.

datadog/tests/resource_datadog_cloud_workload_security_agent_rule_test.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+}
+
+func testAccCheckDatadogCloudWorkloadSecurityAgentRuleCreated(name string) string {
+	return fmt.Sprintf(`
+resource "datadog_cloud_workload_security_agent_rule" "acceptance_test" {
+    name = "%s"
+    description = "an agent rule"
+    enabled = true
+	expression = "exec.file.name == \"java\""
```
skarimo, line 47:
```suggestion
    expression = "exec.file.name == \"java\""
```

datadog/data_source_datadog_cloud_workload_security_agent_rules.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func dataSourceDatadogCloudWorkloadSecurityAgentRules() *schema.Resource {
+	return &schema.Resource{
+		Description: "Use this data source to retrieve information about existing cloud workload security agent rules for use in other resources.",
+		ReadContext: dataSourceDatadogCloudWorkloadSecurityAgentRulesRead,
+
+		Schema: map[string]*schema.Schema{
+			// Computed
+			"agent_rules_ids": {
```
skarimo, line 20: Whats the use case for having this in a separate list? I think the id's should be stored within each of the `agent_rules`.

datadog/data_source_datadog_cloud_workload_security_agent_rules.go
```diff
@@ -0,0 +1,76 @@
...
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"agent_rules": {
+				Description: "List of agent rules.",
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: cloudWorkloadSecurityAgentRuleSchema(),
```
skarimo, line 31: Instead of reusing the schema, I would redefine it in the data source and add the id's within it. Additionally, lets set all of the attributes to computed as well. For example: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/data_source_datadog_monitors.go#L44

datadog/data_source_datadog_cloud_workload_security_agent_rules.go
```diff
@@ -0,0 +1,76 @@
...
+
+		agentRuleTF["name"] = attributes.GetName()
+		agentRuleTF["description"] = attributes.GetDescription()
+		agentRuleTF["expression"] = attributes.GetExpression()
+		agentRuleTF["enabled"] = attributes.GetEnabled()
+
+		agentRules = append(agentRules, agentRuleTF)
+	}
+
+	d.SetId(strings.Join(agentRulesIds, "--"))
```
skarimo, line 71: We can probably set the id to something generic like `cloud-workload-security-agent-rules`

kayayarai (dismissed):
👍  from docs. In the generated .md files, there's a lot of inconsistency about whether Agent is capitalized. If this is referring to the Datadog Agent, it should be capitalized, throughout. Also, Cloud Workload Security should be capitalized (line 11, docs/data-sources/cloud_workload_security_agent_rules.md)

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
+package datadog
+
+import (
+	"context"
+
+	_ "gopkg.in/warnings.v0"
```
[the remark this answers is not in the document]
andpiccione: Yep, it's not needed, it was a leftover from the file I created this.

datadog/resource_datadog_cloud_workload_security_agent_rule.go
```diff
@@ -0,0 +1,168 @@
...
+			Required:    true,
+			Description: "The description of the Agent rule.",
+		},
+
+		"enabled": {
+			Type:        schema.TypeBool,
+			Required:    true,
+			Description: "Whether the Agent rule is enabled.",
+		},
+
```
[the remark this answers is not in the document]
andpiccione: You're right, just made both of them optional and set the default value when they are not set upon creation.

datadog/resource_datadog_cloud_workload_security_agent_rule.go
andpiccione: Done

datadog/resource_datadog_cloud_workload_security_agent_rule.go
andpiccione: This is needed when we update the resource from the response

datadog/tests/data_source_datadog_cloud_workload_security_agent_rules_test.go
andpiccione: I don't think it's needed, we already have a set of default rules that are created by default when the product is activated and I verified these are retrieved when they are read using the datasource, so the test is already not reading an empty list.

datadog/data_source_datadog_cloud_workload_security_agent_rules.go
andpiccione: Yep you're right, just put it inside each agent_rule

datadog/data_source_datadog_cloud_workload_security_agent_rules.go
andpiccione: Done

datadog/resource_datadog_cloud_workload_security_agent_rule.go
andpiccione: Just tried, and I get an unknown provider error during the tests if this line is removed.

kayayarai (approved):

datadog/resource_datadog_cloud_workload_security_agent_rule.go
therve: Right because you use the same method for update and create.

therve (approved):

PR #1343, opened 2022-02-03
AWS Integration: Add support for metrics and resource collections

nikohaa:
Add support metrics and resource collections for `datadog_integration_aws` resource

Added fields:
* `resource_collection_enabled`
* `metrics_collection_enabled`
* `cspm_resource_collection_enabled`


maycmlee (commented):
Small fix

jvanbrie (commented):
Thanks for adding support for these! Could we add cspm_collection_enabled at the same time?

datadog/resource_datadog_integration_aws.go
```diff
@@ -83,6 +83,18 @@ func resourceDatadogIntegrationAws() *schema.Resource {
 				ConflictsWith: []string{"account_id", "role_name"},
 				Optional:      true,
 			},
+			"metrics_collection_enabled": {
+				Description: "Whether Datadog collects metrics for this AWS account.",
+				Type:        schema.TypeBool,
+				Default:     false,
```
skarimo, line 89:
```suggestion
				Default:     true,
```
Let's make this option default to `true` in order to keep it in line with the api and the datadog go client behavior.

datadog/resource_datadog_integration_aws.go
```diff
@@ -83,6 +83,18 @@ func resourceDatadogIntegrationAws() *schema.Resource {
...
+				Optional:    true,
+			},
+			"resource_collection_enabled": {
+				Type:        schema.TypeBool,
+				Description: "Whether Datadog collects a standard set of resources from your AWS account.",
+				Default:     false,
+				Optional:    true,
+			},
```
skarimo, line 97:
```suggestion
			},
			"cspm_resource_collection_enabled": {
				Type:        schema.TypeBool,
				Description: "Whether Datadog collects cloud security posture management resources from your AWS account. This includes additional resources not covered under the general resource_collection.",
				Default:     false,
				Optional:    true,
			},
```

Could we also include the [`cspm_resource_collection_enabled`](https://docs.datadoghq.com/api/latest/aws-integration/#create-an-aws-integration) option in this PR?

skarimo (commented):
Thanks for the pr! Left 2 minor comments

datadog/resource_datadog_integration_aws.go
nikohaa: Sure I'll add that field too 😊

skarimo:
Thanks for the pr made a few small changes and pushed up cassettes. Primary change was not checking if the attributes is set or not. The GetOk method does not work well with zero values.

maycmlee (dismissed):
👍  for docs

therve (approved):

PR #1350, opened 2022-02-07
[ACCESS-597] Support ReadContext for Datadog SAML Authentication Mapping.

alexaliaskovski:
Supporting ReadContext for Datadog SAML Authentication Mapping.

Not adding tests because they should be covered by #1344.


swetha-murali-dd (approved):
👍 💯 🥇  LGTM!

PR #1358, opened 2022-02-11
ACCESS 523: Add terraform support for SAML AuthN Mappings

swetha-murali-dd:


datadog/tests/cassettes/TestAccDatadogAuthNMapping_CreateUpdate.yaml
```diff
@@ -0,0 +1,951 @@
+---
```
alexaliaskovski, line 1: curiosity: what is this and the `.freeze` file for ?

examples/resources/datadog_authn_mapping/resource.tf
```diff
@@ -0,0 +1,11 @@
+# Source the role
+data "datadog_role" "ro_role" {
+  filter = "Datadog Read Only Role"
+}
+
+# Create a new AuthN mapping
+resource "datadog_authn_mapping" "foo" {
```
srosenthal-dd, line 7: Since this is an example for the public docs, let's try to use realistic "good" naming, replacing "foo" with "dev_ro_role_mapping" or something more meaningful like that?

srosenthal-dd (commented):
One missing feature here is enabling/disabling the SAML mappings as a whole for an org, but maybe that's already covered in some other Terraform resource? I don't mind calling that out of scope for this PR/task.

datadog/resource_datadog_authn_mapping.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/DataDog/datadog-api-client-go/api/v2/datadog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
+)
+
+func resourceDatadogAuthnMapping() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a Datadog AuthN Mappings resource. This can be used to create and manage Datadog SAML Mappings.",
```
srosenthal-dd, line 17: Nit: first it's called "Datadog AuthN Mappings", then "Datadog SAML Mappings". 

Could we add a sentence explaining the "who cares?" a bit more? Something like: This feature lets you automatically assign roles to users based on their SAML attributes.

datadog/resource_datadog_authn_mapping.go
```diff
@@ -0,0 +1,203 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"value": {
+				Description: "Identity provider value.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"role": {
+				Description: "The role to assign for key:value mapping.",
```
srosenthal-dd, line 35: Would be good to clarify this is the ID of a role, not the full role resource.

"The ID of a role to attach to all users with the corresponding key and value"?

datadog/resource_datadog_authn_mapping.go
```diff
@@ -0,0 +1,203 @@
...
+		return utils.TranslateClientErrorDiag(err, httpResponse, "error creating authn mapping")
+	}
+	if err := utils.CheckForUnparsed(createResp); err != nil {
+		return diag.FromErr(err)
+	}
+
+	var getAuthNMappingResponse datadog.AuthNMappingResponse
+	var httpResponseGet *http.Response
+
+	err = resource.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
```
srosenthal-dd, line 58: Curious why Create gets wrapped with a retry but other operations don't. Is that some convention you copied from elsewhere in this repo, or did you run into failures/timeout and need to add retries?

datadog/resource_datadog_authn_mapping.go
```diff
@@ -0,0 +1,203 @@
...
+			"role": {
+				Description: "The role to assign for key:value mapping.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+		},
+	}
+}
+
+func resourceDatadogAuthnMappingCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
srosenthal-dd, line 43: I think we're doing Terraform "wrong" by ignoring the ctx variable. The context can contain a cancellation handle and timeout, but I do see that most or all of this repo also ignores ctx because we're already using a separate context to encode the authentication credentials.

So, it looks like a much bigger problem than this PR, so I'm fine with keeping it as-is.

datadog/resource_datadog_authn_mapping.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+func resourceDatadogAuthnMapping() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Provides a Datadog AuthN Mappings resource. This can be used to create and manage Datadog SAML Mappings.",
+		CreateContext: resourceDatadogAuthnMappingCreate,
+		ReadContext:   resourceDatadogAuthnMappingRead,
+		UpdateContext: resourceDatadogAuthnMappingUpdate,
+		DeleteContext: resourceDatadogAuthnMappingDelete,
+
```
srosenthal-dd, line 22: Nice feature to consider adding is support for `terraform import`. Then you can take AuthN mapping defined elsewhere, for example in Datadog's UI, and import them into Terraform like this:

```
terraform import datadog_authn_mapping.dev_ro_mapping "5455f42a-8b9f-11ec-813e-da7ad0900002"
```

As-is, it'll fail with an error `Error: resource datadog_authn_mapping doesn't support import`.

Some other resources but not all implement this feature -- grep for `Importer:`

OK with me if we postpone this feature til later

datadog/tests/resource_datadog_authn_mapping_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+
+	"github.com/terraform-providers/terraform-provider-datadog/datadog"
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
+)
+
+// TODO: Modify CheckDestroy function to testAccCheckDatadogAuthNMappingDestroy after Delete Context is available
```
srosenthal-dd, line 16: Still TODO?

srosenthal-dd (approved):
I reviewed the code somewhat thoroughly, and used the provider thoroughly. Everything worked perfectly except for `terraform import` which requires additional code.

Added a few minor comment inline. Nothing major.

Nice work!

swetha-murali-dd:
> I've tested this pretty thoroughly locally using Terraform that looks like this:
> 
> ```
> My only concern from the user perspective is whether this allows evolution to other mappings that we might add in the future. Would it be awkward to support optional parameters within the same resource like this?
> 
> ```
> resource "datadog_authn_mapping" "role_mapping" {
>   key   = "member-of"
>   value = "admin"
>   role  = data.datadog_role.admin.id
> }
> 
> resource "datadog_authn_mapping" "team_mapping" {
>   key   = "member-of"
>   value = "developers"
>   team  = data.datadog_team.developers.id // <-- role is unset here, instead team provided
> }
> ```
> 
> Next I'll dig into reviewing the code...

I think this can be evolved for other mappings as well. For future expansion to support team mapping for example, I was planning on using `ConflictsWith` schema option (between role & team keys). So the same code can be modified to handle both role & team and `ConflictsWith` check will ensure either role or team is provided and not both while creating a resource.

datadog/tests/cassettes/TestAccDatadogAuthNMapping_CreateUpdate.yaml
swetha-murali-dd: These 2 files were generated by recording server responses while running acceptance test cases `RECORD=true make testall`. 
Going forward, for acceptance tests the idea is to use these server responses as mocks instead of actually CRUDing a resource from the server. 
The .yaml is the server response and the .freeze file is the datetime of when the response was captured.

datadog/resource_datadog_authn_mapping.go
swetha-murali-dd: Will look into this. Will include it in this PR if its not a huge task

datadog/resource_datadog_authn_mapping.go
swetha-murali-dd: Thanks for pointing this out. I didn't know about this. So does the correct way of using context involve binding a retry, cancel and timeout handler to be invoked?

datadog/tests/resource_datadog_authn_mapping_test.go
swetha-murali-dd: Nope. Good catch. Will remove that.

datadog/resource_datadog_authn_mapping.go
srosenthal-dd: Yes and by putting the API & App key into an existing context, we force ourselves to ignore the context provided by Terraform, so Terraform can't cancel or timeout the request to the Datadog provider.

Instead of providing a common, shared context with credentials baked in, I think it would be better to have a helper method like this:
```
func addCredentialsToContext(ctx context.Context) context.Context {
     // takes an existing context (from Terraform)
     // and adds properties to it like Datadog credentials
     // so it can be used in the Datadog Go client
}
```

In this particular API, the API calls are (usually) fast so it's unlikely we'd hit timeouts or cancellations, so don't worry about it for this PR. But maybe we can revisit as we get more experience with Terraform.

datadog/resource_datadog_authn_mapping.go
swetha-murali-dd: Added support for import

srosenthal-dd (approved):
Looks great! No concerns now ✅ 🏁 🚢 

I re-tested and `terraform import datadog_authn_mapping.imported "1e0aec44-8ba5-11ec-813f-da7ad0900002"` worked for me now.

datadog/tests/provider_test.go
```diff
@@ -141,6 +141,7 @@ var testFiles2EndpointTags = map[string]string{
 	"tests/resource_datadog_user_test":                                   "users",
 	"tests/resource_datadog_webhook_custom_variable_test":                "webhook_custom_variable",
 	"tests/resource_datadog_webhook_test":                                "webhook",
+	"tests/resource_datadog_authn_mapping_test":                          "authn_mapping",
```
skarimo, line 144: Nit: Can we move this up to keep the existing ordering style

swetha-murali-dd:
@skarimo incorporated changes from your review. Can you please re-review.

PR #1368, opened 2022-02-24
[Synthetics] Add certificate check option for ssl tests

romainberger:
Add a missing option for Synthetics SSL tests.


AlaricCalmette (approved):
Not sure why you removed all the convertToString but lgtm

kayayarai (approved):

romainberger:
> Not sure why you removed all the convertToString but lgtm

They were all used for the legacy options format that is not used anymore, I supposed it was forgotten to clean when the old format got removed.

skarimo (approved):

Issue #1382, opened 2022-03-15
datadog_monitor: Spurious "changes made outside of Terraform" with newly created monitor


markadev:
### Terraform Version

```
$ terraform -v
Terraform v1.1.7
on darwin_amd64
+ provider registry.terraform.io/datadog/datadog v3.9.0
```

### Affected Resource(s)

-   datadog_monitor

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    datadog = {
      source  = "DataDog/datadog"
      version = "3.9.0"
    }
  }
}

resource "datadog_monitor" "testmon" {
  type = "query alert"
  name = "[staging] too many restarts"
  query = "min(last_5m):per_minute(sum:kubernetes.containers.restarts{datacenter:us1,service:bar}) > 0"
  message = "it's bustered"
  monitor_thresholds {
    critical = 0
  }
  require_full_window = false
  tags = ["team:foo", "service:bar"]
}
```

### Debug Output

Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output

If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior

After the configuration is applied, I expect a `terraform plan` to return that there is no difference between the local and remote configurations.

### Actual Behavior

Terraform says that there were "changes made outside of Terraform" for the monitor that was just created:

```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # datadog_monitor.testmon has changed
  ~ resource "datadog_monitor" "testmon" {
        id                   = "66041677"
        name                 = "[staging] too many restarts"
      + renotify_statuses    = []
      + restricted_roles     = []
        tags                 = [
            "service:bar",
            "team:foo",
        ]
        # (16 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }


Unless you have made equivalent changes to your configuration, or ignored the relevant attributes using ignore_changes, the following plan may include
actions to undo or respond to these changes.
```

### Steps to Reproduce

Using the configuration pasted above:

1. `terraform apply`
2. `terraform plan`

### Important Factoids

I work at Datadog and this configuration is being applied in the internal Datadog org so there may be more feature flags enabled than a typical customer.

### References

[...]

PR #1384, opened 2022-03-16
[datadog_monitor] Fix spurious "changes made outside of terraform"

markadev:
Fix for #1382 

After creating a new `datadog_monitor` with the optional parameters 'renotify_statuses' and 'restricted_roles' not set, plans afterward show a perpetual "changes made outside of terraform" diff for those fields indicating that it was set to an empty list.


jirikuncar (approved):

eflann:
Thank you! This is also affecting me!

PR #1389, opened 2022-03-18
Fix prepare release

therve:
Need to create the branch first.

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -46,6 +46,19 @@ jobs:
               .toString("utf-8")
               .split("\n");
             const newContent = changes.concat(rawContent).join("\n");
+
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -46,6 +46,19 @@ jobs:
               .toString("utf-8")
               .split("\n");
             const newContent = changes.concat(rawContent).join("\n");
+
+            const { data: master } = await github.rest.git.getRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: "heads/master",
```
jirikuncar, line 53: should we use the base from which the `workflow_dispatch` was triggered?

therve: No I don't think we needed that until now.

jirikuncar (approved):

PR #1391, opened 2022-03-18
[datadog_synthetics_test] Add synthetics test data source

sdeprez:
Similar to https://github.com/DataDog/terraform-provider-datadog/pull/1151 which did it for Synthetics Variables, this PR adds a data source for Synthetics Tests. This can be useful for instance to get data from a test that you want to manage in the UI but for which you need some values to terraform other tests.

For now, I only added a few fields on the data source result object, we can add more in the future if needed.


datadog/data_source_datadog_synthetics_test_.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := utils.CheckForUnparsed(tests); err != nil {
+		return diag.FromErr(err)
+	}
+
+	searchedId := d.Get("test_id").(string)
+	var matchedTests []datadog.SyntheticsTestDetails
+
+	for _, test := range *tests.Tests {
+		if test.GetPublicId() == searchedId {
+			matchedTests = append(matchedTests, test)
```
therve, line 66: Can we break here? We don't handle multiple results anyway, and we're testing an ID some presumably it's unique?

datadog/tests/data_source_datadog_synthetics_test_test.go
```diff
@@ -0,0 +1,79 @@
...
+    method = "GET"
+    url    = "https://www.example.com"
+  }
+	options_list {
+		tick_every = 900
+	}
+	assertion {
+    type     = "statusCode"
+    operator = "is"
+    target   = "200"
```
jirikuncar, line 64: can you please reformat the test config?

therve (approved):

PR #1398, opened 2022-03-30
[datadog_synthetics_test] Use a correct regex for variables

AlaricCalmette:


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1504,7 +1504,7 @@ func buildAssertions(attr []interface{}) []datadogV1.SyntheticsAssertion {
 							case
 								datadogV1.SYNTHETICSASSERTIONOPERATOR_LESS_THAN,
 								datadogV1.SYNTHETICSASSERTIONOPERATOR_MORE_THAN:
-								if match, _ := regexp.MatchString("{{\\s*[a-zA-Z0-9]+\\s*}}", v.(string)); match {
+								if match, _ := regexp.MatchString("{{\\s*([^{}]*?)\\s*}}", v.(string)); match {
```
jirikuncar, line 1507: @AlaricCalmette why `*?` instead of `+`?
```suggestion
								if match, _ := regexp.MatchString("{{\\s*([^{}]+)\\s*}}", v.(string)); match {
```

AlaricCalmette: To be consistent with what we check in web-ui [here](https://github.com/DataDog/web-ui/blob/preprod/javascript/datadog/synthetics/lib/constants.ts#L3)

sdeprez (approved):

therve (approved):

PR #1400, opened 2022-03-31
[datadog_monitor] Mark locked as deprecated

phillip-dd:
With monitor restricted roles being GA, this marks the previous locked as deprecated.

@DataDog/monitor-app


gkharkwal (approved):

jtappa (dismissed):
👍🏻 from docs team

therve (approved):

martinb3:
FYI I think this locked attribute may now be rejected by the Datadog monitoring API, and any previous monitor that has locked in the state, even if it's no longer specified, may be failing to update?

https://github.com/DataDog/terraform-provider-datadog/issues/2327

PR #1404, opened 2022-04-04
[resource_datadog_metric_metadata] Clarify supported metric types

keisku:
Clarify supported metric types, `gauge`, `rate` on the `resource_datadog_metric_metadata` document.
`resource_datadog_metric_metadata` depends on [Edit metric metadata](https://docs.datadoghq.com/api/latest/metrics/#edit-metric-metadata), so metric types must be `gauge` or `rate`. That should be documented.
Metrics 2022-04-04 at 2 14 21 PM


maycmlee (dismissed):
👍  from docs.

therve (approved):

PR #1415, opened 2022-04-19
[datadog_resource_dashboard] Implement support for timeseries background in query value widgets

DrkSephy:
Implements support for adding timeseries backgrounds to query value widgets.


jtappa (approved):
👍🏻 from docs team

therve (approved):

PR #1417, opened 2022-04-19
[mute_first_recovery_notification] add mute first recovery notification to monitor options

JoannaYe-Datadog:
add `mute_first_recovery_notification` to monitor options


alai97 (approved):
Looks good, one copy nit!

datadog/resource_datadog_downtime.go
```diff
@@ -262,6 +267,11 @@ func resourceDatadogDowntime() *schema.Resource {
 				Computed:    true,
 				Description: "The id corresponding to the downtime object definition of the active child for the original parent recurring downtime. This field will only exist on recurring downtimes.",
 			},
+			"mute_first_recovery_notification": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "When true the first recovery notification during the downtime will be muted",
```
skarimo, line 273: We should set the default value since it looks like the api defaults to `false`. Otherwise a diff might be produced.
```suggestion
				Description: "When true the first recovery notification during the downtime will be muted",
				Default:     false,
```

skarimo (approved):

PR #1428, opened 2022-04-27
[datadog_integration_gcp] Add example w/ `google_service_account_key` resource

jketcham:
I just had a support issue resolved after updating my `datadog_integration_gcp` resource config to work like the example added in this PR. 

I believe this would be a pretty common scenario for people using this resource (together with a `google_service_account_key`), so I thought adding a straightforward example of that would be nice.

Let me know if any changes need to be made, thanks!


alai97 (approved):
Looks good for docs!

audreyyuan1 (approved):
looks good from the cloud integrations side!

skarimo (approved):

PR #1444, opened 2022-05-16
Exponential backoff period for `5xx` errors and enabled retries by default

skarimo:


maycmlee (dismissed):
👍  for docs.

datadog/internal/transport/custom_transport.go
```diff
@@ -4,22 +4,23 @@ import (
...
 	"net/http"
 	"strconv"
 	"time"
 )
 
 var (
-	defaultHTTPRetryDuration = 5 * time.Second
-	defaultHTTPRetryTimeout  = 60 * time.Second
-	rateLimitResetHeader     = "X-Ratelimit-Reset"
+	defaultBackOffMultiplier float64 = 2
```
rmangi, line 14: nit; format indentation

skarimo: Thanks for the review. This was auto formatted using `goimports` which uses tabs for indentation

datadog/internal/transport/custom_transport.go
```diff
@@ -85,7 +80,7 @@ func (t *CustomTransport) copyRequest(r *http.Request) *http.Request {
 	return &newRequest
 }
 
-func (t *CustomTransport) retryRequest(response *http.Response) (*time.Duration, bool) {
+func (t *CustomTransport) retryRequest(response *http.Response, retryCount int) (*time.Duration, bool) {
```
raymondyou97, line 83: does this code keep retrying the request if its a `5xx` or `429` forever until it works? can we make it so after like 3 retries, it terminates?

skarimo: Yes, it currently retries indefinitely until timeout period is reached or request is successful (whichever is first). Yeah i think we should be able to just cancel the context after x retries

raymondyou97: yeah i think this is the behavior we want (cancelling the context after x retries)

PR #1464, opened 2022-06-15
[synthetics_test] Add rum settings

romainberger:
Add rum settings to Synthetics browser tests.


buraizu:
Thanks, created DOCS-3645 for tracking!

jtappa (dismissed):

dajofo (approved):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1773,6 +1794,19 @@ func buildSyntheticsBrowserTestStruct(d *schema.ResourceData) *datadogV1.Synthet
...
 			options.SetRestrictedRoles(roles)
 		}
+
+		if rum_settings, ok := d.GetOk("options_list.0.rum_settings.0"); ok {
+			settings := rum_settings.(map[string]interface{})
+			applicationId := settings["application_id"]
+			clientTokenId := settings["client_token_id"]
+
+			rumSettings := datadogV1.SyntheticsBrowserTestRumSettings{}
+			rumSettings.SetIsEnabled(true)
```
skarimo, line 1804: I think it's better to add another property for this instead of always defaulting to true.

romainberger: I thought it would be easier to use instead of having  to explicitly have `is_enabled = true` that would seem a bit redundant. 
Basically either you have a configuration and it's enabled, or you don't. But I can add it if you think it can be useful in some case.

skarimo: Yeah i think its better to have it as a configurable property with a default of `true`

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -500,6 +500,32 @@ func syntheticsTestOptionsList() *schema.Schema {
...
+				"rum_settings": {
+					Description: "The RUM data collection settings for the Synthetic browser test.",
+					Type:        schema.TypeList,
+					MaxItems:    1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"is_enabled": {
+								Type:        schema.TypeBool,
+								Description: "Determines whether RUM data is collected during test runs.",
+								Optional:    true,
```
skarimo, line 512:
```suggestion
								Required:    true,
```
Is this required?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1773,6 +1799,29 @@ func buildSyntheticsBrowserTestStruct(d *schema.ResourceData) *datadogV1.Synthet
...
 			options.SetRestrictedRoles(roles)
 		}
+
+		if rum_settings, ok := d.GetOk("options_list.0.rum_settings.0"); ok {
+			settings := rum_settings.(map[string]interface{})
+			isEnabled := settings["is_enabled"]
+
+			rumSettings := datadogV1.SyntheticsBrowserTestRumSettings{}
+
+			if isEnabled == true {
```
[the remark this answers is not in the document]
romainberger: Exactly, it was done in a way that is very strict, the api documentation is more verbose than the one I've put here but if it's confusing we can do something similar: https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/synthetics.yaml#L1484

datadog/resource_datadog_synthetics_test_.go
romainberger: Yes my bad

jtappa (dismissed):

therve (approved):

PR #1497, opened 2022-07-20
[Synthetics] defaults api_step timeout to 60 to avoid it defaulting to 0

AlaricCalmette:
If no timeout is provided, the fact that timeout is of type int makes TF changing it from nil to 0 which is then what is saved by the api. The actual default should be 60.


dajofo (dismissed):
👍

romainberger (dismissed):

therve (approved):

Issue #1487, opened 2022-07-11
Provider produced inconsistent result after apply with datadog_integration_gcp and project_id as empty string


yyovkov:
Hi there,

Trying to create DataDog / GCP integration with empty `project_id` (Service Account Impersonation and Autodiscovery) using terraform returns error. The obvious difference is that the required parameter "project_id" provided is empty stirng.

### Terraform Version

```bash
_$ terraform -v
Terraform v1.2.4
...
+ provider registry.terraform.io/datadog/datadog v3.12.0
...
```

### Affected Resource(s)

-   datadog_integration_gcp

### Terraform Configuration Files

```hcl
locals {
  service_account_key_str = var.service_account_key == null ? var.service_account.private_key : var.service_account_key
  service_account_key     = jsondecode(base64decode(local.service_account_key_str))
}

resource "datadog_integration_gcp" "instance" {
  client_email   = var.service_account.email
  client_id      = local.service_account_key.client_id
  private_key    = local.service_account_key.private_key
  private_key_id = local.service_account_key.private_key_id
  project_id     = ""
}

terraform {
  required_providers {
    datadog = {
      source = "DataDog/datadog"
    }
  }
}
```

### Debug Output

No Debug Output

### Panic Output

No panic produced 

### Expected Behavior

Apply successfully the code, store the result in the state file and not to generate error messages

### Actual Behavior

Applies the change at DD (the resource is visible in DD web interface), however the resource is not added to the state file (on next run the resource "Conflict" error message appears) and the error message is generated:

``` bash
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.datadog.module.datadog_gcp.datadog_integration_gcp.instance, provider
│ "provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Running the same code second time generates "Conflict" error message:

``` bash
╷
│ Error: error creating GCP integration from https://api.datadoghq.eu/api/v1/integration/gcp: 409 Conflict: {"errors":["This Service account already exists in your Datadog account."]}
│ 
│ 
│   with module.datadog.module.datadog_gcp.datadog_integration_gcp.instance,
[...]

PR #1498, opened 2022-07-21
[datadog_integration_gcp] Handle empty project id for GCP integration

skarimo:
Closes: https://github.com/DataDog/terraform-provider-datadog/issues/1487

### DIFF
--- a/datadog/internal/utils/utils.go
+++ b/datadog/internal/utils/utils.go
@@ -132,6 +132,15 @@ func AccountNameAndChannelNameFromID(id string) (string, string, error) {
 	return result[0], result[1], nil
 }
 
+// ProjectIdAndClientEmailFromID returns project id and client email from an ID
+func ProjectIdAndClientEmailFromID(id string) (string, string, error) {
+	result := strings.SplitN(id, ":", 2)
+	if len(result) != 2 {
+		return "", "", fmt.Errorf("error extracting project ID and client email from an GCP integration id: %s", id)
+	}
+	return result[0], result[1], nil
+}
+
 // ConvertResponseByteToMap converts JSON []byte to map[string]interface{}
 func ConvertResponseByteToMap(b []byte) (map[string]interface{}, error) {
 	convertedMap := make(map[string]interface{})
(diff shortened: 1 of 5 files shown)


datadog/internal/utils/utils.go
```diff
@@ -132,6 +132,15 @@ func AccountNameAndChannelNameFromID(id string) (string, string, error) {
 	return result[0], result[1], nil
 }
 
+// ProjectIdAndClientEmailFromID returns project id and client email from an ID
+func ProjectIdAndClientEmailFromID(id string) (string, string, error) {
+	result := strings.SplitN(id, ":", 2)
+	if len(result) != 2 {
+		return "", "", fmt.Errorf("error extracting project ID and client email from an GCP integration id: %s", id)
```
NouemanKHAL, line 139:
```suggestion
		return "", "", fmt.Errorf("error extracting project ID and client email from a GCP integration id: %s", id)
```

datadog/resource_datadog_integration_gcp.go
```diff
@@ -108,35 +111,22 @@ func resourceDatadogIntegrationGcpCreate(ctx context.Context, d *schema.Resource
 		return utils.TranslateClientErrorDiag(err, httpresp, "error creating GCP integration")
 	}
 
-	d.SetId(projectID)
+	d.SetId(fmt.Sprintf("%s:%s", projectID, clientEmail))
 
-	return resourceDatadogIntegrationGcpRead(ctx, d, meta)
+	return updateDatadogIntegrationGCPState(d, meta)
```
NouemanKHAL, line 116:
```suggestion
	return resourceDatadogIntegrationGcpRead(ctx, d, meta)
```
Am I missing something?

skarimo: Why do you think we need the context?  We don't really use it so its an unnecessary extra argument for now

NouemanKHAL (approved):

jtappa (approved):
Looks like docs review not necessary here, but I can't remove us, so approving.

yyovkov:
When/which version is expected the new workflow?

PR #1500, opened 2022-07-22
[datadog_synthetics_test] Add support for grpc subtype

AlaricCalmette:


docs/resources/synthetics_test.md
```diff
@@ -446,6 +446,7 @@ Optional:
 - `number_of_packets` (Number) Number of pings to use per test for ICMP tests (`subtype = "icmp"`) between 0 and 10.
 - `port` (Number) Port to use when performing the test.
 - `servername` (String) For SSL tests, it specifies on which server you want to initiate the TLS handshake, allowing the server to present one of multiple possible certificates on the same IP address and TCP port number.
+- `service` (String) For gRPC tests, service to target for healthcheck.
```
cswatt, line 449:
```suggestion
- `service` (String) For gRPC tests, service to target for health check.
```

docs/resources/synthetics_test.md
```diff
@@ -732,6 +733,7 @@ Optional:
...
```
cswatt, line 736:
```suggestion
- `service` (String) For gRPC tests, service to target for health check.
```

cswatt (approved):
Leaving very nitpicky edit; approving so as not to block.

docs/resources/synthetics_test.md
AlaricCalmette: I purposefully put healthcheck to match what we have in the UI. If we're to change that's a discussion I need to have with my PMs.

therve (approved):

PR #1506, opened 2022-07-27
[datadog_security_monitoring_rule] Check for nil `metrics` values.

juliendoutre:
## What does this PR do?
- Fix a bug leading to an invalid type casting for this field (`metrics` can be set to `nil` leading to issues whenever trying to cast it to a string array)
- Makes its values either set explicitly by the user, or computed by the API's backend


adriacabeza (approved):

muffix (approved):

clementgbcn (dismissed):
LGTM 👌

adriacabeza (approved):

skarimo (approved):

Issue #1418, opened 2022-04-20
datadog_monitor resource "event-v2 alert" Ignore monitor_thresholds { ok = 0 }


vidbaz:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Summary
>We need to migrate from "event alert" to "event-v2 alert" as mentioned here:  https://docs.datadoghq.com/events/guides/migrating_to_new_events_features/
>We did the change on the datadog_monitor resource but we are facing the issue that one of the monitor_thresholds values keeps showing up as a change on the terraform plan. Even if we already apply that change on that resource.

### Terraform Version

```bash
terraform -v       
Terraform v1.1.7
on linux_amd64
+ provider registry.terraform.io/datadog/datadog v3.10.0
```

### Affected Resource(s)

Please list the resources as a list, for example:

-   datadog_monitor

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files

```hcl
resource "datadog_monitor" "watchdog_mysql_monitor" {
  for_each = toset(var.services)
  name     = "${upper(var.team)}: *[Watchdog]* {{event.title}} - ${each.key}-mysql"
  type     = "event-v2 alert"
  message  = <<EOF
  {{#is_alert}}
  {{env}} - @webhook-{{env}}-${var.team}-warnings
  {{/is_alert}} 
EOF

  query = "events(\"source:watchdog story_type:service team:_aggregate service:${each.key}-mysql\").rollup(\"count\").by(\"env\").last(\"30m\") >= 1"

  monitor_thresholds {
    ok       = 0        # <<< This is the value that continues showing up even after applied it
    critical = 1
  }
  require_full_window = false
  notify_no_data      = false
  no_data_timeframe   = 0
  include_tags        = false

}
```

### Debug Output

Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

https://gist.github.com/vidbaz/acbc52d01c5129e1b52edc7f8eefe79b#file-debug_datadog_provider

### Panic Output

If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.
>No Panic output

### Expected Behavior

What should have happened?
[...]

PR #1529, opened 2022-08-16
[datadog_monitor] Suppress diff when using `ok` and `unknown` thresholds in non `service check` monitors.

skarimo:
Closes https://github.com/DataDog/terraform-provider-datadog/issues/1418


unclebconnor (approved):
For reference, pointing to the related validation in dogweb to confirm this is the correct behavior: <see link in thread>

hestonhoffman (approved):

therve (approved):

PR #1530, opened 2022-08-16
[datadog_application_key] Get and list all application keys instead of current user's

fernandocarletti:
In our setup, each user has their own API and APP keys. When user `A` creates a key and user `B` updates the stack the key won't be found and it will be recreated causing a conflict with the existing name.

Should it actually get all the keys instead of current user's?


fernandocarletti:
Never mind. This issue happens in the related Pulumi provider but seems not to be affecting the Terraform's one. Closing it for now.

PR #1542, opened 2022-08-29
[datadog_role] Upgrade provider from old single-permission APIs to newer UpdateRole API

retsguj:
https://datadoghq.atlassian.net/browse/ACCESS-1053

Uses the `UpdateRole` api to handle permission changes in a role instead of the individual `AddPermissionToRole` and `RemovePermissionFromRole` apis.

To test, create a role in terraform or use an existing role name:
```
resource "datadog_role" "my_role" {
  name = "julia gu terraform role testing"
  permission {
    id = "5e605652-dd12-11e8-9e53-375565b8970e"
  }
  permission {
    id = "984d2f00-d3b4-11e8-a200-bb47109e9987"
  }
}
```
Run `terraform plan` to look at changes, than `terraform apply` to make the changes.

Then remove the permissions:
```
resource "datadog_role" "my_role" {
  name = "julia gu terraform role testing"
}
```
Run `terraform plan`. Terraform should tell you something like:
```
  # datadog_role.my_role will be updated in-place
  ~ resource "datadog_role" "my_role" {
        id         = "1120d450-289c-11ed-b5bb-da7ad0900002"
        name       = "julia gu terraform role testing"

      - permission {
          - id   = "5e605652-dd12-11e8-9e53-375565b8970e" -> null
          - name = "logs_read_index_data" -> null
        }
      - permission {
          - id   = "984d2f00-d3b4-11e8-a200-bb47109e9987" -> null
          - name = "standard" -> null
        }
    }

```
Run `terraform apply`. If you re-run `terraform plan` afterwards, you should get a `No changes. Your infrastructure matches the configuration.` message. Or if you go and look at the role in the UI, there should be no permissions except for the 8 restricted ones.


srosenthal-dd:
At a glance, it looks good, but tests are failing. I think you to re-record/re-generate cassettes? https://github.com/DataDog/terraform-provider-datadog/tree/master/datadog/tests/cassettes

Is it ready enough that I could give it a quick local+manual test? I'd like to give it a quick check before approving

retsguj:
> At a glance, it looks good, but tests are failing. I think you to re-record/re-generate cassettes? https://github.com/DataDog/terraform-provider-datadog/tree/master/datadog/tests/cassettes
> 
> Is it ready enough that I could give it a quick local+manual test? I'd like to give it a quick check before approving

It's mostly ready. The cassettes aren't the reason the tests are failing though. I've found a bug, not sure what it should be considered a bug of, though.

In the api implementation in dogweb, if nothing is passed in for `permissions`, the api does not update anything. So there's no way to remove all permissions (that can be removed) from a role using `UpdateRole`. I could fix the bug in this instance by using the `RemovePermissionFromRole` api for when all permissions are removed, but it is a clumsy solution.

If we changed the way the implementation in dogweb works, there could be problems for when one is trying to update the name of a role, but not the permissions.

srosenthal-dd:
Hmm interesting! It looks like it's not easy to distinguish between an empty list of permissions or an unset list of permissions?

Our "main" client of the web UI will always provide a non-empty set of permissions (currently, due to restricted permissions which are sort-of-mandatory on all roles). But other clients might not. If we wanted to be extra careful we could add a metric measuring how often users update roles but only update the name. I suspect it's not often.

Supporting existing APIs can be challenging!

retsguj:
> Our "main" client of the web UI will always provide a non-empty set of permissions (currently, due to restricted permissions which are sort-of-mandatory on all roles)

Oh, is this true for sure? I think the implementation doesn't allow removal of restricted permissions so one wouldn't need to provide them to keep them assigned to the role.

srosenthal-dd:
Yes, the web UI always PATCHES the list of permissions including the ~8 restricted ones, even if you just changed the name. Verified via the browser dev tools.

skarimo (approved):

skarimo (approved):

RaghavK12RK:
where can we get the permission ids?

RaghavK12RK:
Can we have a support feature for multiple dd_user invites through terraform code?

srosenthal-dd:
@RaghavK12RK that doesn't seem relevant to this pull request. I'll response to your issue here: https://github.com/DataDog/terraform-provider-datadog/issues/1558

Issue #1548, opened 2022-09-02
Feature Request: Data Source for Log Pipelines


BennyTheSen:
Hi there,

we currently run into the problem, that we want to control all of the log pipelines with terraform. The pipeline order resource needs information of all log pipelines, including the automatically generated integration pipelines. It is a bit inconvenient to manually import the existing integration pipelines and every time you add a new integration having to import the pipeline as well.
It would be very helpful if there would be Data Sources similar to the API calls to  
* get All log Pipelines,
* get a pipeline with its attributes

Equilibri0:
We have the same kind of problems. Specially if it comes to the order. An Additional data source would help alot !

nkzou:
The feature has been merged and should be available in the next release. Thanks for your FR!

PR #1555, opened 2022-09-07
[datadog_logs_pipelines] Add a pipelines datasource

nkzou:
Addresses #1548. Adds a pipelines data source with a is_read_only filter. Doesn't currently expose processors.


datadog/data_source_datadog_logs_pipelines.go
```diff
@@ -0,0 +1,119 @@
...
+				"is_read_only": pipeline.GetIsReadOnly(),
+				"type":         pipeline.GetType(),
+			})
+		}
+	}
+	if err := d.Set("logs_pipelines", tflogsPipelines); err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId("logs-pipelines")
```
skarimo, line 117: We should calculate a unique hash for the id here based on the filter `is_read_only`. Mainly because users might create multiple resources with different filter settings and ID's would clash. I believe we do it in the monitors datasource which you can get inspiration from.

datadog/tests/data_source_datadog_logs_pipelines_test.go
```diff
@@ -0,0 +1,126 @@
...
+		auth := providerConf.Auth
+		apiInstances := providerConf.DatadogApiInstances
+
+		logsPipelines, _, err := apiInstances.GetLogsPipelinesApiV1().ListLogsPipelines(auth)
+		if err != nil {
+			return err
+		}
+		if err := utils.CheckForUnparsed(logsPipelines); err != nil {
+			return err
+		}
```
skarimo, line 56:
```suggestion
```

datadog/tests/data_source_datadog_logs_pipelines_test.go
```diff
@@ -0,0 +1,126 @@
...
+			if pipeline.GetIsReadOnly() {
+				readOnlyPipelines = append(readOnlyPipelines, pipeline)
+			}
+		}
+		if err != nil {
+			return err
+		}
+		if err := utils.CheckForUnparsed(logsPipelines); err != nil {
+			return err
+		}
```
skarimo, line 114:
```suggestion
```
No need. This check is already done in the resource itself

skarimo (commented):
Could you add an example here as well: https://github.com/DataDog/terraform-provider-datadog/tree/master/examples/data-sources

datadog/data_source_datadog_logs_pipelines.go
nkzou: Good point, done in [a8731f6](https://github.com/DataDog/terraform-provider-datadog/pull/1555/commits/a8731f69a47e4e8797a302854f3c80e474844f89)

skarimo (approved):
Nice!

vinckama (approved):
Thanks, this looks good to me.

PR #1575, opened 2022-09-23
clarify name of an index order

kzw:
Because an index order is always imported from one and only resource at the datadog server and because there does not appear to be a way to query the "name" of this unique index order, clarify what "name" to import from.

### DIFF
--- a/docs/resources/logs_index_order.md
+++ b/docs/resources/logs_index_order.md
@@ -42,4 +42,4 @@ Import is supported using the following syntax:
 # The Datadog Terraform Provider does not support the creation and deletion of index orders. There must be at most one `datadog_logs_index_order` resource
 terraform import <datadog_logs_index_order.name> <name>
 ```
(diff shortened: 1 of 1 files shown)


docs/resources/logs_index_order.md
```diff
@@ -42,3 +42,4 @@ Import is supported using the following syntax:
 # The Datadog Terraform Provider does not support the creation and deletion of index orders. There must be at most one `datadog_logs_index_order` resource
 terraform import <datadog_logs_index_order.name> <name>
 ```
+`<name>` here can be whatever you specify in your code; datadog does not store a name at their server.
```
urseberry, line 45:
```suggestion
`<name>` can be whatever you specify in your code; Datadog does not store a name on the server.
```

urseberry (commented):
Thank you for the PR!

skarimo:
Thanks for the PR. Continuing this in: https://github.com/DataDog/terraform-provider-datadog/pull/1793

PR #1591, opened 2022-10-03
Add Terraform Support for Signal Correlation rules

clementgbcn:
This PR updates the Terraform provider to support `Signal Correlation` rules.

It impact the resources/data:
 - `resource_datadog_security_monitoring_rule`
 - `data_source_datadog_security_monitoring_rules`

To stay consistent with the code generated in Go (see https://github.com/DataDog/datadog-api-client-go/pull/1689), we decided to rely on the root fields `standard_rule` and `signal_rule` to differentiate `SecurityMonitoringStandardRuleSchema` and `SecurityMonitoringSignalRuleSchema`.

Since the Go struct `SecurityMonitoringRuleResponse` does not implement a common interface for `SecurityMonitoringStandardRuleResponse` and `SecurityMonitoringSignalRuleResponse`, most of the methods have been specialised in spite of having common code.

Additionally, this PR homogenises the way the payloads are populated by using the Setters instead of directly accessing to the attributes in `security_monitoring`.

PR #1593, opened 2022-10-04
[datadog_security_monitoring] Add Terraform Support for Signal Correlation Rules

clementgbcn:
## Description
This PR updates the Terraform provider to support `Signal Correlation` rules.

It replaces https://github.com/DataDog/terraform-provider-datadog/pull/1591 which was not backward compatible.

It impact the resources/data:
 - `resource_datadog_security_monitoring_rule`
 - `data_source_datadog_security_monitoring_rules`

To be backward compatible, the fields which are specific to __Signal Correlation__ rules 9(i.e., `ruleId`, `defaultRuleId`, `correlatedByFields`, `correlatedQueryIndex`) are populated in the same Schema of the query.
Since `query` is a List that could have multiple elements, it is not possible to enforce a `DistinctsWith`.

Since the Go struct `SecurityMonitoringRuleResponse` does not implement a common interface for `SecurityMonitoringStandardRuleResponse` and `SecurityMonitoringSignalRuleResponse`, most of the methods have been specialised in spite of having common code.

Additionally, this PR homogenises the way the payloads are populated by using the Setters instead of directly accessing the attributes in `security_monitoring`. It addresses remarks from previous PR about wording.

## How do these changes impact current model

Up to now, Datadog Security Monitoring has been supporting one common type of rule whose Terraform schema is
```
resource "datadog_security_monitoring_rule" "log_detection_example" {
  name = "My log detection rule"
  message = "Rule triggered"
  enabled = false
  has_extended_title = true
  query {
	  name = "first"
	  query = "does not really match much"
	  aggregation = "count"
	  group_by_fields = ["host"]
  }
  case {...}
  options {...}
  filter {...}
  tags = [...]
}
```
This schema is not modified for existing rule type.

On the other hand, signal correlation rules will have a similar schema which will differ by its queries and its type:
```
resource "datadog_security_monitoring_rule" "signal_correlation_example" {
  name = "My Signal Correlation rule"
  message = "Important attack in progress"
  enabled = false
  has_extended_title = true
  query {
	  name = "first"
	  rule_id = "lc2-ol2-ue0"
	  aggregation = "event_count"
	  correlated_fy_fields = ["host"]
          correlated_query_index = ""
  }
  query {
	  name = "first"
	  rule_id = "fgt-40m-kiv"
	  aggregation = "event_count"
	  group_by_fields = ["host"]
          correlated_query_index = "1"
  }
  case {...}
  options {...}
  filter {...}
  tags = [...]
}
```

Although Terraform allows a List of which mixes queries with `query` and queries with `ruleId`, the backend implementation will refuse this kind of request.

## `correlated_query_index`
In spite of being an index (thus an integer), the Terraform Schema uses a string to represent it because this value can be empty.
Having:
- `correlated_query_index=0`: Means correlate on the projected attributes of the first query of the correlated rule
- `correlated_query_index=""`: Means correlate on the non-projected per query attribute of the correlated rule.
Since `TypeInt` object does not support a `nil` value as a Default value, we decided to use a string which is converted to an integer if it is not empty.


maycmlee (approved):
👍 for docs

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -317,35 +426,78 @@ func resourceDatadogSecurityMonitoringRuleCreate(ctx context.Context, d *schema.
 		return diag.FromErr(err)
 	}
 
-	d.SetId(response.SecurityMonitoringStandardRuleResponse.GetId())
+	if response.SecurityMonitoringStandardRuleResponse != nil {
```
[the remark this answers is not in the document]
clementgbcn: However, the next instruction is about getting the id of the response by accessing the non null field of the response. Therefore, it seems to me that this check is mandatory (and then sufficient) for assigning the id.

datadog/data_source_datadog_security_monitoring_rules.go
```diff
@@ -122,16 +122,25 @@ func dataSourceDatadogSecurityMonitoringRulesRead(ctx context.Context, d *schema
 		}
 
 		for _, ruleR := range response.GetData() {
-			if ruleR.SecurityMonitoringStandardRuleResponse == nil {
+			if ruleR.GetActualInstance() == nil {
 				continue
 			}
```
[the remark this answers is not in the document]
clementgbcn: 👌 , I will add a `if` in the `else` to check that the value is not `nil`.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -317,35 +426,78 @@ func resourceDatadogSecurityMonitoringRuleCreate(ctx context.Context, d *schema.
...
+		return err == nil
+	}
+	return false
+}
 
+func checkQueryConsistency(d *schema.ResourceData) error {
+	query := d.Get("query").([]interface{})
+	signalQuery := d.Get("signal_query").([]interface{})
+	if len(query) > 0 && len(signalQuery) > 0 {
+		return fmt.Errorf("query list and signal query list cannot be both populated")
```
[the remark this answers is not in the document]
clementgbcn: I tried but it was not working since, as stated in the documentations: 
```
// Only absolute attribute paths, ones starting with top level attribute
// names, are supported. Attribute paths cannot be accurately declared
// for TypeList (if MaxItems is greater than 1), TypeMap, or TypeSet
// attributes. To reference an attribute under a single configuration block
// (TypeList with Elem of *Resource and MaxItems of 1), the syntax is
// "parent_block_name.0.child_attribute_name".
```
In our case, it is a TypeList whose MaxItems is greater than 1.

skarimo (dismissed):

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -221,7 +223,7 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
 						Type:             schema.TypeString,
 						ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewSecurityMonitoringRuleQueryAggregationFromValue),
 						Optional:         true,
-						Description:      "The aggregation type.",
+						Description:      "The aggregation type. For Signal Correlation rules, it must be event_query.",
```
pietrodll, line 226:
```suggestion
						Description:      "The aggregation type. For Signal Correlation rules, it must be event_count.",
```

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -262,6 +264,50 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
...
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "Queries for selecting logs which are part of the rule.",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"aggregation": {
+						Type:             schema.TypeString,
+						ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewSecurityMonitoringRuleQueryAggregationFromValue),
+						Optional:         true,
+						Description:      "The aggregation type. For Signal Correlation rules, it must be event_query.",
```
pietrodll, line 277:
```suggestion
						Description:      "The aggregation type. For Signal Correlation rules, it must be event_count.",
```

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -317,35 +363,78 @@ func resourceDatadogSecurityMonitoringRuleCreate(ctx context.Context, d *schema.
...
+		payload, err := buildCreateSignalPayload(d)
+		return datadogV2.SecurityMonitoringSignalRuleCreatePayloadAsSecurityMonitoringRuleCreatePayload(&payload), err
+	}
+	payload, err := buildCreateStandardPayload(d)
+	return datadogV2.SecurityMonitoringStandardRuleCreatePayloadAsSecurityMonitoringRuleCreatePayload(&payload), err
+}
+
+func buildCreateStandardPayload(d *schema.ResourceData) (datadogV2.SecurityMonitoringStandardRuleCreatePayload, error) {
 	payload := datadogV2.SecurityMonitoringStandardRuleCreatePayload{}
 	payload.Cases = buildCreatePayloadCases(d)
```
pietrodll, line 416: Is there a reason to change all the other fields to `payload.SetXxx` but not `Cases`?

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -669,6 +893,27 @@ func extractFiltersFromRuleResponse(ruleResponse *datadogV2.SecurityMonitoringSt
 	return filters
 }
 
+func extractRuleCases(responseRulesCases []datadogV2.SecurityMonitoringRuleCase) []interface{} {
+	ruleCases := make([]interface{}, len(responseRulesCases))
```
pietrodll, line 897: Nit to make the output type more explicit
```suggestion
func extractRuleCases(responseRulesCases []datadogV2.SecurityMonitoringRuleCase) []map[string]interface{} {
	ruleCases := make([]map[string]interface{}, len(responseRulesCases))
```

pietrodll (commented):
Nice job 👌 There's a typo in the aggregation type description, otherwise only nits 😄

datadog/resource_datadog_security_monitoring_rule.go
clementgbcn: ✅

datadog/resource_datadog_security_monitoring_rule.go
clementgbcn: ✅

datadog/resource_datadog_security_monitoring_rule.go
clementgbcn: I missed it when doing the update. I will fix that.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -292,14 +338,53 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
...
+			ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewSecurityMonitoringRuleTypeReadFromValue, datadogV2.NewSecurityMonitoringSignalRuleTypeFromValue),
 			Optional:         true,
 			Description:      "The rule type.",
 			Default:          "log_detection",
 		},
 	}
 }
 
+// SecurityMonitoringRuleInterface Common Interface to SecurityMonitoringRuleCreateInterface and SecurityMonitoringRuleReadInterface
+type SecurityMonitoringRuleInterface interface {
```
pietrodll, line 350: I think these interfaces should be private, as they are utilities to factorize code.

pietrodll (approved):
Very nice 🎉 only a couple of nits

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -527,22 +678,75 @@ func buildCreatePayloadQueries(d *schema.ResourceData) []datadogV2.SecurityMonit
...
+
+		if v, ok := query["metrics"]; ok && v != nil {
+			if tfMetrics, ok := v.([]interface{}); ok && len(tfMetrics) > 0 {
+				metrics := make([]string, len(tfMetrics))
+				for i, value := range tfMetrics {
+					metrics[i] = value.(string)
+				}
+				payloadQuery.SetMetrics(metrics)
+			}
+		}
```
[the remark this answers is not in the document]
clementgbcn: ✅ (in all places)

pietrodll (approved):
🎉

skarimo (approved):

PR #1595, opened 2022-10-11
[datadog_dashboard] Add support for the SLO List widget

mmeyer724:
This change will allow our customers to manage their SLO List widgets via Terraform 🙂


DEVELOPMENT.md
```diff
@@ -70,9 +70,9 @@ The Datadog terraform provider uses the standard [Terraform Testing Framework][6
 
 **NOTE** Use the API and APP keys for a sandbox/test organization, never an account hosting production data. This test suite will create/update/delete real resources.
 
--   `DD_API_KEY="<api_key>" DD_APP_KEY=<app_key> make testall` will run the test suite against the real Datadog API.
+-   `DD_TEST_CLIENT_API_KEY="<api_key>" DD_TEST_CLIENT_APP_KEY=<app_key> make testall` will run the test suite against the real Datadog API.
 
-We also use `cassettes` to record API request/responses, which allows the test suite to be reliable and run very quickly. There are a few environment variables that can control this behavior (All commands are assumed to be prefixed with `DD_API_KEY=` and `DD_APP_KEY=`)
+We also use `cassettes` to record API request/responses, which allows the test suite to be reliable and run very quickly. There are a few environment variables that can control this behavior (All commands are assumed to be prefixed with `DD_TEST_CLIENT_API_KEY=` and `DD_TEST_CLIENT_APP_KEY=`)
```
mmeyer724, line 75: While testing this change I discovered that our docs were wrong, so I'm updating them here for posterity (thanks for the tip @skarimo!). I can also pull this out into a separate PR if desired

datadog/resource_datadog_dashboard.go
```diff
@@ -1173,6 +1186,9 @@ func buildTerraformWidget(datadogWidget datadogV1.Widget, k *utils.ResourceDataK
 	} else if widgetDefinition.SLOWidgetDefinition != nil {
 		terraformDefinition := buildTerraformServiceLevelObjectiveDefinition(*widgetDefinition.SLOWidgetDefinition)
 		terraformWidget["service_level_objective_definition"] = []map[string]interface{}{terraformDefinition}
+	} else if widgetDefinition.SLOListWidgetDefinition != nil {
+		terraformDefinition := buildTerraformSloListDefinition(*widgetDefinition.SLOListWidgetDefinition, k.Add("slo_list_definition.0"))
```
zhengshizhao, line 1190: can we unify using either `SLO` or `Slo`? seems makes more sense to SLO as it is used in all other places?

datadog/resource_datadog_dashboard.go
```diff
@@ -4917,6 +4933,156 @@ func buildTerraformServiceLevelObjectiveDefinition(datadogDefinition datadogV1.S
...
+//
+// SLO (Service Level Objective) List Widget Definition helpers
+//
+
+func getSloListDefinitionSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"request": {
+			Description: "A nested block describing the request to use when displaying the widget. Exactly one `request` block is allowed.",
+			Type:        schema.TypeList,
+			Optional:    true,
```
skarimo, line 4945: Should this be required instead (basing it on build definition)?

datadog/resource_datadog_dashboard.go
```diff
@@ -4917,6 +4933,156 @@ func buildTerraformServiceLevelObjectiveDefinition(datadogDefinition datadogV1.S
...
+			ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewWidgetTextAlignFromValue),
+			Optional:         true,
+		},
+	}
+}
+
+func getSloListRequestSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"request_type": {
+			Description:      "The request type for the SLO List request ('slo_list').",
```
skarimo, line 4974:
```suggestion
			Description:      "The request type for the SLO List request.",
```

skarimo (commented):
You will also need to commit the cassettes as well

datadog/resource_datadog_dashboard.go
mmeyer724: Good idea, done

maycmlee (approved):
👍 for docs

datadog/resource_datadog_dashboard.go
mmeyer724: Yeah, this is definitely something I grappled with when writing the code. I ultimately decided to go with the "true" camelCase style after seeing that's how RUM and APM are referred to in the codebase (at least in the small sample size I looked at), for example:

https://github.com/DataDog/terraform-provider-datadog/blob/9ffc78a61b7b61f61d3e4c8f0d73ba36e2ad3772/datadog/resource_datadog_dashboard.go#L7705

https://github.com/DataDog/terraform-provider-datadog/blob/9ffc78a61b7b61f61d3e4c8f0d73ba36e2ad3772/datadog/resource_datadog_dashboard.go#L7826

For SLOs specifically, I see usage of both `"SLO"` and `"Slo"` throughout the codebase, so we're definitely inconsistent there

mmeyer724: I would be happy to switch over to the all-caps version though if that's preferred

skarimo: Agreed that the naming is a bit inconsistent everywhere in this resource. For the purposes of this PR, I am okay with keeping it as is (since it correlates with existing naming convention already used with rum, apm, etc). However, I will create a card on our backlog to make it consistent everywhere since I'm sure this spans across resources and is not just limited to this case.

zhengshizhao: agreed and thanks @skarimo for taking care of it!

zhengshizhao (dismissed):

datadog/resource_datadog_dashboard.go
```diff
@@ -4917,6 +4933,156 @@ func buildTerraformServiceLevelObjectiveDefinition(datadogDefinition datadogV1.S
 	return terraformDefinition
 }
 
+//
+// SLO (Service Level Objective) List Widget Definition helpers
+//
+
+func getSloListDefinitionSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"request": {
```
zhengshizhao, line 4942: is this `request ` or `requests`? asking as it is listed as `requests` in our doc page - https://docs.datadoghq.com/dashboards/widgets/slo_list/#api

datadog/resource_datadog_dashboard.go
```diff
@@ -4917,6 +4933,156 @@ func buildTerraformServiceLevelObjectiveDefinition(datadogDefinition datadogV1.S
...
+			Description: "The size of the widget's title (defaults to 16).",
+			Type:        schema.TypeString,
+			Optional:    true,
+		},
+		"title_align": {
+			Description:      "The alignment of the widget's title.",
+			Type:             schema.TypeString,
+			ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewWidgetTextAlignFromValue),
+			Optional:         true,
+		},
```
zhengshizhao, line 4967: in doc page for APIs, seems `type` is a required field for slo list widget:  https://docs.datadoghq.com/dashboards/widgets/slo_list/#api  
don't we need the `type` field here as well?

datadog/resource_datadog_dashboard.go
mmeyer724: Great question, you're right that it's named `requests` on the API side. The reason why we use the singular form here is due to how lists are defined in terraform. For example, if you wanted to specify multiple request objects your terraform would look something like this:
```
resource "datadog_dashboard" "slo_list_dashboard" {
	title         = "{{uniq}}"
	description   = "Created using the Datadog provider in Terraform"
	layout_type   = "ordered"
	is_read_only  = "true"

	widget {
		slo_list_definition {
			request {
				request_type = "slo_list"
				query {
					query_string = "query one"
					limit = 30
				}
			}
			request {
				request_type = "slo_list"
				query {
					query_string = "query two"
					limit = 30
				}
			}
			request {
				request_type = "slo_list"
				query {
					query_string = "query N"
					limit = 30
				}
			}
			...
			title = "my title"
			title_size = "16"
			title_align = "center"
		}
	}
}
```
_(note: the slo list widget does not support multiple requests, the example above is just for demonstration purposes)_

If we used the plural form, it would look like you're overwriting the requests array with every block, when what you're actually doing is appending to it

datadog/resource_datadog_dashboard.go
mmeyer724: Nah, there is only one valid value for `type` anyways (`slo_list`), so having the ability to configure it would only introduce chance for error

mmeyer724: For context, our API uses `type` to know which type of widget you're trying to create. We don't need that here because we already know what type of widget you're trying to create, so we automatically fill that in for you

mmeyer724: Actually, following that same logic I should remove `request_type` as well because there is only one valid value. Making the user specify it is unnecessary. I'll make that update

zhengshizhao: all good! yeah, I also confirmed datadogV1.NewSLOListWidgetRequestWithDefaults() in go client that it does initialize `type` for slo list definition :)

datadog/resource_datadog_dashboard.go
zhengshizhao: Thanks for the example! all good now! got confused when I read the doc page but then answered myself after read the build method below again :)

skarimo (approved):

datadog/resource_datadog_dashboard.go
mmeyer724: > Actually, following that same logic I should remove `request_type` as well because there is only one valid value. Making the user specify it is unnecessary. I'll make that update

I looked into this and turns out it's normal for this to be the case, so I'm going to leave it as-is for now

PR #1604, opened 2022-10-19
[SEC-5364] Deprecate metric field of Security Monitoring Rules

clementgbcn:
The new field `metrics` have been introduced in Datadog backend. Since it is a list of metrics, the field `metric` could be merged into the `metrics` field as a list of one element. Therefore, this field is doomed to disappear.

In order to prepare the removal of this field, this PR add a `Deprecated` tag to the `metric` field to warm Terraform users.


skarimo (approved):

pietrodll (approved):
👌

Issue #1616, opened 2022-10-28
Priority for synthetic tests doesn't allow `monitor_priority = 0`


fferrin:
Hi there,
the value `0` is not allowed for the `monitor_priority` parameter. Based on the state of my tests, `0` represents priority `Not Defined` but that value can not be assigned.

The problem arises when I try to create my tests programmatically, where some of them have priorities and some do not.

### Terraform Version

```bash
$ terraform -v
Terraform v1.0.6
on darwin_amd64
```

### Affected Resource(s)

- `datadog_synthetics_test`

### Terraform Configuration Files

```hcl
resource "datadog_synthetics_test" "foo" {
  locations = ["aws:ap-northeast-1"]
  name      = "foo"
  status    = "live"
  type      = "api"

  options_list {
    tick_every       = 30
    monitor_priority = 0    # <- THIS VALUE
  }
}
```

### Panic Output

```bash
╷
│ Error: expected options_list.0.monitor_priority to be in the range (1 - 5), got 0
│
│   with datadog_synthetics_test.bla,
│   on synth.tf line 9, in resource "datadog_synthetics_test" "bla":
│    9:     monitor_priority = 0
│
╵
```

### Expected Behavior

`0` should be a valid value since it's the one assigned when you create the resource without specifying it:

```bash
$ terraform state show "datadog_synthetics_test.foo[0]"
# datadog_synthetics_test.foo[0]:
resource "datadog_synthetics_test" "foo" {
    ...
    options_list {
        accept_self_signed   = false
        allow_insecure       = true
        follow_redirects     = false
        min_failure_duration = 0
        min_location_failed  = 1
        monitor_priority     = 0      <- THIS VALUE
        no_screenshot        = false
        tick_every           = 60

        monitor_options {
            renotify_interval = 0
        }

        retry {
            count    = 2
            interval = 300
        }
    }
```

### Actual Behavior

Value `0` to specify `Not Defined` is not allowed.

![image](https://user-images.githubusercontent.com/7387552/198636561-1944af25-efd4-4bb2-9cd4-70d7996366ca.png)

PR #1617, opened 2022-10-28
#1616 Support value `0` for `Not Defined` priority

fferrin:
Proposed solution to address https://github.com/DataDog/terraform-provider-datadog/issues/1616

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -495,7 +495,7 @@ func syntheticsTestOptionsList() *schema.Schema {
 				"monitor_priority": {
 					Type:         schema.TypeInt,
 					Optional:     true,
-					ValidateFunc: validation.IntBetween(1, 5),
+					ValidateFunc: validation.IntBetween(0, 5),
 				},
 				"restricted_roles": {
 					Description: "A list of role identifiers pulled from the Roles API to restrict read and write access.",


romainberger:
Hi, thank you for your work but unfortunately the value `0` is invalid, which is why it is not allowed. To have no priority defined you simply have to remove the key and value from your config file.

fferrin:
Hi @romainberger, I don't understand why this is invalid, as when you don't set a priority, the `monitor_priority` value is `0`.

The use case for this is that you want to create multiple synthetic tests from an array and some of them have priorities while others don't. Is there a way to achieve that?

Here is an example:

```hcl
resource "datadog_synthetics_test" "foo" {
  count     = length(local.tests)
  locations = ["aws:ap-northeast-1"]
  name      = "foo"
  status    = "live"
  type      = "api"

  request_definition {
    url    = local.tests[count.index].url
    method = local.tests[count.index].method
  }
  assertion {
    type     = "statusCode"
    operator = "is"
    target   = try(local.tests[count.index].status, "200")
  }

  options_list {
    tick_every       = 30
    monitor_priority = try(local.tests[count.index].priority, "0") # <- THIS LINE
  }
}
```

nkzou:
Hey there, I think some of the confusion here is because the terraform sdk didn't have the concept of a null value [until Terraform 0.12](https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements) meaning that omitted variables were set to a default value based on type. The `monitor_priority` that you're seeing from `terraform show` is a result of this - it's an unset value that terraform has internally assigned the default int value of 0 to.

Terraform has since added the [null keyword](https://developer.hashicorp.com/terraform/language/expressions/types#null) which represents omission. So something like 
`monitor_priority = try(local.tests[count.index].priority, null)` should work.

Hopefully that helped, I'm going to close this PR and the associated issue. If you run into more trouble feel free to re-open the issue. Thanks!

Issue #1533, opened 2022-08-18
new_group_delay doesn't override new_host_delay


hevcy:
### Terraform Version

1.1.7

### Affected Resource(s)

datadog_monitor

### Expected Behavior

A terraform plan is carried out with no changes required.

### Actual Behavior

Terraform plan produces the following:

~ resource "datadog_monitor" "monitor" {
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;      id                   = "12345"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;       name                 = "somemonitor"
&nbsp;&nbsp;&nbsp;&nbsp;     + new_host_delay       = 300
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;      tags                 = []
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;       # (15 unchanged attributes hidden)
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;      # (1 unchanged block hidden)
  }

### Steps to Reproduce

Define a monitor in terraform such as:

`resource "datadog_monitor" "monitor" {
  ...
  
  new_group_delay = 60
}`

Run a plan.

### Important Factoids

The monitor was created manually before being imported into terraform.

amanda-wee:
I am facing the same issue. If I specify `new_group_delay = 120` only, I end up with a plan that includes:
```
+ new_group_delay     = 120
+ new_host_delay      = 300
```
If I try to specify both `new_group_delay = 120` and `new_host_delay = 0`, I get the warning:
```
Argument is deprecated
(...)
Use `new_group_delay` except when setting `new_host_delay` to zero.
```
and also the error:
```
400 Bad Request: {"errors": ["new_host_delay must not be set when new_group_delay is used"]}
```
My terraform version is 1.2.7

fadimilio:
We are facing the same issue right now, which also breaks our "tfmigrate" state import step in our pipeline, since "tfmigrate" is detecting a change in the imported monitor.

Has anyone found a solution for this issue?

roy-work:
I gave #1626 a quick test today … and it seems like it works? (It at least causes `plan` to converge.) Was there a reason it was closed?

darkn3rd:
[...]

PR #1626, opened 2022-11-01
[datadog_monitor] Properly handle `new_host_delay` when `new_group_delay` is set

skarimo:
Since `new_group_delay` overrides `new_host_delay`, do not set default for the latter and don't include its value in the payload.

Closes: https://github.com/DataDog/terraform-provider-datadog/issues/1533


darkn3rd:
This issue is still happening on `DataDog/datadog` version `3.39.0`.   

I imported a monitor resource that has `new_group_delay=60`, and new_host_delay not even set. But when I do a terraform plan, it shows that it wants to force a `new_host_delay=300`.

PR #1629, opened 2022-11-02
[Synthetics] Add body_type field to SyntheticsTest request_definition

nkzou:
In addition, this PR updates the Go client dependency, and converts all tests'`restricted_roles` fields to use uuids instead of role names


AlaricCalmette:
This won't work without a pr in https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/synthetics.yaml to add body type. I'm working on this ATM. After the PR on api-spec, we will need to update the go client used by the TF provider.

nkzou:
I have a PR open on api-spec with the spec addition https://github.com/DataDog/datadog-api-spec/pull/1922/, asked for a review on it last week. Do you want to supersede it with your own PR?

maycmlee (dismissed):
👍 for docs

skarimo (approved):

PR #1640, opened 2022-11-14
[datadog_security_monitoring_rule] Updating schema validation for field type

symphony-elias:
We'd like to explicitly remove support for `infrastructure_configuration` and `cloud_configuration` rules as their management with TF is broken. See [RFC](https://datadoghq.atlassian.net/wiki/spaces/SA/pages/2697823566/CRUD+API+for+custom+CSPM+rules) for more details.


maycmlee (approved):
👍 for docs.

roughbits01 (approved):

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -338,11 +338,12 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
 		},
 
 		"type": {
-			Type:             schema.TypeString,
-			ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewSecurityMonitoringRuleTypeReadFromValue, datadogV2.NewSecurityMonitoringSignalRuleTypeFromValue),
-			Optional:         true,
-			Description:      "The rule type.",
-			Default:          "log_detection",
+			Type: schema.TypeString,
+			ValidateDiagFunc: validators.ValidateStringEnumValue(datadogV2.SECURITYMONITORINGRULETYPEREAD_LOG_DETECTION,
```
symphony-elias, line 342: I didn't reuse OOTB validator provided by the [TF SDK](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/validation#StringInSlice) because:
* documentation is generated by checking the output of the ValidateDiagFunc when passing an `EnumChecker`
*

symphony-elias: In details: `validation.StringInSlice` takes an array of strings. However, the `datadogV2` constants have a specific type `type SecurityMonitoringRuleTypeRead string` which means we'll have to create same constants of type string or change the type of these constants.
When using the validator, it doesn't update the doc

therve (approved):

Issue #1651, opened 2022-12-02
[datadog_service_definition_yaml] Flaky recreates because of case sensitive service name


marceloboeira:
### Terraform Version

```
Terraform v1.3.5
on darwin_amd64
```

### Affected Resource(s)

Please list the resources as a list, for example:

-   datadog_service_definition_yaml

### Terraform Configuration Files

```hcl
resource "datadog_service_definition_yaml" "foo" {
  service_definition = <<-EOT
  schema-version: v2
  dd-service: Foo
  integrations:
    pagerduty: https://XXX.pagerduty.com/service-directory/XXX
  tags:
  - managed-by:terraform
  team: data-platform
EOT
}
```

### Debug Output

Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Expected Behavior

What should have happened?

### Actual Behavior

```diff
-/+ resource "datadog_service_definition_yaml" "foo" {
      ~ id                 = "foo" -> (known after apply)
      ~ service_definition = <<-EOT # forces replacement
          - dd-service: foo
          + dd-service: Foo
            integrations:
                pagerduty: https://XXX.pagerduty.com/service-directory/XXX
            schema-version: v2
            tags:
                - managed-by:terraform
            team: data-platform
        EOT
    }
```

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:

1. Use the mentioned code
1. `terraform apply`
1.  Resources get created
1. `terraform plan`
1. See the diff with recreation of the service in the catalog

### Important Factoids

* We can work around this in the provider by comparing lowercased versions of the definition and avoid replacing the resource every single time; however, if the API doesn't handle those properly, then it will still be a problem (since case changes won't take effect). The question remains for the datadog folks, is the service name case-sensitive? does it force lower case?

nkzou:
We've looked into it and found that the field conforms to Datadog's universal service tagging rules, found at https://docs.datadoghq.com/getting_started/tagging/#define-tags. This means it not only converts to lowercase, but replaces most special characters as well. We're currently discussing how best to handle/document this - thanks for bringing this to our attention.

PR #1652, opened 2022-12-02
[datadog_service_definition_yaml] Prevent recreating resources on casing-only differences

marceloboeira:
In-depth: https://github.com/DataDog/terraform-provider-datadog/issues/1651

It seems like the API is lower-casing service names, so when your
terraform code has a service called, for instance, "Foo", it would
accept the name, but in the backend convert it to "foo". That causes
problems with terraform, which stored foo locally, and then it tries to
recreate the entire resource over and over:

```diff
-/+ resource "datadog_service_definition_yaml" "foo" {
      ~ id                 = "foo" -> (known after apply)
      ~ service_definition = <<-EOT # forces replacement
          - dd-service: foo
          + dd-service: Foo
            integrations:
                pagerduty: https://XXX.pagerduty.com/service-directory/XXX
            schema-version: v2
            tags:
                - managed-by:terraform
            team: data-platform
        EOT
    }
```

This commit introduces a change to verify if a resource should be
targeted to "replacement" only if the old and new service names are
different in content.

Additionally, it also stores lowercased dd-service values on terraform
state so that it guarantees this won't be showing in all diffs for every
single terraform plan since the file definitions will be consistently
different than the remote state, so it prevents this:

```diff
  ~ resource "datadog_service_definition_yaml" "foo" {
        id                 = "foo"
      ~ service_definition = <<-EOT
          - dd-service: foo
          + dd-service: Foo
            integrations:
                pagerduty: https://XXX.pagerduty.com/service-directory/XXX
            schema-version: v2
            tags:
                - managed-by:terraform
            team: data-platform
        EOT
    }
```

### DIFF
--- a/datadog/resource_datadog_service_definition_yaml.go
+++ b/datadog/resource_datadog_service_definition_yaml.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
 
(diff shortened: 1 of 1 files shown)


marceloboeira:
I'm wondering, since this will still show up in the diff, maybe it's worth writing a custom diff `ValidateChange` and remove the service diff.

nkzou:
Hi, thanks for you contribution - I'm continuing this issue in https://github.com/DataDog/terraform-provider-datadog/pull/1686 since the normalization is a bit more complicated than just lowercasing and also needs to be done in the SLO resource.

PR #1664, opened 2022-12-06
Update downtime.md

falkorichter:
improved documentation based on experience with `timezone` parameter


nkzou:
Hi there, thanks for your contribution - the docs are autogenerated, so modifying the markdown directly will be overwritten. This also seems a bit too specific to put in the docs as is, but we'll look into improving the documentation to specify the format it expects. For now, I believe monitor downtimes use the IANA timezone database names, which you can find on the IANA website or wikipedia.

alai97 (commented):
No docs review needed.

PR #1696, opened 2023-01-03
[datadog_synthetics_test] Add secure field to synthetics config variable

### LINKED ISSUE #1585: Feature Request: Add support for datadog_synthetics_test secure config_variables
On DataDog, it is possible to set a secured local variable for synthetic tests, and hide its content.

"Hide and obfuscate variable value - The value of this variable will be obfuscated in test results."

### Affected Resource(s)

-   `datadog_synthetics_test`

### References

-   https://github.com/DataDog/terraform-provider-datadog/pull/758

nkzou:
SSIA, resolves #1585 

Note - since the backend no longer returns the `secure` variable, setting `secure = true` will result in drift if the variable pattern/example are updated elsewhere (via curl, API client, or web app).


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1075,6 +1075,11 @@ func syntheticsConfigVariable() *schema.Schema {
 					Type:        schema.TypeString,
 					Optional:    true,
 				},
+				"secure": {
+					Description: "Whether the value of this variable will be obfuscated in test results.",
+					Type:        schema.TypeBool,
+					Optional:    true,
```
skarimo, line 1081: Lets set default to false explicitly since we can't differentiate between zero values and unset values in tf.
```suggestion
					Optional:    true,
					Default:     false,
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1049,7 +1049,7 @@ func syntheticsConfigVariable() *schema.Schema {
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"example": {
-					Description: "Example for the variable.",
+					Description: "Example for the variable. When `secure = true`, this value cannot be returned by the backend. Avoid drift by only making updates to this value from within Terraform.",
```
skarimo, line 1052:
```suggestion
					Description: "Example for the variable. This value is not returned by the api when `secure = true`. Avoid drift by only making updates to this value from within Terraform.",
```

docs/resources/synthetics_test.md
```diff
@@ -631,9 +631,10 @@ Required:
 
 Optional:
 
-- `example` (String) Example for the variable.
+- `example` (String) Example for the variable. When `secure = true`, this value cannot be returned by the backend. Avoid drift by only making updates to this value from within Terraform.
 - `id` (String) When type = `global`, ID of the global variable to use.
-- `pattern` (String) Pattern of the variable.
+- `pattern` (String) Pattern of the variable. When `secure = true`, this value cannot be returned by the backend. Avoid drift by only making updates to this value from within Terraform
```
kayayarai, line 636:
```suggestion
- `pattern` (String) Pattern of the variable. When `secure = true`, this value cannot be returned by the backend. Avoid drift by only making updates to this value from within Terraform.
```

kayayarai (approved):
Minor edit, and 👍 from docs

romainberger (approved):
LGTM for Synthetics

skarimo (approved):

PR #1704, opened 2023-01-05
[datadog_service_level_objective] Update terraform-provider-datadog to allow primary timeframe, target, and warning to be specified

ddjamesfrullo:
See https://datadoghq.atlassian.net/browse/SLOS-1142 for details.


phillip-dd (dismissed):
codeowners changes lgtm

datadog/resource_datadog_service_level_objective.go
```diff
@@ -93,6 +93,24 @@ func resourceDatadogServiceLevelObjective() *schema.Resource {
 					},
 				},
 			},
+			"target_threshold": {
+				Description:      "The objective's target in`[0,100]`.",
```
mmeyer724, line 97:
```suggestion
				Description:      "The objective's target in `[0,100]`.",
```
one small typo here

ddjamesfrullo: Oh, thanks. Just pushed a fix.

mmeyer724 (approved):

datadog/resource_datadog_service_level_objective.go
zhengshizhao: 0 and 100 is not valid target value. 
slo target validation is here: https://github.com/DataDog/dogweb/blob/c5d73e5baa851aeaa77ec6188232e4930a397cc1/dogweb/model/service_level_objective.py#L3474-L3475

datadog/resource_datadog_service_level_objective.go
```diff
@@ -93,6 +93,24 @@ func resourceDatadogServiceLevelObjective() *schema.Resource {
 					},
 				},
 			},
+			"target_threshold": {
+				Description:      "The objective's target in `[0,100]`.",
```
zhengshizhao, line 97: `[0,100]` -> `(0,100)`

docs/resources/service_level_objective.md
```diff
@@ -82,6 +90,9 @@ resource "datadog_service_level_objective" "bar" {
 - `query` (Block List, Max: 1) The metric query of good / total events (see [below for nested schema](#nestedblock--query))
 - `tags` (Set of String) A list of tags to associate with your service level objective. This can help you categorize and filter service level objectives in the service level objectives page of the UI. Note: it's not currently possible to filter by these tags when querying via the API
 - `validate` (Boolean) Whether or not to validate the SLO.
+- `timeframe`(String) The time frame for the objective. The mapping from these types to the types found in the Datadog Web UI can be found in the Datadog API documentation page. Valid values are `7d`, `30d`, `90d`.
+- `target_threshold` (Number) The objective's target in`[0,100]`.
```
zhengshizhao, line 94: - in`[0,100]` -> in `(0,100)`
- should we also add it must be the same value as in target value in `thresholds`? 
- `target_threshold` and `timeframe` are added as `required` fields in the schema.  move them under `Required`?

docs/resources/service_level_objective.md
```diff
@@ -82,6 +90,9 @@ resource "datadog_service_level_objective" "bar" {
...
+- `warning_threshold` (Number) The objective's warning value in `[0,100]`. This must be greater than the target value.
```
zhengshizhao, line 96: `[0,100]` -> `(0,100)`

docs/resources/service_level_objective.md
ddjamesfrullo: You're right. I copied this verbatim from the existing documentation for timeframe/target/threshold. This all strikes me as duplicative because the API documents all this to begin with. About referring to `thresholds`, I was avoiding that because there's a long-term plan to remove that and I am being a little lazy. There's a lot I think is missing from the existing docs, but I view that as a separate problem. I'll change the interval notation, though. Right now, `timeframe` and `target_threshold` cannot be required. I agree they will be required in the future, but they are not now in this phase. Again, I'm being lazy about that, but also forward compatible, by avoiding that discussion at this point.

ddjamesfrullo: Oh, I just realized what you mean - they are labeled as required. That is a mistake I need to fix.

zhengshizhao: oh yes, that's what I meant. sorry my comment wasn't that clear.

zhengshizhao: for the second point referring to `thresholds`, to me, it is also ok to not mentioning here but we need to avoid any configuration that can silently fail the slo behavior. it's painful for both sides to fix the misconfiguration later. (well, we can avoid this by adding more restrict validation in api level and return error messages)

ddjamesfrullo: Yeah, I'm trying to avoid documentation tech debt, basically. But I agree we do need the API to prevent getting this wrong. I already do have a ticket to fix the consistency problem. I think you're right that we should fix that before this goes out.

ddjamesfrullo:
Made this a draft for now while I finish my testing. I want to make sure this works as expected because I found and fixed an issue that I nearly missed.

maycmlee (dismissed):
👍 for docs.

nkzou (approved):

nkzou (approved):

ocdrums3:
Looks like these feature got released last week. Is there any documentation showing how to set the primary timeframe?

nkzou:
@ocdrums3 There are docs [here](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/service_level_objective) if you're looking for example terraform configs that use this feature.

ocdrums3:
I must be blind. I don't see anything about setting a primary timeframe in that doc.

nkzou:
Ah, the name of the fields aren't obvious, sorry about that. The primary timeframe/target/warning fields are the respective fields directly under the resource definition, for example:
```
resource "datadog_service_level_objective" "foo" {
  name        = "Example Metric SLO"
  type        = "metric"
  description = "My custom metric SLO"
  query {
    numerator   = "sum:my.custom.count.metric{type:good_events}.as_count()"
    denominator = "sum:my.custom.count.metric{*}.as_count()"
  }

  thresholds {
    timeframe = "7d"
    target    = 99.9
    warning   = 99.99
  }

  thresholds {
    timeframe = "30d"
    target    = 99.9
    warning   = 99.99
  }

  timeframe         = "30d"   # primary timeframe
  target_threshold  = 99.9    # must be the same as the matching timeframe's thresholds above
  warning_threshold = 99.99   # ^
}
```
Will update the docs to make that clear.

PR #1707, opened 2023-01-09
[security_monitoring_default_rule] Fix acceptance tests for default rules

muffix:
The default rule acceptance tests weren't actually testing much previously. The tests did apply the terraform, but did not make any assertions/checks on them.

These checks would have failed because the existing tests are importing an infrastructure configuration rule and attempt to set the "decrease criticality" flag. This, however, is only supported for log detection rules.

This PR changes the acceptance tests to make sure we run the configuration on rule of a supported type and adds a check. It also fixes a bug where the "decrease criticality" flag was returned in the terraform response for rule types for which the flag isn't supported.


datadog/tests/resource_datadog_security_monitoring_default_rule_test.go
```diff
@@ -18,20 +18,22 @@ func TestAccDatadogSecurityMonitoringDefaultRule_Basic(t *testing.T) {
...
 			{
-				Config:            testAccCheckDatadogSecurityMonitoringDefaultNoop(),
-				ResourceName:      tfSecurityDefaultRuleName,
-				ImportState:       true,
-				ImportStateIdFunc: idFromDatasource,
+				Config:             testAccCheckDatadogSecurityMonitoringDefaultNoop(),
+				ResourceName:       tfSecurityDefaultRuleName,
+				ImportState:        true,
+				ImportStateIdFunc:  idFromDatasource,
+				ImportStatePersist: true,
```
muffix, line 31: We need to persist the state so that we can access it in the next step. No checks are run if we're importing a state in a test step.

datadog/tests/resource_datadog_security_monitoring_default_rule_test.go
```diff
@@ -43,26 +45,28 @@ func idFromDatasource(state *terraform.State) (string, error) {
 	return resourceState.Primary.Attributes["rule_ids.0"], nil
 }
 
-func testAccCheckDatadogSecurityMonitoringDefaultDatasource() string {
+func testAccDatadogSecurityMonitoringDefaultDatasource() string {
 	return `
 data "datadog_security_monitoring_rules" "bruteforce" {
-    name_filter = "docker"
+	tags_filter = ["source:cloudtrail"]
```
muffix, line 51: This makes sure we're importing a default rule of type log detection. CloudTrail rules should all be of type log detection. The previous filter returned an infrastructure configuration rule.

pietrodll (approved):
Thanks 🙌

lli-datadog (approved):

nkzou (approved):

PR #1685, opened 2022-12-23
[datadog_service_account] Add service account resource

mnguyendatadog:
Adds the service_account resource, which is a subclass of the User resource.

addresses #1266

PR #1712, opened 2023-01-09
[datadog_application_key] Allow app key management of other users

jorhett:
The DataDog console allows creation of Application Keys for service accounts and users other than your own (given appropriate org permissions).  Terraform as a matter of course is going to be setting up users and their keys. This change should allow creation and management of application keys for other users.

## Working

* `terraform import` of keys shows the correct user
* `terraform state show` now shows the user which owns the key

## Help needed -- how to create key for other user

The problem I'm running into is that the API examples assume that the app key is always my own key. The v2 API only has `CreateCurrentUserApplicationKey` https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/api_key_management.go#L138

V1 API does have an `Owner` field in the key struct https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV1/model_application_key.go#L18 but trying to supply this in the body yields the following error:

`400 Bad Request: {"errors":["Additional properties are not allowed ('owner' was unexpected)"]}`


vjeon:
Hi, 
Creating an application key for a service account uses the API documented here: https://docs.datadoghq.com/api/latest/service-accounts/#create-an-application-key-for-this-service-account. 
Additionally, we do not support creating keys for other users/service accounts. Users can create application keys for service accounts as long as they have sufficient permissions and have to use the "Create an application key for a service account" endpoint to do so. Service account application keys can be modified or revoked using the general application key API by any user with sufficient permissions, but for the sake of consistency within the terraform provider, it probably makes the most sense to use the service account application key edit and delete endpoints. For all of this, support for the service account application key resource needs to be implemented in the Datadog Terraform provider.

jorhett:
> Creating an application key for a service account uses the API documented here: https://docs.datadoghq.com/api/latest/service-accounts/#create-an-application-key-for-this-service-account
>  Additionally, we do not support creating keys for other users/service accounts.

This is the core problem. How does one create a service account and give it an app key without having an app key for that service account?

> Users can create application keys for service accounts as long as they have sufficient permissions and have to use the "Create an application key for a service account" endpoint to do so. Service account application keys can be modified or revoked using the general application key API by any user with sufficient permissions

This seems to conflict with your previous statement.  Unless you are saying that users can create keys for service accounts but a service account can't create a key for another service account? 🤔 

To be clear on the use case here: the effect of this policy is that individuals (who can quit, be fired, move to different teams) can control the service accounts used by an organization... but an organization cannot give a team with SSO login the ability to manage organization keys?  Seems very 1989 management-style to me...

> support for the service account application key resource needs to be implemented in the Datadog Terraform provider.

It was merged in #1685

PR #1728, opened 2023-01-17
[datadog_security_monitoring_default_rule] Warn when the rule has a deprecation date

pietrodll:
# What does this PR do?

Show a warning when a default rule has a deprecation date defined. After the deprecation date, the API blocks rule updates, which can break Terraform pipelines for users. With this change, users are notified in advance so they can remove references to deprecated rules in they Terraform configuration.

# Update process

- Changed the code to return a warning message when the `deprecationDate` field of a default rule is set. Note that this property only exists for default rules, so the code for `datadog_security_monitoring_rule` was not updated.
- Duplicated the test for the default rule resource and recorded a cassette.
- Manually changed the cassette response to include a deprecation date, and changed the test to be skipped if the testing session is not a replay. Unfortunately, the Terraform testing framework does not provide a way to assert on warnings. See issue https://github.com/hashicorp/terraform-plugin-sdk/issues/864.
- Debugged locally to check that the warning is generated as expected.

Note: this PR updates the API client to a non-pinned version.


clementgbcn (approved):
Ok 👌 for me, you will just have to update the docs by running the command `make docs`.

clementgbcn (approved):
👍

nkzou (approved):

estherk15 (approved):

PR #1730, opened 2023-01-20
Terraform fails on capitalised provider source

richard-rvvup:
Terraform 1.3.0 fails to download the provider unless it is lowercase.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -21,7 +21,7 @@ Use the navigation to the left to read about the available resources.
 terraform {
   required_providers {
     datadog = {
-      source = "DataDog/datadog"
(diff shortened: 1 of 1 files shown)


nkzou:
Hi there, could you share some more details about your environment and exact error message? I tested this with terraform 1.3.0 and could not reproduce. `DataDog/datadog` and `datadog/datadog` both work for me.

richard-rvvup:
> Hi there, could you share some more details about your environment and exact error message? I tested this with terraform 1.3.0 and could not reproduce. `DataDog/datadog` and `datadog/datadog` both work for me.

Looks like this was a missing provider configuration on my end - apologies!

Issue #1729, opened 2023-01-20
Feature Request: Add option to toggle additionnal content in monitors notifications


laudibert:
Hello

### Affected Resource(s)

-   datadog_monitor

### Terraform Configuration Files

```hcl
resource "datadog_monitor" "this" {
  notification_preset_name = "hide_all"
}
```

### Expected Behavior

It would be nice to have the new parameter `notification_preset_name` as documented in [the Datadog documentation](https://docs.datadoghq.com/fr/monitors/notify/#toggle-additional-content) that allows to hide content (query and mentions) on the monitor notification messages.

The [API documentation](https://docs.datadoghq.com/api/latest/monitors/) seems limited as you can only find reference to this field in the examples for some reason.

Using the export feature on Datadog, possible values seem to be : `hide_query`, `hide_handles` and `hide_all`.

Thanks for all the hardwork.

PR #1744, opened 2023-01-30
[provider] Support monitor `notification_preset_name`

laudibert:
Implementing issue #1729 

I do not have access to a testing/sandbox Datadog environnement unfortunately so I could not run the acceptance tests.

PR #1749, opened 2023-02-01
[datadog_monitor] Add support for `notification_preset_name`

skarimo:
Supersedes https://github.com/DataDog/terraform-provider-datadog/pull/1744
Closes: https://github.com/DataDog/terraform-provider-datadog/issues/1729


hestonhoffman (dismissed):

docs/resources/monitor.md
```diff
@@ -67,6 +67,7 @@ For example, if the value is set to `300` (5min), the `timeframe` is set to `las
 - `no_data_timeframe` (Number) The number of minutes before a monitor will notify when data stops reporting. Provider defaults to 10 minutes.
 
 We recommend at least 2x the monitor timeframe for metric alerts or 2 minutes for service checks.
+- `notification_preset_name` (String) Toggles the display of additional content sent in the monitor notification. Valid values are: `show_all`, `hide_query`, `hide_handles`, and `hide_all`. Valid values are `show_all`, `hide_query`, `hide_handles`, `hide_all`.
```
nkzou, line 70: The enum already tells tfplugindocs about valid values, so we should remove the valid values from the description to avoid the duplication

datadog/data_source_datadog_monitor.go
```diff
@@ -265,6 +265,11 @@ func dataSourceDatadogMonitor() *schema.Resource {
 					},
 				},
 			},
+			"notification_preset_name": {
+				Description: "Toggles the display of additional content sent in the monitor notification. Valid values are: `show_all`, `hide_query`, `hide_handles`, and `hide_all`.",
+				Type:        schema.TypeString,
+				Computed:    true,
```
nkzou, line 271: Should the datasource schema for this also have the enum validation function? There's one field in the monitor datasource that uses it (renotify_statuses) but since its read-only might not matter too much.

skarimo: Need to be explicit here. Terraform has safeguards for having `Validate` funcs in data sources (at least for primitive types but does seems to ignore it in `TypeSets`)

nkzou (approved):

carlmartensen (approved):
LGTM 👍

PR #1770, opened 2023-02-09
[datadog_dashboards] Docs update for event_size field in dashboard list_stream_definition

nkzou:
Adds note that `event_size` is required when using a list stream with the `event_stream` data source


DrkSephy (approved):

rtrieu (approved):
LGTM

PR #1788, opened 2023-02-22
[datadog_ip_allowlist] Support IP allowlist management in terraform

retsguj:
https://datadoghq.atlassian.net/browse/ACCESS-1400

Some notes:
- ~~Not sure if I should also create a data source (instead of just a resource) for IP allowlist? I read some docs, but still am not understanding if it is necessary here...~~ no data source will be created
- Currently, if you delete an IP allowlist resource, the allowlist gets disabled and all entries (if any) are removed. In some ways, though, it makes more sense to me to not do anything upon deletion. This is not a usual resource like roles; there isn't some obvious notion of what it means to delete the IP allowlist. Would appreciate other thoughts on this.
- ~~If you specify in an entry `cidr_block = "127.0.0.1"`, run `terraform apply`, then (without changing anything in the resource) run `terraform plan`, Terraform will indicate that you are removing the entry with `cidr_block = "127.0.0.1"` and adding an entry with `cidr_block = "127.0.0.1/32"`. I'm not sure yet if there's a way to not trigger a diff in this situation. 
UPDATE: I have added a validator to the `cidr_block` field to require a CIDR block rather than either a CIDR block or IP address. It's limiting, but I haven't been able to come up with a better solution.~~
I used a combination of the diff suppress options and a hashing of set elements to force Terraform to treat an IP address and its CIDR block version as equivalent.
- Like all other terraform resources, you must give the IP allowlist resource a name. This is a little misleading, because there is only one IP allowlist per org, and every person in an org using Terraform to manage it will be modifying the one allowlist, even if they give the resources different names. AFAIK there is no way around this, though.
- ~~Haven't written tests yet, am waiting until we finalize all behavior.~~


srosenthal-dd:
On datasource vs resource:

datasources are useful when you want to build on top of something. For example, you can use a datadog_role datasource to look up a role ID from its name and use it in a restriction policy.

I don't expect anyone is building other resources on top of a Datadog IP allowlist so we should be OK to only implement a resource, no datasource.

srosenthal-dd:
Sorry for the extended delay!

Testing following the instructions in https://github.com/DataDog/terraform-provider-datadog/blob/master/README.md...

Comments:
1. While most users will use IPv4, we also support IPv6 in the UI & API. Looks like the validation code only allows for IPv4. If necessary we could probably live with the normalization behavior you described. It seems to be a known issue for Terraform, ref: https://github.com/hashicorp/terraform-plugin-framework/issues/70.
2. If I provide an entry without a note I'd expect it to default to empty string, but I get `Error: Missing required argument`
3. Now that I see how it works, I'm somewhat regretting the decision to only expose the GET & PATCH (batch) APIs. If we instead had multiple `ip_allowlist_entry` resources, then Terraform plan/apply would report helpful numbers like "10 to add, 1 to delete", but as-is, it's always 1 resource being updated. Let's discuss whether it's worth changing the pattern at all.

So I don't lose it, my test resource was variation on this:

```
resource "datadog_ip_allowlist" "ip_allowlist" {
  enabled = true
  entry {
    cidr_block = "1.2.3.4/32"
    note = "fake IP #1"
  }
  entry {
    cidr_block = "2.4.6.8/32"
    note = "fake IP #2"
  }
  entry {
    cidr_block = "10.0.0.0/8"
    note = "some private IPs"
  }
  entry {
    cidr_block = "76.144.70.0/24"
    note = "IPv4 range"
  }
  entry {
    cidr_block = "2600:100f:b112:56f:8aa:d78e:fd9d:ddb6/128"
    note = "IPv6"
  }
  entry {
    cidr_block = "2600:100f:b112:56f:8aa:d78e:fd9d:0/112"
    note = "IPv6 range"
  }
}
```

srosenthal-dd:
On the problem of IP/CIDR normalization, I thought a bit more:

1. there does seem to be some precedent for normalizing the inputs before sending the request (example: https://github.com/hashicorp/terraform-provider-azurerm/pull/5459/files)
2. Go seems to have good built-in support for all sorts of networking operations. The Azure code linked above uses net.ParseIP.

retsguj:
@srosenthal-dd

> 1. While most users will use IPv4, we also support IPv6 in the UI & API. Looks like the validation code only allows for IPv4. If necessary we could probably live with the normalization behavior you described. It seems to be a known issue for Terraform, ref: [Classifying normalization vs. drift hashicorp/terraform-plugin-framework#70](https://github.com/hashicorp/terraform-plugin-framework/issues/70).

I believe I have now fixed this problem.

> 2. If I provide an entry without a note I'd expect it to default to empty string, but I get `Error: Missing required argument`

I changed the configuration so that a note is optional.

srosenthal-dd:
I re-tested with scenarios:

1. ✅ apply from scratch -> works
2. ✅ enable without current IP in list -> blocked
3. ✅ single entry add/remove -> diff looks good
4. 🆗 modify single entry -> diff shows a delete/add rather than a modification
5. ✅ with or without note
6. ✅ mix of IPv4, IPv6, single entries & ranges
7. 🆗 validation doesn't perfectly match what's done on the server side (ex: `1.2.3.4/8` is OK according to Go but the server API wants it to be `1.0.0.0/8`)
8. Wishlist item: would be nice if the `N to add, M to change, P to destroy` counted entries instead of lists.

To me, it's definitely good enough to ship. We can revisit some of the details in a later iteration if we want, especially if we get feedback from users. For now it'll definitely be useful!

srosenthal-dd (approved):

nkzou (approved):

nkzou (approved):

PR #1797, opened 2023-03-03
Downtime duration

gdlx:
This adds a `duration` downtime parameter allowing to set recurring downtimes the same way Datadog UI does, without having to set an absolute end timestamp:
image

I've seen no reason to restrict `duration` usage to recurring downtimes, so it can actually be used in any case instead of `end` or `end_date`.

I'm also working on a similar change for the start time, the goal being to get rid of the fake diffs on the downtime boundaries, but I splitted them as the duration is easier and done, while start handling is a bit more difficult and I still hesitate on different options.


rtrieu (approved):
LGTM

skarimo:
@gdlx Thanks for creating these PR's. I am hesitant to introduce new terraform specific fields to this resource as it is a bit complicated already.

I would recommend using the [time](https://registry.terraform.io/providers/hashicorp/time/latest/docs) resources by Hashicorp for this instead and adding documentation around the usage of it. For example. something like:
```
resource "time_static" "start" {
  rfc3339 = "2023-03-30T17:50:31Z"
}

resource "time_offset" "end" {
  base_rfc3339 = time_static.start.id
  offset_minutes = 120
}

resource "datadog_downtime" "foo" {
  scope      = ["*"]
  start      = time_static.start.unix
  end        = time_offset.end.unix

  recurrence {
    type   = "days"
    period = 1
  }
}
``` 

The above should be able to handle the functionally that would be introduced by both `beginning_at` and `duration`. What are your thoughts on this?

therve:
> Indeed, we have more than 150 recurrent downtime to implement and this solution would make us create as many "time_static" and "time_offset" resources as downtime.
> This is not viable on a large scale from our window.

Hi, can you elaborate on that point? time resources have basically no cost except a few lines of HCL. Maybe we could document a module encapsulating this logic? Thanks.

gdlx:
Hi @skarimo, any update on this ?

therve:
With downtime v2 being soon released we'll rely on its behavior. Thanks for your contribution.

gdlx:
Thanks Thomas,

Is the v2 ETA a matter of weeks/months... ?

PR #1798, opened 2023-03-03
[datadog_downtime] Add beginning_at

gdlx:
**Merge after #1797**

This adds a `beginning_at` downtime parameter, allowing to set the start time without having to set an absolute date.
The expected format is the time part of a RFC3339 timestamp (after the T), eg. `10:00Z`, `10:00:00+01:00`...
The provider automatically adds the current or next day as a date (next day if the time as already passed on the current day).

Just as #1797, the goal is to avoid fake diffs on recurring downtimes.

This is why this PR is made from the #1797 PR branch: there is no actual need to use `beginning_at` with an absolute end date, but I kept them split because this one isn't as clean as the duration: the time formatting is a little hacky, and I hesitated between this solution, and a natural time parsing when we could use strings like `next monday at 10:00`.

It would have covered a few more cases, like when you set a recurring downtime but you want to delay its first start, but it required young or ~dead packages and it may be harder for users to anticipate what will actually be set.

This is close to the UI behavior except there is no "date only" parameter because we can just use existing  `start` and `start_date` if we want to set a specific date.
image


rtrieu (approved):
LGTM

thepnux:
Hi,
We are looking forward to the passage of this PR.
Is it possible to take it over please ? 
thank's

bashao:
Also looking forward for this one

therve:
With downtime v2 being soon released we'll rely on its behavior. Thanks for your contribution.

PR #1811, opened 2023-03-10
[datadog_downtime] Ignore start/end comparison on recurring downtimes

therve:
We compare with the times of the children, so there is going to be a change after every recurrence. Let's ignore it instead, at the expense of losing UI drifts. This was the intent of the original recurrence support but wasn't done.

### DIFF
--- a/datadog/resource_datadog_downtime.go
+++ b/datadog/resource_datadog_downtime.go
@@ -311,11 +311,17 @@ func downtimeBoundaryNeedsApply(d *schema.ResourceData, tsFrom string, apiTs, co
 
 	if updating {
 		// when updating, we apply when
-		// * API-returned value is different than configured value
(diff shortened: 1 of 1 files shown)


skarimo (approved):

phillip-dd (approved):
Looks good to me. What I did to test:

With a terraform resource of:
```
resource "datadog_downtime" "recurring_downtime" {
  scope        = ["*"]
  start        = 1678479627
  end          = 1678479927
  timezone     = "EST"
  monitor_tags = ["test:foobarbaz"]
  message      = "terraform recurring monitor"

  recurrence {
    type   = "days"
    period = 1
  }
}
```
terraform apply, wait for the downtime to expire, then update the message. Do another terraform apply and the child message is updated.

PR #1813, opened 2023-03-13
[datadog_synthetics_test] Allow empty values for service for Synthetics grpc tests

romainberger:
Synthetics grpc tests allow empty `service` value, but this was not possible with the provider, this PR fixes the issue.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1424,8 +1424,12 @@ func buildSyntheticsAPITestStruct(d *schema.ResourceData) *datadogV1.SyntheticsA
 	if attr, ok := k.GetOkWith("call_type"); ok {
 		request.SetCallType(datadogV1.SyntheticsTestCallType(attr.(string)))
 	}
-	if attr, ok := k.GetOkWith("service"); ok {
-		request.SetService(attr.(string))
+	if syntheticsTest.GetSubtype() == "grpc" {
```
romainberger, line 1427: The `service` field is only valid for `grpc` tests so I manually added a default value here instead of inside the schema directly, to prevent errors with other subtypes.

luissmg (approved):

nkzou (approved):

PR #1820, opened 2023-03-17
[datadog_sensitive_data_scanner_group_order] Add resource and datasource for SDS group order

nkzou:
Adds a resource and datasource for configuring the order of SDS groups.


artslidd (commented):
Looks good to me, but I am not comfortable enough in this repo to approve the PR

PR #1833, opened 2023-03-24
[datadog_service_definition_yaml] Add support for v2.1 schema of service catalog

hyperloglogy:


GalenPickard (approved):

taltcher:
For which terraform version would it be available? is it part of 3.23.0 already?

nkzou:
It will be part of 3.24.0, which is unreleased and planned for next week.

PR #1847, opened 2023-04-03
[datadog_hosts] Add a host list datasource

nkzou:
Adds a data source for querying and filtering hosts tracked by Datadog's infrastructure list.

The datasource is capped at 1000 resources for now.


jeffreychai (approved):
This looks good from the REDAPL-Storage side 👍

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+
+var _ datasource.DataSourceWithConfigure = &hostsDataSource{}
+
+func NewHostsDataSource() datasource.DataSource {
+	return &hostsDataSource{}
+}
+
+type HostListMetadataModel struct {
+	AgentVersion    types.String `tfsdk:"agent_version"`
+	CPU_cores       types.Int64  `tfsdk:"cpu_cores"`
```
skarimo, line 26:
```suggestion
	CPUCores       types.Int64  `tfsdk:"cpu_cores"`
```

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+	Gohai           types.String `tfsdk:"gohai"`
+	Machine         types.String `tfsdk:"machine"`
+	Platform        types.String `tfsdk:"platform"`
+	Processor       types.String `tfsdk:"processor"`
+	PythonV         types.String `tfsdk:"python_version"`
+	Socket_FQDN     types.String `tfsdk:"socket_fqdn"`
```
skarimo, line 32:
```suggestion
	SocketFQDN     types.String `tfsdk:"socket_fqdn"`
```

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+
+type HostsDataSourceModel struct {
+	ID types.String `tfsdk:"id"`
+	// Query Parameters
+	Filter                types.String `tfsdk:"filter"`
+	SortField             types.String `tfsdk:"sort_field"`
+	SortDir               types.String `tfsdk:"sort_dir"`
+	From                  types.Int64  `tfsdk:"from"`
+	IncludeMutedHostsData types.Bool   `tfsdk:"include_muted_hosts_data"`
+	IncludeHostsMetadata  types.Bool   `tfsdk:"include_hosts_metadata"`
```
skarimo, line 66: What new attributes does this return? Is it worth exposing this or is it better to just have it always be true?

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+func (d *hostsDataSource) Configure(_ context.Context, request datasource.ConfigureRequest, response *datasource.ConfigureResponse) {
+	if request.ProviderData == nil {
+		return
+	}
+
+	providerData, ok := request.ProviderData.(*FrameworkProvider)
+	if !ok {
+		response.Diagnostics.AddError(
+			"Unexpected Data Source Configure Type",
+			"")
```
skarimo, line 87:
```suggestion
		response.Diagnostics.AddError("Unexpected Data Source Configure Type", "")
```

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+		parameters.WithFrom(state.From.ValueInt64())
+	}
+
+	ddHostListResponse, _, err := d.Api.ListHosts(d.Auth, parameters)
+	if err != nil {
+		resp.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error getting Hosts"))
+		return
+	}
+
+	state.ID = types.StringValue("datadog-hosts")
```
skarimo, line 212: Lets use the hash of all the filters combined instead for the ID.

datadog/fwprovider/data_source_datadog_hosts.go
```diff
@@ -0,0 +1,273 @@
...
+				Socket_Hostname: basetypes.NewStringValue(meta.GetSocketHostname()),
+			})
+		}
+		hostList = append(hostList, hostListEntry)
+	}
+	state.HostList = hostList
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
skarimo, line 271: Did you mean to append errors when building the host entries? Since no errors are being appended, it seems like this check doesn't do much

datadog/fwprovider/data_source_datadog_hosts.go
nkzou: It adds an object with the attributes in `HostListMetadataModel`. When its false, the `Meta` object's attributes are all zero-value. I don't think it would make a huge difference if we made it always true so thats definitely an option.

nkzou: Actually, since the ObjectType requires all fields to be populated I don't think there's much point in this toggle. I'll change it to be on by default, otherwise its just a bunch of empty strings and zero-values.

skarimo (dismissed):

datadog/fwprovider/data_source_datadog_hosts.go
nkzou: Oh duh. I don't think it makes sense to append errors for every single field in the host schema (a lot of them are sometimes missing depending on the specific host) so I'll just remove that error check and let those values be zero-value if the backend doesn't return it.

skarimo (approved):

Issue #1848, opened 2023-04-03
datadog_dashboard_json has permadrift


jsteinberg-rbi:
### Terraform Version

`1.3.7`

### Affected Resource(s)

-   datadog_dashboard_json

### Terraform Configuration Files

```hcl
resource "datadog_dashboard_json" "dashboard_json" {
  dashboard = <<EOF
  {
  "title": "delete-me",
  "description": "",
  "widgets": [],
  "template_variables": [],
  "layout_type": "ordered",
  "notify_list": [],
  "reflow_type": "fixed",
  "id": "3fa-nkp-wty"
}
EOF
}
```

### Debug Output

https://gist.github.com/jsteinberg-rbi/f043e8bd2d118d9936b9b36c664a45bc

### Expected Behavior

Dashboard plans without drift.

### Actual Behavior

Dashboard plans with drift every time.

### Steps to Reproduce

1. create a new dashboard. doesn't matter if it has anything in it, just create it.
2. export the dashboard as json.
3. put it in:
```
resource "datadog_dashboard_json" "dashboard_json" {
  dashboard = <<EOF
  ...
EOF
}
```
4. apply
5. plan again
6. output will be like: 
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # datadog_dashboard_json.dashboard_json will be updated in-place
  ~ resource "datadog_dashboard_json" "dashboard_json" {
      ~ dashboard               = jsonencode(
          ~ {
              - is_read_only       = false -> null
                # (7 unchanged elements hidden)
            }
        )
        id                      = "snf-7pg-55a"
        # (2 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

despite no changes. I've tried adding different values for `is_read_only` but the drift is always there. Of course the datadog API says this value is deprecated: https://docs.datadoghq.com/api/latest/dashboards/.

therve:
Hi,

As a stopgap if you had `"is_read_only": false` it should remove the drift. We'll look into fix this, sorry for the trouble.

jsteinberg-rbi:
> Hi,
> 
> As a stopgap if you had `"is_read_only": false` it should remove the drift. We'll look into fix this, sorry for the trouble.

like I said in my initial comments: I had already all possible values of `"is_read_only": false`

therve:
[...]

PR #1854, opened 2023-04-04
[datadog_dashboard_json] Handle continuous diff if `is_read_only` is not set

skarimo:
Closes: https://github.com/DataDog/terraform-provider-datadog/issues/1848


nkzou (approved):

therve:
Can we double check with the dashboard team? Given that this field is deprecated it'd be nice not to send it by default.

matt-miller-ddog (approved):

nevir:
Looks like this causes a panic in some cases when refreshing remote state: #1897

PR #1861, opened 2023-04-12
[datadog_synthetics_test] Refactor options for Synthetics tests

romainberger:
This PR puts in common the code used to set the options on Synthetics tests as well as the building of the local state for options. Currently this is done separately for api and browser tests, and this has led to several issues in the past where an option was added somewhere but forgotten in the other type.

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -2481,6 +2481,124 @@ func buildLocalExtractedValues(extractedValues []datadogV1.SyntheticsParsingOpti
 	return localExtractedValues
 }
 
+func buildLocalOptions(actualOptions datadogV1.SyntheticsTestOptions) []map[string]interface{} {
+	localOptionsList := make(map[string]interface{})
+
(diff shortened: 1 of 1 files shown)


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1933,6 +1823,163 @@ func buildAssertions(attr []interface{}) []datadogV1.SyntheticsAssertion {
...
+			ciOptions.SetExecutionRule(datadogV1.SyntheticsTestExecutionRule(testCiOptions["execution_rule"].(string)))
+
+			options.SetCi(ciOptions)
+		}
+
+		if ignoreServerCertificateError, ok := d.GetOk("options_list.0.ignore_server_certificate_error"); ok {
+			options.SetIgnoreServerCertificateError(ignoreServerCertificateError.(bool))
+		}
+
+		// browser tests specific options
```
romainberger, line 1931: I chose not to check the test type here and to blindly set any options put in the configuration. Even though some options are specific to some test types, the server will not raise any error if some invalid options are used.

nkzou (approved):

PR #1879, opened 2023-04-19
[datadog_service_level_objective] Fix edge case removing fields that are both optional and computed

### LINKED ISSUE #1877: Cannot remove the warning value from a primary threshold on a service level correction
### Terraform Version

1.1.3

### Affected Resource(s)

-   datadog_service_level_objective

Original:

```hcl
resource "datadog_service_level_objective" "agent_desktop_availability" {
  name        = "Agent Desktop Availability"
  type        = "monitor"
  monitor_ids = [... omitted]
  thresholds {
    timeframe = "30d"
    target    = "99.9"
    warning   = "99.95"
  }
  timeframe = "30d"
  target_threshold = "99.9"
  warning_threshold = "99.95"
  tags = local.datadog_tags
}
```

Updated - note the goal is to increase the target to 99.9%, rendering a warning impossible. So it is removed.
```hcl
resource "datadog_service_level_objective" "agent_desktop_availability" {
  name        = "Agent Desktop Availability"
  type        = "monitor"
  monitor_ids = [... omitted]
  thresholds {
    timeframe = "30d"
    target    = "99.99"
    warning = null
  }
  timeframe = "30d"
  target_threshold = "99.99"
  warning_threshold = null
  tags = local.datadog_tags
}
```

### Debug Output

plan:
Note the absence of the `warning_threshold`, even though it was explicitly set to `null` in the script.

```
        ]
      ~ target_threshold  = 99.9 -> 99.99
      ~ thresholds {
          ~ target          = 99.9 -> 99.99
          - warning         = 99.99 -> null
            # (3 unchanged attributes hidden)
        }
        # (1 unchanged block hidden)
    }
```

### Expected Behavior

The primary threshold should have been updated to a 99.9% target, and the warning removed.

### Actual Behavior

```
│ Error: error updating service level objective from https://api.datadoghq.com/api/v1/slo/6f077133bca056a590f899a92bf53107: 400 Bad Request: {"errors":["Invalid payload: warning_threshold (99.95) and warning (undefined) for the same timeframe (30d) in the threshold list must match"]}
│ 
│   with module.****.datadog_service_level_objective.all["*****_availability"],
│   on .terraform/modules/*****/datadog_slos.tf line 1, in resource "datadog_service_level_objective" "all":
│    1: resource "datadog_service_level_objective" "all" {
```

### Steps to Reproduce

1. `terraform plan`
2. `terraform apply`

### Important Factoids

Running the terraform script in harness.io

The scripting above is paraphrasing what's actually being used. A TF module is doing the job of creating all DataDog resources at runtime, based on descriptors held in seprate YAML files. It seems easiest to express the issue in terms of the raw TF resources being created by that module.

### References

Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

-   [Unable to update target and warning values on SLO #1849](https://github.com/DataDog/terraform-provider-datadog/issues/1849)

nkzou:
Uses the sdk's experimental `GetRawConfig()` in order to check if the user is trying to remove a field.
In this case, because the field is optional and computed, removing the field in the config doesn't actually remove it in state - the terraform sdk keeps it around assuming it is computed. This causes issues if other fields conflict with the removed field, and there's no way to tell that the field is supposed to be removed (without checking the raw config) because the sdk reports the field as having not changed.

Resolves #1877 

Todo: check other resources for this interaction, good chance this exists elsewhere as well.

### DIFF
--- a/datadog/resource_datadog_service_level_objective.go
+++ b/datadog/resource_datadog_service_level_objective.go
@@ -185,8 +185,6 @@ func resourceDatadogServiceLevelObjective() *schema.Resource {
 
 // ValidateServiceLevelObjectiveTypeString is a ValidateFunc that ensures the SLO is of one of the supported types
 
-// ValidateServiceLevelObjectiveTypeString is a ValidateFunc that ensures the SLO is of one of the supported types
-
 // Use CustomizeDiff to do monitor validation
 func resourceDatadogServiceLevelObjectiveCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
 	if validate, ok := diff.GetOkExists("validate"); ok && !validate.(bool) {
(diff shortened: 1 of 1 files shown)


therve (approved):
Nice!

PR #1904, opened 2023-05-04
[datadog_integration_gcp_sts] Create a GCP STS Terraform Resource

joshhuie:
This pull request includes the Go code changes needed to enable a GCP STS Integration Resource. This Terraform resource sets up a GCP STS Integration within Datadog using a customer’s GCP service account.

Included in this PR are the Create, Read, Update, Delete, and Import operations used to create a standard Terraform resource. 

This PR has a few dependencies that aren’t released yet (in the `datadog-api-client-go` and the `datadog-api-spec` repos).

### DIFF
--- a/datadog/fwprovider/resource_datadog_integration_gcp_sts.go
+++ b/datadog/fwprovider/resource_datadog_integration_gcp_sts.go
@@ -24,7 +24,7 @@ var (
 // datadogIntegrationGCPSTSResource is the resource implementation.
 type datadogIntegrationGCPSTSResource struct {
 	Auth   context.Context
-	GcpApi *datadogV2.GCPIntegrationSTSApi
+	GcpApi *datadogV2.GCPIntegrationApi
 }
 
 // NewDatadogIntegrationGCPSTSResource is a helper function to simplify the provider implementation.
@@ -62,7 +62,7 @@ func (r *datadogIntegrationGCPSTSResource) Configure(_ context.Context, request 
 	}
 
 	r.Auth = providerData.Auth
-	r.GcpApi = providerData.DatadogApiInstances.GetGCPStsIntegrationApiV2()
+	r.GcpApi = providerData.DatadogApiInstances.GetGCPIntegrationApiV2()
 }
 
 // Schema defines the Terraform Resource configuration.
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -93,3 +93,5 @@ require (
 )
 
 go 1.19
+
+replace github.com/DataDog/datadog-api-client-go/v2 => /Users/josh.huie/go/src/github.com/datadog-api-client-go
```
joshhuie, line 97: This line is required for local testing: I will remove it once the spec PRs are merged.

graybjones (approved):
LGTM

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -0,0 +1,428 @@
...
+	// The section below handles optional fields in Terraform.
+	// If an optional field is not specified, then Terraform state stores a nil value.
+	// However, the Datadog GCP API always responds with a value for these optional fields (an empty list, a false boolean, etc).
+	// If these optional fields aren't specified within Terraform Resources, then these fields should always remain nil.
+	if state.HostFilters.IsNull() {
+		if hostFilterCount > 0 {
+			state.HostFilters = outputListValue
+		}
+	} else {
+		state.HostFilters = outputListValue
```
graybjones, line 232: Will outputListValue ever by an empty list? Will Terraform convert that to Null?

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -0,0 +1,211 @@
...
+func integrationGCPSTSAccountExistsHelper(auth context.Context, s *terraform.State, apiInstances *utils.ApiInstances) error {
+	for _, r := range s.RootModule().Resources {
+		if r.Type != "datadog_integration_gcp_sts" {
+			continue
+		}
+		id := r.Primary.ID
+
+		accounts, httpResp, err := apiInstances.GetGCPStsIntegrationApiV2().ListGCPSTSAccounts(auth)
+		if err != nil {
+			return utils.TranslateClientError(err, httpResp, "error retrieving STS enabled GCP service account")
```
graybjones, line 132: Should this error message reflect that getting the list failed?

graybjones (approved):
A couple of questions, but otherwise LGTM

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
joshhuie: Yes, outputListValue will be an empty list if the service account created in Datadog's backend was created without specifying host filters, or if the service account was created with host filters explicitly set to an empty list. 

Terraform won't convert the empty list to null. This isn't an issue, however a user will always see a change in a Terraform plan (from nil -> empty list) if they didn't choose to use the optional host filters field. So this logic section prevents this from showing up

PR #1922, opened 2023-05-18
[datadog_provider] Add resource wrapper and support adding enum values to docs

skarimo:


datadog/internal/fwutils/fw_enrich_schema.go
```diff
@@ -0,0 +1,68 @@
...
+			sep := ""
+			for i := 0; i < v.Len(); i++ {
+				if len(validValuesMsg) > 0 {
+					sep = ", "
+				}
+				validValuesMsg += fmt.Sprintf("%s`%v`", sep, v.Index(i).Interface())
+			}
+			description = fmt.Sprintf("%s Valid values are %s.", description, validValuesMsg)
+			break
+		}
```
nkzou, line 64:
```suggestion
		}
		if validators.Index(i).Elem().Type().Name() == "atLeastValidator" {
			atLeast := validators.Index(i).Elem().FieldByName("min")
			if atLeast.IsValid() {
				description = fmt.Sprintf("%s Value must be at least %d.", description, atLeast.Int())
				break
			}
		}
```
thoughts on adding more validators to this?

skarimo: Good idea, I will work on this in a separate pr. I would need to cross check what type of validation are included in the documentation for resources written using `terraform-plugin-sdk` and implement the same ones for framework.

nkzou (approved):

nkzou (approved):

PR #1936, opened 2023-05-26
[datadog_integration_gcp_sts] generate GCP STS account resource

skarimo:


docs/resources/integration_gcp_sts_account.md
```diff
@@ -0,0 +1,49 @@
...
+## Schema
+
+### Required
+
+- `client_email` (String) Your service account email address.
+
+### Optional
+
+- `automute` (Boolean) Silence monitors for expected GCE instance shutdowns.
+- `host_filters` (Set of String) Your Host Filters.
```
drichards-87, line 36:
```suggestion
- `host_filters` (Set of Strings) Your Host Filters.
```

drichards-87 (approved):
One small bit of feedback from Docs. PR is approved.

docs/resources/integration_gcp_sts_account.md
skarimo: Thanks for the feedback. This part of the documentation is auto generated and we cannot change it without changing the underlying package.

docs/resources/integration_gcp_sts_account.md
```diff
@@ -0,0 +1,50 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_integration_gcp_sts_account Resource - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Provides a Datadog IntegrationGcpStsAccount resource. This can be used to create and manage Datadog integrationgcpsts_account.
```
joshhuie, line 6: should this be `integration_gcp_sts_account`?

docs/resources/integration_gcp_sts_account.md
```diff
@@ -0,0 +1,50 @@
+---
```
joshhuie, line 1: Out of curiosity, is this the documentation that gets displayed on the Terraform website? If not, what controls that? 

Just wondering because I think it could be helpful to include a few examples of setting up this integration (which will also involve setting up some GCP service accounts and their config with Terraform)

datadog/fwprovider/resource_datadog_integration_gcp_sts_account.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (r *IntegrationGcpStsAccountResource) Configure(_ context.Context, request resource.ConfigureRequest, response *resource.ConfigureResponse) {
+	providerData, _ := request.ProviderData.(*FrameworkProvider)
+	r.Api = providerData.DatadogApiInstances.GetGCPIntegrationApiV2()
+	r.Auth = providerData.Auth
+}
+
+func (r *IntegrationGcpStsAccountResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = "integration_gcp_sts_account"
```
joshhuie, line 48: What do you think about renaming the Terraform resource to `integration_gcp_sts`? I think it would make more sense because ultimately this Terraform resource should [overtake this TF resource](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/integration_gcp), and be the new way to set up the GCP integration going forward.

This TF resource does create a GCP STS account, but it mainly "enables" the integration

nkzou (approved):

PR #1947, opened 2023-06-01
[datadog_team_membership] Fix team membership call loop

skarimo:


datadog/fwprovider/resource_datadog_team_membership.go
```diff
@@ -116,11 +116,12 @@ func (r *TeamMembershipResource) Read(ctx context.Context, request resource.Read
 			response.Diagnostics.AddError("response contains unparsedObject", err.Error())
 			return
 		}
-		if len(resp.GetData()) == 0 || len(resp.GetData()) > 100 {
+
+		userTeams = append(userTeams, resp.GetData()...)
+		if len(resp.GetData()) > 100 {
```
nkzou:
```suggestion
		if len(resp.GetData()) < 100 {
```
I think > 100 would infinite loop?

nkzou (approved):

PR #1949, opened 2023-06-02
[datadog_metric_tag_configuration] Adds exclude_tags_mode; updates include_percentiles

jmfurlott:
This PR updates the provider for metric tag configurations, adding support for the new `exclude_tags_mode` field (boolean; default false). And updates the logic for `include_percentiles` which was buggy when users would try to set it to false.

The new logic for the booleans is:
- if set to `true`, the values are then set to true assuming validation passes
- if set to `false` OR completed omitted from the tf apply then it gets to `false`.  This means that if users just remove it, then values could change unexpectedly.  Confirmed with @nkzou on this behavior

#### Changelog
* Updates `include_percentiles` and adds `exclude_tags_mode`
* Adds as new field with description
* Adds validation for when an empty tagset is provided
* Adds a test for empty tagset case
* Adds examples


datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -37,6 +39,9 @@ func resourceDatadogMetricTagConfiguration() *schema.Resource {
 			if includePercentilesOk && *metricTypeValidated != datadogV2.METRICTAGCONFIGURATIONMETRICTYPES_DISTRIBUTION {
 				return fmt.Errorf("cannot use include_percentiles with a metric_type of %s, must use metric_type of 'distribution'", metricType)
 			}
+			if excludeTagsMode.(bool) && len(tags.(*schema.Set).List()) == 0 {
```
efraese, line 42: I think for this we might want to use an `ok` like we have for the other params to check if the user provided the field or not. This case will not error if the field is set to false and they do not provide tags (I think we want to error for that case right?)

datadog/tests/resource_datadog_metric_tag_configuration_test.go
```diff
@@ -47,6 +51,7 @@ func testAccCheckDatadogMetricTagConfigurationIncludePercentilesError(uniq strin
 			metric_name = "%s"
 			metric_type = "%s"
 			tags = ["sport"]
+            exclude_tags_mode = false
```
efraese, line 54: nit: can we fix the formatting so its in line with the other params (same below)

datadog/resource_datadog_metric_tag_configuration.go
jmfurlott: I am a bit confused -- I am trying to test that value of the field being `true` (not the presence of a value in the field) with an empty set should return an error. you can have a validation configuration with `{ exclude_tags_mode: false, tags: [] }` which this shouldn't error on

datadog/tests/cassettes/TestAccDatadogMetricTagConfiguration_Error.yaml
```diff
@@ -0,0 +1,23 @@
...
+    headers:
+      Accept:
+      - application/json
+      Content-Type:
+      - application/json
+    url: https://api.datadoghq.com/api/v2/metrics/tf_TestAccDatadogMetricTagConfiguration_Error_local_1618930539/tags
+    method: POST
+  response:
+    body: |
+      {"status":"error","code":403,"errors":["Forbidden"],"statuspage":"http://status.datadoghq.com","twitter":"http://twitter.com/datadogops","email":"support@datadoghq.com"}
```
jmfurlott, line 17: Not sure why we're getting a 403 forbidden here?

datadog/resource_datadog_metric_tag_configuration.go
efraese: ah sorry I thought we were checking to make sure that when the exclude tags field was provided we also had a tags field too. but if we are testing for the empty list case then this makes sense.

datadog/tests/cassettes/TestAccDatadogMetricTagConfiguration_Error.yaml
efraese: I had issues with cassettes in the past, I think we should probably reach out to `#intg-tools-libs`

drichards-87 (approved):

datadog/tests/cassettes/TestAccDatadogMetricTagConfiguration_Error.yaml
nkzou: This was a tricky one to figure out, turns out this 403 was probably from when you were implementing the new field. The test itself doesn't actually do any api calls, since all the errors are caught by terraform itself, so there was no new cassette to overwrite the original 403 recording.

nkzou (changes requested):
This PR won't work unless the Go client is updated to include the `exclude_tags_mode` field. To do that you need to add that field to the relevant schemas in our [api spec](https://github.com/DataDog/datadog-api-spec/blob/master/spec/v2/metrics.yaml#L138)

docs/resources/metric_tag_configuration.md
```diff
@@ -49,6 +50,7 @@ resource "datadog_metric_tag_configuration" "example_count_metric" {
 ### Optional
 
 - `aggregations` (Block Set) A list of queryable aggregation combinations for a count, rate, or gauge metric. By default, count and rate metrics require the (time: sum, space: sum) aggregation and gauge metrics require the (time: avg, space: avg) aggregation. Can only be applied to metrics that have a `metric_type` of count, rate, or gauge. (see [below for nested schema](#nestedblock--aggregations))
+- `exclude_tags_mode` (Boolean) Toggle to include/exclude tags as queryable for your metric.  Defaults to false.  Can only be applied to metrics that have one or more tags configured. Defaults to `false`.
```
drichards-87, line 53:
```suggestion
- `exclude_tags_mode` (Boolean) Toggle to include/exclude tags as queryable for your metric. Can only be applied to metrics that have one or more tags configured. Defaults to `false`.
```

drichards-87: I think there's a duplicate sentence in the description.

drichards-87 (dismissed):

drichards-87 (approved):

docs/resources/metric_tag_configuration.md
efraese, line 53: Should this match the same text as what we have above for the `exclude_tags_mode` description? also can we remove the double spaces around `Defaults to false.`

efraese (dismissed):

efraese (approved):

drichards-87 (approved):

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -161,20 +173,26 @@ func buildDatadogMetricTagConfiguration(d *schema.ResourceData) (*datadogV2.Metr
...
-		if *metricType != datadogV2.METRICTAGCONFIGURATIONMETRICTYPES_DISTRIBUTION {
-			return nil, fmt.Errorf("include_percentiles field not allowed with metric_type: %s, only with metric_type distribution", *metricType)
-		}
-		attributes.SetIncludePercentiles(includePercentiles.(bool))
+	if includePercentiles.(bool) && *metricType != datadogV2.METRICTAGCONFIGURATIONMETRICTYPES_DISTRIBUTION {
+		return nil, fmt.Errorf("include_percentiles field not allowed with metric_type: %s, only with metric_type distribution", *metricType)
+	}
+
+	if *metricType != datadogV2.METRICTAGCONFIGURATIONMETRICTYPES_DISTRIBUTION {
+		attributes.IncludePercentiles = nil
```
jmfurlott, line 183: Potentially a breaking change; please see updated PR description for more details

nkzou: I don't think this breaking change is actually a problem - the only case I can find where this differs in behavior to the old version is when `include_percentiles = true` on the backend, but is omitted or false in the terraform config. In this case, the old version actually has perpetual drift (terraform keeps trying to set it to null/false, but the provider doesn't actually send a false to backend), and I think its very unlikely a user would leave their config in this state.

datadog/resource_datadog_metric_tag_configuration.go
```diff
@@ -161,20 +173,26 @@ func buildDatadogMetricTagConfiguration(d *schema.ResourceData) (*datadogV2.Metr
 	}
 	attributes.SetMetricType(*metricType)
 
-	includePercentiles, iclFieldSet := d.GetOk("include_percentiles")
+	includePercentiles, _ := d.GetOk("include_percentiles")
```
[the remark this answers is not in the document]
jmfurlott: Easy enough :+1: Updated!

nkzou (approved):

PR #2015, opened 2023-07-21
[datadog_cloud_configuration_rule] Added filter field

symphony-elias:
https://datadoghq.atlassian.net/browse/SEC-9912


docs/resources/cloud_configuration_rule.md
```diff
@@ -66,6 +66,7 @@ resource "datadog_cloud_configuration_rule" "myrule" {
 
 ### Optional
 
+- `filter` (Block List) Additional queries to filter matched events before they are processed. Defaults to empty list (see [below for nested schema](#nestedblock--filter))
```
brett0000FF, line 69:
```suggestion
- `filter` (Block List) Additional queries to filter matched events before they are processed. Defaults to empty list (see [nested schema](#nested-schema-for-filter)).
```

docs/resources/cloud_configuration_rule.md
```diff
@@ -75,6 +76,14 @@ resource "datadog_cloud_configuration_rule" "myrule" {
 
 - `id` (String) The ID of this resource.
 
+<a id="nestedblock--filter"></a>
+### Nested Schema for `filter`
```
brett0000FF, line 80:
```suggestion
### Nested Schema for filter
```

docs/resources/cloud_configuration_rule.md
```diff
@@ -75,6 +76,14 @@ resource "datadog_cloud_configuration_rule" "myrule" {
...
+
+Required:
+
+- `action` (String) The type of filtering action. Valid values are `require`, `suppress`.
```
brett0000FF, line 84:
```suggestion
- `action` (String) The type of filtering action. Valid values are `require` or `suppress`.
```

brett0000FF (approved):
LGTM. Thanks! I offered a few small suggestions.

Also, I wanted to note that our style prefers sentence-case headings for documentation. I didn't change that here since the PR has a smaller scope.

symphony-elias:
@brett0000FF Thanks for your suggestions. However, the documentation is automatically generated and CI fails if there is any discrepancy between the documentation on the branch and the generated one. Please contact @DataDog/integrations-tools-and-libraries team if you have suggestions to make regarding the way the documentation is generated

skarimo (approved):

docs/resources/cloud_configuration_rule.md
skarimo: same as above

PR #2030, opened 2023-07-26
Bump go version to `1.20` and other terraform package versions

skarimo:


buraizu (dismissed):

.github/workflows/test.yml
```diff
@@ -34,15 +34,19 @@ jobs:
...
+          - "1.5.3"
         # platform: [ubuntu-latest, macos-latest, windows-latest]
         platform: [ubuntu-latest]
-        # test old terraform versions only on ubuntu-latest
-        include:
-          - terraform_version: "0.15.5"
-            platform: ubuntu-latest
-          - terraform_version: "0.13.7"
-            platform: ubuntu-latest
+        # # test old terraform versions only on ubuntu-latest
```
therve, line 44: Should we remove that?

skarimo: Ended up removing everything commented out. We can add it back if we ever need to test windows etc

PR #2043, opened 2023-08-03
[data_sources] Relax unparsed object check for datasources

skarimo:
For datasources, a single response item with an unparsed object should not break the entire resource.


nkzou (dismissed):

datadog/fwprovider/data_source_datadog_dashboard_list.go
```diff
@@ -89,6 +85,10 @@ func (d *datadogDashboardListDatasource) Read(ctx context.Context, req datasourc
 		return
 	}
 
+	if err := utils.CheckForUnparsed(listResponse); err != nil {
```
skarimo, line 88:
```suggestion
	if err := utils.CheckForUnparsed(foundList); err != nil {
```

nkzou (approved):

PR #2079, opened 2023-08-25
[datadog_integration_aws] Migrate resource to terraform-plugin-framework 

skarimo:


maycmlee (dismissed):
👍 for docs

nkzou (dismissed):

skarimo:
Placing this as draft for now and rethinking a bit, potentially wait for https://github.com/hashicorp/terraform-plugin-framework/issues/510 to be resolved.

Using StateUpgrade for handling null vs zero values from sdkv2 seems a bit of an overkill and has pitfalls that needs to be reconsidered.

PR #2085, opened 2023-08-28
[datadog_integration_azure] Fix docs for app_service_plan_filters

ava-silver:
We realized the doc comment for app_service_plan_filters in terraform has a typo, and could be generally improved by changing it to the doc we already use in the tile.

### DIFF
--- a/datadog/resource_datadog_integration_azure.go
+++ b/datadog/resource_datadog_integration_azure.go
@@ -49,7 +49,7 @@ func resourceDatadogIntegrationAzure() *schema.Resource {
 					Optional:    true,
 				},
 				"app_service_plan_filters": {
-					Description: "This comma-separated list of tags (in the form `key:value`) defines a filter that Datadog uses when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags are imported into Datadog. The rest, including the apps and functions running on them, are ignored. This also filters the metrics for any App or Function running on the App Service Plan(s).",
+					Description: "This comma-separated list of tags (in the form `key:value,key:value`) defines a filter that Datadog uses when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags are imported into Datadog. The rest, including the apps and functions running on them, are ignored. This also filters the metrics for any App or Function running on the App Service Plan(s).",
 					Type:        schema.TypeString,
 					Optional:    true,
 				},
--- a/datadog/resource_datadog_integration_azure.go
+++ b/datadog/resource_datadog_integration_azure.go
@@ -49,7 +49,7 @@ func resourceDatadogIntegrationAzure() *schema.Resource {
 					Optional:    true,
 				},
 				"app_service_plan_filters": {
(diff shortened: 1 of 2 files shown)


nkzou (approved):

docs/resources/integration_azure.md
```diff
@@ -37,7 +37,7 @@ resource "datadog_integration_azure" "sandbox" {
 
 ### Optional
 
-- `app_service_plan_filters` (String) String of app service plan tag(s) (in the form `key:value,key:value`) defines a filter that Datadog uses when collecting metrics from Azure. Limit the Azure instances that are pulled into Datadog by using tags. Only hosts that match one of the defined tags are imported into Datadog. For example, `env:production,deploymentgroup:red`.
+- `app_service_plan_filters` (String) This comma separated list of tags (in the form key:value) defines a filter that we will use when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags will be imported into Datadog. The rest, including the apps and functions running on them, will be ignored. Note that this will also filter the metrics for any App or Function running on the App Service Plan(s).
```
cswatt, line 40:
```suggestion
- `app_service_plan_filters` (String) This comma-separated list of tags (in the form `key:value`) defines a filter that Datadog uses when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags is imported into Datadog. The rest, including the apps and functions running on them, are ignored. This also filters the metrics for any App or Function running on the App Service Plan(s).
```

datadog/resource_datadog_integration_azure.go
```diff
@@ -49,7 +49,7 @@ func resourceDatadogIntegrationAzure() *schema.Resource {
 					Optional:    true,
 				},
 				"app_service_plan_filters": {
-					Description: "String of app service plan tag(s) (in the form `key:value,key:value`) defines a filter that Datadog uses when collecting metrics from Azure. Limit the Azure instances that are pulled into Datadog by using tags. Only hosts that match one of the defined tags are imported into Datadog. For example, `env:production,deploymentgroup:red`.",
+					Description: "This comma separated list of tags (in the form key:value) defines a filter that we will use when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags will be imported into Datadog. The rest, including the apps and functions running on them, will be ignored. Note that this will also filter the metrics for any App or Function running on the App Service Plan(s).",
```
cswatt, line 52:
```suggestion
					Description: "This comma-separated list of tags (in the form key:value) defines a filter that Datadog uses when collecting metrics from Azure App Service Plans. Only App Service Plans that match one of the defined tags is imported into Datadog. The rest, including the apps and functions running on them, are ignored. This also filters the metrics for any App or Function running on the App Service Plan(s).",
```

cswatt (dismissed):
leaving some feedback for syntax change in accordance with datadog styleguide. approving so as to not block

datadog/resource_datadog_integration_azure.go
ava-silver: > Only App Service Plans that match one of the defined tags is 
"App Service Plans" is plural, so it should be that they `are` imported into Datadog, right? unless I'm missing something.

nkzou (approved):

PR #2097, opened 2023-09-12
[datadog_sensitive_data_scanner_group] Avoid panic when nil returned

mikutas:
related to #2098


skarimo:
Thanks for the PR. For this, we have a schema helper function that will validate the values. Could you update your pr to use:
```
validators.ValidateEnumValue
```

Here is an example usage: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_monitor.go#L244 

Thanks!

PR #2124, opened 2023-09-27
[datadog_synthetics_test] Prevent setting empty values for rum settings

romainberger:
For Synthetics browser tests, rum settings can be set with `is_enabled = true` but currently the provider adds some empty values for the other options (`application_id` and `client_token_id`), but it causes some weird behavior in the frontend. 
This PR fixes that by checking if a value has actually been put.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2010,11 +2010,15 @@ func buildTestOptions(d *schema.ResourceData) *datadogV1.SyntheticsTestOptions {
...
 				if applicationId, ok := settings["application_id"]; ok {
-					rumSettings.SetApplicationId(applicationId.(string))
+					if len(applicationId.(string)) > 0 {
+						rumSettings.SetApplicationId(applicationId.(string))
+					}
 				}
 
 				if clientTokenId, ok := settings["client_token_id"]; ok {
-					rumSettings.SetClientTokenId(int64(clientTokenId.(int)))
+					if clientTokenId.(int) != 0 {
```
romainberger, line 2019: TF uses `0` when a value hasn't been put, but it is not a valid client token so I think it's safe to check for this.

nkzou (approved):

skarimo (approved):

PR #2125, opened 2023-09-27
[datadog_monitor] Improve require_full_window description

### LINKED ISSUE #2122: `require_full_window` property has a different behavior than written in the documentation

3.30.0

### Terraform Version

1.5.6

## Context

We created a monitor via terraform in order to be notified or a sparse metric (the data points have a rate of around 1/hour), which was not triggered even though it should have. We created the monitor manually and it was actually working, the difference being that the manual monitor had the value corresponding to `require_full_window` to `false`, while the one created by terraform had the value set to `true`. 

## Problem

After taking a look at the [doc](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/monitor#require_full_window), we realize that the existing behavior should have the `false` by default. It would be useful to know if this is a bug or an intended behavior. 

```hcl
resource "datadog_monitor" "awesome_monitor" {
  name    = "Awesome name"
  type    = "metric alert"
  message = "{{#is_alert}} Something is terribly wrong {{/is_alert}}"
  query = "max(last_1h):sum:metric.metric_property{env:${local.environment}} >= 1500"

  monitor_thresholds {
    critical = 1500
    warning = 1350
  }

  notify_no_data    = false
  renotify_interval = 3600

  tags = flatten(["tag1", "tag2"])
}
```

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

### Important Factoids

_No response_

### References

_No response_

nkzou:
Description doesn't match api documentation wrt having a conditional truthy value, and there's also an awkward line break that makes it easy to miss the recommendation to set it to false.

Closes #2122


docs/resources/monitor.md
```diff
@@ -76,9 +76,7 @@ We recommend at least 2x the monitor timeframe for metric alerts or 2 minutes fo
 - `renotify_interval` (Number) The number of minutes after the last notification before a monitor will re-notify on the current status. It will only re-notify if it's not resolved.
 - `renotify_occurrences` (Number) The number of re-notification messages that should be sent on the current status.
 - `renotify_statuses` (Set of String) The types of statuses for which re-notification messages should be sent. Valid values are `alert`, `warn`, `no data`.
-- `require_full_window` (Boolean) A boolean indicating whether this monitor needs a full window of data before it's evaluated.
-
-We highly recommend you set this to `false` for sparse metrics, otherwise some evaluations will be skipped. Default: `true` for `on average`, `at all times` and `in total` aggregation. `false` otherwise.
+- `require_full_window` (Boolean) A boolean indicating whether this monitor needs a full window of data before it's evaluated. Defaults to `true`. We highly recommend you set this to `false` for sparse metrics, otherwise some evaluations will be skipped.
```
buraizu, line 79:
```suggestion
- `require_full_window` (Boolean) A boolean indicating whether this monitor needs a full window of data before it's evaluated. Defaults to `true`. Datadog strongly recommends you set this to `false` for sparse metrics, otherwise some evaluations may be skipped.
```

buraizu (approved):
Thanks, approving with a minor wording update per the documentation [style guide](https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#first-and-second-person-pronouns)

chrismdd (approved):

skarimo (approved):

PR #2134, opened 2023-10-04
[datadog_integration_gcp_sts] Add Account tags, ResourceCollectionEnabled and IsSecurityCommandCenterEnabled support to gcp tf module

dtru-ddog:
**Purpose**
Attempting to add new fields for gcp sts tf module to utilize and set account tags for customer accounts.


datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -52,6 +53,11 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 	response.Schema = schema.Schema{
 		Description: "Provides a Datadog Integration GCP Sts resource. This can be used to create and manage Datadog - Google Cloud Platform integration.",
 		Attributes: map[string]schema.Attribute{
+			"account_tags": schema.SetAttribute{
+				Optional:    true,
+				Description: "Tags to be associated with GCP metrics and service checks from your account.",
+				ElementType: types.StringType,
```
ash-ddog, line 59: Hmmm, I wonder why this is `types.StringType` even though above it's `types.Set`. The REST API does have `account_tags` as an optional `array` of strings 🤔

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -240,6 +249,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
 	diags := diag.Diagnostics{}
 	attributes := datadogV2.NewGCPSTSServiceAccountAttributesWithDefaults()
 
+	accountTags := make([]string, 0)
```
ash-ddog, line 252: This is not idiomatic Go. Typically, you initialize slices to `nil`. `nil` slices are valid for just about anything (i.e. iteration, appending, etc).
`var accountTags []string`

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -240,6 +249,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
...
+	if !state.AccountTags.IsNull() {
+		diags.Append(state.AccountTags.ElementsAs(ctx, &accountTags, false)...)
+	}
+	attributes.SetAccountTags(accountTags)
```
ash-ddog, line 256: Feels weird to include `account_tags=[]` in the REST call when it's an optional field.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -267,6 +282,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsUpdateRequestBody(ctx
...
```
ash-ddog, line 289: Same comments apply here.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -267,6 +282,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsUpdateRequestBody(ctx
...
+
 	if !state.Automute.IsNull() {
 		attributes.SetAutomute(state.Automute.ValueBool())
 	}
```
ash-ddog, line 293: All this stuff is super duplicative. Is it worth refactoring like this?

```
func (r *integrationGcpStsResource) buildIntegrationGcpStsCreateRequest(ctx context.Context, state *integrationGcpStsModel) (*datadogV2.GCPSTSServiceAccountCreateRequest, diag.Diagnostics) {
	attrs, diags := r.buildIntegrationGcpStsAccountAttributes(ctx, state)
	req := datadogV2.NewGCPSTSServiceAccountCreateRequestWithDefaults()
	req.Data = datadogV2.NewGCPSTSServiceAccountDataWithDefaults()
	req.Data.SetAttributes(*attrs)
	return req, diags
}

func (r *integrationGcpStsResource) buildIntegrationGcpStsUpdateRequest(ctx context.Context, state *integrationGcpStsModel) (*datadogV2.GCPSTSServiceAccountUpdateRequest, diag.Diagnostics) {
	attrs, diags := r.buildIntegrationGcpStsAccountAttributes(ctx, state)
	req := datadogV2.NewGCPSTSServiceAccountUpdateRequestWithDefaults()
	req.Data = datadogV2.NewGCPSTSServiceAccountUpdateRequestDataWithDefaults()
	req.Data.SetAttributes(*attrs)
	return req, diags
}

func (r *integrationGcpStsResource) buildIntegrationGcpStsAccountAttributes(ctx context.Context, state *integrationGcpStsModel) (*datadogV2.GCPSTSServiceAccountAttributes, diag.Diagnostics) {
	diags := diag.Diagnostics{}
	attributes := datadogV2.NewGCPSTSServiceAccountAttributesWithDefaults()

	accountTags := make([]string, 0)
	if !state.AccountTags.IsNull() {
		diags.Append(state.AccountTags.ElementsAs(ctx, &accountTags, false)...)
	}
	attributes.SetAccountTags(accountTags)

	if !state.Automute.IsNull() {
		attributes.SetAutomute(state.Automute.ValueBool())
	}
	if !state.ClientEmail.IsNull() {
		attributes.SetClientEmail(state.ClientEmail.ValueString())
	}
	if !state.IsCspmEnabled.IsNull() {
		attributes.SetIsCspmEnabled(state.IsCspmEnabled.ValueBool())
	}

	hostFilters := make([]string, 0)
	if !state.HostFilters.IsNull() {
		diags.Append(state.HostFilters.ElementsAs(ctx, &hostFilters, false)...)
	}
	attributes.SetHostFilters(hostFilters)

	return attributes, diags
}
```

Weird thing is that I'm pretty sure that the `update` API does not support updating `ClientEmail`, but it was already in the `update` body before these changes 🥴 .

ash-ddog (dismissed):
Nothing really blocking, but we should talk about some of these comments. 👍 

Also, are there no tests in the terraform stuff? 🤔

jhgilbert (dismissed):

mariuskiessling:
Hey @ash-ddog and @dtru-ddog, do you have any updates on this PR? I was about to hand in my own PR for this missing field when I saw that this one already exists. We rely on setting the account tags in one of our configurations that we provision from Terraform. Anything I can do to help this thing be merged?

NitriKx:
I would also like to see that merged, may I help? For the automated tests we can recycle my closed PR https://github.com/DataDog/terraform-provider-datadog/pull/2248

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
NitriKx: If I'm not wrong (not a Go expert) this is the type of the elements in the set (schema.SetAttribute)
This is inspired from what has been done for host_filters which is has the same type

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
NitriKx: IMO that's pertinent but maybe an a separated PR?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: It's indicating what each element in the set are. (ie: This is a set of strings)

[Reference](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/set)

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -52,10 +57,16 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
...
+			"account_tags": schema.SetAttribute{
+				Optional:    true,
+				Description: "Tags to be associated with GCP metrics and service checks from your account.",
+				ElementType: types.StringType,
+			},
 			"automute": schema.BoolAttribute{
 				Optional:    true,
 				Computed:    true,
 				Description: "Silence monitors for expected GCE instance shutdowns.",
+				Default:     booldefault.StaticBool(false),
```
skarimo, line 69: Whats the reasoning for this change? This can potentially break existing customers if they set `automute` to true in the UI.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -71,17 +82,28 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
...
+				Optional:    true,
+				Description: "Your Host Filters.",
+				ElementType: types.StringType,
+			},
 			"is_cspm_enabled": schema.BoolAttribute{
 				Optional:    true,
 				Computed:    true,
-				Description: "When enabled, Datadog performs configuration checks across your Google Cloud environment by continuously scanning every resource, which may incur additional charges.",
+				Description: "Whether Datadog collects cloud security posture management resources from your GCP project. If enabled, requires `resource_collection_enabled` to also be enabled.",
+				Default:     booldefault.StaticBool(false),
```
skarimo, line 94: Same question here about new default value on TF side. Can we rely on the backend to handle defaults properly since it is already computed-optional field on terraform side?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -283,9 +300,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsUpdateRequestBody(ctx
 	}
 	attributes.SetHostFilters(hostFilters)
 
-	req := datadogV2.NewGCPSTSServiceAccountUpdateRequestWithDefaults()
-	req.Data = datadogV2.NewGCPSTSServiceAccountUpdateRequestDataWithDefaults()
-	req.Data.SetAttributes(attributes)
+	if !state.IsSecurityCommandCenterEnabled.IsNull() {
```
skarimo, line 303:
```suggestion
	if !state.IsSecurityCommandCenterEnabled.IsUnknown() {
```

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: Fair enough. Hadn't really considered the case where customers were using both terraform and the UI to manage their configuration. Will remove.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: Same as above

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -26,12 +28,15 @@ type integrationGcpStsResource struct {
...
-	HostFilters          types.Set    `tfsdk:"host_filters"`
+	ID                             types.String `tfsdk:"id"`
+	AccountTags                    types.Set    `tfsdk:"account_tags"`
+	Automute                       types.Bool   `tfsdk:"automute"`
+	ClientEmail                    types.String `tfsdk:"client_email"`
+	DelegateAccountEmail           types.String `tfsdk:"delegate_account_email"`
+	HostFilters                    types.Set    `tfsdk:"host_filters"`
+	IsCspmEnabled                  types.Bool   `tfsdk:"is_cspm_enabled"`
+	IsSecurityCommandCenterEnabled types.Bool   `tfsdk:"is_security_command_center_enabled"`
+	ResourceCollectionEnabled      types.Bool   `tfsdk:"resource_collection_enabled"`
```
ash-ddog, line 39: Non-blocking: comparing this with https://github.com/DataDog/dogweb/blob/c76c0f9f0ddfd7c2daf210da53bb368c97b4a0ca/dogweb/lib/validation/schemas/gcp_account_validation.py#L2

- What is `DelegateAccountEmail` vs `ClientEmail` ? My guess is `DelegateAccountEmail == client_id` and `ClientEmail == client_email`
  - Update: read the descriptions below, but still, not sure how `DelegateAccountEmail` can have any affect on the API call we make since it's not a request parameter?
- We're missing `cloud_run_revision_filters` here, not sure if that's on purpose or not.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: Thanks! Resolving then.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -71,17 +81,27 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"host_filters": schema.SetAttribute{
+				Optional:    true,
+				Description: "Your Host Filters.",
```
ash-ddog, line 86: Kinda nit: might want to be more specific about what this actually is? maybe `Filters to be applied for GCE instance resources` or something?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: Oh I'm sorry, thought this PR was for our V1 API but it's V2, so comparing to this instead:
https://github.com/DataDog/dogweb/blob/c76c0f9f0ddfd7c2daf210da53bb368c97b4a0ca/dogweb/serializers/integrations/gcp/gcp_account_serializer.py#L45

But still the comments above are valid.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: Looks like we mostly took this (thanks), but still we have `ClientEmail` in the `update` API when that's not supported. Not blocking, but could confuse someone in the future.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: Should we be doing something like the `IsUnknown()` for `AccountTags` (like we're doing below) since it's also a new field? Not familiar enough with terraform to know what's right here.

ash-ddog (approved):
Nothing blocking but most notable comments
- Do we want to add `CloudRunRevisionFilters` ?
- Do we need to do `IsUnknown()` for `AccountTags`

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: I just shuffled it (to be ordered alphabetically).. I didn't actually add this description (It exists currently: https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/integration_gcp_sts#host_filters)

We can update it, but can tackle in a separate PR

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: DelegateAccountEmail is the STS delegate. (Basically, the principal that needs the token creator role on the client Email).

In this schema, it's a computed value that this provider returns, mainly to allow setup purely through terraform.

Basically:

- <terraform to create GCP service account (including permissions)>
- <terraform using this module  to configure the datadog gcp sts integration (using output from previous)>
- <terraform to give output of integration the token creator role to the service account in first>

smuhit: We have an upcoming task to add `cloud_run_revision_filters`.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
smuhit: Agreed.. But didn't want to make functionally different changes (removal of ClientEmail in the Update call) in a R that is essentially additive changes.

skarimo (approved):

PR #2138, opened 2023-10-10
[datadog_integration_azure] Add container_app_filters

ava-silver:
Update the azure integration terraform resources to take in the new setting now available in the API: container_app_filters.


datadog/tests/cassettes/TestAccDatadogIntegrationAzure.yaml
```diff
@@ -3,7 +3,7 @@ version: 1
 interactions:
 - request:
     body: |
-      {"app_service_plan_filters":"","automute":false,"client_id":"testc7f6-1234-5678-9101-3fcbf464test","client_secret":"testingx./Sw*g/Y33t..R1cH+hScMDt","cspm_enabled":false,"custom_metrics_enabled":false,"host_filters":"foo:bar,buzz:lightyear","tenant_name":"aaaaaaaa-bbbb-cccc-dddd-1692885321ee"}
```
aishwarya-ddog, line 6: why did the tenant_name change here? Is that necessary?

datadog/tests/cassettes/TestAccDatadogIntegrationAzure.yaml
```diff
@@ -227,7 +227,7 @@ interactions:
     method: GET
   response:
     body: |
-      [{"tenant_name":"aaaaaaaa-bbbb-cccc-dddd-1692885321ee","client_id":"testc7f6-1234-5678-9101-3fcbf464test","host_filters":"","app_service_plan_filters":"bar:baz,stinky:pete","automute":true,"cspm_enabled":true,"custom_metrics_enabled":true,"errors":[],"warnings":[]}]
+      [{"tenant_name":"aaaaaaaa-bbbb-cccc-dddd-1698416332ee","client_id":"testc7f6-1234-5678-9101-3fcbf464test","host_filters":"","container_app_filters":"bazinga:bazingo,woody:pride","app_service_plan_filters":"bar:baz,stinky:pete","automute":true,"cspm_enabled":true,"custom_metrics_enabled":true,"errors":["Invalid credentials. Datadog could not authenticate with Azure."],"warnings":[]}]
```
aishwarya-ddog, line 230: Is this 'Invalid' because of missing client secret?

datadog/tests/cassettes/TestAccDatadogLogsArchiveAzure_basic.yaml
```diff
@@ -64,11 +64,11 @@ interactions:
     headers:
       Accept:
       - application/json
-    url: https://api.datadoghq.com/api/v2/logs/config/archives/-GWdZ6m2RhKwHu40SqvtKg
+    url: https://api.datadoghq.com/api/v2/logs/config/archives/j0FZmzs-QjusCqvWcPSSsQ
```
aishwarya-ddog, line 67: What are these urls for? Why are we testing them here?

datadog/tests/cassettes/TestAccDatadogIntegrationAzure.yaml
nkzou: this is invalid because we're using a fake tenant name/client id/etc to set up the integration. These tests are mainly to test that the api properly creates the integration, so we don't care if the integration itself doesn't work.

ava-silver: hmm, its possible this is because of the tenant name change? not sure why that happened, maybe someone also removed the auth in the azure portal

datadog/tests/cassettes/TestAccDatadogLogsArchiveAzure_basic.yaml
nkzou: its an integration test that tests that we can create an azure integration, and then create a log archive connected to the same tenant id. since the azure integration was changed, this test was re-recorded since it uses the azure integration.

nkzou (approved):

cswatt (approved):

nkzou:
going to hold off on merging this until mid-next week, we're planning a hotfix release for the provider and this PR bumps the go sdk to a dev version

PR #2141, opened 2023-10-11
[data_source_datadog_service_level_objectives] Add ability to query slo with multiple tags

Thomgrus:
# Datadog Service Level Objectives data source

## What is the goal?

Actually, the `datadog_service_level_objectives` data source return a list of slos in the `slos` attribute.

We can query the data source with:
* ids (List of String) An array of SLO IDs to limit the search.
* metrics_query: (String) Filter results based on SLO numerator and denominator.
* name_query: (String) Filter results based on SLO names.
* tags_query: (String) Filter results based on a **single** SLO tag.

Here is the problem with the `tags_query` parameter. It's not possible to query multiple tags.

If there is no slos, an error is returned.

## Why is this important?

It's not possible to query with multiple tags on the `datadog_service_level_objectives` data source. So we can't use this data source to get a list of slos by `env` and `service` for example.

If there is no slos, an error is returned. It's not possible to get an empty list of slos if we expect to have no slos.

## How can we solve it?

The **single** tag limitation is due to the limitation of the api endpoint of listing slo `GET api/v1/slo` (https://docs.datadoghq.com/fr/api/latest/service-level-objectives/#get-all-slos). We have to change the api endpoint and use the `GET api/v1/slo/search`. In this case, we have work with pagination.

In order to allow to get an empty list of slos, we have to change the behavior of the data source. If there is no slos, we have to return the empty list.

## Implementation details

I propose to change the behavior of the `datadog_service_level_objectives` data source by using the `GET api/v1/slo/search` endpoint and replace parameters by only one parameter `query`.

This allow to combine multiple tags and name filter in the same query.

This implementation manage pagination and call the api endpoint until there is no more slos (100 by 100).

### Parameters

1. The `ids`, `metrics_query`, `name_query` and `tags_query` parameters are removed.
2. The `query` parameter is added.

NOTE: `query` parameter will not replace the `ids` and `metrics_query` parameters, but we can use the `query` parameter to replace the `name_query` and `tags_query` parameters.

NOTE: the second data source `datadog_service_level_objective` is not impacted by this change.

### No slos returned

If there is no slos, an empty list is returned instead of an error.

---

I hope this proposal is clear enough. If you have any questions, don't hesitate to ask me.

Have a nice day!

👋

### DIFF
--- a/datadog/data_source_datadog_service_level_objectives.go
+++ b/datadog/data_source_datadog_service_level_objectives.go
@@ -42,6 +42,17 @@ func dataSourceDatadogServiceLevelObjectives() *schema.Resource {
 					Type:        schema.TypeString,
 					Optional:    true,
 				},
+				"error_on_no_results": {
+					Description: "Throw an error if no results are found.",
+					Type:        schema.TypeBool,
+					Optional:    true,
+					Default:     true,
+				},
+				"q": {
+					Description: "The query string to filter results based on SLO names. Some examples of queries include service:<service-name> and <slo-name>. If you specify this parameter, the name_query, tags_query, and metrics_query parameters are ignored.",
+					Type:        schema.TypeString,
+					Optional:    true,
+				},
 
 				// Computed values
 				"slos": {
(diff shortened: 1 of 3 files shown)


buraizu:
Thanks for the PR, created DOCS-6416 for documentation team review

skarimo:
Hello, thanks for the PR and contribution! Unfortunately as it stands, we are unable to accept this change as this would be a breaking change for all of the existing datadog provider users.

If my understanding is correct,  their are currently 2 limitations for your use case: 
1. Missing ability to filter by multiple tags
2. Not to raise exception when no results are returned

For both of these I think we can figure out a middle ground to not break existing customers and satisfy the above requirements. For example, doing the tag filtering on the terraform side rather than relying on the service side. Additionally, we can add a flag to determine if error should be raised for zero results.

Would the above work for you? If so, I can put a PR together this week and go from there

Thomgrus:
Hello @skarimo, thanks for your response.

I understand that this change would impact all existing Datadog provider users. If I proposed it in the provider, it's because from my perspective, the primary use case for this data source is to obtain a list of SLOs filtered by multiple tags/names. My changes were aimed at modifying the underlying code minimally to reduce maintenance costs. However, I acknowledge that this perspective might not align with the needs of all existing Datadog provider users.

Regarding filtering on the Terraform side, the challenge is that we don't actually have access to the tags of the returned SLOs. This limitation exists in both the `datadog_service_level_objectives` data source and the `datadog_service_level_objective` data source, as indicated in the [datadog_service_level_objectives](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/data-sources/service_level_objectives) and [datadog_service_level_objective](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/data-sources/service_level_objective) documentation.

Concerning the second point, your suggestion of adding a flag to determine whether an error should be raised for zero results is excellent and works for me without introducing any breaking changes.

Regarding the primary issue, if you still wish to incorporate this feature into your provider and maintain backward compatibility, I see two potential solutions:

1. **Introduce a New Data Source**: Create a new data source called `datadog_service_level_objectives_v2` with the new behavior, while keeping the existing `datadog_service_level_objectives` data source with the old behavior. This way, backward compatibility can be maintained while offering the new feature. Additionally, it might be worthwhile to monitor the usage of both the `datadog_service_level_objectives` and `datadog_service_level_objectives_v2` data sources. If the usage of the old data source decreases significantly, it could be considered for removal.

2. **Add a New Parameter**: Introduce a new parameter, such as `q`, similar to other Datadog resources. Bind the new behavior to this parameter. I will modify my PR to incorporate this change. I believe that querying only the necessary data from the Datadog API using the search endpoint is more efficient than retrieving all SLOs and filtering them on the Terraform or provider side. This approach reduces the amount of data transmitted over the network and the processing load on both ends.

I hope this proposal clarifies the situation. I'm here to support and collaborate further on this issue. Please feel free to reach out if you need any additional information or if there are specific aspects you'd like to discuss in more detail. I'm committed to finding the best solution that works for everyone involved.

Best regards

skarimo:
@Thomgrus Thanks for the work on this! I think the new changes you have made is a great middle ground for introducing the new search system + maintaining backward compatibility. 

I will follow up with a more formal review of the PR soon as I am currently clarifying some behaviors with the api owners and looking to see if we can deprecate all of the other search fields as the SearchSLO endpoint query supports complex searches.

skarimo:
👋 I was unable to push to your branch so created: #2172. Let me know if the changes look good to you and i can go ahead and get this in. Main changes are around renaming `q -> query` and test fixes/cleanup.

PR #2157, opened 2023-10-23
[datadog_powerpacks] Implement support for basic widgets

dashashifrina:
This pull request implements the first part of the support for the powerpack resource in Terraform.

Unfortunately, we can't immediately add all widgets since we do not validate powerpacks widgets in the public API spec. As a result, we have to manually do some validation for each new widget type. I initially tried to add all widgets, but the PR got unreasonably large, so I opted to split it out into several parts. This first part is the simplest widgets, which don't have any nested configurations. Widgets with nested configurations require extra validation which I will add on in the following PRs.

Most of the changes in this PR are cassettes and I opted to create a separate test file for each new widget.

## Changes

- Add powerpack resource and support for: 
  - alert_graph
  - check_status
  - free_text
  - iframe
  - image
  - note
  - servicemap
- Add test and recorded cassettes for all the widgets listed above.
- Add doc.

## How to test:

1. Build the provider locally.
2. Try this powerpack resource with all the supported widgets:

```tf
resource "datadog_powerpack" "foo2" {
    name = "Sample Powerpack"
    tags = ["tag:foo1", "tag:foo2"]
    description = "Test Powerpack 2"

  template_variables {
    defaults = ["defaults"]
    name     = "datacenter"
  }
    widget {
      note_definition {
        content          = "note widget"
        background_color = "blue"
      }
      widget_layout {
        height = 12
        width = 6
        x = 5
        y = 5
      }
    }
    widget {
      free_text_definition {
        text       = "free text content"
        color      = "#d00"
        font_size  = "36"
        text_align = "left"
      }
    }
    widget {
      image_definition {
        url    = "https://google.com"
        sizing = "fit"
        margin = "small"
      }
    }
  widget {
    iframe_definition {
      url = "http://google.com"
    }
  }

  widget {
    alert_graph_definition {
      alert_id  = "895605"
      viz_type  = "timeseries"
      title     = "Widget Title"
      title_align = "center"
      title_size = "20"
    }
  }

  widget {
    servicemap_definition {
      service     = "master-db"
      filters     = ["env:prod"]
      title       = "env: prod, datacenter:dc1, service: master-db"
      title_size  = "16"
      title_align = "left"
    }
  }
  widget {
    check_status_definition {
      check     = "aws.ecs.agent_connected"
      grouping  = "cluster"
      group_by  = ["account", "cluster"]
      tags      = ["account:demo", "cluster:awseb-ruthebdog-env-8-dn3m6u3gvk"]
      title     = "Widget Title!!"
[...]

PR #2158, opened 2023-10-24
[datadog_powerpacks] Implement support for live span and more widgets

dashashifrina:
This pull request implements the second part of the support for the powerpack resource in Terraform.

Previous PR:
- https://github.com/DataDog/terraform-provider-datadog/pull/2157

## Changes

- Add support for setting `live_span` in powerpack. In regular dashboard widgets, `live_span` is set at the widget level NOT resource level. For example, `event_stream` and `trace_service` widgets use `live_span`, within a powerpack, we set `live_span` under the `group_widget` object instead. With these changes, users will be redirected to set a `live_span` property at the powerpack resource level if they attempt to set it at the widget level. 
- Add support for widgets with `request` LIST property (note: there are some widgets that define `request` as a single value and will be handled in a follow up)
- Add support for widgets with `custom_link` LIST property
- Add support for: 
  - alert_value
  - change
  - event_stream
  - event_timeline
  - manage_status
  - query_value
  - toplist
  - trace_service
- Add test and recorded cassettes for all the widgets listed above.
- Add doc.

## How to test:

1. Build the provider locally.
2. Try this powerpack resource with all the supported widgets:

```tf
resource "datadog_powerpack" "foo2" {
    name = "Sample Powerpack"
    tags = ["tag:foo1", "tag:foo2"]
    description = "Test Powerpack 2"

  template_variables {
    defaults = ["defaults"]
    name     = "datacenter"
  }
  widget {
    alert_value_definition {
	    alert_id = "895605"
    }
  }
  widget {
    alert_value_definition {
	    alert_id = "895606"
	    precision = 1
	    unit = "b"
	    title_size = "16"
	    title_align = "center"
	    title = "Widget Title"
	    text_align = "center"
    }
  }
   widget {
      change_definition {
		      request {
	      q             = "avg:system.load.1{env:staging} by {account}"
	      change_type   = "absolute"
	      compare_to    = "week_before"
	      increase_good = true
	      order_by      = "name"
	      order_dir     = "desc"
	      show_present  = false
		      }
		      title     = "Widget Title"
      }
   }
  widget {
    event_stream_definition {
      query       = "*"
      event_size  = "l"
      title       = "Widget Title"
      title_size  = 16
      title_align = "right"
    }
  }
  widget {
    manage_status_definition {
	    sort = "triggered,desc"
	    title_size = "20"
	    title = ""
	    title_align = "center"
	    hide_zero_counts = true
	    summary_type = "combined"
	    color_preference = "background"
	    query = "env:prod group_status:alert"
	    show_last_triggered = true
	    display_format = "countsAndList"
    }
    widget_layout {
	    height = 6
	    width = 4
	    x = 5
	    y = 5
    }
  }
  widget {
    trace_service_definition {
      display_format     = "three_column"
      env                = "datadog.com"
      service            = "alerting-cassandra"
      show_breakdown     = true
      show_distribution  = true
      show_errors        = true
      show_hits          = true
      show_latency       = false
      show_resource_list = false
      size_format        = "large"
      span_name          = "cassandra.query"
      title              = "alerting-cassandra #env:datadog.com"
      title_align        = "center"
      title_size         = "13"
    }
  }
  widget {
    event_timeline_definition {
      title = "Widget Title"
      title_align = "right"
      title_size = "16"
      tags_execution = "and"
      query = "status:error"
    }
    widget_layout {
      height = 4
      width = 3
      x = 5
      y = 5
    }
  }
  widget {
    query_value_definition {
      request {
        q          = "avg:system.load.1{env:staging} by {account}"
        aggregator = "sum"
        conditional_formats {
          comparator = "<"
          value      = "2"
          palette    = "white_on_green"
        }
        conditional_formats {
          comparator = ">"
          value      = "2.2"
          palette    = "white_on_red"
        }
      }
      autoscale   = true
      custom_unit = "xx"
      precision   = "4"
      text_align  = "right"
      title       = "Widget Title"
    }
  }
  widget {
    toplist_definition {
      request {
        q = "avg:system.cpu.user{app:general} by {datacenter}"
        conditional_formats {
          comparator = "<"
          value      = "2"
          palette    = "white_on_green"
        }
        conditional_formats {
          comparator = ">"
          value      = "2.2"
          palette    = "white_on_red"
        }
      }
      title = "Widget Title"
    }
  }
}

### CODE REVIEW

**Review by buraizu — APPROVED** [review]

File: datadog/resource_datadog_powerpack.go [hunk 1]
```diff
@@ -589,6 +590,19 @@ func buildDatadogPowerpack(ctx context.Context, d *schema.ResourceData) (*datado
 
 	groupWidget.Definition = definition
 
+	// Set Live span for all powerpack widgets.
+	if v, ok := d.GetOk("live_span"); ok {
```
woodb, line 594: **question:** Should we handle the case for when it's `!ok` as well?

datadog/resource_datadog_powerpack.go
```diff
@@ -602,6 +616,51 @@ func buildDatadogPowerpack(ctx context.Context, d *schema.ResourceData) (*datado
 
 }
 
+func normalizeDashboardWidgetDef(widgetDef map[string]interface{}) (map[string]interface{}, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	// Dashboard widgets set live span at the widget level, we don't allow that for powerpack widgets
+	// where live span is set at the resource level.
```
woodb, line 622: **praise:** Thanks for the comment here, makes sense!

datadog/resource_datadog_powerpack.go
```diff
@@ -602,6 +616,51 @@ func buildDatadogPowerpack(ctx context.Context, d *schema.ResourceData) (*datado
...
+	if widgetDef["live_span"] != nil {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  fmt.Sprintf("live_span must be set at the powerpack level and will be applied to all widgets"),
```
woodb, line 626: **nitpick(non-blocking):** Language tweak for consideration /cc @DataDog/documentation folks
```suggestion
			Summary:  fmt.Sprintf("live_span must be set for all powerpack resources (to be applied to all widgets within the powerpack)"),
```

datadog/resource_datadog_powerpack.go
```diff
@@ -602,6 +616,51 @@ func buildDatadogPowerpack(ctx context.Context, d *schema.ResourceData) (*datado
...
+		})
+		return nil, diags
+	}
+
+	if widgetDef["request"] != nil {
+		// Distribution/change/heatmap widgets have a "requests" field, while API Spec has a "request" field
+		// Here we set the "requests" field and remove "request"
+		widgetDefRequests := *widgetDef["request"].(*[]map[string]interface{})
```
woodb, line 634: **question:** Is there any risk that this won't cast (i.e. should we check)?

datadog/resource_datadog_powerpack.go
```diff
@@ -602,6 +616,51 @@ func buildDatadogPowerpack(ctx context.Context, d *schema.ResourceData) (*datado
...
+		widgetDef["requests"] = widgetDefRequests
+		delete(widgetDef, "request")
+	}
+	return widgetDef, diags
+}
+
+func normalizeTerraformWidgetDef(widgetDef map[string]interface{}) map[string]interface{} {
+	// Dashboard TF widgets have a "requests" field, while API Spec has a "request" field
+	// Here we set the "request" field and remove "requests"
```
woodb, line 643: **praise:** Thanks again, makes sense 👍

datadog/resource_datadog_powerpack.go
```diff
@@ -654,24 +716,41 @@ func ppkWidgetsToDashboardWidgets(ppkWidgets []datadogV2.PowerpackInnerWidgets)
...
+			definition = datadogV1.EventStreamWidgetDefinitionAsWidgetDefinition(buildDatadogEventStreamDefinition(widgetDefinition))
+		case "event_timeline":
+			definition = datadogV1.EventTimelineWidgetDefinitionAsWidgetDefinition(buildDatadogEventTimelineDefinition(widgetDefinition))
 		case "free_text":
 			definition = datadogV1.FreeTextWidgetDefinitionAsWidgetDefinition(buildDatadogFreeTextDefinition(widgetDefinition))
 		case "iframe":
 			definition = datadogV1.IFrameWidgetDefinitionAsWidgetDefinition(buildDatadogIframeDefinition(widgetDefinition))
 		case "image":
 			definition = datadogV1.ImageWidgetDefinitionAsWidgetDefinition(buildDatadogImageDefinition(widgetDefinition))
+		case "manage_status":
```
woodb, line 742: **thought(out-of-scope):** I wonder if we should consider using [the constants from datadogV1](https://github.com/DataDog/datadog-api-client-go/blob/aefcec2d6172891dbba35821a6bb96958f3a8aa7/api/datadogV1/model_monitor_summary_widget_definition_type.go#L18) in this switch statement.

For example, here it would probably make this code a little bit easier to visually check `MONITORSUMMARYWIDGETDEFINITIONTYPE_MANAGE_STATUS` matching:
- `MonitorSummaryWidgetDefinitionAsWidgetDefinition` and,
- `buildDatadogManageStatusDefinition`

Also for others where the string isn't immediately obvious but the constant likely is (added bonus: using constants from our Go client is less to worry about hah)

datadog/resource_datadog_powerpack.go
nkzou: the cast itself won't fail, but we do need to check if request has length > 0. a definition like 
```hcl
  widget {
    change_definition {
      title = "Widget Title"
    }
  }
```
causes the provider to fail with no meaningful error message.
I think the provider is discarding the api error because of how verbose the api error is. As an example, 
```bash
curl -X POST "https://api.datadoghq.com/api/v2/powerpacks" \
-H "Accept: application/json" \
-H "Content-Type: application/json" \
-H "DD-API-KEY: ${DD_API_KEY}" \
-H "DD-APPLICATION-KEY: ${DD_APP_KEY}" \
-d @- << EOF
{
    "data": {
        "attributes": {
            "description": "Test Powerpack 2",
            "group_widget": {
                "definition": {
                    "layout_type": "ordered",
                    "title": "Sample Powerpack",
                    "type": "group",
                    "widgets": [
                        {
                            "definition": {
                                "requests": [],
                                "title": "Widget Title",
                                "type": "change"
                            }
                        }
                    ]
                }
            },
            "name": "Sample Powerpack"
        },
        "type": "powerpack"
    }
}
EOF
```
gives me an api error thats 1.2 megabytes, where the first sentence explains the problem `["Invalid group widget for powerpack. Error: Invalid inner widget definition at position 0 of type change. Error: [] is too short.` followed by 1.2 megabytes of schema-related junk. is there anything that can be done about this on the backend side?

datadog/resource_datadog_powerpack.go
dashashifrina: Do you mean the case where `live_span` is not set? It's not a required field so it's fine to skip if it's not set.

datadog/resource_datadog_powerpack.go
dashashifrina: Good catch! While the cast won't fail, I did add a check for length of `request` - if it is of length 0, then an error will be thrown:
```
resource "datadog_powerpack" "foo2" {
  name = "Sample Powerpack"
  tags = ["tag:foo1", "tag:foo2"]
  description = "Test Powerpack 2"

  template_variables {
    defaults = ["defaults"]
    name     = "datacenter"
  }
  widget {
    change_definition {
      title = "Widget Title"
    }
  }
}
```
returns:
```
datadog_powerpack.foo2: Creating...
╷
│ Error: at least one request should be defined for widget: change
│ 
│   with datadog_powerpack.foo2,
│   on main.tf line 90, in resource "datadog_powerpack" "foo2":
│   90: resource "datadog_powerpack" "foo2" {
│ 
╵
```
This also works for widgets of type `query_value` and `toplist` which have a required `requests` field.

datadog/resource_datadog_powerpack.go
dashashifrina: Oo I did not realize we have constants available for widgets, what do you think about be adding this in a follow up?

woodb: Yep yep, that sounds good as its own follow-up PR (it may turn out to be a bad idea after we dig into it)

datadog/resource_datadog_powerpack.go
dashashifrina: @woodb  good idea - i'll make that adjustment.

woodb (dismissed):
lgtm!

datadog/resource_datadog_powerpack.go
```diff
@@ -654,24 +724,39 @@ func ppkWidgetsToDashboardWidgets(ppkWidgets []datadogV2.PowerpackInnerWidgets)
 		if widgetDefinition == nil {
 			continue
 		}
+		widgetDefinition = normalizeTerraformWidgetDef(widgetDefinition)
 		// Add new powerpack-supported widgets here
 		// We save Powerpack widgets as Dashboard widgets so we need to convert them to the appropriate widget definition object.
 		widgetType := widgetDefinition["type"]
 		switch widgetType {
 		case "alert_graph":
```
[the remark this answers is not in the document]
dashashifrina: Thanks - should be added back in now.

nkzou (approved):

PR #2162, opened 2023-10-27
[datadog_service_level_objectives] Fix test flakiness

nkzou:
two issues:
1. The SLO destroy helper didn't check if the resource id it was using was a SLO resource, so it was sometimes trying to delete using a datasource id.
2. It also exits after the first deleted SLO, so the datasource tests that create two will fail because it's random which one gets deleted.


datadog/tests/data_source_datadog_service_level_objectives_test.go
```diff
@@ -14,7 +14,7 @@ func TestAccDatadogServiceLevelObjectivesDatasource(t *testing.T) {
 	t.Parallel()
 	ctx, accProviders := testAccProviders(context.Background(), t)
 	firstSLOName := strings.ToLower(strings.ReplaceAll(uniqueEntityName(ctx, t), "-", "_"))
-	secondSLOName := strings.ToLower(strings.ReplaceAll(uniqueEntityName(ctx, t), "-", "_"))
+	secondSLOName := firstSLOName
```
skarimo, line 17:
```suggestion
```

Nit for cleanup. Let's remove secondSLOName altogether since its kinda useless.

skarimo (approved):

skarimo (approved):

PR #2173, opened 2023-11-08
[datadog_powerpack] set reasonable default group widget layout value

dashashifrina:
Shouout to @nkzou  for finding this bug - adding a terraformed powerpack onto a dashboard causes it to disappear on refresh. This is happening because the API accepts powerpack inputs without a `group_widget` layout property, which breaks on dashboards.

To fix this, I am adding a reasonable default value for the powerpack group_widget layout.

To test:

1. Create a powerpack TF resource

```
resource "datadog_powerpack" "foo2" {
  name = "Sample Powerpack"
  tags = ["tag:foo1", "tag:foo2"]
  description = "Test Powerpack 2"

  template_variables {
    defaults = ["defaults"]
    name     = "datacenter"
  }
  widget {
    note_definition {
      content          = "note widget"
      background_color = "red"
    }
  }
}
```
3. Add powerpack resource to dashboard. 
4. Refresh dashboard. It should still be there.


docs/resources/powerpack.md
```diff
@@ -18,6 +18,7 @@ Provides a Datadog powerpack resource. This can be used to create and manage Dat
 ### Optional
 
 - `description` (String) The description of the powerpack.
+- `layout` (Block List, Max: 1) The layout of the powerpack on a 'free' dashboard. (see [below for nested schema](#nestedblock--layout))
```
buraizu, line 21:
```suggestion
- `layout` (Block List, Max: 1) The layout of the powerpack on a free-form dashboard. (see [below for nested schema](#nestedblock--layout))
```

docs/resources/powerpack.md
```diff
@@ -28,6 +29,21 @@ Provides a Datadog powerpack resource. This can be used to create and manage Dat
...
+Required:
+
+- `height` (Number) The height of the widget.
+- `width` (Number) The width of the widget.
+- `x` (Number) The position of the widget on the x (horizontal) axis. Should be greater than or equal to 0.
+- `y` (Number) The position of the widget on the y (vertical) axis. Should be greater than or equal to 0.
+
+Optional:
+
+- `is_column_break` (Boolean) Whether the widget should be the first one on the second column in high density or not. Only for the new dashboard layout and only one widget in the dashboard should have this property set to `true`.
```
buraizu, line 44: Would it be possible to clarify what `the new dashboard layout` is referencing?

buraizu (changes requested):
Thanks, just requested some minor edits and posted a question on the doc

docs/resources/powerpack.md
```diff
@@ -28,6 +29,21 @@ Provides a Datadog powerpack resource. This can be used to create and manage Dat
...
 
+<a id="nestedblock--layout"></a>
+### Nested Schema for `layout`
+
+Required:
+
+- `height` (Number) The height of the widget.
+- `width` (Number) The width of the widget.
+- `x` (Number) The position of the widget on the x (horizontal) axis. Should be greater than or equal to 0.
+- `y` (Number) The position of the widget on the y (vertical) axis. Should be greater than or equal to 0.
```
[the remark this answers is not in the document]
dashashifrina: @buraizu This description is generated from an inherited dashboard widget layout schema, so we currently have the same language for dashboard.md [[src](https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/resources/dashboard.md#nested-schema-for-widgetgroup_definitionwidgetwidget_layout)]. What do you think about me addressing this in a follow up where I fix the language in the original dashboard resource?

docs/resources/powerpack.md
dashashifrina: @buraizu This description is generated from an inherited dashboard widget layout schema, so we currently have the same language for dashboard.md [[src](https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/resources/dashboard.md#nested-schema-for-widgetgroup_definitionwidgetwidget_layout)]. What do you think about me addressing this in a follow up where I fix the language in the original dashboard resource?

docs/resources/powerpack.md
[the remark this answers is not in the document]
buraizu: @dashashifrina Thanks for following up to confirm that. If possible, using precise language in accordance with RFC 2119 would be better. Let me know if there is anything I can do to assist.

docs/resources/powerpack.md
buraizu: @dashashifrina Thanks for following up to confirm that. As far as I can tell, there isn't any reference to `the new dashboard layout` anywhere else in the documentation, so a bit of clarification here would indeed be helpful. Let me know if there is anything I can do to assist.

buraizu (approved):
Approving for now, issues raised to be addressed via followup

nkzou (approved):

PR #2194, opened 2023-11-30
[datadog_integration_azure] Add resource_collection_enabled

ava-silver:
Adds Resource Collection field (`resource_collection_enabled`) to Azure Integration.

The new field is optional with no default, as it interacts with `cspm_enabled` so we want to let the API dictate the behavior when not specified by not providing a default value here.


alai97 (dismissed):
Looks good, two suggestions!

ava-silver:
also just realized that the terraform SDK has some really weird behavior around boolean flags, will have to investigate how we want to implement this before we can merge.

ava-silver:
closing because we will have to refactor based on this: https://github.com/DataDog/terraform-provider-datadog/pull/2202

Issue #2200, opened 2023-12-06
api.ddog-gov.com API change breaks dd_integration_aws resources


rjhornsby:
### Datadog Terraform Provider Version

v3.33.0

### Terraform Version

1.6.5

### What resources or data sources are affected?

Specifically, this affects the parts of the provider around the AWS integration, including at least these:

* datadog_integration_aws_lambda_arn
* datadog_integration_aws_log_collection

Specifically, the issue seems to be around the `account_id` argument. There may be other affected resources that I'm not currently aware of.

### Terraform Configuration Files

```hcl
resource "datadog_integration_aws_lambda_arn" "datadog-forwarder-global" {
  # our DD is fedramp, so "account_id" is actually the IAM access key of the dd IAM user.
  account_id = tolist(data.aws_iam_access_keys.datadog-aws-integration.access_keys)[0].access_key_id
  lambda_arn = aws_lambda_function.datadog-forwarder-global.arn
}

resource "datadog_integration_aws_log_collection" "services" {
  # our DD is fedramp, so "account_id" is actually the IAM access key of the dd IAM user
  account_id = tolist(data.aws_iam_access_keys.datadog-aws-integration.access_keys)[0].access_key_id
  services   = ["lambda"]
  depends_on = [datadog_integration_aws_lambda_arn.datadog-forwarder-global]
}
```

### Relevant debug or panic output

Sorry, I can't post the entire output publicly since we're a Fedramp customer and the debug output is extremely noisy, making it difficult to be sure that I've redacted everything. I can provide a selected snippet. (The longer log is attached to a DD support ticket.)

https://gist.github.com/rjhornsby/856d7f6599ea0ce8c538efb6453a6c46

### Expected Behavior

From the support ticket (3664, if it helps any DD folks), we discovered there was a breaking change to the DD fedramp API `api.ddog-gov.com`. Where previously - because FEDRAMP - we had to use an IAM access key (known to DD) to identify the AWS account, the API now expects the actual numeric account ID. The numeric account ID is how the commercial DD behaves, and how the TF documentation is correctly written.

This was a sudden, unexpected, and breaking change to the ddgov API that it appears not even the TF provider is aware of. The provider still accepts the alphanum access key id, but then crashes when talking to the API.

The expected behavior is that the API should respond in a way that coordinates and communicates with the TF provider, so that the provider can pass useful information - ie "the value you provided to me can't be used" to anyone writing TF against the provider. 

[...]

PR #2201, opened 2023-12-07
[datadog_integration_aws_*] Validate AWS account ID and improve error handling

rjhornsby:
Attempting to resolve issue #2200 

Until sometime recently, fedramp DD API required the AWS account ID argument to be an IAM access key ID when the TF provider resource asked for the AWS `account_id`. This seems to have changed to match the commercial behavior where the account id argument is expected to get the actual account ID.

However, the API change broke the provider in a way that the provider cannot handle. It ends up crashing in the middle of the `terraform apply`:

```
│
│ When applying changes to datadog_integration_aws_lambda_arn.datadog-forwarder-global, provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: Root object was present, but now absent.
│
```

This PR does not address the underlying handling of the invalid/unhandled API response, but instead tries to avoid it by ensuring the AWS account ID is a 12 digit string, and providing useful feedback if it does not - ie in the case of code which might still be using the IAM access key.


nkzou:
Hi there, thanks for the PR and raising the issue on the breaking change. 
We're currently verifying some things internally with the aws integration API, and I'm also working on a small refactor + better error handling for aws resources so we can avoid the unhelpful "inconsistent state" errors. Should have an update soon.

skarimo (approved):

andrew-purdin:
@nkzou is everything fixed with this now, or are there more API changes coming? We also had a sudden break in our Datadog/AWS integration IaC from the related issue here.

nkzou:
Things should be stable now. Are you running into issues still?

andrew-purdin:
Yes, but we are stuck on version 3.30.0 of the provider for the time being and are unable to upgrade. I expect we are continuing to see issues due to the recent API changes not being backward compatible with 3.30.0, is that correct?

nkzou:
Yes, unfortunately https://github.com/DataDog/terraform-provider-datadog/pull/2154 is the PR with the key fix for this issue (the PR we're on is just validation/error handling). The earliest version with the bugfix is 3.32.0.

andrew-purdin:
Thanks for clarifying. Just out of curiosity, is there a source we can follow for breaking API changes like this on the DD side? It was tough for us to determine the source of this problem at first.

rjhornsby:
> Thanks for clarifying. Just out of curiosity, is there a source we can follow for breaking API changes like this on the DD side? It was tough for us to determine the source of this problem at first.

+1. I spent a few hours tracking this down by diving into the provider code and then the API's behavior to finally figure out it was an API change. Because we couldn't figure out what was going on, DD support also had to spend time internally running this down.

At minimum the folks responsible for the API need to please communicate with the internal consumers like the folks responsible for the TF provider that an API change is coming so this doesn't send folks like @andrew-purdin and myself running in circles trying to understand what we did wrong.

nkzou:
Breaking changes like this should only happen in very specific and rare cases, so there isn't a specific place to look for updates about them. With that being said, we've received your feedback about this and have passed it along to the relevant team.

PR #2202, opened 2023-12-07
[datadog_integration_azure] Migrate resource to framework

skarimo:


datadog/tests/resource_datadog_integration_gcp_test.go
ava-silver, line 1: why are the gcp and aws tests modified here? it looks like we're just moving the azure resource, right?

jhgilbert (approved):

nkzou (approved):

PR #2222, opened 2023-12-22
[datadog_metric_metadata] correct consistency of type

fl0Lec:
Motivation
---------
It appears that metric API has enable the distribution type on metric metadata but the result between the `PUT` and the `GET` method is not consistent:

```bash
$ export metric_name=name_of_metric_distribution
$ curl -X PUT "https://api.datadoghq.com/api/v1/metrics/${metric_name}" \
-H "Accept: application/json" \
-H "Content-Type: application/json" \
-H "DD-API-KEY: ***" \
-H "DD-APPLICATION-KEY: ***" \
-d @- << EOF
{"type":"gauge"}
EOF
{"description":null,"short_name":null,"type":"gauge","unit":null,"per_unit":null,"statsd_interval":null,"integration":null}
and then the get:
$ curl -X GET "https://api.datadoghq.com/api/v1/metrics/${metric_name}" \
-H "Accept: application/json" \
-H "DD-API-KEY: ***" \
-H "DD-APPLICATION-KEY: ***" \
{"description":null,"short_name":null,"type":"distribution","unit":null,"per_unit":null,"statsd_interval":null,"integration":null}
```

this make metric type always being marked has a diff:
```
terraform plan
~ resource "datadog_metric_metadata" "metric_name" {
        id              = "metric_name"
      ~ type            = "distribution" -> "gauge"
        # (5 unchanged attributes hidden)
    }
```

Scope
----
Set metric type to gauge in terraform state when type return by API is distribution

Test
----
Test no diff locally
```bash
make build
cd terraformdir
terraform plan
...
No changes. Your infrastructure matches the configuration.
```

### DIFF
--- a/datadog/resource_datadog_metric_metadata.go
+++ b/datadog/resource_datadog_metric_metadata.go
@@ -30,7 +30,7 @@ func resourceDatadogMetricMetadata() *schema.Resource {
 					Required:    true,
 				},
 				"type": {
-					Description: "Metric type such as `gauge` or `rate`.",
+					Description: "Metric type such as `count`, `gauge`, or `rate`. Updating a metric of type `distribution` is not supported. If you would like to see the `distribution` type returned, please contact Datadog support.",
 					Type:        schema.TypeString,
 					Optional:    true,
 				},
--- a/datadog/resource_datadog_metric_metadata.go
+++ b/datadog/resource_datadog_metric_metadata.go
@@ -95,8 +95,10 @@ func resourceDatadogMetricMetadataCreate(ctx context.Context, d *schema.Resource
 }
 
 func updateMetricMetadataState(d *schema.ResourceData, metadata *datadogV1.MetricMetadata) diag.Diagnostics {
-	if err := d.Set("type", metadata.GetType()); err != nil {
-		return diag.FromErr(err)
+	if _, ok := d.GetOk("type"); !ok || metadata.GetType() != "distribution" {
+		if err := d.Set("type", metadata.GetType()); err != nil {
+			return diag.FromErr(err)
+		}
 	}
 	if err := d.Set("description", metadata.GetDescription()); err != nil {
 		return diag.FromErr(err)
--- a/datadog/resource_datadog_metric_metadata.go
+++ b/datadog/resource_datadog_metric_metadata.go
@@ -30,7 +30,7 @@ func resourceDatadogMetricMetadata() *schema.Resource {
 					Required:    true,
 				},
 				"type": {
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_metric_metadata.go
```diff
@@ -95,8 +95,10 @@ func resourceDatadogMetricMetadataCreate(ctx context.Context, d *schema.Resource
 }
 
 func updateMetricMetadataState(d *schema.ResourceData, metadata *datadogV1.MetricMetadata) diag.Diagnostics {
-	if err := d.Set("type", metadata.GetType()); err != nil {
-		return diag.FromErr(err)
+	if _, ok := d.GetOk("type"); !ok || metadata.GetType() != "distribution" {
+		if err := d.Set("type", metadata.GetType()); err != nil {
+			return diag.FromErr(err)
+		}
```
jmfurlott, line 101: If I understand correctly, this change allows the type to be changed if the type is valid and if the type is not distribution.

That seems fine to me since `distribution` is feature flagged and  should not be editable

jmfurlott (approved):

datadog/resource_datadog_metric_metadata.go
fl0Lec: That is correct

nkzou (approved):

docs/resources/metric_metadata.md
```diff
@@ -36,7 +36,7 @@ resource "datadog_metric_metadata" "request_time" {
 - `per_unit` (String) Per unit of the metric such as `second` in `bytes per second`.
 - `short_name` (String) A short name of the metric.
 - `statsd_interval` (Number) If applicable, statsd flush interval in seconds for the metric.
-- `type` (String) Metric type such as `gauge` or `rate`.
+- `type` (String) Metric type such as `count`, `gauge`, or `rate`. Updating a metric of type `distribution` is not supported. If you would like to see the `distribution` type returned, please contact Datadog support.
```
maycmlee, line 39:
```suggestion
- `type` (String) Metric type such as `count`, `gauge`, or `rate`. Updating a metric of type `distribution` is not supported. If you would like to see the `distribution` type returned, contact Datadog support.
```

maycmlee: It might be helpful to link "Datadog support" to https://docs.datadoghq.com/help/.

Also, since you mention contacting support, could you also add an entry to this doc so Support knows how to handle the request? https://datadoghq.atlassian.net/wiki/spaces/TS/pages/328404964/The+Public+Docs+say+Contact+Support

maycmlee (commented):
A couple of small suggestions and a request!

nkzou (approved):

PR #2224, opened 2023-12-27
[application_key] Migrate application key resource to framework

HantingZhang2:
Migrate application key resource to use tf framework


apigirl (commented):
no docs team review needed

datadog/fwprovider/data_source_datadog_application_key.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Read implements datasource.DataSource.
+func (d *applicationKeyDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var state applicationKeyDataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if !state.Id.IsNull() {
+		ddResp, _, err := d.Api.GetApplicationKey(d.Auth, state.Id.ValueString())
```
skarimo, line 72: Looks like you are using the wrong operation. Correct one is `GetCurrentUserApplicationKey`. The `GetApplicationKey` doesn't return the actual key. 

Could you also add an integration only test for this case and assert the app key is not empty in state.

nkzou (approved):

Issue #1822, opened 2023-03-20
datadog_integration_pagerduty_service_object missing import


tmidi:
Hi There,
Both `datadog_integration_pagerduty` and `datadog_integration_pagerduty_service_object` don't have import functions.

### Terraform Version

`all versions`

### Affected Resource(s)

-   `datadog_integration_pagerduty_service_object`
-   `datadog_integration_pagerduty`

### Expected Behavior

Both resources should have an import function.

therve:
For security reasons the service key is not returned by the API, so it's impossible to import those. Sorry for the inconvenience.

fernandobarroso-signifyd:
Hi @therve ,
I need the import for this resource not for getting the service key, but just for Terraform knowing that it already exists. Otherwise, it tries to create it again, resulting in this error: `Error: error creating PagerDuty integration service from https://app.datadoghq.com/api/v1/integration/pagerduty/configuration/services: 400 Bad Request: {"errors":["Service already exists"]}`.
Is there a way to prevent this?

skyzyx:
Same issue. We need to be able to import state somehow.

ruben-chainalysis:
I also hit this, [here's a PR I pushed to solve the issue](https://github.com/DataDog/terraform-provider-datadog/pull/2286) (it works for me now, but feel free to add any comments if there's any other thing I missed).

PR #2286, opened 2024-02-19
[datadog_integration_pagerduty_service_object] Add importer

### LINKED ISSUE #513: [Feature Request] Unable to import a PagerDuty service object
### Terraform Version

0.12.24

### Affected Resource(s)

- datadog_integration_pagerduty_service_object

### Expected Behavior

I can define an existing resource and import it.

### Actual Behavior

Import of PagerDuty service objects is not supported.

ruben-chainalysis:
Adds an importer for `datadog_integration_pagerduty_service_object`.

As a way to get unstuck when there's an existing resource that is not reflected in the state.

With this, one can do:

```
terraform import datadog_integration_pagerduty_service_object.<name of the resource> "<service name>"
```

And get rid of `400 Bad Request` errors that might happen when applying on top of an already existing service with the same name.

Fixes #513 and #1822 (which shouldn't be closed, IMHO).

### DIFF
--- a/datadog/resource_datadog_integration_pagerduty_service_object.go
+++ b/datadog/resource_datadog_integration_pagerduty_service_object.go
@@ -19,8 +19,13 @@ func resourceDatadogIntegrationPagerdutySO() *schema.Resource {
 		ReadContext:   resourceDatadogIntegrationPagerdutySORead,
 		UpdateContext: resourceDatadogIntegrationPagerdutySOUpdate,
 		DeleteContext: resourceDatadogIntegrationPagerdutySODelete,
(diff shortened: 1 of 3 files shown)


skarimo:
👋 Thanks for the PR. For better import experience I pushed up a change to allow setting the service key from the env var `SERVICE_KEY`. Let me know if this looks good to you and we can go ahead and get this in. This should still work with the ignore_lifecycles method.

aliciascott (approved):

ruben-chainalysis:
Hi @skarimo, that's a great idea I hadn't thought about, thanks!

skarimo (approved):

nkzou (approved):

PR #2287, opened 2024-02-19
[datadog_metric_metadata] don't send `type` field for distribution metrics

smaeda-ks:
This is similar to https://github.com/DataDog/terraform-provider-datadog/pull/2222

We've enabled the feature flag on our account, but because Datadog APIs don't allow including the `type: "distribution"` in the request payload anyway, update operations could end up with 400 bad request errors due to the upstream API validation.

This PR adds an interim workaround so that the provider doesn't include the `type` filed at all for any `distribution` metrics.

### DIFF
--- a/datadog/resource_datadog_metric_metadata.go
+++ b/datadog/resource_datadog_metric_metadata.go
@@ -81,6 +81,14 @@ func resourceDatadogMetricMetadataCreate(ctx context.Context, d *schema.Resource
 	auth := providerConf.Auth
 
 	id, m := buildMetricMetadataStruct(d)
+
+	// Datadog API only has partial supports for the distribution type in metrics endpoints,
+	// and sending the distribution type field in the request payload will result in a 400 error.
+	// This is an interium workaround so that users can still initialize resources with distribution metric.
+	if m.GetType() == "distribution" {
+		m.SetType("")
+	}
+
 	createdMetadata, httpResponse, err := apiInstances.GetMetricsApiV1().UpdateMetricMetadata(auth, id, *m)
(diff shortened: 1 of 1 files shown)


smaeda-ks:
@nkzou @jmfurlott since you reviewed https://github.com/DataDog/terraform-provider-datadog/pull/2222 before, would be great if you can take a look at this PR as well, thanks in advance!

datadog/resource_datadog_metric_metadata.go
```diff
@@ -81,6 +81,14 @@ func resourceDatadogMetricMetadataCreate(ctx context.Context, d *schema.Resource
 	auth := providerConf.Auth
 
 	id, m := buildMetricMetadataStruct(d)
+
+	// Datadog API only has partial supports for the distribution type in metrics endpoints,
+	// and sending the distribution type field in the request payload will result in a 400 error.
+	// This is an interium workaround so that users can still initialize resources with distribution metric.
+	if m.GetType() == "distribution" {
+		m.SetType("")
```
jmfurlott, line 89: Will an empty string result in a null value in the request? This might get invalidated by the API

jmfurlott (dismissed):
LGTM but left a note about the empty string type

datadog/resource_datadog_metric_metadata.go
smaeda-ks: @jmfurlott https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV1/model_metric_metadata.go#L23-L24

the API client sets `omitempty` so it will be ignored

PR #2311, opened 2024-03-11
[datadog_user_role] Creating a new `datadog_user_role` resource

wangwillson1:
As part of ACCESS-2220, we want to create a `datadog_user_role` resource for user-role assignments via Terraform. These assignments should not overwrite existing ones that aren't managed by Terraform.


drichards-87 (approved):

docs/resources/user_role.md
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+resource "datadog_user" "foo" {
+  email = "new@example.com"
+}
+
+# Create new team_membership resource
+resource "datadog_user_role" "foo" {
+  team_id = datadog_role.foo.id
```
pablito-perez, line 32:
```suggestion
  role_id = datadog_role.foo.id
```
typo?

wangwillson1: hah! good catch lol, forgot to replace it when i copied it over 💀

datadog/fwprovider/data_source_datadog_user_roles.go
```diff
@@ -0,0 +1,163 @@
...
+				Description: "When true, `filter_keyword` string is exact matched against the user's `name`.",
+				Optional:    true,
+			},
+			// Computed values
+			"user_roles": schema.ListAttribute{
+				Computed:    true,
+				Description: "List of users assigned to role.",
+				ElementType: types.ObjectType{
+					AttrTypes: map[string]attr.Type{
+						"id":      types.StringType,
```
skarimo, line 75:
```suggestion
```

Lets omit this since it is not directly related to the object

datadog/fwprovider/data_source_datadog_user_roles.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (r *datadogUserRolesDataSource) Configure(_ context.Context, request datasource.ConfigureRequest, response *datasource.ConfigureResponse) {
+	providerData, _ := request.ProviderData.(*FrameworkProvider)
+	r.Api = providerData.DatadogApiInstances.GetRolesApiV2()
+	r.UsersApi = providerData.DatadogApiInstances.GetUsersApiV2()
+	r.Auth = providerData.Auth
+}
+
+func (d *datadogUserRolesDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = "user_roles"
```
skarimo, line 48: I think we should rename this to `role_users`

datadog/fwprovider/data_source_datadog_user_roles.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (d *datadogUserRolesDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Use this data source to retrieve information about existing Datadog user role assignments.",
```
skarimo, line 53:
```suggestion
		Description: "Use this data source to retrieve information about existing Datadog role users assignments.",
```

datadog/tests/data_source_datadog_user_roles_test.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+}
+
+func testAccDatasourceUserRolesConfig(uniq string) string {
+	return fmt.Sprintf(`
+resource "datadog_user" "foo" {
+	email = "%s@example.com"
+	lifecycle {
+		ignore_changes = [ roles ]
+	}
```
skarimo, line 56:
```suggestion
```

Lets instead mark the `roles` field in datadaog_user as `Optional+Computed`.  And add an additional description stating the roles field and datadog_user_role should not be used together

examples/resources/datadog_user_role/import.sh
```diff
@@ -0,0 +1,3 @@
+# This resource is imported using user_id and role_id seperated by `:`.
+
+terraform import datadog_user_role.foo "${user_id}:${role_id}"
```
skarimo, line 3:
```suggestion
terraform import datadog_user_role.foo "${role_id}:${user_id}"
```
Looks like this needs to be reversed:

docs/resources/user_role.md
```diff
@@ -0,0 +1,57 @@
...
+
+# datadog_user_role (Resource)
+
+Provides a Datadog UserRole resource. This can be used to create and manage Datadog user_role. Note that this resource is in beta and is subject to change.
+
+## Example Usage
+
+```terraform
+resource "datadog_role" "foo" {
+  name = "foo"
```
srosenthal-dd, line 17: nit: it's bugged me for a while that we use foo & bar in our examples. Can name the roles closer to reality? Like "Admin" or "Dashboard Writer" or something?

srosenthal-dd (approved):
Tested locally and it worked well!

I am trying to see how this and https://github.com/DataDog/terraform-provider-datadog/pull/2340 interact, doing some more local testing, will comment on the other PR.

wangwillson1:
> I am trying to see how this and https://github.com/DataDog/terraform-provider-datadog/pull/2340 interact, doing some more local testing, will comment on the other PR.

i wonder if i made those changes obsolete... assuming we don't use `datadog_user` and `datadog_user_role` together, it seems that the `Computed: true` attribute make it so that when the roles attribute is omitted, no changes take effect now (which is what we want!)

srosenthal-dd:
I retested with your latest code, this time not including https://github.com/DataDog/terraform-provider-datadog/pull/2340.

I noted that:
1. ✅ Importing an existing datadog_user works and doesn't change the roles
1. ✅ The new datadog_user_role worked well for creations/deletions/imports
1. ✅ Changing datadog_user to have unset roles always leaves it alone, but setting to empty list removes all roles
1. ⚠️ One edge case I found, not caused by this PR: if you entirely replace the list of roles for a datadog_user, it'll first remove, then add, leaving a brief period where the user has no roles. If the user is the one whose key you are using for Terraform, you get locked out. Proposed fix for that is https://github.com/DataDog/terraform-provider-datadog/pull/2346.
1. ℹ️ I probably missed some discussion, but I'm curious about the decision to make `datadog_role_users` data source instead of a more direct `data datadog_user_role`. I tried using it and it worked OK but I wasn't sure what users would do with it.

srosenthal-dd (approved):

nkzou (approved):

PR #2340, opened 2024-03-27
[datadog_user] Omitting `roles` from `datadog_user` no longer unassigns all roles 

rzyxie:
update `datadog_user` Terraform resource such that:
- if `roles` is omitted/unspecified, do not update existing roles (currently: unassigns all roles)
- if `roles=[]`, unassign all roles

https://github.com/DataDog/terraform-provider-datadog/assets/62684843/f0a1e6f6-5682-40bb-ae6f-292b8ec4d922

### DIFF
--- a/datadog/resource_datadog_user.go
+++ b/datadog/resource_datadog_user.go
@@ -235,7 +235,7 @@ func resourceDatadogUserCreate(ctx context.Context, d *schema.ResourceData, meta
 			return diag.FromErr(err)
 		}
 
-		// Update roles
+		// Update roles if provided
 		if _, ok := d.GetOk("roles"); ok {
 			_, newRolesI := d.GetChange("roles")
 			newRoles := newRolesI.(*schema.Set)
--- a/datadog/resource_datadog_user.go
+++ b/datadog/resource_datadog_user.go
@@ -45,6 +45,7 @@ func resourceDatadogUser() *schema.Resource {
 					Type:        schema.TypeSet,
 					Optional:    true,
 					Elem:        &schema.Schema{Type: schema.TypeString},
+					Deprecated:  "`roles` has been deprecated in favor of the new `datadog_user_role` resource.",
 				},
 				"send_user_invitation": {
 					Description: "Whether an invitation email should be sent when the user is created.",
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_user.go
```diff
@@ -45,6 +45,7 @@ func resourceDatadogUser() *schema.Resource {
 					Type:        schema.TypeSet,
 					Optional:    true,
 					Elem:        &schema.Schema{Type: schema.TypeString},
+					Deprecated:  "`roles` has been deprecated in favor of the new `datadog_user_role` resource.",
```
wangwillson1, line 48: not deprecated yet! i can imagine customers will still want to use this roles attribute if they want TF to be a source of truth

the `datadog_user_role` resource is meant for incremental adoption, and it won't overwrite non-terraform-managed roles, which might not be ideal

datadog/resource_datadog_user.go
```diff
@@ -354,7 +355,9 @@ func resourceDatadogUserUpdate(ctx context.Context, d *schema.ResourceData, meta
 	apiInstances := providerConf.DatadogApiInstances
 	auth := providerConf.Auth
 
-	if d.HasChange("roles") {
+	isRolesOmitted := d.GetRawConfig().AsValueMap()["roles"].IsNull()
```
wangwillson1, line 358: nice 🧠

datadog/resource_datadog_user.go
rzyxie: ooooops okay!

rtrieu (approved):
LGTM

srosenthal-dd (approved):
Didn't test but read the code (so little! nice!) and it totally makes sense to me.

srosenthal-dd (commented):
I tested with importing an existing role and that deleted all my roles. We might need to look into that edge case. TF code like this:

```
resource "datadog_user" "me" {
  email  = "stephen.rosenthal@datadoghq.com"
}

import {
  id = "b40bc50f-2fec-46d2-90ef-06c15e26a914"
  to = datadog_user.me
}
```

rzyxie:
> I tested with importing an existing role and that deleted all my roles. We might need to look into that edge case. TF code like this:

@srosenthal-dd hmm i'm a little bit confused how you got this... i can't reproduce it myself. i can reproduce it with the current state (on master), but with the changes in this PR, the roles don't change (expected?):

but regardless, Willson said this PR isn't needed anymore with his changes 👍 

https://github.com/DataDog/terraform-provider-datadog/assets/62684843/5531812a-c6be-4f02-8999-c8b5ab10556a

PR #2346, opened 2024-04-02
[datadog_user] Re-order role updates

srosenthal-dd:
This PR proposes that we re-order role updates for datadog_user, so additions happen before removals.

I was testing `datadog_user` and ran into a funny case. If you're modifying yourself, for example to change from
```
[ datadog_role.old_admin_role.id ]
```
to
```
[ datadog_role.new_admin_role.id ]
```
... then if we do removals first, then additions, we'll get locked out and won't be able to finish the operation. It seems safer to first add all the new roles, then remove any old roles.

Note: it's still possible to delete your own role and lock yourself out, but only if the target list of roles doesn't contain the user_access_manage permission.

### DIFF
--- a/datadog/resource_datadog_user.go
+++ b/datadog/resource_datadog_user.go
@@ -130,17 +130,6 @@ func updateRoles(meta interface{}, userID string, oldRoles *schema.Set, newRoles
 	rolesToRemove := oldRoles.Difference(newRoles)
 	rolesToAdd := newRoles.Difference(oldRoles)
 
-	for _, roleI := range rolesToRemove.List() {
-		role := roleI.(string)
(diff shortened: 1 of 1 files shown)


srosenthal-dd:
cc @wangwillson1 & @rzyxie ; any opinions on this? I saw this as a possible improvement while testing your PRs in this area

rzyxie (approved):
makes sense to me 😄

srosenthal-dd:
(I tested end-to-end, manually. Happy to walk anyone through it if it'd be useful)

wangwillson1 (approved):
good catch!

nkzou (approved):

PR #2350, opened 2024-04-04
[datadog_restriction_policy] Update test to check new error format

nkzou:
The error message on the backend has changed - it is now `Invalid request: [PrincipalTypeInvalidForRelation] principal \"foo:<redacted>\" is not supported by \"editor\" relation, supported types: \"org, role, team, user\"`

This updates the test to check for the new error string


HantingZhang2 (approved):

datadog/tests/resource_datadog_restriction_policy_test.go
```diff
@@ -69,7 +69,7 @@ func TestAccRestrictionPolicyInvalidInput(t *testing.T) {
 	invalidResourceType := "security_rule"
 
 	invalidResourceTypeError, _ := regexp.Compile("Invalid resource type")
-	invalidPrincipalError, _ := regexp.Compile("not a valid principal")
+	invalidPrincipalError, _ := regexp.Compile("PrincipalTypeInvalidForRelation")
```
vincentyl, line 72: Unrelated to this PR:

curious about your feeling about these standardized error codes instead of freeform error messages. any concerns about it?

nkzou: I don't see anything wrong with it. Is there a page somewhere (preferably public) with a list and explanation of all these error codes? I think standardizing messaging is always good and would be even better if there was a page customers could do a simple lookup to figure out whats wrong.
Historically we've gotten support issues from customers attempting to debug cryptic 400 Bad Requests with minimal explanation, so this is a good step imo.

PR #2357, opened 2024-04-11
[resource_datadog_security_monitoring_rule] Add a validation step in security_monitoring_rule_resource provider.

cpatel24:
[SEC-13266] Add a validation step in security_monitoring_rule_resource provider.

[SEC-13266]: https://datadoghq.atlassian.net/browse/SEC-13266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1433,3 +1443,22 @@ func resourceDatadogSecurityMonitoringRuleDelete(ctx context.Context, d *schema.
 
 	return nil
 }
+
+func resourceDatadogSecurityMonitoringRuleCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	if validate, ok := diff.GetOkExists("validate"); ok && !validate.(bool) {
```
clementgbcn, line 1448: **🎯 suggestion:** ‏If the value does not exist, I think you should skip the validation (to be retrocompatible with existing behaviour).

cpatel24: My bad, there were some glitches regarding the diff. For some tests, it wasn't reaching CustomizeDiff stage or in same case the value wasn't properly coming in ResourceDiff.

aliciascott (approved):
approved for docs

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -439,6 +441,15 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
 			Description: "The rule type.",
 			Default:     "log_detection",
 		},
+		"validate": {
+			Description: "Whether or not to validate the SLO.",
```
clementgbcn, line 445: **💣 chore:** ‏Could you update the comment.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1433,3 +1443,22 @@ func resourceDatadogSecurityMonitoringRuleDelete(ctx context.Context, d *schema.
 
 	return nil
 }
+
+func resourceDatadogSecurityMonitoringRuleCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	if validate, ok := diff.GetOkExists("validate"); !ok || (ok && !validate.(bool)) {
```
clementgbcn, line 1448:
```suggestion
	if validate, ok := diff.GetOkExists("validate"); !ok || !validate.(bool) {
```

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1433,3 +1443,22 @@ func resourceDatadogSecurityMonitoringRuleDelete(ctx context.Context, d *schema.
...
+		// Explicitly skip validation
+		log.Printf("[DEBUG] Validate is %v, skipping validation", validate.(bool))
+		return nil
+	}
+
+	providerConf := meta.(*ProviderConfiguration)
+	apiInstances := providerConf.DatadogApiInstances
+	auth := providerConf.Auth
+
+	if payload, err := buildCreatePayload(diff); err == nil {
```
clementgbcn, line 1458: What should we do in case `err` is not `nil`?

docs/resources/security_monitoring_rule.md
```diff
@@ -69,6 +69,7 @@ resource "datadog_security_monitoring_rule" "myrule" {
 - `tags` (Set of String) Tags for generated signals.
 - `third_party_case` (Block List, Max: 10) Cases for generating signals for third-party rules. Only required and accepted for third-party rules (see [below for nested schema](#nestedblock--third_party_case))
 - `type` (String) The rule type. Valid values are `application_security`, `log_detection`, `workload_security`, `signal_correlation`. Defaults to `"log_detection"`.
+- `validate` (Boolean) Whether or not to validate the SLO.
```
clementgbcn, line 72: **🤓 nitpick:** ‏It will be updated when the comment in the Go file is updated.

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -510,15 +521,15 @@ func resourceDatadogSecurityMonitoringRuleCreate(ctx context.Context, d *schema.
 	return nil
 }
 
-func isSignalCorrelationSchema(d *schema.ResourceData) bool {
+func isSignalCorrelationSchema(d utils.Resource) bool {
```
clementgbcn, line 524: **❔ question:** ‏Why did you have to update the method signature?

datadog/resource_datadog_security_monitoring_rule.go
cpatel24: Done.

datadog/resource_datadog_security_monitoring_rule.go
cpatel24: The method is called from `buildCreatePayload` and it is used from multiple places. One where we want to create payload from `ResourceData`, and other where we want to create from `ResourceDiff`. Since they are separate (and not one extending other), I found `utils.Resource` being a parent interface. With that, we can pass either `ResourceData/ResourceDiff` and it accepts them.

datadog/resource_datadog_security_monitoring_rule.go
cpatel24: Thanks, not sure when this was changed, i had this while coding :(

datadog/resource_datadog_security_monitoring_rule.go
cpatel24: Sure, lets log it as Debug.

docs/resources/security_monitoring_rule.md
cpatel24: Done.

datadog/tests/resource_datadog_security_monitoring_rule_test.go
```diff
@@ -1021,6 +1089,7 @@ resource "datadog_security_monitoring_rule" "acceptance_test" {
 	name = "%s"
 	message = "acceptance rule triggered"
 	enabled = false
+	validate = false
```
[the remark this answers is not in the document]
cpatel24: Done.

clementgbcn (dismissed):

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -439,6 +441,15 @@ func datadogSecurityMonitoringRuleSchema() map[string]*schema.Schema {
 			Description: "The rule type.",
 			Default:     "log_detection",
 		},
+		"validate": {
```
skarimo, line 444: Could we update the function signature to accept a boolean. Something along the lines of `datadogSecurityMonitoringRuleSchema(includeValidate bool)` and add validate block based on this. Mainly because this schema is reused in the data-source which doesn't make use of this

skarimo (approved):

clementgbcn (approved):

PR #2365, opened 2024-04-16
[datadog_integration_aws] Add extended resource collection field

nkzou:
Adds a new `extended_resource_collection_enabled` field that replaces `resource_collection_enabled`. If `cspm_resource_collection_enabled` is true, `extended_resource_collection_enabled` will automatically be set to true as well.


docs/resources/integration_aws.md
```diff
@@ -37,10 +37,11 @@ resource "datadog_integration_aws" "sandbox" {
...
 - `filter_tags` (List of String) Array of EC2 tags (in the form `key:value`) defines a filter that Datadog uses when collecting metrics from EC2. Wildcards, such as `?` (for single characters) and `*` (for multiple characters) can also be used. Only hosts that match one of the defined tags will be imported into Datadog. The rest will be ignored. Host matching a given tag can also be excluded by adding `!` before the tag. e.x. `env:production,instance-type:c1.*,!region:us-east-1`.
 - `host_tags` (List of String) Array of tags (in the form `key:value`) to add to all hosts and metrics reporting through this integration.
 - `metrics_collection_enabled` (String) Whether Datadog collects metrics for this AWS account.
-- `resource_collection_enabled` (String) Whether Datadog collects a standard set of resources from your AWS account.
+- `resource_collection_enabled` (String, Deprecated) Whether Datadog collects a standard set of resources from your AWS account. **Deprecated.** Deprecated in favor of 'extended_resource_collection_enabled'.
```
cswatt, line 44:
```suggestion
- `resource_collection_enabled` (String, Deprecated) Whether Datadog collects a standard set of resources from your AWS account. **Deprecated.** Deprecated in favor of `extended_resource_collection_enabled`.
```

datadog/resource_datadog_integration_aws.go
```diff
@@ -95,6 +95,14 @@ func resourceDatadogIntegrationAws() *schema.Resource {
...
 					Type:         schema.TypeString,
 					Description:  "Whether Datadog collects a standard set of resources from your AWS account.",
+					Deprecated:   "Deprecated in favor of `extended_resource_collection_enabled`.",
+					Computed:     true,
+					Optional:     true,
+					ValidateFunc: validation.StringInSlice([]string{"true", "false"}, true),
+				},
+				"extended_resource_collection_enabled": {
+					Type:         schema.TypeString,
+					Description:  "Whether Datadog collects additional attributes and configuration information about the resources in your AWS account. If `cspm_resource_collection` is enabled, this will be automatically enabled as well.",
```
jvanbrie, line 105: The wording we used in the api-spec was "Required for `cspm_resource_collection`" which I think gets across the idea that we'll fail if it's set to false while cspm is set to true. I don't feel strongly about the wording, but it'd be nice for them to be consistent. Do you think this wording should be in the api-spec too?

datadog/resource_datadog_integration_aws.go
```diff
@@ -175,6 +183,11 @@ func buildDatadogIntegrationAwsStruct(d *schema.ResourceData) *datadogV1.AWSAcco
 		iaws.SetCspmResourceCollectionEnabled(vBool)
 	}
 
+	if v, ok := d.GetOk("extended_resource_collection_enabled"); ok && v.(string) != "" || iaws.GetCspmResourceCollectionEnabled() {
+		vBool, _ := strconv.ParseBool(v.(string))
+		iaws.SetExtendedResourceCollectionEnabled(vBool || iaws.GetCspmResourceCollectionEnabled())
```
jvanbrie, line 188: Not sure I'm reading this right, but this seems weird to me - does this mean if the customer sets extended=false, and cspm=true the request will succeed as the terraform provider will send extended=true and cspm=true to our api? I feel like in that case the request should fail since the parameters provided were invalid, rather than fixing the parameters.

datadog/resource_datadog_integration_aws.go
nkzou: I can change the tf description to match api spec, I wanted to mention that the value might be enabled automatically depending on the cspm setting, but since the field is computed I think that might not be necessary to explicitly point out.

datadog/resource_datadog_integration_aws.go
nkzou: Looks like I messed up while attempting to factor out some nested if-statements into boolean ops. I’ll fix it up to fail if extended is explicitly false when cspm is true.

datadog/resource_datadog_integration_aws.go
```diff
@@ -175,6 +183,13 @@ func buildDatadogIntegrationAwsStruct(d *schema.ResourceData) *datadogV1.AWSAcco
 		iaws.SetCspmResourceCollectionEnabled(vBool)
 	}
 
+	if d.GetRawConfig().GetAttr("extended_resource_collection_enabled").IsNull() && iaws.GetCspmResourceCollectionEnabled() {
+		iaws.SetExtendedResourceCollectionEnabled(true)
```
jvanbrie, line 187: Do we need this case?

nkzou: We need to check the raw config because Terraform has no way of differentiating between a unset value being interpreted as false and an explicit false value. Without this case, a user who enables `cspm` on an existing resource will get 400s because terraform thinks they're explicitly setting `extended` to false as well, when it's actually the computed value being carried forward from the previous state.

jvanbrie (approved):

skarimo (approved):

PR #2383, opened 2024-04-24
[datadog_security_monitoring_suppression] Handle an empty string for expiration date

clementgbcn:
An empty expiration date should be a valid argument and process as a not having an expiration date.

This PR addresses this issue by handling this case and by updating the unit tests to cover this case.


pietrodll (approved):
Thanks 🙌

nkzou (approved):

Issue #1795, opened 2023-03-02
datadog_security_monitoring_rule | panic: interface conversion: interface {} is nil, not string


marcossv9:
### Terraform Version

```
Terraform v1.3.9
on linux_amd64
+ provider registry.terraform.io/datadog/datadog v3.21.0
```

### Affected Resource(s)

Please list the resources as a list, for example:

-   datadog_security_monitoring_rule

### Terraform Configuration Files

I've created my own module to create `datadog_security_monitoring_rule`, here is the `main.tf` for the module:

```hcl
resource "datadog_security_monitoring_rule" "rule" {
  enabled            = var.rule_enabled
  name               = var.rule_name
  message            = var.rule_message
  has_extended_title = var.rule_has_extended_title

  dynamic "query" {
    for_each = var.rule_queries

    content {
      name  = query.key
      query = query.value.query

      aggregation     = query.value.aggregation != null ? query.value.aggregation : null
      group_by_fields = query.value.group_by_fields != null ? query.value.group_by_fields : null
    }
  }

  dynamic "case" {
    for_each = var.rule_cases

    content {
      name          = case.key
      status        = case.value.status
      condition     = case.value.condition
      notifications = case.value.notifications
    }
  }

  options {
    evaluation_window   = var.rule_options_evaluation_window
    keep_alive          = var.rule_options_keep_alive
    max_signal_duration = var.rule_options_max_signal_duration
  }

  tags = var.rule_tags
}
```
And here is how I'm using the module to create a security rule:

```hcl

module "my-rule" {
  source = "git::ssh://git@my_module_repo"

  rule_enabled            = var.rule_enabled
  rule_name               = var.rule_name
  rule_message            = var.rule_message
  rule_has_extended_title = var.rule_has_extended_title

  rule_queries = var.rule_queries

  rule_cases = var.rule_cases

  rule_options_evaluation_window   = var.rule_options_evaluation_window
  rule_options_keep_alive          = var.rule_options_keep_alive
  rule_options_max_signal_duration = var.rule_options_max_signal_duration

  rule_tags = var.rule_tags

  datadog_api_key = var.datadog_api_key
  datadog_app_key = var.datadog_app_key
}
```

### Panic Output
https://gist.github.com/marcossv9/a29f7cb211e7d119cd90f3c1c90a38ef

### Expected Behavior

a security monitoring rule should be created.

### Actual Behavior

A panic in the terraform plugin

### Steps to Reproduce

[...]

PR #2393, opened 2024-05-03
[datadog_synthetics_test] Ensure tags are non-empty and castable

nkzou:
SSIA, bugfix in response to customer ticket


Drarig29 (commented):
Does it mean the `tag` will be silently ignored if not castable to string?

Can you give an example of something non-castable that the user was passing?

skarimo (approved):

Drarig29 (approved):
LGTM

For context, a customer had `tags = ["foo:bar", "", "hello:world"]` and it was returning the same error as https://github.com/DataDog/terraform-provider-datadog/issues/1795:

```
interface {} is nil, not string
```

PR #2396, opened 2024-05-06
[datadog_organization_settings] Add Security Contacts support

alexandre-pocquet:
[Security Contacts](https://www.datadoghq.com/security/?tab=platform-safety#security-contacts) can be configured on a Datadog organization to receive security notification.
Currently users must manually set this setting in the Datadog web UI.

This pull request makes it possible to do it from Terraform as well within a `datadog_organization_settings` resource.

Example:
```hcl
resource datadog_organization_settings my_org {
    name = "My Organization"
    security_contacts = ["foo@example.org", "bar@example.org"]
}
```


maycmlee (approved):

datadog/internal/validators/validators.go
```diff
@@ -232,6 +232,29 @@ func ValidateAWSAccountID(v any, p cty.Path) diag.Diagnostics {
 	return diags
 }
 
+var _ schema.SchemaValidateDiagFunc = ValidateBasicEmail
+var basicEmailRe = regexp.MustCompile("^[^@]+@[^@]+\\.[^@.]+$")
```
daisukixci, line 236: FYI this regex catch some special char like carriage return https://regex101.com/r/yGRf01/1

alexandre-pocquet: My goal here is only to catch simple usage errors, with a very basic regex.
The full restrictive email regex is a lot more [hairy](https://stackoverflow.com/a/201378), and doesn't really bring more value since the only reliable way to test an email address is to send something to it.

daisukixci (approved):
LGTM from T&S perspective

nkzou (approved):

PR #2399, opened 2024-05-09
[datadog_security_monitoring_default_rule] Add support for custom tags in security monitoring default rules

paul-hf:
Adds support for custom tags for default rules. Tested updating the tags in staging to confirm it works as expected.

https://datadoghq.atlassian.net/browse/SEC-12971


datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -100,6 +100,13 @@ func resourceDatadogSecurityMonitoringDefaultRule() *schema.Resource {
 					Computed:    true,
 					Description: "The rule type.",
 				},
+
+				"tags": {
+					Type:        schema.TypeSet,
+					Optional:    true,
+					Description: "Tags for generated signals.",
```
clementgbcn, line 107: **🎯 suggestion:** Could you add the fact that: `Tags must contain all OOTB rule tags`.‏

clementgbcn (dismissed):
I left two small comments.

hestonhoffman (dismissed):

datadog/tests/resource_datadog_security_monitoring_default_rule_test.go
```diff
@@ -110,6 +115,14 @@ resource "datadog_security_monitoring_default_rule" "acceptance_test" {
     options {
         decrease_criticality_based_on_env = true
     }
+
+	tags = [
+		"iaas:aws",
```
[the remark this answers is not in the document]
paul-hf: Tabs vs spaces 🫠 Seems like gofmt does not run automatically here? Fixed

datadog/resource_datadog_security_monitoring_default_rule.go
paul-hf: Added to the description. One thing I realized is that anyone upgrading to this version of our terraform provider will have to import all these tags initially, and they'll also have to import them any time we change the tags on a default rule. I just want to confirm that's expected and there's no better way to do this?

clementgbcn: Yes, this API is working like that.

clementgbcn: Good point.
You are right, it could be an issue for rule update.
I think we should append to existing ones (by doing a get before doing the put).

datadog/internal/utils/api_instances_helper.go
```diff
@@ -347,12 +347,12 @@ func (i *ApiInstances) GetAuthNMappingsApiV2() *datadogV2.AuthNMappingsApi {
 	return i.authNMappingsApiV2
 }
 
-// GetCloudWorkloadSecurityApiV2 get instance of CloudWorkloadSecurityApi
-func (i *ApiInstances) GetCloudWorkloadSecurityApiV2() *datadogV2.CloudWorkloadSecurityApi {
-	if i.cloudWorkloadSecurityApiV2 == nil {
-		i.cloudWorkloadSecurityApiV2 = datadogV2.NewCloudWorkloadSecurityApi(i.HttpClient)
+// GetCSMThreatsApiV2 get instance of CSMThreatsApi
+func (i *ApiInstances) GetCSMThreatsApiV2() *datadogV2.CSMThreatsApi {
```
paul-hf, line 351: The underlying datadog package had a rename from datadogV2.CloudSecurityWorkloadApi to datadogV2.CSMThreatsApi so I did a find and replace to update the name

paul-hf:
@clementgbcn As we discussed, I've updated this PR to use the new default tags field from the datadog api and I've also made two necessary changes to accommodate the package version bump which I've explained in comments above. This is ready to be reviewed again

clementgbcn (approved):

hestonhoffman (approved):

SalahEddineBC (approved):

skarimo (approved):

PR #2422, opened 2024-06-05
[datadog_synthetics_test] support file upload for api test

etnbrd:
This PR implements the upload file feature for API tests (and steps) for the terraform provider.

⚠️ One gotcha is that we aren't sure how to upload the file only when the file content is updated, because it implies to know when the terraform config was updated locally, which isn't trivial and might lead to more edge cases than expected.
Instead, we accept re-uploading the file every time the test is updated.


Drarig29 (approved):
Looks good to me

HantingZhang2 (approved):

PR #2455, opened 2024-06-24
[datadog_monitor] Handle zero-based priority values

carlmartensen:
Add support for the upcoming zero-based monitor priority values. Terraform was storing `priority = null` as `0` in the state which means that a monitor priority changed from `null` to `0` would not detect a state change. To work around this, the schema type is switched from int to string so the change is detected. Users can still enter integer values for priority despite the schema change. 

This solution has the drawback of `terraform apply` on `priority=null` pre-existing monitors stating a change was applied as the terraform state is updated from `0` to `""`. `terraform plan` correctly states no change in this case. 

The implemented solution was discussed with the Api Clients team and determined to be the best path forward.


datadog/resource_datadog_monitor.go
```diff
@@ -763,17 +763,24 @@ func buildMonitorStruct(d utils.Resource) (*datadogV1.Monitor, *datadogV1.Monito
...
-	u.SetPriority(int64(d.Get("priority").(int)))
 	u.SetOptions(o)
 
+	if attr, ok := d.GetOk("priority"); ok {
+		x, _ := strconv.ParseInt(attr.(string), 10, 64)
+		m.SetPriority(x)
+		u.SetPriority(x)
+	} else {
+		m.SetPriorityNil()
+		u.SetPriorityNil()
```
carlmartensen, line 781: Needs publishing updated Golang client as this function doesn't yet exist.

rtrieu (approved):
LGTM

yaodingyd (approved):

datadog/resource_datadog_monitor.go
```diff
@@ -84,7 +84,7 @@ func resourceDatadogMonitor() *schema.Resource {
 				},
 				"priority": {
 					Description: "Integer from 1 (high) to 5 (low) indicating alert severity.",
```
nkzou, line 86:
```suggestion
					Description: "Integer from 0 (high) to 5 (low) indicating alert severity.",
```
Should this be updated to say 0 - 5?

carlmartensen: I had the same inclination as you did however I spoke with PM about this and they requested to not change the documentation as this is a private beta feature.

datadog/resource_datadog_monitor.go
```diff
@@ -84,7 +84,7 @@ func resourceDatadogMonitor() *schema.Resource {
...
-					Type:        schema.TypeInt,
+					Type:        schema.TypeString,
 					Optional:    true,
```
nkzou, line 88:
```suggestion
					Optional:    true,
					ValidateFunc: validation.StringInSlice([]string{"", "0", "1", "2", "3", "4", "5"}, true),

```
sorry for the back and forth, we should add some validation on the string so that customers can't put in invalid values. apart from this I think we're good to go

nkzou (changes requested):

datadog/resource_datadog_monitor.go
carlmartensen: The issue with that is that we may allow arbitrary ranges later that orgs can configure. The API will 4XX priorities out of range based on the orgs custom range.

nkzou (approved):

PR #2445, opened 2024-06-18
[datadog_synthetics_test] Support `unary` gRPC steps in multistep API tests

Drarig29:
This is a PR stack, here is how it's organized:
- https://github.com/DataDog/terraform-provider-datadog/pull/2445 (← **you are here**)
- https://github.com/DataDog/terraform-provider-datadog/pull/2457
- https://github.com/DataDog/terraform-provider-datadog/pull/2406
- https://github.com/DataDog/terraform-provider-datadog/pull/2459

---

This PR adds support for `unary` gRPC steps in multistep API tests.

What was already supported:
- `healthcheck` gRPC test
- `unary` gRPC test
- Multistep API test
  - `healthcheck` gRPC step

PR #2406, opened 2024-05-15
[datadog_synthetics_test] Add default value to `http_version`

Drarig29:
This is a PR stack, here is how it's organized:
- https://github.com/DataDog/terraform-provider-datadog/pull/2445
- https://github.com/DataDog/terraform-provider-datadog/pull/2457
- https://github.com/DataDog/terraform-provider-datadog/pull/2406 (← **you are here**)
- https://github.com/DataDog/terraform-provider-datadog/pull/2459

---

This PR adds sets a default value for `http_version` for both:
- An API test
- An API step

PR #2457, opened 2024-06-25
[datadog_synthetics_test] Fix `renotify_interval` drift when saving from UI

Drarig29:
This is a PR stack, here is how it's organized:
- https://github.com/DataDog/terraform-provider-datadog/pull/2445
- https://github.com/DataDog/terraform-provider-datadog/pull/2457 (← **you are here**)
- https://github.com/DataDog/terraform-provider-datadog/pull/2406
- https://github.com/DataDog/terraform-provider-datadog/pull/2459

---

When we save a test in the UI, the frontend saves the `monitor_options` with an additional property: `notification_preset_name`. Because of this, the `monitor_options` is not `nil` (`HasMonitorOptions` returns true) and as a result, `monitor_options.renotify_interval` is `nil`, which is saved as `0` in the local state. This causes drift with what's returned by the backend, which doesn't contain a `monitor_options.renotify_interval` at all.

This PR only saves the `monitor_options` in the local state when we do have actual properties to set (e.g. `renotify_interval`).


etnbrd (dismissed):
LGTM 👍

urseberry (approved):

nkzou (approved):

PR #2459, opened 2024-06-25
[datadog_synthetics_test] Remove `request_definition.timeout` default value

Drarig29:
This is a PR stack, here is how it's organized:
- https://github.com/DataDog/terraform-provider-datadog/pull/2445 
- https://github.com/DataDog/terraform-provider-datadog/pull/2457
- https://github.com/DataDog/terraform-provider-datadog/pull/2406
- https://github.com/DataDog/terraform-provider-datadog/pull/2459 (← **you are here**)

---

In https://github.com/DataDog/terraform-provider-datadog/pull/1497, we added the same value in TF as in the UI to fix a bug where the timeout was set to `0`.

But always hardcoding the default value has a side effect: the UI says that 1 option is configured, which isn't exactly right when users do not have a `request_definition.timeout` set in TF.

image

Our backend now ignores `0` for a timeout, so we can remove the default value from TF and ensure the change is backwards compatible (i.e. doesn't generate diffs when planning/applying)


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -168,7 +168,13 @@ func syntheticsTestRequest() *schema.Resource {
 				Description: "Timeout in seconds for the test.",
 				Type:        schema.TypeInt,
 				Optional:    true,
-				Default:     60,
+				DiffSuppressFunc: func(key, old, new string, d *schema.ResourceData) bool {
```
etnbrd, line 171: Let's remove this `DiffSuppressFunc` to avoid any unintended consequences. Having a one-time drift because of this change should not be an issue for our customers.

etnbrd (dismissed):
LGTM up to removing the `DiffSuppressFunc` to avoid adding unnecessary complexity.

datadog/resource_datadog_synthetics_test_.go
Drarig29: I agree, and it turns out that having a "configured timeout by default" is a bug on its own: that configured timeout would override an org level default timeout.

So let's accept this one-time drift to enforce the fixing of that bug 👍

Drarig29: I changed the label of the PR to be a `changelog/bugfix` instead

michaelcretzman (approved):

nkzou (approved):

PR #2465, opened 2024-06-27
[datadog_application_key] Add deprecation warning to datadog_application_key data source

jackakeller:
Add new deprecation message to datadog_application_key data source.

Followed these guidelines: https://developer.hashicorp.com/terraform/plugin/framework/deprecations#provider-data-source-or-resource-removal

Tested locally.

**Warning when deprecated data source is used:**

```
=> terraform plan
datadog_application_key.jackakeller-foo: Refreshing state... [id=ad77b716-c871-4e94-bb53-cef5ed244551]
data.datadog_application_key.foo: Reading...
data.datadog_application_key.foo: Read complete after 0s [id=ad77b716-c871-4e94-bb53-cef5ed244551]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
╷
│ Warning: Deprecated
│
│   with data.datadog_application_key.foo,
│   on main.tf line 24, in data "datadog_application_key" "foo":
│   24: data "datadog_application_key" "foo" {
│
│ The datadog_application_key data source is deprecated and will be removed in a future release with prior notice. Securely store your application key using a secret management
│ system or use the datadog_application_key resource to manage application keys in your Datadog account.
│
```

This also is printed for `terraform refresh` and `terraform apply`. On either `refresh` or `apply` the key is still stored in Terraform state.

**Error when deprecated source is used and application key is not provided from the datadog API:**

```
=> terraform plan
data.datadog_application_key.foo: Reading...
datadog_application_key.jackakeller-foo: Refreshing state... [id=ad77b716-c871-4e94-bb53-cef5ed244551]

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Warning: Deprecated
│
│   with data.datadog_application_key.foo,
│   on main.tf line 24, in data "datadog_application_key" "foo":
│   24: data "datadog_application_key" "foo" {
│
│ The datadog_application_key data source is deprecated and will be removed in a future release with prior notice. Securely store your application key using a secret management
│ system or use the datadog_application_key resource to manage application keys in your Datadog account.
│
│ (and one more similar warning elsewhere)
╵
╷
│ Error: Deprecated
│
│   with data.datadog_application_key.foo,
│   on main.tf line 24, in data "datadog_application_key" "foo":
│   24: data "datadog_application_key" "foo" {
│
│ The datadog_application_key data source is deprecated and will be removed in a future release. Securely store your application key using a secret management system or use the
│ datadog_application_key resource to manage application keys in your Datadog account.
```

This also is printed for `terraform refresh` and `terraform apply`.

**Testing note:**

If in the future we stop returning the `key` field, then the integration tests will start failing. This is okay because at that point the data source is no longer functional and the tests can be removed.

**Note:**

We are _not_  going to simply preserve pre-existing data sources by not overwriting the key value like for the [resource here](https://datadoghq.atlassian.net/browse/AUTHN-4201).

Right now, if you configure a data source based on the name, then it reads the current state from that name and overwrites key and UUID. In other words, it's always up to date with Datadog.

This means if you delete an app key that is already configured as a data source, and then create a new one with the same name, you'll get the new app key UUID and key.

If we preserve the original key's raw value, you could get the new UUID but have the old key value. Similarly, if we were to stop updating both UUID and key, then the name also wouldn't match current state. This would cause state drift.

(This is not a problem for imported resources because those import based solely on UUID, and name can still be read on future refreshes even without the key field)

### DIFF
--- a/datadog/fwprovider/data_source_datadog_application_key.go
+++ b/datadog/fwprovider/data_source_datadog_application_key.go
@@ -37,7 +37,7 @@ func (d *applicationKeyDataSource) Metadata(_ context.Context, req datasource.Me
 // Schema implements datasource.DataSource.
 func (d *applicationKeyDataSource) Schema(_ context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
 	resp.Schema = schema.Schema{
-		Description: "Use this data source to retrieve information about an existing application key.",
+		Description: "Use this data source to retrieve information about an existing application key. Deprecated. This will be removed in a future release with prior notice. Securely store your application keys using a secret management system or use the datadog_application_key resource to manage application keys in your Datadog account.",
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
 				Description: "Id for Application Key.",
@@ -57,6 +57,7 @@ func (d *applicationKeyDataSource) Schema(_ context.Context, req datasource.Sche
 				Sensitive:   true,
 			},
 		},
+		DeprecationMessage: "The datadog_application_key data source is deprecated and will be removed in a future release with prior notice. Securely store your application key using a secret management system or use the datadog_application_key resource to manage application keys in your Datadog account.",
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tyjet (dismissed):

buraizu (dismissed):

datadog/fwprovider/data_source_datadog_application_key.go
```diff
@@ -80,7 +81,9 @@ func (d *applicationKeyDataSource) Read(ctx context.Context, req datasource.Read
 			return
 		}
 		apiKeyData := ddResp.GetData()
```
adityab-datadog, line 83:
```suggestion
		appKeyData := ddResp.GetData()
```

jackakeller: Haha, looks like this has been there a long time. I can fix it.

adityab-datadog (approved):
lgtm!

buraizu (approved):

tyjet (approved):

nkzou (approved):

PR #2469, opened 2024-06-28
[datadog_synthetics_test] Add validation on synthetics test header

### LINKED ISSUE #2462: Synthetic test fails with INVALID_CHAR_IN_HEADER if entry in request_headers comes from data.local_file

v3.37.0

### Terraform Version

v1.5.7

datadog_synthetics_test

```hcl
terraform {
  required_version = ">= 0.14.11, < 2.0.0"
}


locals {
  app_name = "app_name"
  supported_events = [
    "x",
    "y",
    "z",
  ]
  # NOTE: these files need to exist locally
  fixtures = {
    for idx, event_name in local.supported_events: idx => {
      name = event_name,
      payload = jsonencode(jsondecode(file("${path.module}/fixtures/event/${event_name}.json"))),
      sha256_path = "${path.module}/fixtures/event/${event_name}.sha256.txt"
    }
  }
}

resource "aws_lambda_function_url" "this" {
  function_name      = "y"
  authorization_type = "NONE"
}

data "aws_secretsmanager_secret_version" "current" {
  secret_id = "k"
}

resource "terraform_data" "calculate_sha256" {
  for_each = local.fixtures

  # Since the file won't be stored in remote state, we have to enforce TF to always
  # generate it.
  triggers_replace = [timestamp()]

  provisioner "local-exec" {
    interpreter = ["bash", "-c"]

    environment = {
      payload = each.value.payload
      secret  = data.aws_secretsmanager_secret_version.current.secret_string
      sha256_path = each.value.sha256_path
    }

    command = <<EOF
      echo -n "$payload" | openssl dgst -sha256 -hmac "$secret" -binary | xxd -p -c 256 | xargs -I{} echo "sha256={}" > "$sha256_path"
    EOF
  }
}

data "local_file" "sha256" {
  for_each = local.fixtures
  filename = each.value.sha256_path
  depends_on = [terraform_data.calculate_sha256]
}

resource "datadog_synthetics_test" "this" {
  for_each  = local.fixtures
  name      = "[${local.app_name}] Publish '${each.value.name}' event"
  type      = "api"
  status    = "live"
  locations = ["aws:${var.region}"]
  message   = "Notify ${local.formatted_slack_channels}"
  tags      = local.monitor_tags

  options_list {
    tick_every = 30
  }

  request_definition {
    method    = "POST"
    url       = aws_lambda_function_url.this.function_url
    body      = each.value.payload
    body_type = "application/json"
    timeout   = 10
  }

  request_headers = {
    x-hub-signature-256                    = data.local_file.sha256[each.key].content
    x-github-event                         = each.value.name
    x-github-delivery                      = "X"
    x-github-hook-id                       = "X"
    x-github-hook-installation-target-id   = "X"
    x-github-hook-installation-target-type = "X"
    Content-Type                           = "application/json"
  }

  assertion {
    type     = "statusCode"
    operator = "is"
    target   = 200
  }

  assertion {
    type     = "responseTime"
    operator = "lessThan"
    target   = 4 * 1000 # 4 seconds
  }
}
```

N/A - it applies successfully.

### Expected Behavior

Synthetic tests should run successfully, without the need to edit the `x-hub-signature-256` header value and save it.

### Actual Behavior

Synthetic tests are failing with `INVALID_CHAR_IN_HEADER`

### Steps to Reproduce

1. Create empty files in `./fixtures/event/*.json` - these need to map to the `local.fixtures`
2. `terraform apply`
3. Go to synthetic test page in DataDog and run them

### Important Factoids

_No response_

### References

_No response_

HantingZhang2:
The synthetics resource now validates that the provided headers don't have new line characters 
fix #2462

### DIFF
--- a/datadog/internal/validators/validators.go
+++ b/datadog/internal/validators/validators.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-framework-validators/helpers/validatordiag"
@@ -304,26 +305,41 @@ func Float64Between(min, max float64) validator.String {
 	}
 }
 
-func ValidateHeaders(val any, p cty.Path) diag.Diagnostics {
-	var diags diag.Diagnostics
-	value := val.(map[string]interface{})
-	for _, v := range value {
-		valueString, ok := v.(string)
+func ValidateHttpRequestHeader(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(map[string]interface{})
+	for headerField, headerValue := range value {
+		if !isValidToken(headerField) {
+			errors = append(errors, fmt.Errorf("invalid value for %s (header field must be a valid token)", k))
+			return
+		}
+		headerStringValue, ok := headerValue.(string)
 		if !ok {
-			diags = append(diags, diag.Diagnostic{
-				Severity:      diag.Error,
-				Summary:       "Invalid value",
-				Detail:        "Synthetics test header must be a string",
-				AttributePath: p,
-			})
-		} else if strings.Contains(valueString, "\n") || strings.Contains(valueString, "\r\n") {
-			diags = append(diags, diag.Diagnostic{
-				Severity:      diag.Error,
-				Summary:       "Invalid value",
-				Detail:        "Synthetics test header must not contain newline characters",
-				AttributePath: p,
-			})
+			errors = append(errors, fmt.Errorf("expected type of %s to be string", k))
+			return
+		} else {
+			for _, r := range headerStringValue {
+				if (unicode.IsControl(r) && r != '\t') || (r == '\r' || r == '\n') {
+					errors = append(errors, fmt.Errorf("invalid value for %s (header value must not contain invisible characters)", k))
+					return
+				}
+			}
 		}
 	}
-	return diags
+	return
+}
+
+func isValidToken(token string) bool {
+	for _, r := range token {
+		if !isTokenChar(r) {
+			return false
+		}
+	}
+	return true
+}
+
+func isTokenChar(r rune) bool {
+	if r >= '!' && r <= '~' && !strings.ContainsRune("()<>@,;:\\\"/[]?={} \t", r) {
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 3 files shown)


datadog/internal/validators/validators.go
```diff
@@ -303,3 +303,27 @@ func Float64Between(min, max float64) validator.String {
...
+	for _, v := range value {
+		valueString, ok := v.(string)
+		if !ok {
+			diags = append(diags, diag.Diagnostic{
+				Severity:      diag.Error,
+				Summary:       "Invalid value",
+				Detail:        "Synthetics test header must be a string",
+				AttributePath: p,
+			})
+		} else if strings.Contains(valueString, "\n") || strings.Contains(valueString, "\r\n") {
```
nkzou, line 319:
```suggestion
		} else if strings.Contains(valueString, "\n") {
```
I think the `\n` case is a superset of `\r\n`. Also, we should get a review from Synthetics confirming if there are any other invalid characters we should blacklist.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -259,9 +259,10 @@ func syntheticsTestRequest() *schema.Resource {
 
 func syntheticsTestRequestHeaders() *schema.Schema {
 	return &schema.Schema{
-		Description: "Header name and value map.",
-		Type:        schema.TypeMap,
-		Optional:    true,
+		Description:      "Header name and value map.",
```
nkzou, line 262: could you look into why the description of this field got removed from the generated docs? I guess it's something to do with the validate function.

nkzou (changes requested):

buraizu (approved):

datadog/internal/validators/validators.go
etnbrd: http headers don't work with a blocklist, but with an allowlist. Basically, any visible character is allowed, everything else is forbidden: https://stackoverflow.com/a/75998796
But since this schema is not enforce on the backend, I guess it's fine to accept anything here as well 🤷

etnbrd (dismissed):
LGTM modulo getting back the documentation as mentioned by @nkzou.

datadog/resource_datadog_synthetics_test_.go
HantingZhang2: Need to use ValidateFunc instead of ValidateDiagFunc for some reason, although ValidateFunc is marked as deprecated https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatefunc

etnbrd (approved):

nkzou (approved):

PR #2509, opened 2024-07-29
[datadog_ip_allowlist] add ip allowlist example

afrieds:
Add IP Allowlist example. Merge when approved.


drichards-87 (approved):

amaskara-dd (approved):

docs/resources/ip_allowlist.md
```diff
@@ -10,7 +10,23 @@ description: |-
...
 
-
+## Example Usage
+
+```terraform
+resource "datadog_ip_allowlist" "example" {
+  enabled = false
+
+  entry {
+    cidr_block = "172.0.0.1/32"
```
srosenthal-dd, line 20: @afrieds this is a real public IP; can we switch to a private IP address which won't be in use?

One of the following would work great, and show that /32 isn't the only option possible. All are reserved for documentation, ref: https://en.wikipedia.org/wiki/Reserved_IP_addresses
* 192.0.2.0/24	
* 198.51.100.0/24
* 203.0.113.0/24

afrieds: Oops I meant to do 127. Yes of course. I will reopen. Sorry about that.

PR #2522, opened 2024-08-06
[datadog_synthetics_test] allow variable in port

etnbrd:
This PR changes the type of the `port` in the request definition from `int` to `string`, to allow setting with variable like: `{{ PORT }}`.


HantingZhang2 (dismissed):

docs/resources/synthetics_test.md
```diff
@@ -737,7 +737,7 @@ Optional:
 - `number_of_packets` (Number) Number of pings to use per test for ICMP tests (`subtype = "icmp"`) between 0 and 10.
 - `persist_cookies` (Boolean) Persist cookies across redirects.
 - `plain_proto_file` (String) The content of a proto file as a string.
-- `port` (Number) Port to use when performing the test.
+- `port` (String) Port to use when performing the test.
```
estherk15, line 740: Non-blocking, verifying for consistency, does the `dns_server_port` run multiple tests while the 'port' only run one test?

estherk15 (dismissed):
Approved with one question on consistency

docs/resources/synthetics_test.md
etnbrd: For context, I guess you're referring to the difference of wording:
- `dns_server_port (String) DNS server port to use for DNS tests.`
- `port (String) Port to use when performing the test.`

`port` is used for http test, which is the default type of API test. Singular test here refers to the test that's being configured.
`dns_server_port` is used for dns tests, which was introduced later, and is not the default type of API test. The plural tests refers to all of the dns tests that could hypothetically be configured.
For both of these tests, they run only one test in a single API test. (http is available in multi-step API test, but that's probably out of the scope here).

Given all that, I'm not sure how to change the wording. Do you have any suggestion?

estherk15 (approved):

amaskara-dd (approved):

PR #2526, opened 2024-08-08
[datadog_team] Add avatar to schema.

jmcarp:


datadog/fwprovider/resource_datadog_team.go
```diff
@@ -75,6 +76,10 @@ func (r *teamResource) Schema(_ context.Context, _ resource.SchemaRequest, respo
 				Required:    true,
 				Description: "The name of the team.",
 			},
+			"avatar": schema.StringAttribute{
+				Required:    false,
+				Description: "The avatar of the team.",
```
pgoldrbx, line 81: **Suggestion (non-blocking)**: maybe the description should specify that we expect a URL to an image asset?

jmcarp: It looks like this is just an emoji string and not a URL:
Screenshot 2024-09-10 at 3 51 52 PM

pgoldrbx: Oh, you are correct. Sorry about that. I forgot about that.

jmcarp:
@pgoldrbx does anything need to be revised, or is this ok to merge?

pgoldrbx (approved):
No issues here. 👍

datadog/fwprovider/resource_datadog_team.go
```diff
@@ -75,6 +76,10 @@ func (r *teamResource) Schema(_ context.Context, _ resource.SchemaRequest, respo
 				Required:    true,
 				Description: "The name of the team.",
 			},
+			"avatar": schema.StringAttribute{
+				Required:    false,
```
fpighi, line 80: Hello @jmcarp. Sorry for the late reply.
This should be:
```suggestion
				Optional:    true,
```

Also please regenerate the test cassettes (see [DEVELOPMENT.md](https://github.com/DataDog/terraform-provider-datadog/blob/master/DEVELOPMENT.md#testing-the-provider)) so we can approve this PR.

fpighi (changes requested):

PR #2545, opened 2024-08-28
[datadog_restriction_policy] Allow policy deletions outside of terraform to prompt resource recreation

dchayes-dd:
By not setting `state.Bindings` to an empty list if the attributes list is empty, we have a bug where if the policy was removed from a resource outside of terraform (manually in the UI, through the API, etc), it wouldn't be registered, and attempts to re-apply it would not detect a change.

Steps to reproduce the bug are:
1. Create a restriction policy using terraform
2. Delete the restriction policy in the UI
3. Run `terraform apply` again
  Expected: The policy gets recreated
  Actual: No differences are found, terraform succeeds with an empty apply.

Tested manually in a test env - confirmed that I could reproduce the original bug report, and that with this change, the policy was correctly reapplied in the same case.

### DIFF
--- a/datadog/fwprovider/resource_datadog_restriction_policy.go
+++ b/datadog/fwprovider/resource_datadog_restriction_policy.go
@@ -197,7 +197,7 @@ func (r *RestrictionPolicyResource) updateState(ctx context.Context, state *Rest
 	data := resp.GetData()
 	attributes := data.GetAttributes()
 
-	if bindings, ok := attributes.GetBindingsOk(); ok && len(*bindings) > 0 {
+	if bindings, ok := attributes.GetBindingsOk(); ok {
 		state.Bindings = []*BindingsModel{}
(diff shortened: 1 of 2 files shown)


sanjaythakkar-dd:
changes look good
can we please also add a test in https://github.com/DataDog/terraform-provider-datadog/blob/699cecb5df0c4c969528f63c829aff32c36f0b86/datadog/tests/resource_datadog_restriction_policy_test.go

dchayes-dd:
> changes look good can we please also add a test in https://github.com/DataDog/terraform-provider-datadog/blob/699cecb5df0c4c969528f63c829aff32c36f0b86/datadog/tests/resource_datadog_restriction_policy_test.go

Added a unit test.

agnesjang98 (approved):

jack-edmonds-dd (approved):

PR #2570, opened 2024-09-10
Set more specific GitHub token permissions on workflows.

jack-edmonds-dd:


.github/workflows/sdkv2_deprecation.yaml
```diff
@@ -1,5 +1,10 @@
 name: SDKv2 Deprecation Check
 
+permissions:
+  contents: write
```
nkzou, line 4: does this need `contents: write`? probably just `read`, unless i've overlooked something

jack-edmonds-dd: I think you're right. In fact I think we can remove permissions here since the job already specifies permissions.

ecdatadog (approved):

PR #2572, opened 2024-09-10
[data_source_datadog_service_account] fix exact_match pointer bug

### LINKED ISSUE #2571: Data source service_account: fails on exact_match filter

v3.41.0

### Terraform Version

v1.8.3

- data_source_datadog_service_account

```hcl
data "datadog_service_account" "foo" {
  filter               = "Foo"
  exact_match = true
}
```

_No response_

### Expected Behavior

Assuming the response body on `/api/v2/users?filter=Foo` is something like:
```
{
  "data": [
    {
      "id": 1234,
      "attributes: {
        "name": "Foo",
      }
    },
    {
      "id": 5678,
      "attributes: {
        "name": "Foo - Plus",
      }
    },
  ]
}
```

We've got two service accounts that match the filter, but only one that matches **exactly**.

So my expectation is that service account `1234` will get picked.

### Actual Behavior

The provider doesn't complain about finding more than one match, but it will pick the last match in the list returned by the API. This is caused by this code: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/fwprovider/data_source_datadog_service_account.go#L169-L192

The problem in that loop is that `userData` gets assigned a pointer to `serviceAccount`. Exact match will only happen once, but the loop will go on and the underlying value of `serviceAccount` keeps getting reassigned as you iterate.

`userData` will end up pointing to the last item in `serviceAccounts` no matter what.

### Steps to Reproduce
Assuming these service account resources have been created previously:
(and assuming they'll be returned in this order by the users list API, unsure what determines that)
```
resource "datadog_service_account" "foo" {
  email = "foo@example.com"
  name  = "Foo"
}

resource "datadog_service_account" "foo_plus" {
  email = "foo@example.com"
  name  = "Foo - Plus"
}
```

Sourcing the first SA by name:
```
data "datadog_service_account" "foo" {
  filter = "Foo"
  exact_match = true
}

resource "datadog_role" "team_foo" {}

resource "datadog_user_role" "foo_team_foo" {
  role_id = datadog_role.team_foo.id
  user_id = data.datadog_service_account.foo.id
}
```

This will attach the Role to the second Service Account ("Foo - Plus"), regardless of exact_match.

### Important Factoids

_No response_

### References

_No response_

pablito-perez:
Fixes #2571 

TL;DR: select the correct item when specifying an exact_match filter on the service_account data source

Passes the test added in the previous PR

Longer explanation: when iterating over multiple service account objects returned by the API, the `userData` variable is assigned to the pointer of the iterator `serviceAccount`. The underlying value keeps changing as we iterate over all items. It always ends up pointing to the last item in the collection, even if we did hit an exact match (`matchCount == 1`).

### DIFF
--- a/datadog/fwprovider/data_source_datadog_service_account.go
+++ b/datadog/fwprovider/data_source_datadog_service_account.go
@@ -122,7 +122,7 @@ func (d *datadogServiceAccountDatasource) Read(ctx context.Context, req datasour
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	var userData *datadogV2.User
+	var userData datadogV2.User
 	if !state.ID.IsNull() {
 		serviceAccountID := state.ID.ValueString()
 		ddResp, _, err := d.Api.GetUser(d.Auth, serviceAccountID)
@@ -135,7 +135,7 @@ func (d *datadogServiceAccountDatasource) Read(ctx context.Context, req datasour
 			resp.Diagnostics.AddError("Obtained entity was not a service account", "")
 			return
 		}
-		userData = ddResp.Data
+		userData = *ddResp.Data
 	} else {
 		optionalParams := datadogV2.ListUsersOptionalParameters{}
 		filter := state.Filter.ValueString()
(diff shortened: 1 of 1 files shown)


pablito-perez:
* **#2572** <a href="https://app.graphite.dev/github/pr/DataDog/terraform-provider-datadog/2572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @pablito-perez and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

pablito-perez:
The issue is fixed "accidentally" in latest versions of the provider, which include this golang change: https://go.dev/blog/loopvar-preview

PR #2591, opened 2024-09-24
[datadog_software_catalog] v3 terraform needs to hydrate the current state from raw content

christinexlee:
[SER-611](https://datadoghq.atlassian.net/browse/SER-611) Made changes so that the Read function uses `raw_schema` instead of `schema`. 

[SER-611]: https://datadoghq.atlassian.net/browse/SER-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


datadog/fwprovider/resource_datadog_software_catalog.go
```diff
@@ -41,7 +42,8 @@ type Included struct {
 }
 
 type IncludedAttributes struct {
-	Schema *Entity `json:"schema"`
+	Schema    *Entity `json:"schema"`
```
hyperloglogy, line 45: do we still need `Schema` ?

christinexlee: I thought so, because it's that's used in updating the entity (in `resourceEntityUpsert()`), which I thought is separate from the read. Would my understanding be correct?

datadog/fwprovider/resource_datadog_software_catalog.go
```diff
@@ -279,13 +290,21 @@ func (r *catalogEntityResource) Read(ctx context.Context, request resource.ReadR
 		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error unmarshalling entity"))
 	}
 
-	if len(entityResp.Included) != 1 || entityResp.Included[0].Attributes == nil || entityResp.Included[0].Attributes.Schema == nil {
+	if len(entityResp.Included) != 2 || entityResp.Included[1].Attributes == nil || entityResp.Included[1].Attributes.RawSchema == "" {
```
hyperloglogy, line 293: len(included) should be 1 since we are only requesting `raw_schema`

datadog/fwprovider/resource_datadog_software_catalog.go
```diff
@@ -254,6 +242,15 @@ func (r *catalogEntityResource) ImportState(ctx context.Context, request resourc
...
 }
 
+func decodeBase64String(data string, response *resource.ReadResponse) []byte {
+	bytes, err := base64.StdEncoding.DecodeString(data)
+	if err != nil {
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error decoding base64 string"))
+		return nil
+	}
+	return bytes
+}
```
skarimo, line 252: Nit, Could we remove the response argument and just return the error from the function so the execution stops rather than continuing with nil in the caller

Bunk (approved):

skarimo (approved):

PR #2611, opened 2024-10-08
[datadog_dashboard] Add support for x and y axis controls on distribution widgets

RobThomas02:
I'd like to add support for x and y axis controls on Distribution widgets in the Terraform provider. This is possible to do in the UI, but not via Terraform yet.


brtu:
Can you also add a test in https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/tests/resource_datadog_dashboard_distribution_test.go and record test cassettes?

brett0000FF (dismissed):

brtu:
Updated tests and cassettes and also tried this locally - was able to create distribution widgets with x & y coords, and omitting those fields in the terraform config worked too.  This change shouldn't impact existing terraformed widgets.

brtu (approved):

amaskara-dd (approved):

PR #2612, opened 2024-10-08
[datadog_integration_gcp_sts] Add CloudRevisionRunFilters field to GCPStsServiceAccount Tf module

tedkahwaji:
## Summary 

The following change adds support for `CloudRevisionRunFilters` for the following [terraform module](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/integration_gcp#example-usage). 

CloudRevisionRunFilters is included in our [api-spec ](https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/gcp.yaml#L64) but wasn't supported in tf yet. 

Note, we recently updated GCP-ICS requests/responses; the dogweb api that these TF resources hit now do not expect `client-email` as part of the request for `UpdateGCPStsAccount` ([reference](https://github.com/DataDog/dogweb/pull/128793)), I've updated the Update request in this PR to not pass that field to the v2 update endpoint. 

### Testing 

Update the tests & recorded results: 

```
DD_TEST_CLIENT_APP_KEY="<prod-app-key>" DD_TEST_CLIENT_API_KEY="<prod-api-key>" 
DD_TEST_SITE_URL="https://api.datadoghq.com" DD_TEST_ORG="<public_org_id>"  TESTARGS="-run TestAccDatadogIntegrationGCP" RECORD=true make testacc

DD_TEST_CLIENT_APP_KEY="<prod-app-key>" DD_TEST_CLIENT_API_KEY="<prod-api-key>" 
DD_TEST_SITE_URL="https://api.datadoghq.com" DD_TEST_ORG="<public_org_id>" TESTARGS="-run TestAccIntegrationGcpStsBasic" RECORD=true make testacc 
 
DD_TEST_CLIENT_APP_KEY="<prod-app-key>" DD_TEST_CLIENT_API_KEY="<prod-api-key>" 
DD_TEST_SITE_URL="https://api.datadoghq.com" DD_TEST_ORG="<public_org_id>" TESTARGS="-run TestAccIntegrationGcpStsDefault" RECORD=true make testacc

```


datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -88,6 +89,11 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 				Description: "Your Host Filters.",
 				ElementType: types.StringType,
 			},
+			"cloud_run_revision_filters": schema.SetAttribute{
+				Optional:    true,
+				Description: "Tags to filter which Cloud Run revisions are imported into Datadog. Only revisions that meet specified criteria will be monitored.",
```
buraizu, line 94:
```suggestion
				Description: "Tags to filter which Cloud Run revisions are imported into Datadog. Only revisions that meet specified criteria are monitored.",
```

docs/resources/integration_gcp_sts.md
```diff
@@ -48,6 +48,7 @@ resource "datadog_integration_gcp_sts" "foo" {
 
 - `account_tags` (Set of String) Tags to be associated with GCP metrics and service checks from your account.
 - `automute` (Boolean) Silence monitors for expected GCE instance shutdowns.
+- `cloud_run_revision_filters` (Set of String) Tags to filter which Cloud Run revisions are imported into Datadog. Only revisions that meet specified criteria will be monitored.
```
buraizu, line 51:
```suggestion
- `cloud_run_revision_filters` (Set of String) Tags to filter which Cloud Run revisions are imported into Datadog. Only revisions that meet specified criteria are monitored.
```

buraizu (approved):
Approving with a couple of minor updates requested

dtru-ddog:
Don't forget to update the non-sts module: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/fwprovider/resource_datadog_integration_gcp.go

buraizu (approved):

datadog/fwprovider/resource_datadog_integration_gcp.go
```diff
@@ -299,6 +303,9 @@ func (r *integrationGcpResource) updateState(ctx context.Context, state *integra
 	if privateKeyId, ok := resp.GetPrivateKeyIdOk(); ok {
 		state.PrivateKeyId = types.StringValue(*privateKeyId)
 	}
+	if runFilters, ok := resp.GetCloudRunRevisionFiltersOk(); ok && len(*runFilters) > 0 {
+		state.CloudRunRevisionFilters, _ = types.SetValueFrom(ctx, types.StringType, *runFilters)
```
ash-ddog, line 307: Why ignore the `err` here? Not familiar with the terraform library, but seems unsafe?

datadog/fwprovider/resource_datadog_integration_gcp.go
```diff
@@ -341,12 +348,22 @@ func (r *integrationGcpResource) addRequiredFieldsToBody(body *datadogV1.GCPAcco
...
 	body.SetAutomute(state.Automute.ValueBool())
 	body.SetIsCspmEnabled(state.CspmResourceCollectionEnabled.ValueBool())
 	body.SetIsSecurityCommandCenterEnabled(state.IsSecurityCommandCenterEnabled.ValueBool())
 	body.SetHostFilters(state.HostFilters.ValueString())
+
+	runFilters := make([]string, 0)
+	if !state.CloudRunRevisionFilters.IsNull() {
+		diags.Append(state.CloudRunRevisionFilters.ElementsAs(ctx, &runFilters, false)...)
+	}
+	body.SetCloudRunRevisionFilters(runFilters)
```
ash-ddog, line 362: Is this safe? Won't this send an empty array even if the state does not have filters set? FYI the dogweb v2 API takes whatever you give it and sets it. So, if you send an empty array, you'll overwrite any existing filters.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -174,6 +180,10 @@ func (r *integrationGcpStsResource) Create(ctx context.Context, request resource
 	state.DelegateAccountEmail = types.StringValue(delegateEmail)
 
 	attributes, diags := r.buildIntegrationGcpStsRequestBody(ctx, &state)
+	if !state.ClientEmail.IsNull() {
+		attributes.SetClientEmail(state.ClientEmail.ValueString())
+	}
```
ash-ddog, line 185: Just FYI: the create endpoint will fail without `client_email` (it's required), so maybe we don't even send it?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -272,6 +282,9 @@ func (r *integrationGcpStsResource) updateState(ctx context.Context, state *inte
 	if hostFilters, ok := attributes.GetHostFiltersOk(); ok && len(*hostFilters) > 0 {
 		state.HostFilters, _ = types.SetValueFrom(ctx, types.StringType, *hostFilters)
 	}
+	if runFilters, ok := attributes.GetCloudRunRevisionFiltersOk(); ok && len(*runFilters) > 0 {
+		state.CloudRunRevisionFilters, _ = types.SetValueFrom(ctx, types.StringType, *runFilters)
```
ash-ddog, line 286: Same: ignoring error

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -296,9 +309,6 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
 	if !state.Automute.IsNull() {
 		attributes.SetAutomute(state.Automute.ValueBool())
 	}
-	if !state.ClientEmail.IsNull() {
-		attributes.SetClientEmail(state.ClientEmail.ValueString())
-	}
```
ash-ddog, line 301: 👍 🙏

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -309,6 +319,12 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
 	}
 	attributes.SetHostFilters(hostFilters)
 
+	runFilters := make([]string, 0)
+	if !state.CloudRunRevisionFilters.IsNull() {
+		diags.Append(state.CloudRunRevisionFilters.ElementsAs(ctx, &runFilters, false)...)
+	}
+	attributes.SetCloudRunRevisionFilters(runFilters)
```
ash-ddog, line 326: Same: sending an empty array.

ash-ddog (changes requested):
Mostly looks good, just had a couple of questions. 🙏

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
tedkahwaji: So if the CloudRunRevisionFilters is set, `ElementsAs(ctx, &runFilters, false)...)`

Will populate the TF Set with the values of the runFilters; in the event that `IsNull` is true then the intended behavior is to send an empty array

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
tedkahwaji: That's the intention we verify no errors exist in the diagnostics prior to making this call [here](https://github.com/DataDog/terraform-provider-datadog/blob/b8a14093ba2ef0fcdcfbac3417fc7b0af9898a5e/datadog/fwprovider/resource_datadog_integration_gcp_sts.go#L227-L227)

datadog/fwprovider/resource_datadog_integration_gcp.go
ash-ddog: Spoke offline. The 2nd return param is not `err` but is `diag.Diagnostics`, which I _think_ can represent an error. However, looking all over this repo, it seems the norm is to ignore the 2nd return param. Will look into more later but good for now.

datadog/fwprovider/resource_datadog_integration_gcp.go
ash-ddog: Spoke offline. I think we're good here. If `cloud_run_revision_filters` is absent from the terraform object, then we should delete them. 👍

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: Spoke offline. We prob could make an optimization here in bailing on the request since this will fail. But this is what existed, so good with it for now.

Some follow-ups on the dogweb side:
- `client_email` is required, but can be blank, and ICS will reject that request, which means dogweb returns 500. Not good, should be 400
- dogweb should probably do the validation for `client_email` so we don't even go to ICS

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: https://github.com/DataDog/terraform-provider-datadog/pull/2612#discussion_r1797400708

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog: https://github.com/DataDog/terraform-provider-datadog/pull/2612#discussion_r1797401223

ash-ddog (approved):
After talking through comments, LGTM, thanks! 🙏

nkzou (approved):

PR #2621, opened 2024-10-16
[datadog_synthetics_test] Add mobile support

teodor2312:
This PR adds support for mobile tests


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -278,7 +285,7 @@ func syntheticsTestRequestHeaders() *schema.Schema {
 		Description:  "Header name and value map.",
 		Type:         schema.TypeMap,
 		Optional:     true,
-		ValidateFunc: validators.ValidateHttpRequestHeader,
+		ValidateFunc: validators.ValidateHttpRequestHeader, // TODO we might need one of these for initialApplicationArguments
```
etnbrd, line 288: No, the `ValidateHttpRequestHeader` is to check that the values are compatible with http (e.g. no invisible characters). There is no need for such checks in application arguments, I think.
```suggestion
		ValidateFunc: validators.ValidateHttpRequestHeader, // TODO we might need one of these for initialApplicationArguments
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -819,6 +826,164 @@ func syntheticsTestOptionsList() *schema.Schema {
 	}
 }
 
+func syntheticsMobileTestOptionsList() *schema.Schema {
+	return &schema.Schema{
```
etnbrd, line 830: Maybe later we could reuse the `syntheticsTestOptionsList`  and add just the fields we need?
```
schema := syntheticsTestOptionsList()
schema.Elem.Schema["allow_application_crash"] = &schema.Schema{
  Type:     schema.TypeBool,
  Optional: true,
}
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2375,6 +2855,66 @@ func buildDatadogSyntheticsBrowserTest(d *schema.ResourceData) *datadogV1.Synthe
 	return syntheticsTest
 }
 
+func buildDatadogSyntheticsMobileTest(d *schema.ResourceData) *datadogV1.SyntheticsMobileTest {
+	syntheticsTest := datadogV1.NewSyntheticsMobileTestWithDefaults()
+	if attr, ok := d.GetOk("message"); ok {
+		syntheticsTest.SetMessage(attr.(string))
+	}
+	// if attr, ok := d.GetOk("monitor_id"); ok { // TODO uncomment and find out why it's unexpected
```
etnbrd, line 2863: ?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2375,6 +2855,66 @@ func buildDatadogSyntheticsBrowserTest(d *schema.ResourceData) *datadogV1.Synthe
...
+			config.SetInitialApplicationArguments(make(map[string]string))
+		}
+		for k, v := range initialApplicationArguments {
+			config.GetInitialApplicationArguments()[k] = v.(string)
+		}
+	}
+
+	syntheticsTest.SetConfig(config)
+
+	// if attr, ok := d.GetOk("device_ids"); ok { // TODO we don't want to build this because the response doesn't contain it and if you build it the local state goes all whack
```
etnbrd, line 2893: You mention the response and the state in this comment, but it's in the `buildDatadogSyntheticsMobileTest` function, which reads the state to build the request.

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3386,6 +4074,155 @@ func buildTerraformTestOptions(actualOptions datadogV1.SyntheticsTestOptions) []
 	return localOptionsLists
 }
 
+func buildTerraformMobileTestOptions(actualOptions datadogV1.SyntheticsMobileTestOptions) []map[string]interface{} {
```
etnbrd, line 4077: Let's move these functions to have the following outline:
```
...
buildDatadadogTestOptions
buildTerraformTestOptions
buildDatadogMobileTestOptions
buildTerraformMobileTestOptions
...
```

etnbrd (dismissed):
LGTM 👍 
Left a few comments to remove dead code, TODO comments and reorganise a bit the file, but nothing blocking.

datadog/resource_datadog_synthetics_test_.go
teodor2312: I think i asked you about this to confirm a few days ago, but i thought about it a bit more and the id is generated from the backend so doesn't really make sense to put in in the request 🤔

datadog/resource_datadog_synthetics_test_.go
teodor2312: If i remember correctly the issue here was that if put it in the request then in the end of the test it complained that this was not parsed in the response 🤔

datadog/resource_datadog_synthetics_test_.go
teodor2312: Yes that would be nice :D

cswatt (dismissed):

PR #2633, opened 2024-10-28
Add Resources to Cloudflare Terraform Provider

apiazza-dd:
This PR allows customers to provision resources for their Cloudflare Accounts using Terraform.


docs/resources/integration_cloudflare_account.md
```diff
@@ -33,6 +33,7 @@ resource "datadog_integration_cloudflare_account" "foo" {
 ### Optional
 
 - `email` (String) The email associated with the Cloudflare account. If an API key is provided (and not a token), this field is also required.
+- `resources` (List of String) An allowlist of resources to restrict pulling metrics for including 'web', 'dns', 'lb' (load balancer), 'worker')
```
maycmlee, line 36: Are `web`, `dns`, etc. tags on the metrics? Or something else?

```suggestion
- `resources` (List of String) An allowlist of resources to restrict pulling metrics for including `web`, `dns`, `lb` (load balancer), 'worker')
```

maycmlee (commented):
Small suggestions and a question.

docs/resources/integration_cloudflare_account.md
apiazza-dd: These are enabling which resources to collect when provisioning a cloudflare account via terraform.

apiazza-dd: These docs are also autogenerated

docs/resources/integration_cloudflare_account.md
```diff
@@ -33,6 +33,7 @@ resource "datadog_integration_cloudflare_account" "foo" {
 ### Optional
 
 - `email` (String) The email associated with the Cloudflare account. If an API key is provided (and not a token), this field is also required.
+- `resources` (List of String) An allowlist of resources to restrict pulling metrics for including `web`, `dns`, `lb` (load balancer), `worker`)
```
maycmlee, line 36:
```suggestion
- `resources` (List of String) An allowlist of resources, such as `web`, `dns`, `lb` (load balancer), `worker`, to pull metrics for.
```

maycmlee (commented):
How about this?

alpinet (dismissed):

docs/resources/integration_cloudflare_account.md
apiazza-dd: I think the restriction part is important here, by default, they're set to all, but if they want one turned off, then they use the resources list. Phrasing as an allow list makes it seem like if they specify one, they'll collect metrics for it, when it's the opposite.

maycmlee: @apiazza-dd Ah okay, thanks for the clarification. I updated the suggestion, what do you think?

maycmlee: @apiazza-dd actually rereading your comment, I might have misunderstood.

> but if they want one turned off, then they use the resources list. 

Is the allowlist of resources, the resources that users want to pull metrics for or the ones they don't want pulling metrics for?

ZacharyAnderson (dismissed):

docs/resources/integration_cloudflare_account.md
apiazza-dd: @maycmlee These docs are autogenerated and have been updated to match our other repository. Does this look okay, or should I make a change in the other repo?

maycmlee: DM'd. Docs are autogenerated so the suggestion will be updated in the other repo.

maycmlee (approved):

ZacharyAnderson (dismissed):

docs/resources/integration_cloudflare_account.md
apiazza-dd: https://github.com/DataDog/datadog-api-spec/pull/3270/files

gunterd (approved):

maycmlee (dismissed):

gunterd (approved):

gunterd (approved):

skarimo (approved):

PR #2644, opened 2024-11-04
Update the version.

jack-edmonds-dd:


drichards-87 (approved):

datadog/resource_datadog_dashboard.go
```diff
@@ -8068,7 +8068,6 @@ func buildDatadogToplistStyle(terraformToplistStyle map[string]interface{}) data
 				if t == "stacked" {
 					datadogToplistStyle.SetDisplay(datadogV1.ToplistWidgetDisplay{
 						ToplistWidgetStacked: datadogV1.NewToplistWidgetStacked(
-							datadogV1.TOPLISTWIDGETLEGEND_AUTOMATIC,
```
Drarig29, line 8071: Since the `legend` doesn't have a default value, the request will lose `"legend":"automatic"`? Is it safe to remove it, or should we `.SetLegend(datadogV1.TOPLISTWIDGETLEGEND_AUTOMATIC)`? Does the backend default to automatic or none? cc. @brtu

skarimo: Reverted this change and added back the old defaults

skarimo (approved):

PR #2667, opened 2024-11-13
Make recommended keywords an opt-out option in terraform

artslidd:
## Description

Make recommended keywords an opt-out option from terraform.

See the full explanation of what we are doing at https://datadoghq.atlassian.net/wiki/spaces/SDS/pages/4373446920/Use+opt-out+strategy+for+keyword+config+in+Terraform

We decided to go for option 1, option where we don't expose the new field called `use_recommended_keywords` in terraform directly, but depending on what the user decides to do, let's make the field true, or false, or leave it empty


HantingZhang2 (approved):

fuchsnj (approved):

michaelcretzman (approved):

m-shykula (approved):
👍

Issue #2471, opened 2024-07-02
Synthetics Test Multi-steps client certificate config broken when applied second time


ibmboy19:
### Datadog Terraform Provider Version

v3.39.0

### Terraform Version

v1.8.3

### What resources or data sources are affected?

datadog_synthetics_test - multi-steps

### Expected Behavior

The client certificate config should still working

### Actual Behavior

The client certificate config broken and the API test can't pass the client cer authentication

### Steps to Reproduce

1. Newly terraform apply the Multi-steps synthetic test
2. Modify something on the Multi-steps synthetic test and apply again
3. Trigger the synthetic test and the test start to fail

Drarig29:
Hi @ibmboy19! Is this bug still happening in the latest version of the provider?

ibmboy19:
@Drarig29 Yes, this bug also still exists in v3.46.0.

Drarig29:
Thanks, we added a card in our backlog. We'll get back to you once it's fixed.

AntoineDona:
Hi @ibmboy19! 
Thanks for for reporting this bug.
We made a fix that will be present in the next release of the terraform provider (probably next week).

ibmboy19:
Hi @AntoineDona, 
Thank you so much for your efforts!

AntoineDona:
Hey @ibmboy19 ! The fix was just released in [v3.50.0](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.50.0) of the terraform-provider. Thanks again for reporting the issue!

PR #2683, opened 2024-11-19
[datadog_synthetics_test] Fix multistep client certificate

AntoineDona:
### Description

This PR was created in response to [this issue](https://github.com/DataDog/terraform-provider-datadog/issues/2471) where applying a multistep test with a client certificate multiple times would result in the certificate being deleted in the backend.
The tricky thing here is that the certificate is never stored in the tf state for security reasons, and we pass an empty value to the backend if its value does not change. Because we don't have the step ids in terraform, the backend can't partially update the step and just overrides everything, thus the certificate is destroyed.

As a fix, we decided to pass the certificate on each tf update.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3783,3 +3792,71 @@ func validateSyntheticsAssertionOperator(val interface{}, key string) (warns []s
 	}
 	return
 }
+
+func getConfigCertificate(rawConfig cty.Value, stepIndex int) (*string, *string) {
+
```
Drarig29, line 3796: We don't have blank lines at the beginning of functions in the file
```suggestion
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3783,3 +3792,71 @@ func validateSyntheticsAssertionOperator(val interface{}, key string) (warns []s
...
+	// Convert certValue and keyValue to strings
+	certString := certValue.AsString()
+	keyString := keyValue.AsString()
+
+	return &certString, &keyString
+}
+
+func overrideStateCertificate(requestClientCertificates []interface{}, configCert, configKey string) error {
+
```
Drarig29, line 3832: ditto
```suggestion
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3783,3 +3791,71 @@ func validateSyntheticsAssertionOperator(val interface{}, key string) (warns []s
...
+	if len(requestClientCertificates) == 0 {
+		return fmt.Errorf("requestClientCertificates is empty")
+	}
+	requestClientCertificate, ok := requestClientCertificates[0].(map[string]interface{})
+	if !ok {
```
Drarig29, line 3837: It's idiomatic in Go to have the assignments in the `if` statement:
```suggestion
	if requestClientCertificate, ok := requestClientCertificates[0].(map[string]interface{}); !ok {
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3783,3 +3791,71 @@ func validateSyntheticsAssertionOperator(val interface{}, key string) (warns []s
...
+	}
+
+	// Convert certValue and keyValue to strings
+	certString := certValue.AsString()
+	keyString := keyValue.AsString()
+
+	return &certString, &keyString
+}
+
+func overrideStateCertificate(requestClientCertificates []interface{}, configCert, configKey string) error {
```
Drarig29, line 3831: You don't do anything with the returned error - is it expected?

Drarig29 (commented):
Great work! Do you have an idea of how to capture this in unit tests?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3783,3 +3791,71 @@ func validateSyntheticsAssertionOperator(val interface{}, key string) (warns []s
...
+		return fmt.Errorf("key is not a valid list or is empty")
+	}
+	key, ok := keyList[0].(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("key[0] is not a map")
+	}
+	key["content"] = configKey
+
+	return nil
+}
```
etnbrd, line 3861: Mutation is to be avoided as much as possible. Instead, let's have this function return the `requestClientCertificates` interface, and override it in the caller.
```suggestion
func buildDatadogRequestClientCertificate(configCert, configKey string)  {
	requestClientCertificates := datadogV1. ... TODO

	certList, ok := requestClientCertificate["cert"].([]interface{})
	cert, ok := certList[0].(map[string]interface{})
	if !ok {
		return fmt.Errorf("cert[0] is not a map")
	}
	cert["content"] = configCert

	keyList, ok := requestClientCertificate["key"].([]interface{})
	if !ok || len(keyList) == 0 {
		return fmt.Errorf("key is not a valid list or is empty")
	}
	key, ok := keyList[0].(map[string]interface{})
	if !ok {
		return fmt.Errorf("key[0] is not a map")
	}
	key["content"] = configKey

	return requestClientCertificates
}
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2123,6 +2125,12 @@ func buildDatadogSyntheticsAPITest(d *schema.ResourceData) *datadogV1.Synthetics
 						}
 					}
 				}
+				// Override the request client certificate with the one from the config
+				rawConfig := d.GetRawConfig()
+				configCert, configKey := getConfigCertificate(rawConfig, i)
+				if configCert != nil && configKey != nil {
+					overrideStateCertificate(stepMap["request_client_certificate"].([]interface{}), *configCert, *configKey)
```
etnbrd, line 2132: Let's keep this function pure, just returning the right structure, and overriding here.

etnbrd (approved):
Overall LGTM, great work!
Left a few comments on structure to keep the provider implementation moving into the right direction.

datadog/resource_datadog_synthetics_test_.go
AntoineDona: Doing so means requestClientCertificate is undefined in the rest of the code I think... It's only defined in the if, and I wanted to avoid having nested ifs

datadog/resource_datadog_synthetics_test_.go
AntoineDona: Agreed that we should avoid the mutation. However, `completeSyntheticsTestRequest` is not expecting a nicely formatted `requestClientCertificates`, but rather a raw `stepMap["request_client_certificate"]` with a more complexe structure. So the `buildDatadogRequestClientCertificate` does not really make sense, even more that we already have the `buildDatadogRequestCertificates` function which kinda does the same thing, inside of `completeSyntheticsTestRequest`.

etnbrd (approved):
I understand it's not the cleanest, but it's good enough like that. With `buildDatadogRequestCertificates` aligned, we'll try to align further the retrieval later.
Great job! Thanks a lot for taking into account our reviews 🙇

etnbrd (approved):
🚀

PR #2703, opened 2024-12-04
[datadog_monitor] Allow default_tags values to contain colons.

Amaury-Behague:
Only separate the key before the first colon, all subsequent colons are
considered to be part of the value as stated in datadog docs.


datadog/provider.go
```diff
@@ -470,12 +470,14 @@ func tagDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{}) erro
 		kv := strings.Split(tag.(string), ":")
```
LorisFriedel, line 470: You could probably use `strings.Cut` (or even `strings.SplitN`) and avoid having to `strings.Join` afterwards. This would probably allow to simplify the whole check logic!

LorisFriedel: Like so:
```go
    tagStr := tag.(string)
    if tagStr == "" {
        return fmt.Errorf("invalid tag: '%s'", tag)
    }
    key, value, found := strings.Cut(tagStr, ":")
    if !found {
        value = ""
    }
    tags[key] = value
```

Amaury-Behague: Good point - I always forget about this function.

Amaury-Behague: Done - I've even removed the special case for empty tags because it does not break the code.

urseberry (approved):

nkzou (approved):

PR #2705, opened 2024-12-04
[resource_datadog_webhook] Fix terraform import for datadog_webhook resource

fl0ge:
The `terraform import datadog_webhook.foo example-webhook` provided example was not working because in the `datadog_webhook` resource, Name was used and it is not suitable for import.

Indeed, since the name and the id of datadog_webhook resource are the same once the resource is created, and that during the import phase, only id is given as a parameter, replacing Name by ID will make it work for both cases.


amaskara-dd:
thank you for bringing this to our attention — this was handled in https://github.com/DataDog/terraform-provider-datadog/pull/2742

PR #2710, opened 2024-12-09
[resource_datadog_role] ACCESS-2867 - Support `default_permissions_opt_out` attribute in Terraform

wangwillson1:
Updates the datadog_role Terraform resource to support the `default_permissions_opt_out` request attribute. This enables TF users to manage all permissions on roles, including restricted permissions.

Tested with the following config and verified that the role configs were correct:
```
resource "datadog_role" "perms" {
  name = "restricted perms"
}

resource "datadog_role" "noperm" {
  name = "no perms"
  default_permissions_opt_out = true
}

resource "datadog_role" "dashboards_read_role" {
  name = "dashboards read only"
  permission {
    id = data.datadog_permissions.bar.permissions.dashboards_read
  }
  default_permissions_opt_out = true
}
```

also validated the following error cases:
```
data "datadog_permissions" "unrestricted_perms" {}

resource "datadog_role" "role_invalid_perm" {
  name = "dashboards read only"
  permission {
    id = "invalid id"
  }
  default_permissions_opt_out = true
}

resource "datadog_role" "role_invalid_restricted_perm" {
  name = "dashboards read only"
  permission {
    id = data.datadog_permissions.unrestricted_perms.permissions.dashboards_read
  }
  default_permissions_opt_out = true
}
```


docs/resources/role.md
```diff
@@ -37,6 +37,7 @@ resource "datadog_role" "foo" {
 
 ### Optional
 
+- `default_permissions_opt_out` (Boolean) If set to `true`, the role will not have default (restricted) permissions unless they are explicitly set. The `include_restricted` attribute for the datadog_permissions data source must be set to `true` to manage default permissions in Terraform
 - `permission` (Block Set) Set of objects containing the permission ID and the name of the permissions granted to this role. (see [below for nested schema](#nestedblock--permission))
 - `validate` (Boolean) If set to `false`, skip the validation call done during plan.
```
neko-dd, line 42:
```suggestion
- `default_permissions_opt_out` (Boolean) - If set to `true`, the role does not have default (restricted) permissions unless they are explicitly set. The `include_restricted` attribute for the `datadog_permissions` data source must be set to `true` to manage default permissions in Terraform.
- `permission` (Block Set) - Set of objects containing the permission ID and the name of the permissions granted to this role. (See below for [nested schema](#nestedblock--permission).)
- `validate` (Boolean) - If set to `false`, skip the validation call done during plan.
```

neko-dd (approved):
A few suggested edits :)

docs/resources/role.md
wangwillson1: i think most of these are autogenerated so i can't directly modify the markdown, but i made the wording changes to the `default_permissions_opt_out` description as requested (will -> does, added period)

neko-dd (dismissed):

scripts/generate-docs.sh
```diff
@@ -2,7 +2,9 @@
 # Generate the documentation using tfplugindocs and remove changes to files that shouldn't change
 
 # Add here the files to be excluded from the doc generation
-exclude_files=()
+exclude_files=(
+  "docs/data-sources/role.md"
```
skarimo, line 6:
```suggestion
  "docs/resources/role.md"
```

datadog/tests/resource_datadog_role_test.go
```diff
@@ -171,16 +222,26 @@ resource "datadog_role" "foo" {
...
 	return fmt.Sprintf(`
 data "datadog_permissions" foo {}
 
 resource "datadog_role" "foo" {
   name      = "%s"
+  default_permissions_opt_out = true
 }`, uniq)
 }
 
-func testAccCheckDatadogRoleConfigRestrictedPerm(uniq string) string {
```
bi1yeu, line 183: hmm this test name is curious. it wasn't configuring with a restricted permission but a nonexistent permission. tho i guess if the permission data source didn't include restricted permissions they would've been "invalid" permission names

datadog/resource_datadog_role.go
```diff
@@ -228,14 +250,15 @@ func updateRolePermissionsState(ctx context.Context, d *schema.ResourceData, rol
...
+func appendPerm(perms []map[string]string, permID string, permIDToAttributes map[string]PermAttributes, isOptedOutOfDefaultPermissions bool) []map[string]string {
+	if permAttributes, ok := permIDToAttributes[permID]; ok {
 		permR := map[string]string{
 			"id":   permID,
-			"name": permName,
+			"name": permAttributes.Name,
+		}
+		if isOptedOutOfDefaultPermissions || !permAttributes.IsRestricted {
+			perms = append(perms, permR)
 		}
```
bi1yeu, line 261: just want to make sure i understand this part. if it's (a) _not_ opted out of default permissions (i.e., the role will include all restricted permissions), and (b) the permission is restricted, then that permission will _not_ be added to the request. that's because the backend will automatically add _all_ restricted permissions, so they shouldn't be explicitly provided. did i get that right? might be worth a code comment bc it's a little counterintuitive

datadog/resource_datadog_role.go
```diff
@@ -15,8 +15,13 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-// validPermissions is a map of all unrestricted permission IDs to their name
-var validPermissions map[string]string
+type PermAttributes struct {
+	Name         string
+	IsRestricted bool
```
bi1yeu, line 20: jw why did we decide to call the attribute `restricted` but the role opt-out field is called `default`? i'd imagine these should be the same word. especially now since the permission is no longer fully restricted

datadog/resource_datadog_role.go
```diff
@@ -36,6 +41,11 @@ func resourceDatadogRole() *schema.Resource {
 					Required:    true,
 					Description: "Name of the role.",
 				},
+				"default_permissions_opt_out": {
+					Type:        schema.TypeBool,
+					Optional:    true,
+					Description: "If set to `true`, the role does not have default (restricted) permissions unless they are explicitly set. The `include_restricted` attribute for the `datadog_permissions` data source must be set to `true` to manage default permissions in Terraform.",
```
bi1yeu, line 47: > The `include_restricted` attribute for the `datadog_permissions` data source must be set to `true` to manage default permissions in Terraform.

interesting. it seems to me like we might want to make the datadog permissions datasource always include restricted permissions now. is there any reason to omit them?

datadog/resource_datadog_role.go
wangwillson1: yeah this is a good point, i called it `restricted` here because in the `datadog_permissions` resource we also use the term "restricted"

i think we _should_ be using `default` here but i worry about the consistency w/ our other resource(s). if this isn't a major concern tho i'll change it here

datadog/resource_datadog_role.go
wangwillson1: agree, this seems like a backwards-compatible change so it's probably not a major risk?

datadog/resource_datadog_role.go
wangwillson1: yes your understanding is correct, i'll add code comment to make it more clear

datadog/tests/resource_datadog_role_test.go
wangwillson1: yeah lol, i was also initially confused by the test name (hence the rename + addition of a new test)

datadog/resource_datadog_role.go
bi1yeu: yeah if we can gain confidence it's safe to make that change i think it would be preferable as it simplifies the interface

nkzou:
it seems like the role resource drifts continuously when trying to manage a restricted permissions. For example, with 
```hcl
resource "datadog_role" "name" {
    name = "test"
    default_permissions_opt_out = true
    permission {
        id = "f6e917aa-8502-11ec-bf20-da7ad0900002" // monitors_read
    }
}
```
on apply, the backend seems to be adding in a bunch of other restricted permissions, and on the next terraform plan, terraform thinks it needs to remove the other ones.

wangwillson1:
> on apply, the backend seems to be adding in a bunch of other restricted permissions, and on the next terraform plan, terraform thinks it needs to remove the other ones.

discussed in slack, this seems to be a result of the FF not being enabled in the org

datadog/resource_datadog_role.go
wangwillson1: from discussion in slack:
> changing the data source to expose restricted permissions by default is riskier to do, especially if not all customers have access to this outright

since we're currently in a state where an FF needs to be enabled for these changes to work, we probably shouldn't expose all the permissions _yet_ since it might lead them to confusion when a role gets created w/ restricted perms anyways

datadog/resource_datadog_role.go
```diff
@@ -123,9 +133,19 @@ func resourceDatadogRoleCustomizeDiff(ctx context.Context, diff *schema.Resource
...
+
+		if !permissionExists {
 			return fmt.Errorf(
-				"permission with ID %s is restricted and cannot be managed by terraform or does not exist, remove it from your configuration",
+				"permission with ID %s does not exist, remove it from your configuration",
+				permID,
+			)
+		}
+
+		if permAttributes.IsDefaultPermission && !defaultPermissionsOptOut.(bool) {
```
skarimo, line 146: Nit. Should this change be reverted to avoid new users from trying to use it

skarimo (approved):

datadog/resource_datadog_role.go
wangwillson1: good call! will keep this line of code for now and assess after code freeze. if we haven't started any of the rollout process yet, i'll remove it.

bi1yeu (approved):

neko-dd (approved):
Looks like most of what there was for docs to review is gone now. :) The rest LGTM

skarimo (approved):

PR #2713, opened 2024-12-10
fix(github-workflows): use official GitHub action for token generation

bthuilot:
# Changes

- Replaces usage of `tibdex/github-app-token` with official GitHub action


bthuilot:
Dismissing "pin by tag" due to sensitivity of action and lack of dependabot

jack-edmonds-dd (approved):

PR #2716, opened 2024-12-11
[resource_datadog_logs_index] Fix having zero retention logs indexes with flex

Prepager:
Motivation
---
The resource currently has two issues:
- On creation: You can not create an index with only flex retention and no standard retention defined.
- On update: You can not remove the standard retention to have only the flex retention.

Changes
---
Context: While the endpoint is marked as a `PUT` it acts as a `PATCH` ([docs](https://docs.datadoghq.com/api/latest/logs-indexes/#update-an-index:~:text=Using%20the%20PUT%20method%20updates%20your%20index%E2%80%99s%20configuration%20by%20replacing%20your%20current%20configuration%20with%20the%20new%20one%20sent%20to%20your%20Datadog%20organization.))

**On creation**
If the resource has a `flex_retention_days` set, but no `retention_days`, it will default to `0` for `retention_days`.

**On update**
The previous `GetOk` check would turn `0` into `null`, making it impossible to remove the `retention_days`.
The new condition does an explicit IsNull check to allow through `0` values.


brunorenier (approved):

corentinmarc (approved):

arc013 (approved):
I'm doing a blind approval to not block you but I'm a little unsure why this change requires logs-forwarding's approval as it seems to be about index (which logs-forwarding team has close to zero context on).

jack-edmonds-dd (approved):

PR #2718, opened 2024-12-16
[datadog_action_connection] ACTP-104 Terraform provider for Connections API

embeaken:
This is the TF provider for the new [public Connections API](https://docs.datadoghq.com/api/latest/action-connection/).

It includes a data source for using Connections outside of resource definitions and a resource for managing Connections. It also includes support for importing existing connections as resources.

Todo:
- [x] ~~Deploy https://github.com/DataDog/dd-source/pull/173632~~
- [x] Update `datadog-api-client-go` to a specific version once it's released
- [x] Re-record test cassettes with proper test org (that's why tests don't pass now)

For follow-up PR:
* Prepare release as per RELEASING.md in this repo


docs/data-sources/connection.md
```diff
@@ -0,0 +1,99 @@
...
+Read-Only:
+
+- `assume_role` (Block, Read-only) Configuration for an assume role AWS connection (see [below for nested schema](#nestedblock--aws--assume_role))
+
+<a id="nestedblock--aws--assume_role"></a>
+### Nested Schema for `aws.assume_role`
+
+Read-Only:
+
+- `account_id` (String) AWS account the connection is created for
```
drichards-87, line 40:
```suggestion
- `account_id` (String) AWS account that the connection is established for
```

docs/data-sources/connection.md
```diff
@@ -0,0 +1,99 @@
...
+- `external_id` (String) External ID used to scope which connection can be used to assume the role
```
drichards-87, line 41:
```suggestion
- `external_id` (String) External ID that specifies which connection can be used to assume the role
```

docs/data-sources/connection.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+
+<a id="nestedblock--http"></a>
+### Nested Schema for `http`
+
+Read-Only:
+
+- `base_url` (String) Base HTTP url for the integration
+- `token_auth` (Block, Read-only) Configuration for an HTTP connection using token auth (see [below for nested schema](#nestedblock--http--token_auth))
```
drichards-87, line 53:
```suggestion
- `token_auth` (Block, Read-only) Configuration for an HTTP connection that uses token auth (see [below for nested schema](#nestedblock--http--token_auth))
```

docs/resources/connection.md
```diff
@@ -0,0 +1,105 @@
...
+Optional:
+
+- `assume_role` (Block, Optional) Configuration for an assume role AWS connection (see [below for nested schema](#nestedblock--aws--assume_role))
+
+<a id="nestedblock--aws--assume_role"></a>
+### Nested Schema for `aws.assume_role`
+
+Optional:
+
+- `account_id` (String) AWS account the connection is created for
```
drichards-87, line 43:
```suggestion
- `account_id` (String) AWS account that the connection is created for
```

drichards-87 (approved):
Left some suggestions from Docs and approved the PR.

datadog/fwprovider/data_source_datadog_connection.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (d *connectionDatasource) Configure(_ context.Context, request datasource.ConfigureRequest, response *datasource.ConfigureResponse) {
+	providerData := request.ProviderData.(*FrameworkProvider)
+	d.Api = providerData.DatadogApiInstances.GetActionConnectionApiV2()
+	d.Auth = providerData.Auth
+}
+
+func (d *connectionDatasource) Metadata(_ context.Context, request datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = "connection"
```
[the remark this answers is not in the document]
embeaken: renamed it today!

datadog/fwprovider/resource_datadog_connection.go
```diff
@@ -0,0 +1,758 @@
...
+			len(conn.HTTP.TokenAuth.URLParameters) == 0 &&
+			conn.HTTP.TokenAuth.Body == nil {
+			response.Diagnostics.AddAttributeError(
+				path.Root("http").AtName("token_auth"),
+				"Credential information required",
+				"You must specify at least one of: tokens, headers, URL parameters, body.",
+			)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
embeaken: i changed this today to check for at least one token to be defined + at least one of headers/url params/body.

for base URL i can add a custom check

datadog/fwprovider/resource_datadog_connection.go
```diff
@@ -0,0 +1,758 @@
...
+}
+
+func (r *connectionResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = "connection"
+}
+
+func (r *connectionResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "A connection that can be used in Actions, including in the Workflow Automation and App Builder products.",
+		Attributes: map[string]schema.Attribute{
```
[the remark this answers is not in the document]
embeaken: sadly no cause the types are different. one is a `resources.Schema` and the other one is `datasources.Schema`. i might look into some workarounds for this once we're getting more connection types implemented

datadog/fwprovider/resource_datadog_connection.go
```diff
@@ -0,0 +1,758 @@
...
+	conn, httpResponse, err := r.Api.CreateActionConnection(r.Auth, *createRequest)
+	if err != nil {
+		if httpResponse != nil {
+			// error body may have useful info for the user
+			body, err := io.ReadAll(httpResponse.Body)
+			if err != nil {
+				response.Diagnostics.AddError("Could not read error response", err.Error())
+				return
+			}
+			response.Diagnostics.AddError("Could not create connection", string(body))
```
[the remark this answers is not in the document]
embeaken: there's a return on line 354 (in the current code) that will catch this

datadog/fwprovider/resource_datadog_connection.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+
+	data := datadogV2.NewActionConnectionDataUpdate(*attributes, datadogV2.ACTIONCONNECTIONDATATYPE_ACTION_CONNECTION)
+	req := datadogV2.NewUpdateActionConnectionRequest(*data)
+
+	return req, nil
+}
+
+// The connections API handles deletions of tokens, headers, and URL params with a "deleted" flag instead of
+// just exclusion from the request body as is common in a PUT request. So some more work is required to
```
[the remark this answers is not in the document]
embeaken: TF doesn't act as a `PATCH` since on every `apply` it plans out the entire new state, not just the subset of the state that changed. So `apply` acts like a `PUT` because we're building the update request with the entire connection, not just the updates.

So if we want to do this without a delete flag we would need the API to act as more of a `PUT` where what we give as the input is essentially what the updated connection should look like. But I think that goes against how the connections API was designed in the first place.

merchristK (commented):
looks fancy !

datadog/fwprovider/data_source_datadog_action_connection.go
```diff
@@ -0,0 +1,171 @@
...
+										},
+										"value": schema.StringAttribute{
+											Description: "Token value",
+											Computed:    true,
+											Sensitive:   true,
+										},
+									},
+								},
+							},
+							"header": schema.ListNestedBlock{
```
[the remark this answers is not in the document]
embeaken: Good question! It comes down to how the TF config is written.

Here's an example TF config for an HTTP connection:
```terraform
resource "datadog_action_connection" "ilya-tf-conn-httpmin" {
  name = "minimal HTTP conn TF"

  http {
    base_url = "https://catfact.ninja"

    token_auth {
      token {
        type  = "SECRET"
        name  = "t1"
        value = "${var.token1}"
      }

      header {
        name  = "Authorization"
        value = "{{ t1 }}"
      }

      header {
        name  = "my-header"
        value = "my-value"
      }
    }
  }
}
```

You can see every `header` block is one header. This reads and writes more naturally.

If you look at the model (`httpTokenAuthConnectionModel`) you can see there we use plural `Headers`. That's where the config gets serialized into for use in the code.

So it works out for both customer and developer. Customers use the singular form which makes writing the config more intuitive. And devs use the plural form which makes writing code more intuitive.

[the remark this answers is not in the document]
merchristK: i see. nice... thks for explaining

datadog/fwprovider/resource_datadog_action_connection.go
Waleed101, line 1: @samrcarr 

Discussed in person; trying to decide if it's better to have validation in both the API and Terraform or if we can consolidate all of that into the API. imo it makes more sense to have it in the API, but @embeaken highlighted that there might be a slightly more annoying UX when errors do come back.

Waleed101 (approved):
Looks great - thank you for getting this out so quickly!

datadog/fwprovider/resource_datadog_action_connection.go
embeaken: I included only surface-level validation in this PR just cause I thought it was easy. No fancy business logic. But I can get rid of that too if y'all want.

And we do return the error body from the API to the terraform CLI so users will still see any validation errors that come back anyway.

Waleed101 (dismissed):

drichards-87 (approved):

examples/data-sources/datadog_action_connection/data-source.tf
```diff
@@ -0,0 +1,3 @@
+data "datadog_action_connection" "my_connection" {
+  id = "1d05af79-db0e-4eca-b3b2-98c4a673c8cf"
```
tyffical, line 2: nit: is this an existing id? it might be better to use a dummy id like `11111111-2222-3333-4444-555555555555`

tyffical (approved):
looks great :)

examples/data-sources/datadog_action_connection/data-source.tf
embeaken: This repo dismisses all reviews if I make any changes so I'd rather leave this alone :grimacing:

Waleed101 (dismissed):

tyffical (approved):

samrcarr (dismissed):

samrcarr (approved):

jack-edmonds-dd (approved):
LGTM.

PR #2723, opened 2024-12-16
[datadog_app_builder_app] Add App Builder Terraform Support

tyffical:
## Motivation
- Adds the TF provider for the new [App Builder Public API](https://docs.datadoghq.com/api/latest/app-builder/).
- Includes
  - A data source for using Apps outside of resource definitions
  - A resource for managing Apps, with an `app_json` input
  - Support for importing existing Apps as resources.
  - Support for dynamically specifying Connection IDs using an optional `action_query_names_to_connection_ids` map to take precedence over the JSON
  - Support for optional top-level fields like `name`, `description`, `root_instance_name` to take precedence over the JSON
  - Support for optional `published` field that publishes an app when `true` and unpublishes it when `false`.
- Usage: 
  - Copy and paste App definition JSON from App Builder's JSON editor into Terraform either directly into the `app_json` field or load it from a file.
  - Alternatively, we will eventually have an "Import/Export Terraform" button that will write the TF resource definition.
  - We don't expect users to really create Apps from scratch using TF syntax (too complex). If users come to want this, we will put it in future work.
- Ticket: https://datadoghq.atlassian.net/browse/APPS-2248
- Epic: https://datadoghq.atlassian.net/browse/APPS-2144

### Design
See [App Builder App Terraform Resource Design Doc](https://datadoghq.atlassian.net/wiki/spaces/APPS/pages/4794745706/2025-03-05+App+Builder+App+Terraform+Resource+Design).
```
resource "datadog_app_builder_app" "example_app_inline_optional" {
  name               = "Example Terraform App - Inline - Optional"
  description        = "Created using the Datadog provider in Terraform."
  root_instance_name = "grid0"
  published          = false

  action_query_names_to_connection_ids = {
    "listTeams0" = var.connection_id
  }

  app_json = jsonencode(
    {
      ...some json here
    }
  )
}
```

## References
- ADR/Project doc: https://datadoghq.atlassian.net/wiki/spaces/APPS/pages/4285301372/App+Builder+Terraform
- Terraform Provider dev guide: https://datadoghq.atlassian.net/wiki/spaces/API/pages/3881239106/Getting+Started+-+Terraform+Provider
- Synthetics' guide to Terraform Provider dev: https://datadoghq.atlassian.net/wiki/spaces/SYN/pages/2729051036/datadog-api-spec+terraform-provider-datadog
- Ilya's PR for Connections Terraform Provider: https://github.com/DataDog/terraform-provider-datadog/pull/2718
- Gabriel's PR for Workflow Terraform Provider: https://github.com/DataDog/terraform-provider-datadog/pull/2833
- Slack channel for Actions org public API & TF dev: [#app-platform-public-apis-and-terraform](https://dd.enterprise.slack.com/archives/C06T1CLA2DR)

## Testing
- [x] Local manual testing (CRUD)
- [x] Record test cassettes with test org 

## Wait Before Merging
- [x] This Public API PR is merged and deployed: https://github.com/DataDog/dd-source/pull/178396
- [x] Other upcoming Public API changes for App Builder are done, expected to be ~ 4 weeks
  - [x] https://github.com/DataDog/datadog-api-spec/pull/3543
- [x] Update `datadog-api-client-go` to latest version to include above changes
- [ ] Prepare release as per `RELEASING.md`: https://github.com/DataDog/terraform-provider-datadog/blob/master/RELEASING.md

## Open Questions
**UPDATE: name will be `datadog_app_builder_app` b/c we decided offline we'll just keep this one resource and add extra TF optional fields if customer feedback suggests it is necessary**
~~**RESOLVED: name will be `datadog_app_builder_app_json`**~~
- ~~Currently, this resource is called `datadog_app`. The plan is, in the future, updating this resource with extra structs in order to build an app in TF with TF syntax, strongly typing structs like `queries` and `components`.~~
  - ~~Alternatively, what about creating a separate resource? This would help with making the code cleaner and disambiguate the two use cases. For instance, this resource would instead be `datadog_app_json` similar to how dashboards has a `datadog_dashboard_json`, and the non-json usage would be called `datadog_app`, like how dashboards also has `datadog_dashboard`.~~
  
## Files to review:
- `datadog/`
  - `fwprovider/`
    - `data_source_datadog_app_builder_app.go`
    - `resource_datadog_app_builder_app.go`
  - `internal/`
    - `utils/`
      - `utils.go`
      - `utils_test.go`
  - `tests/`
    - `data_source_datadog_app_builder_app_test.go`
    - `resource_datadog_app_builder_app_test.go`
-  `examples/resources/datadog_app_builder_app/resource.tf`


tyffical:
* **#2723** <a href="https://app.graphite.dev/github/pr/DataDog/terraform-provider-datadog/2723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/terraform-provider-datadog/2723?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

datadog/fwprovider/data_source_datadog_app.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (d *appDataSource) Configure(_ context.Context, request datasource.ConfigureRequest, response *datasource.ConfigureResponse) {
+	providerData := request.ProviderData.(*FrameworkProvider)
+	d.Api = providerData.DatadogApiInstances.GetAppBuilderApiV2()
+	d.Auth = providerData.Auth
+}
+
+func (d *appDataSource) Metadata(_ context.Context, request datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = "app"
```
embeaken, line 33: maybe worth naming the resource `app_builder_app` across this PR since `app` is such an overloaded word

datadog/fwprovider/framework_provider.go
```diff
@@ -409,6 +411,16 @@ func defaultConfigureFunc(p *FrameworkProvider, request *provider.ConfigureReque
 	ddClientConfig.SetUnstableOperationEnabled("v2.DeleteAWSAccount", true)
 	ddClientConfig.SetUnstableOperationEnabled("v2.GetAWSAccount", true)
 
+	// Enable unstable operations for the App Builder API
```
embeaken, line 414: before public release let's make sure the endpoints are on `v2`

datadog/fwprovider/resource_datadog_app.go
```diff
@@ -0,0 +1,292 @@
...
+	createRequest, err := appModelToCreateApiRequest(plan)
+	if err != nil {
+		response.Diagnostics.AddError("Error building create app request", err.Error())
+		return
+	}
+
+	resp, httpResp, err := r.Api.CreateApp(r.Auth, *createRequest)
+	if err != nil {
+		if httpResp != nil {
+			// error body may have useful info for the user
```
embeaken, line 89: just an observation from my testing - you should make sure your API doesn't expose any internal errors or they will show up in TF errors as well

datadog/fwprovider/resource_datadog_app.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return appModel, nil
+}
+
+func appModelToCreateApiRequest(appModel appResourceModel) (*datadogV2.CreateAppRequest, error) {
+	attributes := datadogV2.NewCreateAppRequestDataAttributesWithDefaults()
+
+	// decode encoded json into string and then decode that into the attributes struct
+	var appJsonString string
+	err := json.Unmarshal([]byte(appModel.AppJson.String()), &appJsonString)
```
embeaken, line 229: In providers, you should use `ValueString()` to get a string instead of `String()` which is for debugging. when i used `String()` it made the strings all double-quoted

this comment applies to all usage of `String()` in this PR

datadog/fwprovider/resource_datadog_app.go
```diff
@@ -0,0 +1,292 @@
...
+	req.Data = datadogV2.NewCreateAppRequestDataWithDefaults()
+	req.Data.SetAttributes(*attributes)
+
+	return req, nil
+}
+
+func appModelToUpdateApiRequest(plan appResourceModel) (*datadogV2.UpdateAppRequest, error) {
+	attributes := datadogV2.NewUpdateAppRequestDataAttributesWithDefaults()
+
+	// decode encoded json into string and then decode that into the attributes struct
```
embeaken, line 251: do you need two `Unmarshal`s here? Seems like you could just do

```
err = json.Unmarshal([]byte(plan.AppJson.ValueString()), attributes)
```

embeaken (commented):
great work!

docs/data-sources/app.md
```diff
@@ -0,0 +1,24 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_app Data Source - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Use this data source to retrieve information about an existing Datadog App from the App Builder product, for use in other resources.
```
drichards-87, line 6:
```suggestion
  Use this data source to retrieve information about an existing Datadog App from App Builder for use in other resources.
```

docs/data-sources/app.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# datadog_app (Data Source)
+
+Use this data source to retrieve information about an existing Datadog App from the App Builder product, for use in other resources.
```
drichards-87, line 11:
```suggestion
Use this data source to retrieve information about an existing Datadog App from App Builder for use in other resources.
```

docs/data-sources/app.md
drichards-87: An alternative suggestion to avoid using the word "Use" twice in the same sentence: "This data source retrieves information about an existing Datadog App from App Builder for use in other resources."

docs/resources/app.md
```diff
@@ -0,0 +1,62 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_app Resource - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Provides a Datadog App resource. This can be used to create and manage a Datadog App from the App Builder product.
```
drichards-87, line 6:
```suggestion
  Provides a Datadog App resource for creating and managing Datadog Apps from App Builder.
```

docs/resources/app.md
```diff
@@ -0,0 +1,62 @@
...
+---
+
+# datadog_app (Resource)
+
+Provides a Datadog App resource. This can be used to create and manage a Datadog App from the App Builder product.
```
drichards-87, line 11:
```suggestion
Provides a Datadog App resource for creating and managing Datadog Apps from App Builder.
```

docs/resources/app.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Example Usage
+
+```terraform
+# Example app json - loaded from a file (necessary if includes ${} syntax)
```
drichards-87, line 16:
```suggestion
# Example app JSON - loaded from a file (required for ${} syntax)
```

drichards-87 (approved):
Left some suggestions from the Docs and approved the PR.

oliverli (commented):
Looks great! I don't have much more to add beyond what Ilya has already covered

Strongly in favor of naming the resource `datadog_app_json` so that we have the option of creating a new `datadog_app` resource in the future with higher level CDK style mechanics, especially more so since dashboards has `datadog_dashboard_json` so we'll be in line with them

drichards-87 (dismissed):

datadog/fwprovider/data_source_datadog_app_builder_app.go
```diff
@@ -0,0 +1,218 @@
...
+				Computed:    true,
+				Description: "The human-readable description of the App.",
+			},
+			"root_instance_name": schema.StringAttribute{
+				Computed:    true,
+				Description: "The name of the root component of the app. This is a grid component that contains all other components.",
+			},
+			"published": schema.BoolAttribute{
+				Computed:    true,
+				Description: "Whether the app is published or unpublished. Published apps are available to other users. To ensure the app is accessible to the correct users, you also need to set a [Restriction Policy](https://docs.datadoghq.com/api/latest/restriction-policies/) on the app if a policy does not yet exist.",
```
GabeWM, line 77: One caveat on the restriction policies is they do not currently support scoped keys. Yet we require scoped keys. So this could be really tedious for a customer. Just calling this out.

tyffical: Do you have any suggestions for how to change this description to reflect that? I was following the description in the Publish API doc https://docs.datadoghq.com/api/latest/app-builder/#publish-app

GabeWM: I honestly don't just want to make you aware. AAA is working on changing this so hopefully will not be a problem in the future.

tyffical: makes sense thanks!

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+			}
+		}
+	}
+
+	// Create copies of the JSON structures to modify without affecting originals
+	planCopy := make(map[string]interface{})
+	configCopy := make(map[string]interface{})
+	stateCopy := make(map[string]interface{})
+	maps.Copy(planCopy, planJSON)
+	maps.Copy(configCopy, configJSON)
```
embeaken, line 397: i don't think we should be modifying the config since that's basically the user input and messing with it could bypass TF's consistency rules. what's the reason it's needed here?

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+	if !plan.OverrideActionQueryNamesToConnectionIDs.IsNull() {
+		overrides := plan.OverrideActionQueryNamesToConnectionIDs.Elements()
+
+		// Process queries in all JSONs to remove connectionId fields that will be overridden
+		processQueries := func(queries []interface{}, isState bool) {
+			for _, q := range queries {
+				if query, ok := q.(map[string]interface{}); ok {
+					if queryName, ok := query["name"].(string); ok {
+						if _, exists := overrides[queryName]; exists {
+							if properties, ok := query["properties"].(map[string]interface{}); ok {
```
embeaken, line 427: style: prefer `any` to `interface{}` (applies to the whole PR)

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+	// set computed values
+	plan.ID = types.StringValue(appID.String())
+	attributes := createReq.GetData().Attributes
+	plan.ActionQueryNamesToConnectionIDs, err = buildActionQueryNamesToConnectionIDsMap(attributes.GetQueries())
+	if err != nil {
+		response.Diagnostics.AddError("error building action_query_names_to_connection_ids map", err.Error())
+		return
+	}
+
+	// if optional fields are not set, compute them from the app json
```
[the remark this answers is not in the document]
GabeWM: Ahh I see this is handled in modifyPlan. Nice.

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+}
+
+// replace the connection ids in the queries with the ones provided in the plan, as specified by {action_query_name: connection_id}
+func replaceConnectionIDsInActionQueries(overrideActionQueryNamesToConnectionIDs types.Map, appJsonMap map[string]interface{}) error {
+	// skip if overrideActionQueryNamesToConnectionIDs is empty
+	if overrideActionQueryNamesToConnectionIDs.IsNull() || overrideActionQueryNamesToConnectionIDs.IsUnknown() {
+		return nil
+	}
+
+	// Handle connection ID overrides
```
embeaken, line 609: nit: comment doesn't seem to apply to the code

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+		return
+	}
+
+	appID := createResp.Data.GetId()
+
+	// if the published attribute is not set, set it to false
+	if plan.Published.IsNull() || plan.Published.IsUnknown() {
+		plan.Published = types.BoolValue(false)
+	}
+	err = handleAppBuilderAppPublishUnpublish(r.Auth, response.Diagnostics, r.Api, &plan, appID)
```
GabeWM, line 157: Since these operations are not atomic in this case if an app is created successfully, but then the unpublished fails the app would remain created. Do we not want to do cleanup?

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+			response.Diagnostics.AddError("error updating app", err.Error())
+		}
+		return
+	}
+
+	// if the published attribute is not set, set it to false
+	if plan.Published.IsNull() || plan.Published.IsUnknown() {
+		plan.Published = types.BoolValue(false)
+	}
+	err = handleAppBuilderAppPublishUnpublish(r.Auth, response.Diagnostics, r.Api, &plan, appID)
```
GabeWM, line 269: Same question about atomic above although this does seem more difficult for update.

GabeWM (dismissed):
Wow, nice job!

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+			if !ok {
+				continue
+			}
+
+			spec["connectionId"] = connectionID
+			delete(unusedOverrides, queryName)
+		}
+	}
+
+	// Return error if any overrides weren't used
```
embeaken, line 662: suggestion: this could just be a no-op instead of an error

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+		if err != nil {
+			if httpResp != nil {
+				// error body may have useful info for the user
+				body, err := io.ReadAll(httpResp.Body)
+				if err != nil {
+					diags.AddError("error reading error response", err.Error())
+					return err
+				}
+
+				// if error is related to the app already being published, we can ignore it
```
embeaken, line 683: ditto for this being a no-op instead of an error

datadog/tests/resource_datadog_app_builder_app_test.go
```diff
@@ -0,0 +1,1035 @@
...
+	if err != nil {
+		return fmt.Errorf("error parsing ID as UUID: %s", err)
+	}
+	if _, _, err := apiInstances.GetAppBuilderApiV2().GetApp(ctx, id); err != nil {
+		return fmt.Errorf("error getting app: %s", err)
+	}
+	return nil
+}
+
+func testComplexAppBuilderApp(name string) string {
```
embeaken, line 390: can you pull this from the example resource JSON file? i think that's the same as this and it would de-duplicate like 600 lines from this PR.

datadog/internal/customtypes/app_builder_app_json_string_value.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var temp interface{}
+	if err := dec.Decode(&temp); err != nil {
+		return "", err
+	}
+
+	// feature specific to AppBuilderAppStringValue:
+	// we only want to compare fields that matter to Create/Update requests when comparing JSON strings
+	if jsonMap, ok := temp.(map[string]interface{}); ok {
+		fieldsToKeep := []string{"components", "description", "name", "queries", "rootInstanceName"}
```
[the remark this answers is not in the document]
tyffical: fields that would get excluded in this comparison would be "id", "favorite", "selfService", "tags", "connections", "deployment". these fields are included in the App JSON but don't make a difference when calling Create/Update endpoints

datadog/fwprovider/data_source_datadog_app_builder_app.go
```diff
@@ -0,0 +1,218 @@
...
+
+		queryName := actionQuery.GetName()
+
+		// since we are reading the response from the API, we can ignore validation errors
+		specObj := actionQuery.Properties.GetSpec().ActionQuerySpecObject
+		if specObj.HasConnectionId() {
+			connectionID := specObj.GetConnectionId()
+			elementsMap[queryName] = connectionID
+		} else {
+			elementsMap[queryName] = ""
```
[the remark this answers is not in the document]
tyffical: hm I can see that - I'll remove it

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+					stringvalidator.LengthAtLeast(1),
+				},
+				CustomType: customtypes.AppBuilderAppStringType{},
+			},
+			"override_action_query_names_to_connection_ids": schema.MapAttribute{
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "If specified, this will override the Action Connection IDs for the specified Action Query Names in the App JSON.",
+			},
+			"action_query_names_to_connection_ids": schema.MapAttribute{
```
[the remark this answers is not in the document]
tyffical: that's a good point - makes sense to have override when it's specified and just build mapping otherwise

datadog/fwprovider/resource_datadog_app_builder_app.go
tyffical: yeah update would be particularly difficult. i think i am fine with errors in publishing state b/c that can likely be resolved with another `apply` (as Ilya has commented below), and it's more important that an app is created / updated correctly than for it to be published/unpublished correctly (b/c the latter is easily fixed)

datadog/fwprovider/resource_datadog_app_builder_app.go
tyffical: ah good catch! this section of code is just copying for sake of comparison and not modifying the config itself. i think the config copies are leftover from previous iterations actually, since we end up only comparing the plan and state anyway. i'll remove the config stuff

datadog/fwprovider/resource_datadog_app_builder_app.go
tyffical: hm, I thought this would be helpful for users who may have accidentally misspelled a query name and want to quickly figure out why a connectionID wasn't changed as expected

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+				// error body may have useful info for the user
+				body, err := io.ReadAll(httpResp.Body)
+				if err != nil {
+					diags.AddError("error reading error response", err.Error())
+					return err
+				}
+
+				// if error is related to the app already being unpublished, we can ignore It
+				if !strings.Contains(string(body), ErrAppAlreadyDisabled) {
+					diags.AddError("error unpublishing app", string(body))
```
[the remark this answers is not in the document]
tyffical: to clarify, do you mean we should ignore any errors related to unpublishing/publishing as long as the HTTP error body is readable? because this section of code already ignores specific errors like app already being disabled or published

```golang
				// if error is related to the app already being unpublished, we can ignore It
				if !strings.Contains(string(body), ErrAppAlreadyDisabled) {
					diags.AddError("error unpublishing app", string(body))
				} else {
					return nil
				}
```

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+	}
+
+	appID := createResp.Data.GetId()
+
+	// if the published attribute is not set, set it to false
+	if plan.Published.IsNull() || plan.Published.IsUnknown() {
+		plan.Published = types.BoolValue(false)
+	}
+	err = handleAppBuilderAppPublishUnpublish(r.Auth, response.Diagnostics, r.Api, &plan, appID)
+	if err != nil {
```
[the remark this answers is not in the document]
tyffical: i handled response diags in the helper func itself - i can change the return statement to a bool instead, like `fail` or `success`  just so we know whether to exit this TF CRUD operation early

datadog/fwprovider/resource_datadog_app_builder_app.go
```diff
@@ -0,0 +1,720 @@
...
+	// Clean up any remaining empty values that might have been created during processing
+	removeEmptyValues(planCopy)
+	removeEmptyValues(configCopy)
+	removeEmptyValues(stateCopy)
+
+	// Marshal back to JSON for final comparison
+	planBytes, _ := json.Marshal(planCopy)
+	stateBytes, _ := json.Marshal(stateCopy)
+
+	planAndStateEq, _ := customtypes.NewAppBuilderAppStringValue(string(planBytes)).StringSemanticEquals(ctx, customtypes.NewAppBuilderAppStringValue(string(stateBytes)))
```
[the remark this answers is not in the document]
tyffical: i think it's probably possible to make it just a helper function but using a custom type makes more sense to me and keeps all that logic separate / away from this already really long file

datadog/fwprovider/resource_datadog_app_builder_app.go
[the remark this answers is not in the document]
embeaken: basically if i change the published flag in my TF config and run `terraform apply`, i'd want it to not give me an error if the publish state already matches what i put in my config. is that doable with the API?

datadog/fwprovider/resource_datadog_app_builder_app.go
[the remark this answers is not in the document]
embeaken: ah ok

datadog/fwprovider/resource_datadog_app_builder_app.go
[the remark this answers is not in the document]
embeaken: the custom type itself is a lot of boilerplate though. if you can have the comparison be a helper func (and you can put it somewhere else), that would get rid of the boilerplate and also would help anyone reading this code since they wouldn't have to also understand terraform custom types to work on it.

datadog/fwprovider/resource_datadog_app_builder_app.go
embeaken: sounds reasonable to me

datadog/fwprovider/resource_datadog_app_builder_app.go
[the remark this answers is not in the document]
tyffical: "if the publish state already matches what I put in my config" -> that's what happens when the App Builder API returns `ErrAlreadyDisabled` or `ErrDeploymentExists`, hence if we reach that error it will no-op and complete the CRUD operation normally (verified by manual and unit testing). I don't believe the other API errors have to do with publishing state already matching what we want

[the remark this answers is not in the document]
tyffical: hopefully the latest version makes this more clear:
```golang
				// if error is related to the app already being published, we can ignore it
				if strings.Contains(string(body), ErrDeploymentExists) {
					return true
				}
				diags.AddError("error publishing app", string(body))
				return false
```

drichards-87 (approved):

datadog/fwprovider/data_source_datadog_app_builder_app.go
```diff
@@ -0,0 +1,197 @@
...
+
+func NewDatadogAppBuilderAppDataSource() datasource.DataSource {
+	return &appBuilderAppDataSource{}
+}
+
+func (d *appBuilderAppDataSource) Configure(_ context.Context, request datasource.ConfigureRequest, response *datasource.ConfigureResponse) {
+	providerData := request.ProviderData.(*FrameworkProvider)
+	d.Api = providerData.DatadogApiInstances.GetAppBuilderApiV2()
+	// Used to identify requests made from Terraform
+	d.Api.Client.Cfg.AddDefaultHeader("X-Datadog-App-Builder-Source", "terraform")
```
fpighi, line 31: Not sure if this is required. You could use `User-Agent` to identify requests being made by Terraform clients. I see a similar approach to yours is already used in `workflow_automation` resource/datasource, but that's the only occurrence.

Is `User-Agent` not enough for your use case?

docs/data-sources/app_builder_app.md
```diff
@@ -0,0 +1,35 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_app_builder_app Data Source - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources.
+---
+
+# datadog_app_builder_app (Data Source)
+
+This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources.
```
fpighi, line 11: Is there any use case where customers might want to rely on this data source? Might be worth having an example here.

embeaken (dismissed):
:balloon:

docs/data-sources/app_builder_app.md
tyffical: Does this suffice?
```text
This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources, such as embedding Apps in Dashboards or Notebooks.
```

tyffical: Does this suffice?
```text
This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources, such as embedding Apps in Dashboards or Notebooks.
```

datadog/fwprovider/data_source_datadog_app_builder_app.go
tyffical: Ah yes I was copying the approach done in `workflow_automation`. As discussed offline, you are right that `User-Agent` should be enough to identify Terraform requests, and it's already included by default.

https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/fwprovider/framework_provider.go#L388-L394

https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/internal/utils/utils.go#L230-L236

datadog/fwprovider/data_source_datadog_app_builder_app.go
```diff
@@ -0,0 +1,195 @@
...
+	d.Auth = providerData.Auth
+}
+
+func (d *appBuilderAppDataSource) Metadata(_ context.Context, request datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = "app_builder_app"
+}
+
+func (d *appBuilderAppDataSource) Schema(_ context.Context, request datasource.SchemaRequest, response *datasource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources, such as embedding Apps in Dashboards or Notebooks.",
```
fpighi, line 39:
```suggestion
		Description: "This data source retrieves the definition of an existing Datadog App from App Builder for use in other resources, such as embedding Apps in Dashboards.",
```
Notebooks doesn't exist as a Terraform resource.
As for dashboards, I'm ok with this example mentioned but customers will only be able to embed it when using the [dashboard JSON-based resource](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/dashboard_json), since the non-JSON one doesn't seem to support this right now.
I'm not sure this data source is extremely useful right now, but it could be kept for future integrations.
As long as the change in description is accepted, I'm happy to approve.

fpighi (changes requested):

embeaken (approved):

fpighi (approved):

GabeWM (approved):

oliverli (approved):

PR #2731, opened 2024-12-18
[datadog_synthetics_test] Prevent updating `files` with the backend response to keep the plan stable

etnbrd:
This PR prevents updating the `files` param from the upload file step in browser tests, to keep next plans stable.

Without this, the next plans keep trying to align the state returned by the backend and the local config:
```
~ files      = jsonencode(
    ~ [
        ~ {
            - bucketKey = "browser-upload-file-step/72e-vim-wme/2024-12-17T13:05:48.286579_f235f491-4dfd-4fbd-96ae-ed1b233b3e41.json"
            + content   = "this is the new new test content"
              name      = "test file"
          },
      ]
  )
```

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -2150,6 +2150,11 @@ func updateSyntheticsBrowserTestLocalState(d *schema.ResourceData, syntheticsTes
 				// keep the element from the local state instead
 				element := d.Get(fmt.Sprintf("browser_step.%d.params.0.element", stepIndex))
 				localParams["element"] = element
+			} else if key == "files" {
+				// prevent overriding `files` in the local state with the one received from the backend, and
+				// keep the files from the local state instead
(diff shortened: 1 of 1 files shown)


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -4651,6 +4656,7 @@ func decompressAndDecodeValue(value string) string {
 	return string(compressedProtoFile)
 }
 
+// TODO
```
amaskara-dd, line 4659:
```suggestion
```

Drarig29 (commented):
In the example plan you shared in the PR description, it looks like the `bucketKey` also needs to be removed from the local state (given the `-` character before the property name). Is this PR sufficient to entirely fix the plan instability?

etnbrd:
> In the example plan you shared in the PR description, it looks like the `bucketKey` also needs to be removed from the local state (given the `-` character before the property name). Is this PR sufficient to entirely fix the plan stability?

Yes. The snippet I shared was the plan instability before this fix.

Drarig29 (approved):

PR #2740, opened 2024-12-23
[datadog_logs_archives] Handle encryption field for S3 destinations for Logs Archives

michelledeng30:
We want to handle encryption configuration for S3 destinations in logs archives in Terraform. We are supporting the ability for users to control their encryption settings for their S3 archives through Logs Archives public API. 

These changes allow TF users to control how their data is encrypted when they upload to their S3 buckets. The possible encryption types are `NO_OVERRIDE` (either no encryption or using the encryption set on their own S3 bucket), `SSE_S3`, and `SSE_KMS`. There is an optional encryption key field to specify an Amazon Resource Name, which identifies an encryption key only for the `SSE_KMS` type. 

The encryption settings, which consist of just the encryption type and key, are not required, so no breaking changes are introduced. 

This PR is dependent on this [datadog_api_spec PR](https://github.com/DataDog/datadog-api-spec/pull/3311). The dependencies including the datadog_api_clients_go, logs-api, and dogweb implementations are all merged. 

[Jira](https://datadoghq.atlassian.net/browse/LOGSF-1125)


datadog/resource_datadog_logs_archive.go
```diff
@@ -44,6 +44,8 @@ func resourceDatadogLogsArchive() *schema.Resource {
 								ValidateDiagFunc: validators.ValidateAWSAccountID,
 							},
 							"role_name": {Description: "Your AWS role name", Type: schema.TypeString, Required: true},
+							"encryption_type": {Description: "The type of encryption on your archive.", Type: schema.TypeString, Required: true},
```
EdwinSri, line 47: If this new field is required wouldn't users need to add them to their terraform configuration? Let's say a user updates the terraform provider version, wouldn't it break their setup until they add it? 

If yes then what can/should we do about it? 🤔

datadog/resource_datadog_logs_archive.go
```diff
@@ -421,9 +426,28 @@ func buildS3Destination(dest interface{}) (*datadogV2.LogsArchiveDestinationS3,
 	if !ok {
 		path = ""
 	}
+	encryptionType, ok := d["encryptionType"]
```
EdwinSri, line 429: d["encryptionType"] is a different thing than "encryption_type" added elsewhere in the PR? Just making sure this isn't a typo

datadog/resource_datadog_logs_archive.go
```diff
@@ -421,9 +426,28 @@ func buildS3Destination(dest interface{}) (*datadogV2.LogsArchiveDestinationS3,
...
+	if !ok {
+		return &datadogV2.LogsArchiveDestinationS3{}, fmt.Errorf("encryption type is not defined")
+	}
+	encryptionKey, ok := d["encryptionKey"]
+	var LogsArchiveEncryptionS3 encryption
+
+	if !ok {
```
EdwinSri, line 436: I assume from reading the code that if the user sets SSE_KMS as the encryption type but doesn't provide an encryption key we do not want to handle this error here and just pass the values along is that correct?

docs/resources/logs_archive.md
```diff
@@ -87,6 +88,8 @@ Required:
 Optional:
 
 - `path` (String) Path where the archive is stored.
+- `encryption_type` (String) The type of server-side encryption to use when uploading data to your s3 bucket. `NO_OVERRIDE`, `SSE_S3`, and `SSE_KMS` are the possible types. `NO_OVERRIDE` is used most commonly to leave data unencrypted or leave encryption to the default encryption set on s3 bucket settings.
```
EdwinSri, line 91: I don't really understand the NO_OVERRIDE value, where does it come from? Is it set by default for users who don't define the encryption_type explicitely?
Also is that really true "most commonly to leave data unencrypted" ? I thought AWS encrypted all s3 buckets by default?

datadog/resource_datadog_logs_archive.go
```diff
@@ -421,9 +426,28 @@ func buildS3Destination(dest interface{}) (*datadogV2.LogsArchiveDestinationS3,
 	if !ok {
 		path = ""
 	}
+	encryptionType, ok := d["encryptionType"]
+	if !ok {
+		return &datadogV2.LogsArchiveDestinationS3{}, fmt.Errorf("encryption type is not defined")
+	}
+	encryptionKey, ok := d["encryptionKey"]
```
EdwinSri, line 433: Same comment here

EdwinSri (commented):
I have a few questions, I don't know much about Go or this repository so I want to make sure I understand everything properly.

datadog/resource_datadog_logs_archive.go
michelledeng30: I switched the encryption type to be optional instead of required, it should now be implemented in the code

datadog/resource_datadog_logs_archive.go
michelledeng30: It should be the same thing

datadog/resource_datadog_logs_archive.go
michelledeng30: Yes that's right, technically setting SSE_KMS with no encryption key is also a valid method (just not done very commonly)

docs/resources/logs_archive.md
michelledeng30: Yes the NO_OVERRIDE value is the default type if users don't choose another type. 
It looks like SSE_S3 is the default encryption for S3 buckets, so I will revise this statement to make this clear

datadog/resource_datadog_logs_archive.go
EdwinSri: @michelledeng30  Nice, although I wasn't necessarily saying to change it to be optional. I was wondering if we had a good reason for it to be either required or optional and if we were aware of the consequences of choosing either option? These weren't rhetorical questions in any way

datadog/resource_datadog_logs_archive.go
EdwinSri: I see, I didn't know that. Thanks for explaining!

docs/resources/logs_archive.md
EdwinSri: @michelledeng30 Linking this message my other message i posted today, if NO_OVERRIDE is set by default for the encryption-type field, does it make sense for for the encryption-type field to be optional? I really don't know much about the terraform provider code so I'm only talking hypothetically but if encryption-type is made to be required wouldn't it be set to NO_OVERRIDE by default if users don't explicitly set a value for it in their configuration? 

I'm trying to understand which way of doing it makes the most sense but if we really don't know maybe we can ask someone who is more knowledgeable about this.

datadog/resource_datadog_logs_archive.go
michelledeng30: I think it makes sense for the field to be optional. 

If encryption type is required, then every time users interact with their s3 archives via Terraform, they must deliberately set `encryption_type` in their Terraform configuration. Most users don't need this added functionality of encrypting their data at an object level, since they use default bucket encryption set in s3. I don't think it's necessary for it to be required, and I could see how making it required would be misleading/confusing to users who aren't familiar with/don't need the ability to set encryption in their PUT requests rather than the bucket. 

If the field is optional, we can allow users to only set this field if they need the extra level of encryption, otherwise it's automatically set to `NO_OVERRIDE` for most users. 

Also, comparing the [Terraform resource](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/logs_archive#azure_archive-1) and the [logs archives API reference](https://docs.datadoghq.com/api/latest/logs-archives/), it looks like the same fields are required or optional. In the public API, we support creating/updating archives without any encryption type, so it makes sense to do the same in Terraform. 

> if encryption-type is made to be required wouldn't it be set to NO_OVERRIDE by default if users don't explicitly set a value for it in their configuration

I believe if a required parameter is not set in a Terraform configuration, it will fail `terraform validate` and the user will be forced to define it to pass the check 

@EdwinSri

docs/resources/logs_archive.md
michelledeng30: Response is in this comment ! - https://github.com/DataDog/terraform-provider-datadog/pull/2740#discussion_r1913903912

datadog/resource_datadog_logs_archive.go
EdwinSri: Thanks a lot for the explanations, I understand better now and I agree this makes sense like this 🙇

EdwinSri (dismissed):

michelledeng30:
The failed test in the CI is due to an issue with a generated AWS Account Id in the [testcase](https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/tests/resource_datadog_logs_archive_test.go#L232) not being associated with an actual AWS account in Datadog. This happens in the api call and fails pretty consistently (when I started testing and re-recording the test cassettes, it was passing occasionally but not anymore). I'm following up about this in this [slack thread](https://dd.slack.com/archives/CHXRF0UG3/p1737740567303819).

antonio-ramadas-dd (approved):
Looks good to me,

hestonhoffman (approved):

hestonhoffman (approved):

thibaultkrebs (approved):

corentinmarc (dismissed):

antonio-ramadas-dd (dismissed):

EdwinSri (approved):

datadog/resource_datadog_logs_archive.go
```diff
@@ -43,7 +43,9 @@ func resourceDatadogLogsArchive() *schema.Resource {
 								Required:         true,
 								ValidateDiagFunc: validators.ValidateAWSAccountID,
 							},
-							"role_name": {Description: "Your AWS role name", Type: schema.TypeString, Required: true},
+							"role_name":       {Description: "Your AWS role name", Type: schema.TypeString, Required: true},
+							"encryption_type": {Description: "The type of encryption on your archive.", Type: schema.TypeString, Optional: true, Default: "NO_OVERRIDE"},
```
skarimo, line 47:
```suggestion
							"encryption_type": {Description: "The type of encryption on your archive.", Type: schema.TypeString, Optional: true, Default: datadogV2.LOGSARCHIVEENCRYPTIONS3TYPE_NO_OVERRIDE, ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewLogsArchiveEncryptionS3TypeFromValue)},
```

skarimo: Does sending a default value all the time without an encryption key change the behavior in any way?

skarimo, line 47: I think these should be in its own `encryption` block since it is an optional field. This would also avoid the issues without default value always being sent:
```
"encryption": {
	Type:     schema.TypeList,
	MaxItems: 1,
	Optional: true,
	Elem: &schema.Resource{
			"encryption_type": {
				Description:      "The type of encryption on your archive.",
				Optional:         true,
				Default:          datadogV2.LOGSARCHIVEENCRYPTIONS3TYPE_NO_OVERRIDE,
				ValidateDiagFunc: validators.ValidateEnumValue(datadogV2.NewLogsArchiveEncryptionS3TypeFromValue)},
			"encryption_key": {Description: "The AWS KMS encryption key.", Type: schema.TypeString, Optional: true},
		},
	},
},
```

skarimo (commented):
The `encryption_key` is also not returned by the api so we need to handle this otherwise terraform will be in a continuous diff

datadog/resource_datadog_logs_archive.go
michelledeng30: If the `encryption` object is optional, wouldn't this make `encryption_type` required? @skarimo

skarimo: It should not conflict. `encryption_type` can still be kept as optional. The main thing with this change would be, we would have a new `encryption` block that is optional.

michelledeng30: I created the optional encryption block. I was originally getting errors when testing a resource that didn't have any encryption information, since it's not required to provide it (the first step in [`TestAccDatadogLogsArchiveS3Update_basic`](https://github.com/DataDog/terraform-provider-datadog/pull/2740/files#diff-eabb4fdfa2a6aa777f41584eba6bb3440785b44ba4821e127855a821fb18e85fR262)). 

The terraform error output displays `encryption_type = "NO_OVERRIDE" -> null`. Since in the Datadog API, if no encryption is provided in the s3 destination, it will return an encryption object with type set to `NO_OVERRIDE`. If I understand correctly, because encryption is not in the Terraform plan, this causes the mismatch. To resolve this, I added the `Computed: true` field to the resource, let me know if I'm using it correctly in this case.

skarimo (dismissed):

datadog/resource_datadog_logs_archive.go
skarimo: I undid the changes suggested above as I did not have full context during the original review

skarimo (approved):

PR #2741, opened 2024-12-24
update the ML in the steps when reordering

etnbrd:


docs/resources/synthetics_test.md
```diff
@@ -1002,6 +1002,7 @@ Optional:
 - `exit_if_succeed` (Boolean) Determines whether or not to exit the test if the step succeeds.
 - `force_element_update` (Boolean) Force update of the "element" parameter for the step
 - `is_critical` (Boolean) Determines whether or not to consider the entire test as failed if this step fails. Can be used only if `allow_failure` is `true`.
+- `local_key` (String) A unique identifier allowing to track steps after reordering.
```
estherk15, line 1005:
```suggestion
- `local_key` (String) A unique identifier used to track steps after reordering.
```

estherk15 (dismissed):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -4773,6 +4751,160 @@ func getCertificateStateValue(content string) string {
...
+			stateSteps = rawState.GetAttr("browser_step")
+			stateStepCount = stateSteps.LengthInt()
+		}
+
+		if stateStepCount > 0 {
+			for i := range stateStepCount {
+				stateStep := stateSteps.Index(cty.NumberIntVal(int64(i)))
+				localKeyValue := stateStep.GetAttr("local_key")
+				if localKeyValue.IsNull() {
+					continue
```
teodor2312, line 4815: so this mean that until people add a local_key to their step definitions their steps won't update right? 🤔
Is there some way we'll be communicating this to users? Cuz we run the risk of people not noticing it immediately and then eventually wondering what's going on 🤔 
How about making the local_key a required field? It will make this change not be backwards compatible but it will also be a very clear signal to people that otherwise their steps won't be updating 🤔

teodor2312 (approved):
Some question but otherwise LGTM 👍

amaskara-dd (approved):

datadog/resource_datadog_synthetics_test_.go
amaskara-dd: Making this field required would be a big breaking change. Perhaps we could just output a warning if the field is unset?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -4773,6 +4751,160 @@ func getCertificateStateValue(content string) string {
...
+		stateStepCount := 0
+		stateSteps := cty.ListValEmpty(cty.DynamicPseudoType)
+		if !rawState.IsNull() {
+			stateSteps = rawState.GetAttr("browser_step")
+			stateStepCount = stateSteps.LengthInt()
+		}
+
+		if stateStepCount > 0 {
+			for i := range stateStepCount {
+				stateStep := stateSteps.Index(cty.NumberIntVal(int64(i)))
```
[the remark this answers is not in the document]
etnbrd: Yes, this just mean `stateSteps[i]` 😅

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -4773,6 +4751,160 @@ func getCertificateStateValue(content string) string {
 	return utils.ConvertToSha256(content)
 }
 
+func getStepParams(stepMap map[string]interface{}, d *schema.ResourceData) map[string]interface{} {
+	stepType := datadogV1.SyntheticsStepType(stepMap["type"].(string))
```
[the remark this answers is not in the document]
etnbrd: Excellent question.
With the new implementation, `getStepParams` now requires `stepMap` and `d`.
The `step` you're referencing here is a `datadogV1.SyntheticsStep{}` populated outside of this function, from `stepMap`: `step.SetType(datadogV1.SyntheticsStepType(stepMap["type"].(string)))`.
To use `step.GetType()` here it would need to be passed as a parameter as well, which would be a bit redundant with `stepMap`, and it would be used only here, which could be confusing.
Given this, I decided to keep `step` outside of this function, and just recompute the `stepType` here.

PR #2760, opened 2025-01-10
[datadog_application_key] Support scoped app key management

wangwillson1:
This PR allows users to manage (create/update) scoped app keys in Terraform

Tested manually with:
```
resource "datadog_application_key" "app_key_1" {
  name = "willson tf scoped app key test"
  scopes = ["dashboards_read", "dashboards_write"]
}
```

Also verified that an invalid scope in the list results in a 400.

### DIFF
--- a/datadog/fwprovider/data_source_datadog_application_key.go
+++ b/datadog/fwprovider/data_source_datadog_application_key.go
@@ -56,6 +56,11 @@ func (d *applicationKeyDataSource) Schema(_ context.Context, req datasource.Sche
 				Computed:    true,
 				Sensitive:   true,
 			},
+			"scopes": schema.SetAttribute{
+				Description: "Authorization scopes for the Application Key.",
+				Optional:    true,
+				ElementType: types.StringType,
+			},
 		},
 		DeprecationMessage: "The datadog_application_key data source is deprecated and will be removed in a future release with prior notice. Securely store your application key using a secret management system or use the datadog_application_key resource to manage application keys in your Datadog account.",
 	}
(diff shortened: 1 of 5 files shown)


jhgilbert (approved):

retsguj (commented):
Generally looks good to me. I tested and it works fine, though weirdly I keep getting an error that says
```
Error: Provider returned invalid result object after apply
After the apply operation, the provider still indicated an unknown value for datadog_application_key.app_key_1.key. All values must be known after apply, so this is always a bug in the provider and should be reported in the provider's own repository. Terraform will still save the other known object values in the state.
```
after running `terraform apply`. The application key does look to be correctly updated in Datadog, though, so I'm not sure what the error message is referring to.

I'm assuming you didn't get this while testing?

amaskara-dd:
> Generally looks good to me. I tested and it works fine, though weirdly I keep getting an error that says
> 
> ```
> Error: Provider returned invalid result object after apply
> After the apply operation, the provider still indicated an unknown value for datadog_application_key.app_key_1.key. All values must be known after apply, so this is always a bug in the provider and should be reported in the provider's own repository. Terraform will still save the other known object values in the state.
> ```
> 
> after running `terraform apply`. The application key does look to be correctly updated in Datadog, though, so I'm not sure what the error message is referring to.
> 
> I'm assuming you didn't get this while testing?

I get the same error when I tried updating the scopes (although the change was reflected in the UI)

retsguj (approved):
The error I got in my comment above occurs on the `master` branch too, so I don't think it's something introduced by this PR.

So it looks good to merge (for core access)

amaskara-dd:
this apply error is an existing bug that occurs with any update to an application key  — will address in follow up

amaskara-dd (approved):

PR #2763, opened 2025-01-13
[datadog_application_key] use plan modifiers for app keys

amaskara-dd:
App keys are not returned by the API after creation, so we should save the previous state values to avoid the `Provider returned invalid result object after apply` and overwrite of state value with null

Tested by creating an application key and updating its name

### DIFF
--- a/datadog/fwprovider/resource_datadog_api_key.go
+++ b/datadog/fwprovider/resource_datadog_api_key.go
@@ -7,8 +7,6 @@ import (
 	frameworkPath "github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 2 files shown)


nkzou (approved):

jackakeller:
@amaskara-dd Why is this needed if we already check whether the response from the API has a copy of the key [here](https://github.com/DataDog/terraform-provider-datadog/blob/18af0813b0fece9a86de64842b9d70c72589612d/datadog/fwprovider/resource_datadog_application_key.go#L185-L187)?

amaskara-dd:
@jackakeller We noticed the overwrite when updating the name of the application key after creation. The check in `updateState` was not sufficient as the Create and Update functions use the [Plan.Get](https://github.com/DataDog/terraform-provider-datadog/blob/18af0813b0fece9a86de64842b9d70c72589612d/datadog/fwprovider/resource_datadog_application_key.go#L132) to update the state prior to the API call

PR #2772, opened 2025-01-16
fix(synthetics): duplication of default info

alexis974:
This will remove duplicated default information for `[elementsoperator]`

### DIFF
--- a/docs/resources/synthetics_test.md
+++ b/docs/resources/synthetics_test.md
@@ -743,7 +743,7 @@ Required:
 
 Optional:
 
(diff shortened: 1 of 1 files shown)


aliciascott (approved):
:) looks good, good catch

alexis974:
Hello,

Is there a maintainer available to trigger the workflow ?

skarimo:
@alexis974 thanks for the PR. The description also needs to be removed from the schema as well: https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_synthetics_test_.go#L541
Otherwise docs will be overriden on the next regeneration. We use tfplugindocs for generating docs which is the reason for duplicate being inserted.

skarimo:
@alexis974 The terraform docs are autogenerated and the default value gets automatically injected based on the attributes [default value](https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_synthetics_test_.go#L544) during generation. It combines the description and adds a string for default value hence the reason for original duplication. I cherry-picked your commit into https://github.com/DataDog/terraform-provider-datadog/pull/2803 . Closing this PR in favor of above. Thanks!

PR #2774, opened 2025-01-16
[data_datadog_application_key] remove scopes from schema

amaskara-dd:
fixes regression caused by https://github.com/DataDog/terraform-provider-datadog/pull/2760, which added scopes to the Schema but not the model or state update


skarimo (dismissed):

wangwillson1 (dismissed):

amaskara-dd:
opting to remove scopes since data source will be deprecated soon

skarimo (approved):

PR #2788, opened 2025-01-21
[datadog_integration_aws] EXTERNAL_ID environment variable is now optional for import

### LINKED ISSUE #2257: datadog_aws_integration resource should support "import" block

resource_datadog_aws_integration

### Feature Request

At the moment, [datadog_aws_integration](https://registry.terraform.io/providers/DataDog/datadog/3.35.0/docs/resources/integration_aws) resource cannot be imported from configuration using [`import`](https://developer.hashicorp.com/terraform/language/import) block, because of requirement of providing `external_id` attribute as environment variable.

Such semantics is not common for Terraform providers and breaks in configuration-driven imports, where it's not possible to specify environment variables per `import` block.

The workaround is, obviously, to run `terraform import` manually, but it may be difficult or even impossible in Terraform Cloud (Enterprise) remote runs, when the configuration depends on variables provided by Terraform Cloud. That is because `terraform import` always runs locally, so if some sensitive input variables are provided by TFC, you'd have to somehow extract them from TFC and provide for a local `terraform import` via `-var` command-line argument.

The desired outcome would be ability to import `datadog_aws_integration` resource without all this hassle, like this:
```hcl
import {
  id = "111111111111:DatadogIntegrationRole/477dc7a8-6daa-4be4-b5e3-6a2c39c5d56b"
  to = datadog_aws_integration.this
}
```

### References

_No response_

ecdatadog:
Fixes #2257 

The `external_id` field is a computed field that is : 
* never set outside of create => no risk of drift 
* never used for requesting Datadog API => no risk at the execution of actions 

With the [import block](https://developer.hashicorp.com/terraform/language/import) (tf 1.5+), needing an environment variable can be an issue => we put this environment variable as optional.

### DIFF
--- a/datadog/resource_datadog_integration_aws.go
+++ b/datadog/resource_datadog_integration_aws.go
@@ -361,7 +361,8 @@ func resourceDatadogIntegrationAwsImport(ctx context.Context, d *schema.Resource
 	if d.Id() == "" {
 		return nil, fmt.Errorf("error importing aws integration resource. Resource with id `%s` does not exist", originalId)
 	}
-
-	d.Set("external_id", os.Getenv("EXTERNAL_ID"))
+	if v, ok := os.LookupEnv("EXTERNAL_ID"); ok {
+		d.Set("external_id", v)
+	}
 	return []*schema.ResourceData{d}, nil
 }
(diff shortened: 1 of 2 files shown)


cswatt (approved):

agulen:
Closing this PR since it is over a year old

PR #2795, opened 2025-01-23
[datadog_synthetics_test] Fix steps reordering without `local_key`

Drarig29:
Terraform applying a browser test with provider version [v3.51.0](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.51.0) and steps that do not have a `local_key` overwrites the `element` of all steps to the last step’s `element`.

This is caused by https://github.com/DataDog/terraform-provider-datadog/pull/2741, which compares `local_key`s together, but doesn't check that they are different than `""`, which is the default value.

You can see the proof that updated tests would fail without the change: [here](https://github.com/DataDog/terraform-provider-datadog/actions/runs/12930278229/job/36061431016?pr=2795)


datadog/tests/resource_datadog_synthetics_test_test.go
```diff
@@ -744,14 +744,14 @@ func createSyntheticsBrowserStepsConfigWithLocalML(uniq string) string {
 			modifiers = []
 			value     = "step %[1]s value"
 		}
-	}`, uniq, createSyntethicsBrowserStepMLElement(uniq))
+	}`, uniq, createSyntheticsBrowserStepMLElement(uniq))
 }
 
-func createSyntethicsBrowserStepMLElement(uniq string) string {
```
teodor2312, line 750: lol i was looking at the diff and it still took me like 30sec to figure out what's the difference 😂

teodor2312 (approved):
I didn't look through all of the cassettes but the main changes look good 👍

nkzou (approved):

Issue #2452, opened 2024-06-21
Issues with Javascript browser variable and related extractFromJavascript browser step


fabrideci:
### Datadog Terraform Provider Version

v3.39.0

### Terraform Version

v1.8.3

### What resources or data sources are affected?

datadog_synthetics_test

### Terraform Configuration Files

```hcl
browser_variable {
  type    = "javascript"
  name    = "PWD_EXAMPLE"
}

browser_variable {
  type    = "javascript"
  name    = "PWD_EXAMPLE_2"
  secure  = true // applicable only for variables of type "text"
}

browser_step {
  name = "Create new random password"
  type = "extractFromJavascript"
  params {
    code = <<EOT
const minLength = 8;
const lowerCaseChars = 'abcdefghijklmnopqrstuvwxyz';
const upperCaseChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
const numericChars = '0123456789';
const specialChars = '$!@#$%^&*)(+=._-';
const allChars = lowerCaseChars + upperCaseChars + numericChars + specialChars;

function getRandomCharFromString(str) {
  return str.charAt(Math.floor(Math.random() * str.length));
}

let randomString = getRandomCharFromString(lowerCaseChars)
  + getRandomCharFromString(upperCaseChars)
  + getRandomCharFromString(numericChars)
  + getRandomCharFromString(specialChars);

while (randomString.length < minLength) {
  randomString += getRandomCharFromString(allChars);
}

randomString = randomString.split('').sort(() => 0.5 - Math.random()).join('');

return randomString;
EOT
    variable {
      name = "PWD_EXAMPLE_3"
    }
  }
}
```

### Expected Behavior

In a browser test, I should have the capability to declare a JavaScript variable and utilize the corresponding `extractFromJavaScript` step to assign it a value.

### Actual Behavior

When attempting to set a simple JavaScript `browser_variable` in Terraform, such as `PWD_EXAMPLE`, an error occurs even if the 'Create new random password’ step points to `PWD_EXAMPLE` instead of `PWD_EXAMPLE_3`, as shown in the previous code example. The error message is as follows:

> Error: error updating synthetics browser test from /api/v1/synthetics/tests/browser/8ss-wcy-akr: 400 Bad Request: {"errors":["'variables' value '{'example': '', 'name': 'PWD_EXAMPLE', 'pattern': '', 'type': 'javascript'}' is invalid"]}" 

[...]

PR #2818, opened 2025-02-07
[datadog_synthetics_test] Add secure field for extractFromJavascript browser step variables

AntoineDona:
The secure field was missing in the `extractFromJavascript` step variables, making it impossible to create obfuscated javascript variables, as highlighted in [this issue](https://github.com/DataDog/terraform-provider-datadog/issues/2452).


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1377,6 +1377,12 @@ func syntheticsBrowserStepParams() schema.Schema {
 								Type:        schema.TypeString,
 								Optional:    true,
 							},
+							"secure": {
+								Description: "Whether the value of this variable will be obfuscated in test results.",
```
Drarig29, line 1381: **nit** Just to indicate that it's the user that wants it
```suggestion
								Description: "Whether the value of this variable should be obfuscated in test results.",
```

docs/resources/synthetics_test.md
```diff
@@ -1066,6 +1066,7 @@ Optional:
 
 - `example` (String) Example of the extracted variable. Defaults to `""`.
 - `name` (String) Name of the extracted variable.
+- `secure` (Boolean) Whether the value of this variable will be obfuscated in test results. Defaults to `false`.
```
Drarig29, line 1069:
```suggestion
- `secure` (Boolean) Whether the value of this variable should be obfuscated in test results. Defaults to `false`.
```

Drarig29 (approved):
LGTM

datadog/resource_datadog_synthetics_test_.go
AntoineDona: Hmmm we should modify it everywhere then, because I copied this from existing variables that already support the `secure` parameter...

estherk15 (approved):

PR #2834, opened 2025-02-14
Update resource_datadog_logs_metric.go

jtg-github:
Using `TypeList` instead of TypeMap to avoid no-op changes in terraform when the order is different.

Because `group_by` is a TypeMap, the order depends on the hashcode of its attributes and should not matter. However, when running `terraform plan`, the order of `group_by` can sometimes be different. This causes Terraform to try and apply changes, when there are no changes. It leads to noise in the plan output.


jtg-github:
Unit tests fail. Closing for now.

PR #2832, opened 2025-02-13
[synthetics_global_variable] Add is_totp

AntoineDona:
### What & Why ?

It is currently [impossible](https://github.com/DataDog/terraform-provider-datadog/issues/2693) create a MFA/TOTP  global variable using terraform. This was partially implemented in the past, but `is_totp` was missing.
This PR adds `is_totp`.

### How ?

- Add a new `is_totp` parameter in the global variable schema.
- Change `secure` default behaviour for totp: it should always be true, and we should not have to set it.

Issue #2693, opened 2024-11-25
Add support for TOTP/MFA and FIDO variable types in datadog_synthetics_global_variable


mguillet-payfit:
### What resources or data sources are affected?

- datadog_synthetics_global_variable

### Feature Request

A synthetics global variable can have different types (e.g., raw value, MFA token, Virtual Authenticator).
Currently, the `datadog_synthetics_global_variable` resource only supports defining a variable with a raw value. Although the `options.totp_parameters` attribute exists, it doesn't set the variable as an MFA/TOTP token.

The [Datadog API documentation](https://docs.datadoghq.com/api/latest/synthetics/#create-a-global-variable) indicates that the `is_fido` and `is_totp` boolean parameters can be used in the request body to define the type of a variable. However, these parameters are not currently configurable in the `datadog_synthetics_global_variable` resource.

FYI, I tested defining a synthetic global variable with the resource and manually changing its type afterward. The modification is reverted to a raw value on next change applied.

Please add support for defining TOTP/MFA and FIDO variables in the `datadog_synthetics_global_variable` resource.

AntoineDona:
Hi @mguillet-payfit !
Thanks a lot for bringing this problem to our attention.
We just release a [new version](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.55.0) of the terraform-provider adding the `is_totp` parameter. The pr adding `is_fido` variable is ready, the change should be shipped in the  next release.

ecdatadog:
In [3.56.0](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.56.0), both `is_totp` and `is_fido` are now both available. 
Please confirm the issue is solved for you ! (reaction to this post will be great)

Without additional elements, this PR will be closed in one week

mguillet-payfit:
> In [3.56.0](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.56.0), both `is_totp` and `is_fido` are now both available.
> Please confirm the issue is solved for you ! (reaction to this post will be great)

Perfect, the issue is solved.
Thanks 🚀

PR #2836, opened 2025-02-14
[synthetics_global_variable] Add is_fido

AntoineDona:
### What & Why ?

- #2832 
- #2836 (← **you are here**)

It is currently [impossible](https://github.com/DataDog/terraform-provider-datadog/issues/2693) create a FIDO  global variable using terraform. This PR adds `is_fido`.

### How ?

- Add a new `is_fido` parameter in the global variable schema.
- Change `value` default behaviour for fido: it should always absent from the request to the backend


datadog/resource_datadog_synthetics_global_variable.go
```diff
@@ -46,18 +46,24 @@ func resourceDatadogSyntheticsGlobalVariable() *schema.Resource {
 					Elem:        &schema.Schema{Type: schema.TypeString},
 				},
 				"value": {
-					Description: "The value of the global variable.",
+					Description: "The value of the global variable. This setting is ignored if `is_fido` is set to `true` and required otherwise.",
```
Drarig29, line 49: Did you confirm dogweb rejects the request if `value` is not set for non-FIDO variables?

estherk15 (dismissed):

datadog/resource_datadog_synthetics_global_variable.go
AntoineDona: Yep, we get this message:

Screenshot 2025-02-17 at 10 31 09

AntoineDona:
Sorry had to rebase manually, the auto rebase didn't work as expected because /merge squashes the commits

datadog/resource_datadog_synthetics_global_variable.go
amaskara-dd: nit: this error message seems unclear for the case `value` was unset (seems to imply `value` was included when it was not)

PR #2838, opened 2025-02-17
[datadog_shared_dashboard] New resource

### LINKED ISSUE #876: [Feature Request] Support Public Dashboard URLs
### Affected Resource(s)

-   datadog_dashboard_public_url

### Important Factoids

It would be really nice if it were possible to configure and retrieve the public URL for sharing dashboards. In my use case, I add these as a Pagerduty Add-On so that incident reports include this URL in an iFrame, which is really very useful for incident responders. I'm sure there are many other uses!

### References

https://www.pagerduty.com/docs/guides/datadog-add-on-guide/

ecdatadog:
resolves #876 

Change compared to OpenAPI schema : 
* `created_at` field is not transmitted as the body response has field `created` with the openAPI spec says `created_at` 
* `author` structure is split in 2 fields `author_handle` and `author_name` : it is linked to a limitation at terraform level of SingleNestedBlock with only Computed field (limit of using v5 provider with framework lib)


datadog/fwprovider/resource_datadog_shared_dashboard.go
```diff
@@ -0,0 +1,431 @@
...
+
+type sharedDashboardModel struct {
+	ID                          types.String                   `tfsdk:"id"`
+	DashboardId                 types.String                   `tfsdk:"dashboard_id"`
+	DashboardType               types.String                   `tfsdk:"dashboard_type"`
+	GlobalTimeSelectableEnabled types.Bool                     `tfsdk:"global_time_selectable_enabled"`
+	PublicUrl                   types.String                   `tfsdk:"public_url"`
+	ShareType                   types.String                   `tfsdk:"share_type"`
+	Token                       types.String                   `tfsdk:"token"`
+	ShareList                   types.List                     `tfsdk:"share_list"`
```
michlee1337, line 34: Hey! This field will be marked deprecated soon sorry. We're in the process of updating the API contract for this API.

jhgilbert (approved):

PR #2839, opened 2025-02-17
[generator] fixes for generate_resource

nizox:
This PR includes the following changes:
- create the target directory when it doesn't exist
- introduce a `--no-go-fmt` option to disable go formatting
- use `subprocess.call` instead of `os.system` to avoid escaping issues


.generator/src/generator/templates/utils/request_builder_helper.j2
```diff
@@ -2,22 +2,22 @@
     {%- set primitiveAttr, primitiveArrAttr, nonPrimitiveListAttr, nonPrimitiveObjAttr = schema|tf_sort_properties_by_type %}
 
     {%- for attr, attrSchema in primitiveAttr.items() %}
-    {%- set isRequired = attr in schema.get("required", []) %}
```
ecdatadog, line 5: I am not sure it is correct : 
* required can be at two levels : 
  * at [attribute level](https://swagger.io/docs/specification/v3_0/describing-parameters/#required-and-optional-parameters)
  * at [object level](https://json-schema.org/understanding-json-schema/reference/object#required)  
My alternative solution is [here](https://github.com/DataDog/terraform-provider-datadog/pull/2838/files#diff-4485e304a3c764acddf7365223bc8ee96a7650899876d3c64d262c77dd9ed4b9R51-R59)

nizox: good catch! I will remove this part of the change then.

ecdatadog: I extracted the change linked to the generator so from https://github.com/DataDog/terraform-provider-datadog/pull/2838 to https://github.com/DataDog/terraform-provider-datadog/pull/2841

ecdatadog (approved):

PR #2773, opened 2025-01-16
LOGSC-2078 Add tags and description to logs pipelines

joboccara:

PR #2845, opened 2025-02-18
[datadog_logs_custom_pipeline] Add support for default tags in log pipelines

shuheiktgw:
[Since pipeline tags are already supported](https://github.com/DataDog/terraform-provider-datadog/pull/2773), this change adds default tags support to log pipelines for consistency. Thanks for the review!


datadog/tests/cassettes/TestAccDatadogLogsPipeline_basic.yaml
```diff
@@ -13,7 +13,7 @@ interactions:
         remote_addr: ""
```
shuheiktgw, line 13: I'm not sure why, but the test fails if I don't update these mock responses. Since the test passes when using the "real" API, I suppose we don't need to worry about it too much.

ecdatadog:
Great Feature. Could you please also modify https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/index.md?plain=1#L55-L68 so people it is available not only for monitors but also for your resource.

shuheiktgw:
The CI failure does not seem to be related to the change. I've created another PR to address the issue.
https://github.com/DataDog/terraform-provider-datadog/pull/2850

buraizu (approved):

PR #2857, opened 2025-02-21
[datadog_api_key] Fix bug introduced in 3.55.0 for organisation that doesn't have Remove config enabled

### LINKED ISSUE #2847: datadog_api_key  Error: Provider produced inconsistent result after apply

v.3.55.0

### Terraform Version

v1.5.7

 `datadog_api_key`

```hcl

resource "datadog_api_key" "api-key" {
  name = "my-new-key"
}

```

With version 3.55.0 that was just published, I'm running into the following error when trying to deploy a new `datadog_api_key` resource.

```
When applying changes to module.operator.datadog_api_key.api-key, provider "module.operator.provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: .remote_config_read_enabled: was cty.True, but now cty.False.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

I see that the docs say: `Warning : default value is true for backwards compatibility Defaults to true.` Should this be part of a major version release? 

### Expected Behavior

deploys a new `datadog_api_key`

### Actual Behavior

The apply fails. 

```
When applying changes to module.operator.datadog_api_key.api-key, provider "module.operator.provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: .remote_config_read_enabled: was cty.True, but now cty.False.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Steps to Reproduce

terraform apply

### Important Factoids

_No response_

### References

https://github.com/DataDog/terraform-provider-datadog/pull/2831

ecdatadog:
Fixes #2847 

Remove config is configured globally at org level (see /organization-settings/remote-config). 
* When it is not enabled, the `remote_config_read_enabled` argument of api_key is always set to `false` 
* When it is enabled, the `remote_config_read_enabled` argument is set by default to `true` but can be set explicitly to `false`

Proposed fix : 
* When the user don't set `remote_config_read_enabled` , it will be the default value (true if RC is enabled, false if not) . Similar behavior as in <3.55.0
* When the user set `remote_config_read_enabled` to `false` , it will always works 
* When the user set `remote_config_read_enabled` to `true`, 
  * it will succeeds if RC is enabled at org level 
  * it will fails if RC is not enabled at org level

You will get this error if you set the argument while RC is not enabled at org level : 
```
╷
│ Error: remote_config_read_enabled is true but Remote config is not enabled at org level
│ 
│   with datadog_api_key.rctrue,
│   on api.tf line 5, in resource "datadog_api_key" "rctrue":
│    5: resource "datadog_api_key" "rctrue" {
│ 
│ Please either remove remote_config_read_enabled from the resource configuration or enable Remote config at org level
╵
```

### DIFF
--- a/docs/resources/api_key.md
+++ b/docs/resources/api_key.md
@@ -28,7 +28,7 @@ resource "datadog_api_key" "foo" {
 
 ### Optional
 
-- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Should be put to true only if remote config is enabled in /organization-settings/remote-config
+- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Set to true only if remote config is enabled in `/organization-settings/remote-config`.
 
 ### Read-Only
 
--- a/docs/resources/api_key.md
+++ b/docs/resources/api_key.md
@@ -28,7 +28,7 @@ resource "datadog_api_key" "foo" {
 
 ### Optional
 
-- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Warning : default value is true for backwards compatibility Defaults to `true`.
+- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Should be put to true only if remote config is enabled in /organization-settings/remote-config
 
 ### Read-Only
 
--- a/datadog/fwprovider/resource_datadog_api_key.go
+++ b/datadog/fwprovider/resource_datadog_api_key.go
@@ -62,7 +62,7 @@ func (r *apiKeyResource) Schema(_ context.Context, _ resource.SchemaRequest, res
 				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
 			},
 			"remote_config_read_enabled": schema.BoolAttribute{
-				Description: "Whether the API key is used for remote config. Should be put to true only if remote config is enabled in /organization-settings/remote-config",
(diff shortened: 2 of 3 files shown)


Supam (approved):

tyjet (dismissed):

aliciascott (changes requested):
Hi @ecdatadog looks good just a small suggestion, let me know when this is ready for re-review thanks!

docs/resources/api_key.md
```diff
@@ -28,7 +28,7 @@ resource "datadog_api_key" "foo" {
 
 ### Optional
 
-- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Warning : default value is true for backwards compatibility Defaults to `true`.
+- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Should be put to true only if remote config is enabled in /organization-settings/remote-config
```
hestonhoffman, line 31:
```suggestion
- `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Should be set to `true` only if remote config is enabled in `/organization-settings/remote-config`.
```

hestonhoffman (commented):
Left you a small suggestion

aliciascott (approved):
looks good

Supam (approved):

PR #2867, opened 2025-02-25
[Generator] (fix) Support date-time format enrichment

Supam:
## Motivation
Strings with the 'format' fields set to 'date-time' were being read as strings.
Meanwhile the Go client automatically converts them to `*time.Time` and breaks the updateState function.
Such fields are now properly converted back to strings when updating the state.

## Changes
- Added a new formatting function and filter to support date-time enrichment
- Changed the generation to use this filter
- Tidied `.generator/src/generator/utils.py`

### DIFF
--- a/.generator/src/generator/utils.py
+++ b/.generator/src/generator/utils.py
@@ -48,8 +48,9 @@ def is_primitive(schema):
         return True
     return False
 
+
 def is_required(schema, attr=None):
-    req_args =schema.get("required")
+    req_args = schema.get("required")
     if req_args is None:
         return False
     if isinstance(req_args, bool):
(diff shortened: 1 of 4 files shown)


.generator/src/generator/utils.py
```diff
@@ -58,12 +59,20 @@ def is_required(schema, attr=None):
...
     return "enum" in schema
 
+
 def is_nullable(schema):
-    return schema.get("nullable", False)
\ No newline at end of file
+    return schema.get("nullable", False)
+
+
+def print_filter(value):
```
ecdatadog, line 76: As it seems to be a debug function, perhaps rename it to `debug_filter`

ecdatadog (approved):

PR #2882, opened 2025-03-04
[datadog_dashboard] Remove duplicate accepted values in table cell display mode docs

janine-c:
I noticed after https://github.com/DataDog/terraform-provider-datadog/pull/2856 got merged that we had five instances of a line that read:

```
- `cell_display_mode` (List of String) A list of display modes for each table cell. List items one of `number`, `bar`. Valid values are `number`, `bar`, `trend`.
```

I thought it didn't make sense that `trend` would be a valid value, but not a list item. This PR ~adds `trend` as one of the items in all five instances~ removes those duplicate lists.

### DIFF
--- a/docs/resources/dashboard.md
+++ b/docs/resources/dashboard.md
@@ -5484,7 +5484,7 @@ Optional:
 - `alias` (String) The alias for the column name (defaults to metric name).
 - `apm_query` (Block List, Max: 1) The query to use for this widget. (see [below for nested schema](#nestedblock--widget--group_definition--widget--query_table_definition--request--apm_query))
 - `apm_stats_query` (Block List, Max: 1) (see [below for nested schema](#nestedblock--widget--group_definition--widget--query_table_definition--request--apm_stats_query))
-- `cell_display_mode` (List of String) A list of display modes for each table cell. List items one of `number`, `bar`, `trend`. Valid values are `number`, `bar`, `trend`.
+- `cell_display_mode` (List of String) A list of display modes for each table cell. List items one of `number`, `bar`. Valid values are `number`, `bar`, `trend`.
 - `conditional_formats` (Block List) Conditional formats allow you to set the color of your widget content or background, depending on the rule applied to your data. Multiple `conditional_formats` blocks are allowed using the structure below. (see [below for nested schema](#nestedblock--widget--group_definition--widget--query_table_definition--request--conditional_formats))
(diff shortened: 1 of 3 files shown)


docs/resources/dashboard.md
```diff
@@ -5484,7 +5484,7 @@ Optional:
 - `alias` (String) The alias for the column name (defaults to metric name).
 - `apm_query` (Block List, Max: 1) The query to use for this widget. (see [below for nested schema](#nestedblock--widget--group_definition--widget--query_table_definition--request--apm_query))
 - `apm_stats_query` (Block List, Max: 1) (see [below for nested schema](#nestedblock--widget--group_definition--widget--query_table_definition--request--apm_stats_query))
-- `cell_display_mode` (List of String) A list of display modes for each table cell. List items one of `number`, `bar`. Valid values are `number`, `bar`, `trend`.
+- `cell_display_mode` (List of String) A list of display modes for each table cell. List items one of `number`, `bar`, `trend`. Valid values are `number`, `bar`, `trend`.
```
fpighi, line 5487: Docs are auto-generated. See https://github.com/DataDog/terraform-provider-datadog/blob/master/DEVELOPMENT.md#generating-documentation for more details.

You might want to change the description at [this line](https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/resource_datadog_dashboard.go#L4522).
When doing that, I think the right fix is to remove the hardcoded values in the description, since they are provided by the Datadog Go SDK and as you can see they are repeated twice here.

fpighi (changes requested):

docs/resources/dashboard.md
janine-c: Thank you, I didn't realize they were auto-generated 😅 I've made the change to the line you suggested. I appreciate your help!

fpighi: Looks good, but you forgot to re-run the docs generation so the linter is complaining they are not up to date 😅
All you need to do for this is run `make docs` and commit the output.

estherk15 (dismissed):

docs/resources/dashboard.md
janine-c: Thanks for your patience; this is very obviously my first time in this repo 😅 I think I got it now!

fpighi (approved):

damfleu (approved):

aditir12 (approved):

PR #2886, opened 2025-03-06
[synthetics_global_variable] Migrate synthetics resource and data_source to framework

AntoineDona:
This PR migrates the `synthetics_global_variable` resource and data_source to the new terraform framework.


docs/resources/synthetics_global_variable.md
```diff
@@ -3,16 +3,18 @@
 page_title: "datadog_synthetics_global_variable Resource - terraform-provider-datadog"
 subcategory: ""
 description: |-
-  Provides a Datadog synthetics global variable resource. This can be used to create and manage Datadog synthetics global variables.
+  Provides a Datadog SyntheticsGlobalVariable resource. This can be used to create and manage Datadog synthetics_global_variable.
```
joepeeples, line 6:
```suggestion
  Provides a Datadog SyntheticsGlobalVariable resource. This can be used to create and manage a Datadog synthetics_global_variable.
```

docs/resources/synthetics_global_variable.md
```diff
@@ -3,16 +3,18 @@
...
 ---
 
 # datadog_synthetics_global_variable (Resource)
 
-Provides a Datadog synthetics global variable resource. This can be used to create and manage Datadog synthetics global variables.
+Provides a Datadog SyntheticsGlobalVariable resource. This can be used to create and manage Datadog synthetics_global_variable.
```
joepeeples, line 11:
```suggestion
Provides a Datadog SyntheticsGlobalVariable resource. This can be used to create and manage a Datadog synthetics_global_variable.
```

joepeeples (approved):

docs/resources/synthetics_global_variable.md
fpighi: Note this needs to be in code, docs are auto-generated.

AntoineDona: Oh I actually didn't mean to change this description. Will revert back to the original!

fpighi (approved):

MintsInc (approved):

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+	Id               types.String                                    `tfsdk:"id"`
+	Name             types.String                                    `tfsdk:"name"`
+	Description      types.String                                    `tfsdk:"description"`
+	Tags             types.List                                      `tfsdk:"tags"`
+	Value            types.String                                    `tfsdk:"value"`
+	Secure           types.Bool                                      `tfsdk:"secure"`
+	ParseTestId      types.String                                    `tfsdk:"parse_test_id"`
+	ParseTestOptions []syntheticsGlobalVariableParseTestOptionsModel `tfsdk:"parse_test_options"`
+	Options          []syntheticsGlobalVariableOptionsModel          `tfsdk:"options"`
+	RestrictedRoles  types.List                                      `tfsdk:"restricted_roles"`
```
ecdatadog, line 47: Even if this field is deprecated, there is a breaking change. It was previously a Set and now is a List.
Could we use a Set so we have no breaking change at the terraform schema/HCL level.

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+		}
+
+		if len(state.Options) > 0 {
+			var options datadogV1.SyntheticsGlobalVariableOptions
+
+			if len(state.Options[0].TotpParameters) > 0 {
+				var totpParameters datadogV1.SyntheticsGlobalVariableTOTPParameters
+
+				if !state.Options[0].TotpParameters[0].Digits.IsNull() {
+					totpParameters.SetDigits(int32(state.Options[0].TotpParameters[0].Digits.ValueInt64()))
```
teodor2312, line 479: Would there be an issue here since you're casting to a `int32` but the value is `ValueInt64`? 🤔

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+	// If `is_fido` is `false` and `value` is not set, return an error.
+	if !isFido && config.Value.IsNull() {
+		resp.Diagnostics.AddAttributeError(
+			frameworkPath.Root("value"),
+			"Invalid Configuration",
+			"`value` must be set.",
+		)
+	}
+
+	// If `is_fido` or `is_totp`, is `secure` should not be `false`
```
teodor2312, line 524:
```suggestion
	// If `is_fido` or `is_totp` are `true`, `secure` should be `true`
```

teodor2312 (dismissed):
I'm still not a expert in the new terraform format but looks good overall. Left some questions just to understand better how things work now 👍

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+	"context"
+	"regexp"
+
+	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV1"
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	frameworkPath "github.com/hashicorp/terraform-plugin-framework/path"
```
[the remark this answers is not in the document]
AntoineDona: This is to align with all other framework resources. I guess the reason for the different naming is to not mix it up with the go path package

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
AntoineDona: Yes,I actually changed it to align with the private location, but this is technically breaking. It would only result in a non empty plan on migration but agreed that this should not be done here.

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+			"name": schema.StringAttribute{
+				Description: "Synthetics global variable name.",
+				Required:    true,
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(regexp.MustCompile(`^[A-Z][A-Z0-9_]+[A-Z0-9]$`), "must be all uppercase with underscores"),
+				},
+			},
+			"description": schema.StringAttribute{
+				Description: "Description of the global variable.",
+				Computed:    true,
```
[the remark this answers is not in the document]
AntoineDona: Yes, this is to handle the case where these are not provided by the user.
In that case, the backend still returns an empty string and the new framework returns an error saying the Plan and the actual new State value are different.
The previous sdkv2 provider was more tolerant and allowed this, so that's why we did not get the error before.

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+				Default:     stringdefault.StaticString(""),
+			},
+			"tags": schema.ListAttribute{
+				Description: "A list of tags to associate with your synthetics global variable.",
+				ElementType: types.StringType,
+				Computed:    true,
+				Optional:    true,
+				Default:     listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})),
+			},
+			"value": schema.StringAttribute{
```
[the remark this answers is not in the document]
AntoineDona: the logic got put in the `ValidateConfig` as we dont allow to pass value when is_fido is defined. Previously, passing a `value` with fido would result in a backend error. Now we return the error earlier.

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+		return
+	}
+
+	r.updateState(ctx, &state, &resp)
+
+	// Save data into Terraform state
+	response.Diagnostics.Append(response.State.Set(ctx, &state)...)
+}
+
+func (r *syntheticsGlobalVariableResource) Create(ctx context.Context, request resource.CreateRequest, response *resource.CreateResponse) {
```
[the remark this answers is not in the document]
AntoineDona: Yes, I removed them as they used a deprecated sdkv2 Retry library, which has no equivalent in the new framework. I don't think they are needed anymore.

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
```diff
@@ -0,0 +1,551 @@
...
+				localTotpParameters.Digits = types.Int64Value(int64(totpParameters.GetDigits()))
+				localTotpParameters.RefreshInterval = types.Int64Value(int64(totpParameters.GetRefreshInterval()))
+				localVariableOptions.TotpParameters = []syntheticsGlobalVariableTotpParametersModel{localTotpParameters}
+			}
+			state.Options = []syntheticsGlobalVariableOptionsModel{localVariableOptions}
+		}
+	}
+
+	if attributes, ok := resp.GetAttributesOk(); ok {
+		state.RestrictedRoles, _ = types.ListValueFrom(ctx, types.StringType, attributes.GetRestrictedRoles())
```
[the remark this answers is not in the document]
AntoineDona: its a diagnostic, but not needed here

datadog/fwprovider/resource_datadog_synthetics_global_variable.go
AntoineDona: Oh ok so I phrased it like this, because if `is_fido` or `is_totp` is set to true, we don't want secure to be set  at all, or set to true. But the phrasing is not ideal you are right! I will change the comment and the error message.

MintsInc (approved):

Supam (approved):

PR #2891, opened 2025-03-07
[datadog_spans_metric] Fix meaningless drifts with include_percentiles

fpighi:
The field shouldn't be a computed field, otherwise its state might be unknown when not passed explicitly. We never compute it anyway, it's only passed in by the practictioner.


fpighi:
closing in favor of https://github.com/DataDog/terraform-provider-datadog/pull/2906

PR #2881, opened 2025-03-03
Migrate `synthetics_private_location` to framework

AntoineDona:

PR #2892, opened 2025-03-10
[synthetics_private_location] Fix unremovable metadata bug

AntoineDona:
### What and why

- #2881
- #2892  (← **you are here**)

This PR fixes a bug where the metadata field would not be removable, by always sending a metadata object, even if null in the state.
It was initially in the migration to framework PR but we decided to split it to a separate one for separation of concerns.

### Steps to reproduce

Terraform apply this resource:
```
resource "datadog_synthetics_private_location" "private_location" {
  name        = "My private location"
  description = "My private location description"
  tags        = ["team:synthetics"]
  metadata {
    restricted_roles = ["role"]
  }
}
```

Then remove the metadata field and apply.

```
resource "datadog_synthetics_private_location" "private_location" {
  name        = "My private location"
  description = "My private location description"
  tags        = ["team:synthetics"]
}
```

- On the sdkv2 version: When doing another terraform plan, you should see that the state was not modified, and the metadata is still there.
- On the framework version, you should see this error:

```
│ 
│ When applying changes to datadog_synthetics_private_location.private_location, provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: .metadata: block count changed from 0 to 1.
│ 
╵
````


teodor2312 (dismissed):

fpighi (changes requested):
I realize this is deprecated feature, so having a specific test for this is not that valuable.

fpighi (approved):

Alexis-Jacob-datadog (approved):

teodor2312 (approved):

PR #2899, opened 2025-03-11
[datadog_rum_retention_filter] Add support for the resource.

kwangjunkim:
Adds terraform support for RUM retention filters. Leverages this public API:
https://docs.datadoghq.com/api/latest/rum-retention-filters/

https://datadoghq.atlassian.net/browse/RUM-8778


fpighi (approved):

hemeryar (changes requested):
This patch differs a bit in the approach taken in the RFC, how do you guarantee the order of the RFs you define? What if you add a new one in the middle of existing filters? I think terraform would probably create a new RF and it would end up at the end. Although unlikely, the same thing could happen with RFs for the same app in different files, I'm not sure the parsing order is deterministic.

I would still recommend having a top level "retention_filterS" resource, for which you define all the RFs for an app, in blocks that are order-aware.

kwangjunkim:
> This patch differs a bit in the approach taken in the RFC, how do you guarantee the order of the RFs you define? What if you add a new one in the middle of existing filters? I think terraform would probably create a new RF and it would end up at the end. Although unlikely, the same thing could happen with RFs for the same app in different files, I'm not sure the parsing order is deterministic.
> 
> I would still recommend having a top level "retention_filterS" resource, for which you define all the RFs for an app, in blocks that are order-aware.

Thanks for pinpointing that. My thought was that ordering is not feasible because of default retention filters created internally when new application is created. Currently, order API requires the list of all retention filter IDs whether is created as default or by user, and Terraform is unclear on how to reorder regarding default retention filters especially which edition mode is COMPLETE.

examples/resources/datadog_rum_retention_filter/resource.tf
```diff
@@ -0,0 +1,10 @@
+# Create new rum_retention_filter resource
+
+resource "datadog_rum_retention_filter" "testing_rum_retention_filter" {
+  app_id      = "<APP_ID>"
```
hemeryar, line 4: Should this be application_id as well to harmonize?

docs/resources/rum_retention_filters_order.md
```diff
@@ -0,0 +1,52 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_rum_retention_filters_order Resource - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Provides a Datadog RumRetentionFiltersOrder resource. This is used to manage the order of Datadog RUM retention filters.Please note that retention_filter_ids should contain all IDs of retention filters, including the default ones created internally for a given RUM application.
```
hemeryar, line 6: nit: missing space, filters.Please -> filters. Please

docs/resources/rum_retention_filters_order.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+# datadog_rum_retention_filters_order (Resource)
+
+Provides a Datadog RumRetentionFiltersOrder resource. This is used to manage the order of Datadog RUM retention filters.Please note that retention_filter_ids should contain all IDs of retention filters, including the default ones created internally for a given RUM application.
+
+## Example Usage
```
hemeryar, line 13: I think we want a more involved example, because this one isn't really realistic, and the setup will be complex. Is the expected use case to have rum_application resource -> retention_filters data source of said application -> other retention_filters resources -> order resources? Do you need depends_on markers to avoid having race conditions?

Also how do you realistically specify the ordering including the default resources? I guess you need to check for their name and then grab their id?

datadog/tests/data_source_datadog_rum_retention_filters_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+const RumRetentionFiltersDataSourceTestAppId = "6d6b8210-efde-4a1d-bc4d-0f9773f21b0c"
```
hemeryar, line 12: nit: comment saying which app it is and where it lives

datadog/tests/data_source_datadog_rum_retention_filters_test.go
```diff
@@ -0,0 +1,37 @@
...
+	_, _, accProviders := testAccFrameworkMuxProviders(context.Background(), t)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV5ProviderFactories: accProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: dataDatadogRumRetentionFilters(),
+				Check: resource.TestMatchResourceAttr("data.datadog_rum_retention_filters.testing_rum_retention_filters",
+					"retention_filters.#", regexp.MustCompile(`^(0|[1-9]\d*)$`)), // check the size of list is >= 0
```
hemeryar, line 26: This doesn't really test much, but unfortunately we cannot guarantee what default retention filters will be there 😢

datadog/tests/resource_datadog_rum_retention_filter_test.go
```diff
@@ -0,0 +1,254 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/fwprovider"
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
+)
+
+const AppId = "17a2877d-5a77-406e-9039-9da24714936e"
```
hemeryar, line 16: Same here

datadog/tests/resource_datadog_rum_retention_filters_order_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func validDatadogRumRetentionFiltersOrder() string {
+	return fmt.Sprintf(`data "datadog_rum_retention_filters" "my_retention_filters" {
+		application_id = %q
+	}
+
+	resource "datadog_rum_retention_filters_order" "testing_rum_retention_filters_order" {
+		application_id = %q
+	    retention_filter_ids = concat([
+		    data.datadog_rum_retention_filters.my_retention_filters.retention_filters[0].id
```
hemeryar, line 27: We're not really re-ordering anything here no? We take the first element returned by the data source and put it back in first position 🤔

datadog/tests/resource_datadog_rum_retention_filters_order_test.go
```diff
@@ -0,0 +1,101 @@
...
+	t.Parallel()
+	_, providers, accProviders := testAccFrameworkMuxProviders(context.Background(), t)
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV5ProviderFactories: accProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: validDatadogRumRetentionFiltersOrder(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDatadogRumRetentionFiltersOrderResourceMatch(providers.frameworkProvider,
```
hemeryar, line 53: This test will fail if any of the default retention filters is deleted and we don't have at least 4 ids right? This sounds like it is going to be hard to maintain, maybe we should instead create a few RFs instead, put them first in a given order, and only assess the ordering on these.

hemeryar (commented):
Added some comments, I think the most important is having a clear idea on how to do a full terraform setup

docs/resources/rum_retention_filters_order.md
kwangjunkim: I can't think of realistic and practical scenario with complexity involving default RFs either 😅 

But based on your feedback, I will update this example covering all path `rum_application creation` -> `rum_retention_filter creation` and `rum_retention_filters retrieval` -> `rum_retention_filters_order creation`. Definitely, user needs to care about race condition between `rum_retention_filter creation` and `rum_retention_filters retrieval` so I will put how to user `depends_on` syntax for it in example.

kwangjunkim:
> Added some comments, I think the most important is having a clear idea on how to do a full terraform setup

Updated example as following. I believe this is good enough for user's understanding.
```
# Create new rum_application resource
resource "datadog_rum_application" "my_rum_application" {
  name = "my-rum-application-test"
  type = "browser"
}
# Retrieve rum_retention_filters for rum_application created above
data "datadog_rum_retention_filters" "my_retention_filters" {
  application_id = resource.datadog_rum_application.my_rum_application.id
}
# Create new rum_retention_filter resource.
# 'depends_on' is to prevent creating rum_retention_filter and retrieving rum_retention_filters from running in parallel for race condition.
resource "datadog_rum_retention_filter" "new_rum_retention_filter" {
  application_id = resource.datadog_rum_application.my_rum_application.id
  name = "testing.rum.retention_filter"
  event_type = "action"
  sample_rate = 60
  query = "@session.has_replay:true"
  enabled = true
  depends_on = [data.datadog_rum_retention_filters.my_retention_filters]
}
# Create new rum_retention_filters_order resource for reordering
resource "datadog_rum_retention_filters_order" "my_rum_retention_filters_order" {
  application_id = resource.datadog_rum_application.my_rum_application.id
  retention_filter_ids = concat([
    for rf in data.datadog_rum_retention_filters.my_retention_filters.retention_filters :
      rf.id if startswith(rf.id, "default")
    ], [datadog_rum_retention_filter.new_rum_retention_filter.id])
}
```

docs/resources/rum_retention_filters_order.md
hemeryar: I'm not entirely convinced by the example you added. The depends_on doesn't really change anything because we're only ever going to be using the default RFs from the data source. The major thing that is missing is a clear explanation that the default retention filters will have an id starting with "default_" and even maybe listing all the actual default RFs so that people can place them accordingly.

hemeryar (commented):
Some more comments!

docs/resources/rum_retention_filters_order.md
kwangjunkim: Basically, data source is supposed to capture resources created not by Terraform, and my intention of using `depends_on` is for data source not to capture resource created through Terraform. I don't think we need to provide a complicated example assuming some user RFs are created externally which is not best practice of using Infrastructure as Code (IaC) template like Terraform. But the updated example like below covering both default and non-default RFs from data source would be enough for user's understanding. What do you think?

```
resource "datadog_rum_retention_filters_order" "my_rum_retention_filters_order" {
  application_id = resource.datadog_rum_application.my_rum_application.id
  retention_filter_ids = concat([
    for rf in data.datadog_rum_retention_filters.my_retention_filters.retention_filters :
    rf.id if startswith(rf.id, "default")],
  [datadog_rum_retention_filter.new_rum_retention_filter.id],
  [
    for rf in data.datadog_rum_retention_filters.my_retention_filters.retention_filters :
    rf.id if !startswith(rf.id, "default")
  ])
}
```

hemeryar: > data source is supposed to capture resources created not by Terraform
Imagine you want to update an existing RF and you just change the rate. You will go through the data source, which will return the already existing RF, which will return more data than initially, which will re-trigger the ordering with your retention directly with its ID and as part of rf.id if !startswith(rf.id, "default"). Or am I getting confused 🤔 ?

hemeryar:
We discussed a bit offline about the shape of the terraform provider. It mimics what has been done for APM, but existing APIs seem extremely confusing and error prone for end customers. Some of my concerns:

- The ordering is particularly messy, as you need to do filtering on a list to get what you want into another list. It is possible for the users to shoot themselves in the foot with their ordering code, and to make it work initially but not on updates.

- The user needs to guess the exact names of the "default_" retention filters if they want to order stuff properly. We cannot provide documentation for these as long as they are changing, as it would prove even more confusing if we have obsolete doc at any point. Changing these IDs will break customer setups.

- An alternative would be having hardcoded and documented default retention filters in a data source, but it is not possible unless we have a guarantee on default retention filters, which will probably never happen. Any change would break customer setups.

With those reserves, I don't want to block that work from going through, and will approve!

hemeryar (dismissed):

docs/resources/rum_retention_filters_order.md
kwangjunkim: Fair point. 1 and 2 are to be added twice in reordering input in update use-case.
According to [the official doc of Terraform about data source](https://developer.hashicorp.com/terraform/language/data-sources),

> Data sources allow Terraform to use information defined outside of Terraform, defined by another separate Terraform configuration, or modified by functions.

we need to include RFs created by separate Terraform files as well as others, but within the scope of the RF data source, there is no way to determine which RFs were created by the current Terraform file. Filtering out RFs based on whether they were created by Terraform does not help us distinguish which RFs were created by the user's current Terraform file, which makes providing the entire list of RFs the better decision.

So although it is getting uglier 😓 , we might end up with adding additional condition in our example like below.
```
    [for rf in data.datadog_rum_retention_filters.my_retention_filters.retention_filters :
      rf.id if !startswith(rf.id, "default") && rf.id != datadog_rum_retention_filter.new_rum_retention_filter.id
    ]                                                                                                  
``` 
With this, I don't think using `depends_on` is meaningless in the example.

datadog/tests/resource_datadog_rum_retention_filters_order_test.go
kwangjunkim: It is more like manually editing test data not manipulating cassette recording, so what is recorded in cassette is intact. We still have limited edition mode imposed, so we can't change the position for the first default RF for cassette recording.

hemeryar (dismissed):
See above comment for concerns, unblocking 👍

janine-c (approved):

buraizu (approved):

ecdatadog (approved):

PR #2901, opened 2025-03-12
[datadog_integration_aws_account] Change namespace_filters attributes from List to Set

raymondeah:
We're currently generating meaningless diffs in some cases in the `datadog_integration_aws_account` resource for `metrics_config.namespace_filters.exclude_only` and `metrics_config.namespace_filters.include_only` because of the list ordering, for example: 
```
 ~ metrics_config {
            # (4 unchanged attributes hidden)

          ~ namespace_filters {
              ~ exclude_only = [
                  - "AWS/ApiGateway",
                    "AWS/AutoScaling",
                  + "AWS/ApiGateway",
                ]
            }
        }
```
Order doesn't matter here and values are unique so we can change both of these from `List` to `Set`


baer95:
Related issues: #2903 #2904

raymondeah:
Closing this as it may break list indexing and opting for a different solution

PR #2906, opened 2025-03-14
[datadog_spans_metric] Suppress unknown include_percentiles state

### LINKED ISSUE #2888: datadog_spans_metric resource with count aggregation always show diff even if resources have not changed

v3.56.0

### Terraform Version

1.9.8

datadog_spans_metric resources with an aggregation_type of count

```hcl
resource "datadog_spans_metric" "alarm_name" {
...
compute {
   aggregation_type = "count"
 }
...
```

_No response_

### Expected Behavior

If the resource has not changed in either the terraform configuration or in the configuration on DataDog itself terraform plan should not show any changes to the resource. 

### Actual Behavior

Every time we run a plan with a datadog_spans_metric that has a count aggregation it always shows the resource as being updated in place:

```
# datadog_spans_metric.alarm_name will be updated in-place
~ resource "datadog_spans_metric" "alarm_name" {

    ~ compute {
        + include_percentiles = (known after apply)
      }

      # (4 unchanged blocks hidden)
  }
```

In the apply step nothing happens. This makes the terraform plans noisy since we have to double check if things are actually changing or if its datadog_spans_metric resource. 

### Steps to Reproduce

Create a datadog_spans_metric with the count compute aggregation type and apply a plan without changing the resource. 

### Important Factoids

_No response_

### References

https://github.com/DataDog/terraform-provider-datadog/issues/2854

fpighi:
This occurs in the event of tag_name being normalized (e.g. from uppercase to lowercase). When that occurs, TF will mark include_percentiles as unknown, since it's a calculated field. We know that in this case nothing should happen though.

Should fix #2888

### DIFF
--- a/datadog/fwprovider/resource_datadog_spans_metric.go
+++ b/datadog/fwprovider/resource_datadog_spans_metric.go
@@ -20,8 +20,9 @@ import (
 )
 
 var (
-	_ resource.ResourceWithConfigure   = &spansMetricResource{}
-	_ resource.ResourceWithImportState = &spansMetricResource{}
+	_ resource.ResourceWithConfigure      = &spansMetricResource{}
+	_ resource.ResourceWithImportState    = &spansMetricResource{}
+	_ resource.ResourceWithValidateConfig = &spansMetricResource{}
 )
 
 type spansMetricResource struct {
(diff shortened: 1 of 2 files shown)


Supam (dismissed):

Supam (approved):

datadog/fwprovider/resource_datadog_spans_metric.go
```diff
@@ -143,6 +147,58 @@ func (r *spansMetricResource) Schema(_ context.Context, _ resource.SchemaRequest
 	}
 }
 
+type includePercentilesUnknownSuppressor struct{}
+
+func SuppressUnknowns() planmodifier.Bool {
```
ecdatadog, line 152: Have a less generic name for this function that will be used only in for this resource or perhaps not even have it this function and directly use `includePercentilesUnknownSuppressor{}` in the PlanModifiers definition

ecdatadog (dismissed):

skarimo (approved):

PR #2917, opened 2025-03-18
Enrich documentation is now working with multi level nested Blocks 

ecdatadog:
Previously : the enrichment of documentation (ex : Default or Valid values) were only managed for Attributes and attributes at the first level Blocks 

With this PR, it works for all attributes whatever the block level.


Supam (approved):

fpighi (approved):
🚀

Issue #2663, opened 2024-11-12
unable to import datadog_integration_azure


dhduvall:
### Datadog Terraform Provider Version

v3.47.0

### Terraform Version

tofu 1.8.5

### What resources or data sources are affected?

datadog_integration_azure

### Terraform Configuration Files

```hcl
terraform {
    required_providers {
      datadog = {
        source  = "DataDog/datadog"
        version = "~> 3.47.0"
      }
    }

    required_version = ">= 1.1.0"
  }

  provider "datadog" {
    api_key = var.datadog_api_key
    app_key = var.datadog_app_key
  }

  variable "datadog_api_key" {
    type = string
  }

  variable "datadog_app_key" {
    type = string
  }

  resource "datadog_integration_azure" "testing" {
    tenant_name = "tenant-uuid"
    client_id = "client-app-uuid"
    client_secret = "mumble"
  }
```

### Relevant debug or panic output

https://gist.github.com/dhduvall/4bdd469d9a9d8824550a479a35891282

### Expected Behavior

The integration resource should have been imported into the state.

### Actual Behavior

Error complaining about a non-existent object, despite the JSON representing the object having just been returned from the DataDog API.

### Steps to Reproduce

Make sure you have an Azure integration, and all the variables populated
```
TF_LOG=TRACE tofu import -no-color -var datadog_app_key=$DD_APP_KEY -var datadog_api_key=$DD_API_KEY \
  datadog_integration_azure.testing tenant-uuid:client-app-uuid
```

fpighi:
@dhduvall Closing this assuming the linked PR fixed it. If you still have problems please do let us know.

PR #2924, opened 2025-03-19
[datadog_azure_integration] Fix import command

agulen:
Azure import resource to terraform is failing with `Error: Cannot import non-existent remote object`

`tenant_name` and `client_id` need to be explicitly set instead of using `ImportStatePassthroughID` based on the request's ID


cswatt (approved):

skarimo (approved):

dhduvall:
Fixes #2663?

PR #2930, opened 2025-03-24
[Doc] (feat) added  description enrichment for more validators

Supam:
### Motivation
Some validators did not have an description enrichment set up. This PRs aims to add description enrichment that takes into account more validators.

### Changes

Enrichment added for the following validators: 
- `oneOfValidator`
-  `regexMatchesValidator`
- `ValidEntityYAMLValidator`
- `cidrIpValidator`
- `lengthAtLeastValidator`
- `betweenValidator`
- `atLeastValidator`

### DIFF
--- a/datadog/internal/fwutils/fw_enrich_schema.go
+++ b/datadog/internal/fwutils/fw_enrich_schema.go
@@ -98,7 +98,7 @@ func buildEnrichedSchemaDescription(rv reflect.Value) {
 					// If we would use "only" Index(i) we would have { 2 <VALUE> }
 					validValuesMsg += fmt.Sprintf("%s`%v`", sep, allowedValues.Index(i).Field(1))
 				}
-				currentDesc = fmt.Sprintf("%s valid values are %s.", currentDesc, validValuesMsg)
+				currentDesc = fmt.Sprintf("%s Valid values are %s.", currentDesc, validValuesMsg)
 				break
 			}
 
@@ -107,11 +107,11 @@ func buildEnrichedSchemaDescription(rv reflect.Value) {
 				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "validEntityYAMLValidator") ||
 				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "cidrIpValidator") ||
 				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "lengthAtLeastValidator") ||
-				// BetweenValidator is a "homemade" validator and does not come from Hashicorp, it lives in out validators package as Float64Between
+				// BetweenValidator is a custom validator and does not come from Hashicorp, it lives in out validators package as Float64Between
 				// It validates a string representation of a float64
 				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "BetweenValidator") {
 				validationMessage := rv_validators.Index(i).Elem().Interface().(validator.String).Description(context.Background())
-				currentDesc = fmt.Sprintf("%s %s.", currentDesc, validationMessage)
+				currentDesc = fmt.Sprintf("%s %s", ensureTrailingPoint(currentDesc), formatDescription(validationMessage))
 				break
 			}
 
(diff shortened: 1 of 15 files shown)


michaelcretzman (approved):

datadog/internal/fwutils/fw_enrich_schema.go
```diff
@@ -81,9 +81,62 @@ func buildEnrichedSchemaDescription(rv reflect.Value) {
...
+				currentDesc = fmt.Sprintf("%s. Valid values are %s.", currentDesc, validValuesMsg)
+				break
+			}
+			if strings.HasPrefix(validators.Index(i).Elem().Type().Name(), "regexMatchesValidator") {
+				validationMessage := validators.Index(i).Elem().FieldByName("message").String()
+				currentDesc = fmt.Sprintf("%s %s", currentDesc, validationMessage)
+				break
+			}
+			if strings.HasPrefix(validators.Index(i).Elem().Type().Name(), "validEntityYAMLValidator") {
+				currentDesc = fmt.Sprintf("%s entity must be a valid entity YAML/JSON structure.", currentDesc)
```
fpighi, line 108: I think we could use the `Description` method from the validator struct here rather than re-writing this description.
This is also true for the other validators.
This way we can probably make all the `if`s in here behave in the same way and refactor quite a bit.

docs/resources/synthetics_global_variable.md
```diff
@@ -28,7 +28,7 @@ resource "datadog_synthetics_global_variable" "test_variable" {
 
 ### Required
 
-- `name` (String) Synthetics global variable name.
+- `name` (String) Synthetics global variable name. must be all uppercase with underscores
```
fpighi, line 31: Needs to be capitalized.

fpighi (changes requested):

skedida (dismissed):

datadog/internal/fwutils/fw_enrich_schema.go
```diff
@@ -81,9 +83,46 @@ func buildEnrichedSchemaDescription(rv reflect.Value) {
...
+				currentDesc = fmt.Sprintf("%s valid values are %s.", currentDesc, validValuesMsg)
+				break
+			}
+
+			// String validators
+			if strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "regexMatchesValidator") ||
+				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "validEntityYAMLValidator") ||
+				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "cidrIpValidator") ||
+				strings.HasPrefix(rv_validators.Index(i).Elem().Type().Name(), "lengthAtLeastValidator") ||
+				// BetweenValidator is a "homemade" validator and does not come from Hashicorp, it lives in out validators package as Float64Between
```
fpighi, line 110:
```suggestion
				// BetweenValidator is a custom validator and does not come from Hashicorp, it lives in our validators package as Float64Between
```

datadog/internal/validators/validators.go
```diff
@@ -266,7 +268,7 @@ func (v BetweenValidator) Description(ctx context.Context) string {
 }
 
 func (v BetweenValidator) MarkdownDescription(_ context.Context) string {
-	return fmt.Sprintf("value must be between %f and %f", v.min, v.max)
+	return fmt.Sprintf("value must be between %.2f and %.2f", v.min, v.max)
```
fpighi, line 271:
```suggestion
	return fmt.Sprintf("Value must be between %.2f and %.2f", v.min, v.max)
```

docs/resources/action_connection.md
```diff
@@ -146,36 +146,36 @@ Optional:
...
 ### Nested Schema for `http.token_auth.token`
 
 Optional:
 
-- `name` (String) Token name
-- `type` (String) Token type
-- `value` (String, Sensitive) Token value
+- `name` (String) Token name string length must be at least 1.
+- `type` (String) Token type valid values are `SECRET`.
+- `value` (String, Sensitive) Token value string length must be at least 1.
```
fpighi, line 169: Given the inconsistencies in using `.` after the `currentDesc` and then not knowing if we need upper or lower case for the validation string, what about adding a small utility function to always append `.` at the end of the client-provided description and always upper casing the validation message?

Something like:

```go
func appendDotIfMissing(s string) string {
	if !strings.HasSuffix(s, ".") {
		return s + "."
	}
	return s
}
```

fpighi (changes requested):

fpighi (approved):

PR #2934, opened 2025-03-27
[datadog_rum_metric] Set `tag_name` parameter as computed

### LINKED ISSUE #2929: Support default tag_name in group_by for datadog_rum_metric resource

v3.58.0

### Terraform Version

v1.11.2

datadog_rum_metric

https://github.com/glenwinters/datadog-terraform-rum-metric-bug/blob/main/main.tf

```
datadog_rum_metric.test: Creating...
╷
│ 
│ When applying changes to datadog_rum_metric.test, provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: .group_by: planned set element
│ cty.ObjectVal(map[string]cty.Value{"path":cty.StringVal("env"), "tag_name":cty.NullVal(cty.String)}) does not correlate with any element in actual.
│ 
```

### Expected Behavior

The [docs](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/rum_metric) say `By default, path is used as the tag name.`. However, not setting tag_name is crashing with an error.

I would like to avoid passing tag_name to avoid duplicating text for every group by.

For example, you could accidentally have a typo that renames the tag:
```
  group_by {
    path = "env"
    tag_name = "evn"
  }
```

I'd prefer to just pass the path if I want the tag name the same:
```
  group_by {
    path = "env"
  }
```

Ideally, this would automatically strip unsupported characters like from RUM attributes (e.g. `@context.dc` becomes `context.dc`). This is the default behavior in the UI.

### Actual Behavior

The apply fails with the provided error output.

### Steps to Reproduce

Run "terraform apply" on the example repo I shared

### Important Factoids

_No response_

### References

_No response_

ecdatadog:
Fixes #2929 

As described in the [documentation](https://docs.datadoghq.com/api/latest/rum-metrics/#create-a-rum-based-metric), the `group_by.tag_name` is optional and by default has the value of `path` 

After discussing with @hemeryar, ModifyPlan is prefered rather than Computed field : 
* Drawbacks of Computed field : depending on the path you can get differents results which is at least disturbing
   * if you create directly `{"path": "test"}`, you get `{"path": "test", "tag_name": "test"}`
   * if you create a resource with `{"path": "test", "tag_name": "test_name"}` then update it to `{"path": "test"} you get `{"path": "test", "tag_name": "test_name"}` 

The main issue of ModifyPlan is it is not explicit in the schema and less visible than a Computed field but it provides more consistent results.


Supam (approved):

fpighi (approved):

hemeryar (approved):
Discussed offline the possibility of using ModifyPlan instead, but letting the final decision to experts in the area!

fpighi (approved):

hemeryar (approved):
lgtm, thanks!

Supam (approved):

PR #2936, opened 2025-03-28
[datadog_rum_metric] Fix issue with rum metric resource.

hemeryar:
While having a look at https://github.com/DataDog/terraform-provider-datadog/pull/2934, I realised we have another issue, which is the inability to completely remove one of the optional blocks (filter, group_by). We also are getting cryptic errors when trying to remove the uniqueness attribute (which is not permitted as an update, because we also need to change the event type).

This is due to the fact that the PATCH syntax doesn't support removal of object attributes, because null is considered a no update. To circumvent it, we use an empty array for group_by, and force a recreate for filter and uniqueness.

Also needed to add workarounds for older terraform versions that had a bug with single nested blocks being removed (https://github.com/hashicorp/terraform/issues/32460)


datadog/tests/resource_datadog_rum_metric_test.go
```diff
@@ -85,7 +85,7 @@ func TestAccRumMetricAttributes(t *testing.T) {
 			{
 				Config: distributionDatadogRumMetricUpdate(uniq),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckDatadogSpansMetricExists(providers.frameworkProvider),
+					testAccCheckDatadogRumMetricExists(providers.frameworkProvider),
```
hemeryar, line 88: Bad copypaste that somehow never triggered anything?

datadog/tests/resource_datadog_rum_metric_test.go
```diff
@@ -113,6 +113,12 @@ func TestAccRumMetricAttributes(t *testing.T) {
 						"datadog_rum_metric.testing_rum_metric", "filter.query", "@service:another-service"),
 				),
 			},
+			{
+				Config: filterDatadogRumMetricDelete(uniq),
+				Check: resource.ComposeTestCheckFunc(
```
kwangjunkim, line 118: Maybe, worth checking value of `filter` in state as intended after deleting?

datadog/fwprovider/resource_datadog_rum_metric.go
```diff
@@ -139,11 +150,16 @@ func (r *rumMetricResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
-						},
 					},
 				},
+				// Setting this at the object level ensures that the resource will be recreated if the uniqueness
+				// block is emptied. There is no valid use case for this (the event type needs to be changed as well,
+				// forcing a recreate anyway), but the user trying it would cause a cryptic provider state mismatch
+				// error, because the API does not support emptying an object attribute. Instead they will receive a
+				// 400 from the API during the recreate.
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
```
kwangjunkim, line 161: Just realized that we don't have testcase for `uniqueness` unlike `filter`. Is it ok since there is no valid use case as described in comment?

kwangjunkim (dismissed):

datadog/tests/resource_datadog_rum_metric_test.go
hemeryar: Done!

datadog/fwprovider/resource_datadog_rum_metric.go
hemeryar: Done, and allowed me to catch more bugs (the resource cannot be updated in place, I was missing a RequiresReplace from the beginning). I only test that the plan does not cause an error when deleting the attribute.

hemeryar (commented):
Addressed your comments @kwangjunkim and also added a workaround for old tf versions test failures. Took me while to find the offending bug, maybe it will be useful for you as well :)

fpighi (dismissed):

datadog/fwprovider/resource_datadog_rum_metric.go
```diff
@@ -324,12 +335,32 @@ func (r *rumMetricResource) ModifyPlan(ctx context.Context, req resource.ModifyP
...
+	// This is workaround for https://github.com/hashicorp/terraform/issues/32460,
+	// which is fixed in Terraform 1.4.0 and above. Otherwise, removing the filter after
+	// it was created will cause an internal error.
+	if config.Filter == nil {
+		plan.Filter = nil
+
+		// The RequiresReplaceIf() for filter is not run due to the same bug, so force it here.
+		resp.RequiresReplace = append(resp.RequiresReplace, frameworkPath.Root("filter"))
+	}
+	resp.Diagnostics.Append(resp.Plan.Set(ctx, &plan)...)
```
ecdatadog, line 355: Nit : this Append is useless (and could potentially duplicates errors linked to Filter creation)

ecdatadog (approved):

datadog/fwprovider/resource_datadog_rum_metric.go
hemeryar: Addressed the nit, but lost the approval :)

ecdatadog (approved):

PR #2942, opened 2025-04-01
[datadog_monitor] Update docs for on_missing_data compatibility

phillip-dd:
Support for metric monitors with on_missing_data was rolled out in the Fall. Updating the docs here to reflect that only a few monitor types are _not_ supported.


datadog/resource_datadog_monitor.go
```diff
@@ -173,7 +173,7 @@ func resourceDatadogMonitor() *schema.Resource {
...
+					Description:   "Controls how groups or monitors are treated if an evaluation does not return any data points. The default option results in different behavior depending on the monitor query type. For monitors using `Count` queries, an empty monitor evaluation is treated as 0 and is compared to the threshold conditions. For monitors using any query type other than `Count`, for example `Gauge`, `Measure`, or `Rate`, the monitor shows the last known status. This option is not available for Service Check, Composite or SLO monitors. Valid values are: `show_no_data`, `show_and_notify_no_data`, `resolve`, and `default`.",
```
drichards-87, line 176:
```suggestion
					Description:   "Controls how groups or monitors are treated if an evaluation does not return any data points. The default option results in different behavior depending on the monitor query type. For monitors using `Count` queries, an empty monitor evaluation is treated as 0 and is compared to the threshold conditions. For monitors using any query type other than `Count`, for example `Gauge`, `Measure`, or `Rate`, the monitor shows the last known status. This option is not available for Service Check, Composite, or SLO monitors. Valid values are: `show_no_data`, `show_and_notify_no_data`, `resolve`, and `default`.",
```

drichards-87 (approved):
Left a very tiny grammatical suggestion from Docs and approved the PR.

fpighi (dismissed):

fpighi (approved):

PR #2947, opened 2025-04-03
[datadog_synthetics_tests] Fix encoding bug on plain_proto_file / compressedProtoFile

ecdatadog:
A past bug in Datadog UI corrupted the value of `compressedProtoFile` for gRPC tests or multistep API tests with gRPC step(s).

This field should be zlib-compressed then base64-encoded. Unfortunately, the past UI bug made it so that when updating an synthetics test in the UI, the `compressedProtoFile` was only base64-encoded.

While the issue is now fixed at UI level, we have to handle corrupted tests by accepting without error both a b64 or a zlib+b64 field.

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	_nethttp "net/http"
 	"regexp"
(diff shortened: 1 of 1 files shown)


Drarig29 (dismissed):
LTGM.

The UI bug was fixed on the 28th of January. So we are introducing this fix for tests which got corrupted by a manual UI update before that date.

Confirmed it works as expected:
- Imported a corrupted test into a new TF file with `terraformer import`, **using the patched provider**
- Created the test from scratch with `terraform apply`
- In the UI, removed the proto file, re-imported the file, and saved the test
  - The test still passes ✅ 
- Re-updated the test with `terraform apply` and accepted the drift
  - The test still passes ✅ 
- The test is now **not** corrupted (version history shows `compressedProtoFile` going from a buffer starting with `0x73`, to a buffer starting with `0x78` which is a zlib header)

fpighi (dismissed):

fpighi (approved):

PR #2950, opened 2025-04-03
[datadog_integration_slack_channel] Add mute_button argument

### LINKED ISSUE #67: Add support for Slack integration

ecdatadog:
Linked to #67 

About the implementation : 
* I set the argument as `Default:true` to have the same behavior compared to the other parameters (no surprise for the user) but the argument seems to be default false at the API level 
* Name of the argument : `mute_buttons` is the same as the API level. In Datadog UI, it is named `muting`


jhgilbert (dismissed):

Tim-Vuong (approved):
Looks good to me! Agree with defaulting mute_buttons to true to enable mute buttons by default.

Supam (approved):

PR #2975, opened 2025-04-16
provider for custom frameworks

nkonjeti:


docs/resources/custom_framework.md
```diff
@@ -0,0 +1,59 @@
...
+- `id` (String) The ID of the custom framework resource.
+
+<a id="nestedblock--requirements"></a>
+### Nested Schema for `requirements`
+
+Required:
+
+- `name` (String) The name of the requirement.
+
+Optional:
```
nkonjeti, line 41: It auto generates as optional even though I have a validator in the schema requiring controls

datadog/tests/resource_datadog_custom_framework_test.go
```diff
@@ -0,0 +1,636 @@
...
+					}),
+					resource.TestCheckTypeSetElemAttr(path, "requirements.*.controls.*.rules_id.*", "def-000-be9"),
+					resource.TestCheckTypeSetElemAttr(path, "requirements.*.controls.*.rules_id.*", "def-000-cea"),
+				),
+			},
+		},
+	})
+}
+
+func TestCustomFramework_delete(t *testing.T) {
```
nkonjeti, line 70: I will delete this test since delete is already tested in another tests

brett0000FF (approved):

docs/resources/custom_framework.md
```diff
@@ -0,0 +1,59 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
```
nkonjeti, line 2: This was auto generated and the cassettes were also auto generated. Everything else was written

datadog/fwprovider/resource_datadog_custom_framework.go
```diff
@@ -0,0 +1,358 @@
...
+func NewCustomFrameworkResource() resource.Resource {
+	return &customFrameworkResource{}
+}
+
+func (r *customFrameworkResource) Metadata(_ context.Context, _ resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = "custom_framework"
+}
+
+func (r *customFrameworkResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
```
vbarth2, line 45: can you use the term custom compliance framework instead of custom framework throughout these files?

datadog/fwprovider/resource_datadog_custom_framework.go
```diff
@@ -0,0 +1,358 @@
...
+				Description: "The URL of the icon representing the framework.",
+				Optional:    true,
+			},
+			"description": schema.StringAttribute{
+				Description: "The description of the framework.",
+				Optional:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"requirements": schema.SetNestedBlock{
```
nkonjeti, line 74: Can't set a nested block Requirements required as true and can't set nested block Controls required as true. When generating the docs Controls is set as optional due to this.

datadog/fwprovider/resource_datadog_custom_framework.go
```diff
@@ -0,0 +1,358 @@
...
+	// 3. run terraform plan
+	// 4. terraform will see that the framework does not exist so it will remove it from the state
+	// 5. no changes are calculated by terraform so nothing is updated
+
+	// UPDATE example:
+	// 1. create the framework in terraform
+	// 2. delete the framework in the UI
+	// 3. update the framework in terraform
+	// 4. run terraform plan
+	// 5. terraform will see that the framework does not exist so it will remove it from the state
```
vbarth2, line 178: Can you make this comment a bit more concise? My takeaway from this was if something was modified or deleted in the UI since the terraform was last applied, the state (for that resource) needs to be cleared. Is there something else that I’m missing?

datadog/fwprovider/resource_datadog_custom_framework.go
```diff
@@ -0,0 +1,358 @@
...
+}
+func readStateFromDatabase(data datadogV2.GetCustomFrameworkResponse, handle string, version string) customFrameworkModel {
+	// Set the state
+	var state customFrameworkModel
+	state.ID = types.StringValue(handle + "-" + version)
+	state.Handle = types.StringValue(handle)
+	state.Version = types.StringValue(version)
+	state.Name = types.StringValue(data.GetData().Attributes.Name)
+	state.Description = types.StringValue(data.GetData().Attributes.Description)
+	state.IconURL = types.StringValue(data.GetData().Attributes.IconUrl)
```
vbarth2, line 264: will this fail if they don’t provide the optional attributes?

datadog/internal/validators/duplicate_control_validator.go
```diff
@@ -0,0 +1,56 @@
...
+func (v controlNameValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v controlNameValidator) ValidateSet(ctx context.Context, req validator.SetRequest, resp *validator.SetResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	// Get all control names from the configuration
```
vbarth2, line 27: you can remove

datadog/internal/validators/duplicate_control_validator.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, control := range req.ConfigValue.Elements() {
+		controlObj := control.(types.Object)
+		name := controlObj.Attributes()["name"].(types.String).ValueString()
+		log.Printf("Found control name in config: %s", name)
+		controlNames = append(controlNames, name)
+	}
+
+	log.Printf("Found %d control names in config", len(controlNames))
+
+	// Check for duplicates in the list
```
vbarth2, line 38: can remove

datadog/internal/validators/duplicate_requirement_validator.go
```diff
@@ -0,0 +1,56 @@
...
+func (v requirementNameValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v requirementNameValidator) ValidateSet(ctx context.Context, req validator.SetRequest, resp *validator.SetResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	// Get all requirement names from the configuration
```
vbarth2, line 27: can remove

datadog/fwprovider/resource_datadog_custom_framework.go
nkonjeti: nope you are right I wrote out the steps that would happen - realized this is not needed!

datadog/fwprovider/resource_datadog_on_call_schedule.go
```diff
@@ -151,17 +151,17 @@ func (r *onCallScheduleResource) Schema(_ context.Context, _ resource.SchemaRequ
 							NestedObject: schema.NestedBlockObject{
```
nkonjeti, line 151: NOTE TO REVIEWER: this was a temp fix change I had to make before the official fix goes in - this won't be merged in so you can ignore this!!!

docs/resources/custom_framework.md
nkonjeti: the fix for this: manually editing the file since the api-spec channel ack'ed this and said a solution still needs to go out

datadog/fwprovider/resource_datadog_custom_framework.go
nkonjeti: yup - changed the logic to account for this

datadog/fwprovider/resource_datadog_custom_framework.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// Convert requirements to set
```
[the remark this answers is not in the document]
nkonjeti: this comment was old - removed it

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
```diff
@@ -0,0 +1,355 @@
...
+	diags := request.State.Get(ctx, &state)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	data, _, err := r.Api.GetCustomFramework(r.Auth, state.Handle.ValueString(), state.Version.ValueString())
+	// If the framework does not exist, remove it from terraform state
+	// This is to avoid the provider to return an error when the framework is deleted in the UI prior
+	if err != nil && err.Error() == "400 Bad Request" {
```
vbarth2, line 176: Are there any other scenarios that throw a 400? Can we case on something else?

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	state.Requirements = setRequirements(requirements)
+	return state
+}
+
+// ImportState is used to import a resource from an existing framework so we can update it if it exists in the database and not in terraform
+func (r *complianceCustomFrameworkResource) ImportState(ctx context.Context, request resource.ImportStateRequest, response *resource.ImportStateResponse) {
+	// Split the ID into handle and version
+	// The last hyphen separates handle and version
+	lastHyphenIndex := strings.LastIndex(request.ID, "-")
```
vbarth2, line 280: can we directly pass the handle and version instead of splitting the id

datadog/internal/validators/duplicate_requirement_validator.go
```diff
@@ -0,0 +1,54 @@
...
+func (v requirementNameValidator) ValidateSet(ctx context.Context, req validator.SetRequest, resp *validator.SetResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	var requirementNames []string
+	for _, requirement := range req.ConfigValue.Elements() {
+		reqObj := requirement.(types.Object)
+		name := reqObj.Attributes()["name"].(types.String).ValueString()
+		log.Printf("Found requirement name in config: %s", name)
```
vbarth2, line 31: can you add `[INFO] ` in front of the non-error logs

examples/resources/datadog_compliance_custom_framework/resource.tf
```diff
@@ -0,0 +1,24 @@
+resource "datadog_compliance_custom_framework" "example" {
+  version  = "1"
+  handle   = "new-terraform-framework-handle"
+  name     = "new-terraform-framework"
+  icon_url = "https://example.com/icon.png"
+  requirements {
+    name = "requirement1"
+    controls {
+      name     = "control1"
+      rules_id = ["aaa-000-ccc", "bbb-000-ddd"]
```
vbarth2, line 10: Can we use real rule IDs here

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
nkonjeti: for GetCustomFramework the other ways to get a 400 is if the handle or version is empty but the validator already checks that - so the only way to get a 400 if the framework doesn't exist 

But I will try to use another specific case instead of 400 bad request

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
nkonjeti: unfortunately the only way to do it using the id. This can be a problem if the handle and version can contain '-'. In the UI it says the version should only have numbers and letters but I will double check on the backend if we validate this.

nkonjeti: also I realized the terraform provider using the ID to uniquely identify the framework

so if we have 
handle "neha-framework"
version "2"

this would be the same as
handle "neha"
version "framework-2"

nkonjeti: I think instead of importing the resource - because we can't directly pass in the handle and version I will change the create function to update the framework if the handle and version exists.

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
```diff
@@ -0,0 +1,341 @@
...
+	data, httpResp, err := r.Api.GetCustomFramework(r.Auth, state.Handle.ValueString(), state.Version.ValueString())
+	// If the framework does not exist, remove it from terraform state
+	// This is to avoid the provider to return an error when the framework is deleted in the UI prior
+	if err != nil && httpResp != nil && httpResp.StatusCode == 400 {
+		// 400 could only mean the framework does not exist
+		// because terraform would have already validated the framework in the create function
+		response.State.RemoveResource(ctx)
+		return
+	}
+	if err != nil {
```
vbarth2, line 188: can we switch the order of this and the if block on 182

nkonjeti: yup will just check that it's not 400 before

nkonjeti: actually I can't switch the ordering because if I check if err is not nil without checking the status code first it would return before checking if the status code is 400

datadog/internal/validators/duplicate_requirement_validator.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (v requirementNameValidator) Description(context.Context) string {
+	return "checks for duplicate requirement names"
+}
+
+func (v requirementNameValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v requirementNameValidator) ValidateList(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) {
```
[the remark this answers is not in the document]
nkonjeti: I just wanted different error messaging and I thought this approach was cleaner. Instead I could check for duplicate control names in the requirement validator

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
```diff
@@ -0,0 +1,388 @@
...
+	if data.GetData().Attributes.IconUrl != nil {
+		state.IconURL = types.StringValue(*data.GetData().Attributes.IconUrl)
+	}
+	// since the requirements and controls from the API response might be in a different order than the state
+	// we need to sort them to match the state so terraform can detect the changes
+	// without taking order into account
+	apiReqMap := make(map[string]datadogV2.CustomFrameworkRequirement)
+	apiCtrlMap := make(map[string]map[string]datadogV2.CustomFrameworkControl)
+
+	for _, req := range data.GetData().Attributes.Requirements {
```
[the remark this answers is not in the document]
nkonjeti: I didn't separate into functions before because the logic isn't reused anywhere else. However, the loop did get a bit long so will separate it out

datadog/fwprovider/resource_datadog_compliance_custom_framework.go
```diff
@@ -0,0 +1,390 @@
...
+			"icon_url": schema.StringAttribute{
+				Description: "The URL of the icon representing the framework",
+				Optional:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"requirements": schema.ListNestedBlock{
+				Description: "The requirements of the framework.",
+				Validators: []validator.List{
+					validators.DuplicateRequirementControlValidator(),
```
nkonjeti, line 102: change: merged the two validators into one

NaOHman (approved):

PR #2978, opened 2025-04-17
[datadog_synthetics_test] SYN-4083: Update tick-every description with value ranges

aleks-dog:
**What's New / Different**
SYN-4083: Update tick-every description with value ranges for different test types

### DIFF
--- a/datadog/resource_datadog_synthetics_test_.go
+++ b/datadog/resource_datadog_synthetics_test_.go
@@ -679,7 +679,7 @@ func syntheticsTestOptionsList() *schema.Schema {
 				"allow_insecure":   syntheticsAllowInsecureOption(),
 				"follow_redirects": syntheticsFollowRedirectsOption(),
 				"tick_every": {
-					Description:  "How often the test should run (in seconds). Valid range is `30-604800 for API tests and `60-604800` for browser tests.",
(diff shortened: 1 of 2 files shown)


janine-c (approved):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -848,7 +848,7 @@ func syntheticsMobileTestOptionsList() *schema.Schema {
 				},
 				"retry": syntheticsTestOptionsRetry(),
 				"tick_every": {
-					Description:  "How often the test should run (in seconds).",
+					Description:  "How often the test should run (in seconds). Valid range is `300-604800` for mobile tests.",
```
Alexis-Jacob-datadog, line 851: I'm wondering if you need to specify for mobile tests since you are already in the section of MobileTestOptionsList 🤔 ? 
Except that LGTM for me

Alexis-Jacob-datadog (approved):

datadog/resource_datadog_synthetics_test_.go
Drarig29: I think it's a good thing to emphasize that each test type has different settings

amaskara-dd (approved):

PR #2980, opened 2025-04-18
[datadog_monitor_notification_rule] Add support for Monitor Notification Rule

shaneyuandd:
This PR introduces a new Terraform resource: `datadog_monitor_notification_rule`, which supports managing [Datadog Monitor Notification Rules](https://docs.datadoghq.com/api/latest/monitors/?code-lang=curl#get-a-monitor-notification-rule).

This resource allows users to define reusable notification rules that can be applied to multiple monitors via tag filters. It simplifies and centralizes notification configuration, reducing the need to manage alert channels individually across many monitors.

#### Example
```
resource "datadog_monitor_notification_rule" "foo" {
  name       = "A notification rule name"
  recipients = ["slack-test-channel", "jira-test"]
  filter {
    tags = ["env:abc"]
  }
}
```


datadog/fwprovider/resource_datadog_monitor_notification_rule.go
```diff
@@ -0,0 +1,240 @@
...
+		Blocks: map[string]schema.Block{
+			"tags_filter": schema.SingleNestedBlock{
+				Attributes: map[string]schema.Attribute{
+					"tags": schema.SetAttribute{
+						Required:    true,
+						ElementType: types.StringType,
+						Description: "Tags that all target monitors must match.",
+					},
+				},
+				Validators: []validator.Object{objectvalidator.IsRequired()},
```
shaneyuandd, line 79: This can be potentially confusing. The generated doc listed this field as `Optional` for now (which matches our low level API `oneOf`). However, given that we only support `tags_filter` right now, in practice this is actually a required field.

docs/resources/monitor_notification_rule.md
```diff
@@ -0,0 +1,56 @@
...
+### Read-Only
+
+- `id` (String) The ID of this resource.
+
+<a id="nestedblock--filter"></a>
+### Nested Schema for `filter`
+
+Required:
+
+- `tags` (Set of String) Tags that all target monitors must match.
```
maycmlee, line 48:
```suggestion
- `tags` (Set of String) All tags that target monitors must match.
```

maycmlee (commented):
Just a small suggestion

carlmartensen (dismissed):
Nice 👍

datadog/fwprovider/resource_datadog_monitor_notification_rule.go
```diff
@@ -0,0 +1,251 @@
...
+			"recipients": schema.SetAttribute{
+				Required:    true,
+				ElementType: types.StringType,
+				Description: "List of recipients to notify.",
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"filter": schema.SingleNestedBlock{
+				Attributes: map[string]schema.Attribute{
+					"tags": schema.SetAttribute{
```
carlmartensen, line 82: Consider setting a min size `setvalidator.SizeAtLeast(1` with recipients and tags

datadog/fwprovider/framework_provider.go
```diff
@@ -437,6 +438,12 @@ func defaultConfigureFunc(p *FrameworkProvider, request *provider.ConfigureReque
 	ddClientConfig.SetUnstableOperationEnabled("v2.UpdatePipeline", true)
 	ddClientConfig.SetUnstableOperationEnabled("v2.DeletePipeline", true)
 
+	// Enable MonitorNotificationRule
+	ddClientConfig.SetUnstableOperationEnabled("v2.CreateMonitorNotificationRule", true)
```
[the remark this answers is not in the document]
shaneyuandd: Preview endpoints are set by default as "Unstable Operation" on [Go client](https://github.com/DataDog/datadog-api-client-go/blob/65afec6a446323c465509119879cd3d702498a92/api/datadog/configuration.go#L367). Thus, the override is required, until status is cleared.

datadog/fwprovider/resource_datadog_monitor_notification_rule.go
shaneyuandd: Will add these checks

datadog/fwprovider/resource_datadog_monitor_notification_rule.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	r.updateState(ctx, &state, &resp)
+
+	// Save data into Terraform state
+	response.Diagnostics.Append(response.State.Set(ctx, &state)...)
+}
+
+func (r *MonitorNotificationRuleResource) Delete(ctx context.Context, request resource.DeleteRequest, response *resource.DeleteResponse) {
+	var state MonitorNotificationRuleModel
+	response.Diagnostics.Append(request.State.Get(ctx, &state)...)
```
MintsInc, line 183: Any reason why this is not at the end of the function ?

shaneyuandd: This line appends the prior `request.State` into `response.Diagnostics`, which allows provider to know if it is ok to proceed with the deletion. The following line `response.Diagnostics.HasError()` won't catch any error if this is moved towards the end. ([Reference](https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#define-delete-method))

maycmlee (approved):

skarimo (approved):

fpighi (approved):

PR #2986, opened 2025-04-24
Update CODEOWNERS

nguyening:
Metrics Query doesn’t own public API endpoints => Timeseries Query can help triage whether this needs input from deeper in the stack.

points-aggregation => metrics-storage-platform

Metric tag endpoints are owned by metrics-index
Metric Metadata endpoints are owned by metrics-experience


fpighi (approved):

Supam (approved):
LGTM

PR #3000, opened 2025-05-02
[on_call_schedule] Add on-call schedule resource

daniel-m-dd:
Create new resource `datadog_on_call` to manage schedules used for on-call
https://docs.datadoghq.com/api/latest/on-call/#create-on-call-schedule

A schedule can have "empty member"; gap/void users, I would have liked to express it as `member{}` but this does not seem supported by old terraform version [see [here](https://github.com/hashicorp/terraform/pull/32463)]; ~the current workaround is `member{ user{} }` but there might be a better API here.~ ~Edit 1: updated to be `member {}` or `member {user_id=null}}`~
~Edit 2: updated to be `member_ids=[id, null,...]`~
Edit 3: updated to be `users=[id, null, ...]`


datadog/fwprovider/resource_datadog_on_call_schedule.go
```diff
@@ -0,0 +1,803 @@
...
+		Restrictions:         restrictionsModels,
+		Interval:             &intervalModel{Days: types.Int32Value(int32(interval.GetDays())), Seconds: types.Int64Value(interval.GetSeconds())},
+	}
+}
+
+func formatTime(t time.Time) string {
+	return t.Format(time.RFC3339)
+}
+
+func (r *onCallScheduleResource) buildOnCallScheduleRequestBody(ctx context.Context, state *onCallScheduleModel) (*datadogV2.ScheduleCreateRequest, diag.Diagnostics) {
```
daniel-m-dd, line 495: a lot of code is duplicated between Create and Update because the golang sdk use different go types for both actions. This might be simplified by potentially sharing some types in between the 2 operations. (Types are not the same as some minor differences exsit between the 2 operations, for example an PUT/Update might include a layer.id field; a POST/Create can't.

datadog/fwprovider/resource_datadog_on_call_schedule.go
```diff
@@ -0,0 +1,803 @@
...
+										Description: "The weekday when the restriction period starts (Monday through Sunday).",
+									},
+									"start_time": schema.StringAttribute{
+										Optional:    true,
+										Description: "The time of day when the restriction begins (hh:mm:ss).",
+									},
+								},
+							},
+						},
+						"interval": schema.SingleNestedBlock{
```
daniel-m-dd, line 203: we might want to create a special type here to express that at least one of seconds or days should be > 0
we can start with a validator

datadog/fwprovider/resource_datadog_on_call_schedule.go
```diff
@@ -0,0 +1,802 @@
...
+								"days": schema.Int32Attribute{
+									Optional:    true,
+									Computed:    true,
+									Description: "The number of full days in each rotation period.",
+									Default:     int32default.StaticInt32(int32(0)),
+								},
+								"seconds": schema.Int64Attribute{
+									Optional:    true,
+									Computed:    true,
+									Description: "Extra seconds that may be added to extend the rotation beyond whole days.",
```
bndbsh, line 214:
```suggestion
									Description: "For intervals that are not expressible in whole days, this will be added to `days`.",
```

From the doc it's not immediately clear that interval can also be less than one day

Supam (approved):

urseberry (approved):

Bnjis (approved):

bndbsh (approved):

PR #3027, opened 2025-05-26
[datadog_synthetics_test] Change target and targetvalue to be number or string

teodor2312:
We had report from customers that setting the value of `target` to be a number was causing errors so this PR fixes this


AntoineDona (approved):
LGTM 👍

Supam (approved):

PR #3041, opened 2025-06-03
[datadog_synthetics_test] Add support for all api subtypes for multistep tests

teodor2312:
We have enabled in the UI the option to create multistep api tests from every subtype so now we also want to support this through terraform


teodor2312:
This PR is waiting on https://github.com/DataDog/datadog-api-spec/pull/3893 to be merged so all the checks can pass 👍

teodor2312:
We used this file to double check exactly what was needed for each subtype
https://github.com/DataDog/web-ui/blob/b9a64ddf5e3c11f1e300da27b1898864ec57dad5/packages/apps/synthetics/private/runtime/page-recorder-api-multistep/store/helpers/format-recorder-step-to-server.ts#L98

Supam (approved):

Issue #3039, opened 2025-06-02
Regression due to change to synthetic test assertions


schans:
### Datadog Terraform Provider Version

v3.64.0

### Terraform Version

1.11.2

### What resources or data sources are affected?

datadog_synthetics_test

### Terraform Configuration Files

```hcl
resource "datadog_synthetics_test" "tls_check" {
  name      = var.name
  type      = "api"
  subtype   = "ssl"
  status    = "live"
  message   = var.message
  locations = var.locations
  tags      = var.tags

  request_definition {
    host = var.hostname
    port = 443
  }

  assertion {
    type     = "certificate"
    operator = "isInMoreThan"
    target   = 2
  }

  assertion {
    operator = "moreThanOrEqual"
    target   = "1.2"
    type     = "minTlsVersion"
  }

  assertion {
    operator      = "lessThan"
    target        = "20000"
    timings_scope = "all"
    type          = "responseTime"
  }

}
```

### Relevant debug or panic output

Error: error creating synthetics API test from /api/v1/synthetics/tests/api: 400 Bad Request: {"errors":["'assertions' value '{'operator': 'moreThanOrEqual', 'target': 1.2, 'type': 'minTlsVersion'}' is invalid"]}
│ 
│   with module.tls_check["myhostname"].datadog_synthetics_test.tls_check,
│   on ../../modules/synthetic_checks_tls/main.tf line 1, in resource "datadog_synthetics_test" "tls_check":
│    1: resource "datadog_synthetics_test" "tls_check" {

### Expected Behavior

The check gets created successfully like with 3.63.0 and before with the minTlsVersion set to "1.2".

### Actual Behavior

It fails to create the check. It looks like a string is expected but due to the change a float/double is provided.

### Steps to Reproduce

See above code. 

### References

This change broke it:
https://github.com/DataDog/terraform-provider-datadog/pull/3027/files

Listed as bug fix in the changelog:
https://github.com/DataDog/terraform-provider-datadog/compare/v3.63.0...v3.65.0#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR14

donhector:
I have the same 3 types of assertions in my SSL `datadog_synthetics_test` as the OP, and today I went on to update the certificate expiration value using provider "3.65.0" and faced the same error:

[...]

PR #3042, opened 2025-06-03
[datadog_synthetics_test] Exclude TLS version types from being converted to number

teodor2312:
We fixed a bug recently where assertion targets were not being set as numbers correctly, but fixing that caused another bug with TLS versions specifically which can't be numbers


AntoineDona (approved):

AntoineDona:
Citing the [issue](https://github.com/DataDog/terraform-provider-datadog/issues/3039) just for referencing

Issue #2964, opened 2025-04-14
datadog_synthetics_test: There is no way to add form data when choosing body_type multipart/form


hajali-amine:
### What resources or data sources are affected?

datadog_synthetics_test

### Feature Request

Allow us to add form data, for example:

``` hcl
request_definition {
  url = "example.com"
  method = "POST"
  body_type = "multipart/form-data"
  body_form_data {
    type = "file"
    value = "<FILENAME FROM request_file attribute>"
  }
  body_form_data {
    type = "text"
    value = "blahblah"
  }
....
}
```

AntoineDona:
Hi @hajali-amine ! We just released a [new version](https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.66.0) of the provider that fixes your issue. I will close this issue for now, but please let us know if you have any other issue in the mean time.

PR #3048, opened 2025-06-10
[datadog_synthetics_test] Add `form` for `multipart/form-data` API tests

Drarig29:
Requires https://github.com/DataDog/datadog-api-spec/pull/3929

This PR adds a `form` property for API tests with `multipart/form-data` header.


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2626,6 +2631,15 @@ func buildDatadogSyntheticsAPITest(d *schema.ResourceData) *datadogV1.Synthetics
...
 		request.SetBodyType(datadogV1.SyntheticsTestRequestBodyType(attr.(string)))
 	}
+	if attr, ok := d.GetOk("request_definition.0.form"); ok {
+		form := attr.(map[string]interface{})
+		if len(form) > 0 {
+			request.SetForm(make(map[string]string))
+		}
+		for k, v := range form {
+			request.GetForm()[k] = v.(string)
+		}
```
teodor2312, line 2641: Should this be inside the `if`? I guess in that case `form` would still have 0 items so the `for` would not really loop over anything but 🤔

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2924,12 +2947,15 @@ func buildDatadogSyntheticsBrowserTest(d *schema.ResourceData) *datadogV1.Synthe
 	if attr, ok := d.GetOk("request_definition.0.url"); ok {
 		request.SetUrl(attr.(string))
 	}
+
+	// XXX: `body` and `body_type` are accepted by the backend, but ignored by the browser test runner
```
teodor2312, line 2951: is XXX supposed to be a ticket number 🤔

datadog/resource_datadog_synthetics_test_.go
teodor2312: and same for the one below

Drarig29: The loop would not loop indeed. And this aligns with https://github.com/DataDog/terraform-provider-datadog/blob/67b5adf812f8bf2d9b15ccb0dade6f9af3a70974/datadog/resource_datadog_synthetics_test_.go#L2989-L2994

datadog/resource_datadog_synthetics_test_.go
Drarig29: No, let's not remove it now that it's here because it would be a breaking change

Supam (approved):

AntoineDona:
Linked issue: https://github.com/DataDog/terraform-provider-datadog/issues/2964

PR #3049, opened 2025-06-10
[datadog_synthetics_test] Add QoL feedback when converting multistep API test

Drarig29:
This PR adds diagnostics for common mistakes when migrating an API test that was previously terraformed to do 1 thing, into a multistep API test that does this same thing + some extra steps.

For example, here I copied the previous logic and pasted it into an `api_step`, but forgot to remove the leftover `request_definition` and `assertion` at the root:

image

Another common mistake is forgetting to update the `subtype` to `multi`. In that case, here is the error:

image

This PR only impacts `type = "api"` tests, in `resourceDatadogSyntheticsTestCreate` and `resourceDatadogSyntheticsTestUpdate`.

### DIFF
--- a/docs/resources/synthetics_test.md
+++ b/docs/resources/synthetics_test.md
@@ -750,7 +750,7 @@ resource "datadog_synthetics_test" "test_grpc_health" {
 
 ### Optional
 
-- `api_step` (Block List) Steps for multi-step api tests (see [below for nested schema](#nestedblock--api_step))
+- `api_step` (Block List) Steps for multistep api tests (see [below for nested schema](#nestedblock--api_step))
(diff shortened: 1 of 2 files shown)


docs/resources/synthetics_test.md
```diff
@@ -750,7 +750,7 @@ resource "datadog_synthetics_test" "test_grpc_health" {
 
 ### Optional
 
-- `api_step` (Block List) Steps for multi-step api tests (see [below for nested schema](#nestedblock--api_step))
+- `api_step` (Block List) Steps for multistep api tests (see [below for nested schema](#nestedblock--api_step))
```
joepeeples:
```suggestion
- `api_step` (Block List) Steps for multistep API tests (see [below for nested schema](#nestedblock--api_step))
```

AntoineDona (approved):
LGTM 👍

PR #3054, opened 2025-06-13
[app_key_registration] Add App Key Registration Resource 

GabeWM:
[Ticket: 10685](https://datadoghq.atlassian.net/browse/WF-10685)

Motivation: https://docs.google.com/document/d/1GxqhcJLP2RTxqZZBed5lLp2mEehDA1K_Lu_FNgKH0qM/edit?tab=t.0#heading=h.g96a7txsgj2p

Dependent on public API [release](https://github.com/DataDog/datadog-api-spec/pull/3941/files). This PR adds the new app key registration resource.


go.mod
```diff
@@ -102,4 +102,6 @@ require (
 	google.golang.org/protobuf v1.36.6 // indirect
 )
 
-go 1.23.0
\ No newline at end of file
+go 1.23.0
+
+replace github.com/DataDog/datadog-api-client-go/v2 => ../datadog-api-client-go
```
GabeWM: Remove once public API PR gets merged.

datadog/fwprovider/resource_datadog_app_key_registration.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var (
+	_ resource.ResourceWithConfigure   = &appKeyRegistrationResource{}
+	_ resource.ResourceWithImportState = &appKeyRegistrationResource{}
+)
+
+type appKeyRegistrationResource struct {
```
embeaken, line 24: should we have a data source for this too? up to you if it makes sense or not.

datadog/fwprovider/resource_datadog_app_key_registration.go
embeaken, line 1: can you add tests for this?

datadog/fwprovider/resource_datadog_app_key_registration.go
```diff
@@ -0,0 +1,164 @@
...
+	response.TypeName = "app_key_registration"
+}
+
+func (r *appKeyRegistrationResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "Registers App Keys to be used for Action Connection, App Builder, and Workflow Automation. This registration is required to enable API and Terraform use in these products.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "The Application Key ID to register.",
+				Required:    true,
```
embeaken, line 53: shouldn't this be set as `Computed` instead? unless the user is expected to provide their own UUID.

datadog/fwprovider/resource_datadog_app_key_registration.go
```diff
@@ -0,0 +1,164 @@
...
+	appKeyRegistration, httpResp, err := r.Api.GetAppKeyRegistration(r.Auth, state.ID.ValueString())
+	if err != nil {
+		if httpResp != nil {
+			body, err := io.ReadAll(httpResp.Body)
+			if err != nil {
+				response.Diagnostics.AddError("Error reading error response when getting app key registration", err.Error())
+				return
+			}
+
+			// If the app key registration is not found, we log a warning and remove the resource from state. This may be due to changes in the UI or attempting to import an app key that is not registered.
```
embeaken, line 116: I think having a "read" remove the resource from state is unintuitive. I think we should keep it as a 404 and it'll be on the customer to remove any non-existent registrations from their TF config.

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: It doesn't make sense there is nothing to read the only property is the ID

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: Yea

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: This should not be computed. We need them to enter the ID it is the only property.

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: No I don't think so. This actually standard I believe. The reason is if the customer deletes it in the UI and you run any terraform command (like plan) your commands will fail and it is unrecoverable unless they reregister in the UI. The plan will display the user that the terraform resource will be deleted on the next apply.

datadog/fwprovider/resource_datadog_app_key_registration.go
embeaken: It might be useful if someone wants to reference an App Key that isn't TF-managed from another TF resource. If that's not useful then yea a data source wouldn't make sense.

datadog/fwprovider/resource_datadog_app_key_registration.go
embeaken: kk that's fine we just usually generate IDs on the server that's why I asked

datadog/fwprovider/resource_datadog_app_key_registration.go
embeaken: sounds good to me

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: My thought was that this would be done by the app key data source then, not a app key registration data source.

GabeWM:
@embeaken thanks for taking a look.

datadog/fwprovider/data_source_datadog_action_connection.go
```diff
@@ -31,7 +31,7 @@ func (d *actionConnectionDatasource) Metadata(_ context.Context, request datasou
 
 func (d *actionConnectionDatasource) Schema(_ context.Context, request datasource.SchemaRequest, response *datasource.SchemaResponse) {
 	response.Schema = schema.Schema{
-		Description: "A connection that can be used in Actions, including in the Workflow Automation and App Builder products.",
+		Description: "A connection that can be used in Actions, including in the Workflow Automation and App Builder products. This data source requires a [registered application key](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/app_key_registration).",
```
urseberry, line 34: This link gives me a Page Not Found error on the Terraform site. I assume the link is not ready yet?

GabeWM: Yes the link will only work once this is merged.

mg-dd (approved):

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: Done

datadog/fwprovider/resource_datadog_app_key_registration.go
embeaken: i think for that you would still need `app key registration` to be a resource. unless there's a mechanism in the `app key` resource to also create an `app key registration` on it (which seems like a convenient way to do that!)

datadog/fwprovider/resource_datadog_app_key_registration.go
```diff
@@ -0,0 +1,164 @@
...
+	appKeyRegistrationModel := &appKeyRegistrationResourceModel{
+		ID: types.StringValue(appKeyRegistration.Data.Id.String()),
+	}
+
+	diags = response.State.Set(ctx, appKeyRegistrationModel)
+	response.Diagnostics.Append(diags...)
+}
+
+func (r *appKeyRegistrationResource) Update(ctx context.Context, request resource.UpdateRequest, response *resource.UpdateResponse) {
+	// No-op: Changes to the only field `id` require replacement, so Update will never be called.
```
embeaken: very optional suggestion - implement this with an error as a sanity check:

```go
response.Diagnostics.AddError("Update should not be called", "Updating this resource should replace it.")
```

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: The app key registration only provides the app key ID, which makes it somewhat redundant in the context of a read-only resource like a data source. Since the only attribute returned is the app key ID, it doesn’t add meaningful value on its own.

I’m happy to include it if there’s a specific request, but to create the data source, you’d already need the app key ID. The only advantage of including it would be to fail early if the app key isn’t registered—though that can be checked more easily through the UI or by calling get directly.

So, in short, I’m open to adding it in the future if there’s demand, but at the moment, I don’t see a compelling use case that isn’t already handled more simply.

As for including it in the app key resource itself—that would be neat. But since we use a different endpoint, sticking with the current approach (as agreed with Max) keeps things simpler and we can get this out faster.

datadog/fwprovider/resource_datadog_app_key_registration.go
GabeWM: Agreed will make the change.

datadog/tests/resource_datadog_app_key_registration_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/fwprovider"
+)
+
+var testAppKeyRegistrationID = "14852f74-9cc7-4a4c-94c5-0f97cb33f810"
```
GabeWM, line 13: Is this an issue? It is a hard coded app key id.

PR #3066, opened 2025-06-23
[datadog_synthetics_test] Rewrite all `log.Printf` into `diags`

Drarig29:
This PR changes all `log.Printf` calls into diagnostics that are now returned and **actually shown** to the user.

For example:
image

The conventions I came up with:
- Always declare `diags := diag.Diagnostics{}` at the top of the function and use `append()`
- Never use `nil` for `diags`: it should always be initialized
- Use `HasError()` to exit early
- Every single `log.Printf` should be replaced
- Logs which used to be `[ERROR]` are now considered as `diag.Error` diagnostics, which **will exit early**
  - I think it's a good thing because those errors would likely cause backend validation errors anyway, which are harder to understand


AntoineDona (approved):
LGTM, thank you so much for fixing this 🙏

Drarig29:
I'll merge this in 2 weeks after my PTO. This PR is meant to unblock me on https://github.com/DataDog/terraform-provider-datadog/pull/3049, and neither is urgent.

PR #3070, opened 2025-06-24
[datadog_synthetics_test] Support `dns_server_port` as string or number

teodor2312:
Making dns server port accept both string and number

### DIFF
--- a/datadog/tests/resource_datadog_synthetics_test_test.go
+++ b/datadog/tests/resource_datadog_synthetics_test_test.go
@@ -6406,7 +6406,7 @@ func createSyntheticsMultistepAPITestConfigAllStepSubtypes(testName string) stri
 				request_definition {
 					host       = "troisdizaines.com"
 					dns_server = "8.8.8.8"
-					dns_server_port = 53
+					dns_server_port = "53"
 				}
 				retry {
 					count    = 0


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -5236,63 +5259,6 @@ func getParamsKeysForStepType(stepType datadogV1.SyntheticsStepType) []string {
...
-	case datadogV1.SYNTHETICSMOBILESTEPTYPE_TYPETEXT:
-		return []string{"value", "element", "delay", "with_enter"}
-
-	case datadogV1.SYNTHETICSMOBILESTEPTYPE_WAIT:
-		return []string{"value"}
-	}
-
-	return []string{}
-}
-
```
Supam, line 5295: Is this deletion intended ?

teodor2312: Yes, the function is not used 👍

Supam (approved):

datadog/tests/resource_datadog_synthetics_test_test.go
```diff
@@ -6402,6 +6406,7 @@ func createSyntheticsMultistepAPITestConfigAllStepSubtypes(testName string) stri
 				request_definition {
 					host       = "troisdizaines.com"
 					dns_server = "8.8.8.8"
+					dns_server_port = 53
```
Drarig29: Can you make this a string? We already have a numeric `dns_server_port` somewhere else in the file, and what you added is the support for `dns_server_port` passed as a string to support variables

teodor2312: No it's the other way around it's to support number as well, as it already supported string 👍

teodor2312: But fair enough i will change this to string so it has both number and string 👍

PR #3091, opened 2025-07-02
[datadog_synthetics_test] Add extract body email step 

Alexis-Jacob-datadog:
The `extractFromEmailBody` step was not available in the provider for browser test. This PR add this feature. 

Example conf:
```hcl
  browser_step {
    name = "Test email"

    params {
      pattern {
        type  = "regex"
        value = "hello world"
      }

      variable {
        example = "email_body"
        name    = "OTP_FROM_EMAIL"
        secure  = "false"
      }
    }
    type    = "extractFromEmailBody"
  }
```

But keep in mind that you also need to put an browser variable in your test to make it work
```hcl
  browser_variable {
    example = "jd8-afe-ydv.4546132139@synthetics.dtdg.co"
    name    = "MY_EMAIL_VAR"
    pattern = "jd8-afe-ydv.{{ numeric(10) }}@synthetics.dtdg.co"
    secure  = "false"
    type    = "email"
  }
  ```

related to #3072

### CODE REVIEW

File: docs/resources/synthetics_test.md [hunk 1]
```diff
@@ -1151,6 +1152,15 @@ Optional:
 
 
 
+<a id="nestedblock--browser_step--params--pattern"></a>
+### Nested Schema for `browser_step.params.pattern`
+
+Optional:
+
+- `type` (String) Type of the pattern to use for the step.
```
maycmlee:
```suggestion
- `type` (String) Type of pattern to use for the step.
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1399,6 +1400,26 @@ func syntheticsBrowserStepParams() schema.Schema {
...
+					MaxItems:    1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"value": {
+								Description: "Pattern to use for the step.",
+								Type:        schema.TypeString,
+								Optional:    true,
+							},
+							"type": {
+								Description: "Type of pattern to use for the step.",
```
AntoineDona: As this is in fact an enum, I think it would be better to add the possible values. They would be auto generated if we had the explicited in the api spec, but we don't...
```suggestion
								Description: "Type of pattern to use for the step. Valid values are `regex`, `x_path`.",
```

AntoineDona (approved):
LGTM 👍 Just a small nit

PR #3096, opened 2025-07-03
[datadog_downtime_schedule] Validate recurrence start

gmeligio:
## Context

Validation errors for the `recurring_schedule.recurrence.start` variable are being caught only during `apply` time. This causes bad developer experience because the plan works OK but it fails during apply.

## Solution

Add a date validator to the `recurring_schedule.recurrence.start` variable. This will validate the condition stated in the description 
> ISO-8601 Datetime to start the downtime. Must not include a UTC offset.

This will allow to validate the input variable during the `plan` phase and catch errors earlier. A similar validation is already available in `one_time_schedule.start` and `one_time_schedule.end`.


tioresko (approved):
LGTM

MintsInc (approved):

PR #3106, opened 2025-07-08
[datadog_synthetics_test] Add append_to_content option for text input

Alexis-Jacob-datadog:
This PR adds the append_to_content option for text input in browser step.
It was already available in the UI but never in the terraform provider.


joepeeples (approved):

docs/resources/synthetics_test.md
```diff
@@ -1105,6 +1105,7 @@ Read-Only:
 
 Optional:
 
+- `append_to_content` (Boolean) Select this option to append value to existing text input content. By default, content is cleared before text input.
```
Drarig29: **nit** Just aligning with the `click_with_javascript` (Boolean) below, which starts with "Whether". And added a note for what type of step that option applies to.

(For context, the tooltip we show in the UI does say "Select this option to append the value to the existing text input content. By default, content is cleared before text input.")

```suggestion
- `append_to_content` (Boolean) Whether to append the `value` to existing text input content for a "typeText" step. By default, content is cleared before text input.
```

Drarig29 (approved):
LGTM, just a nit on the description

PR #3118, opened 2025-07-15
[synthetics_test] Fix file upload issue in the terraform provider

mtalec:
### Problem description 
When users upload files via the Datadog UI for Synthetics tests, the content of the file is not passed by the API so metadata (like the size) would change on the UI and be sent to terraform which would create a non null plan because of the unchanged content and thus [overriding false values in the UI](https://datadoghq.atlassian.net/browse/SYNTH-20509). 

#### Root cause
API limitation by not returning the file content and inconsistent state creation when uploading a file on the UI that leads to preserve old content but change new API values. The result is phantom diffs causing unwanted update plans

### Solution 
Instead of detecting and suppressing the diffs after they occur with a DiffSupressFunc we prevent inconsistency in the state by modifying the `Read` operation logic. 
#### Implementation Details

1. Added `isReadOperation` parameter to distinguish between Read vs Create/Update operations

1. UI Change Detection Logic in buildTerraformBodyFiles:
- During Read operations: Check if content exists in state but metadata changed from API
- If UI change detected: Preserve config values instead of overwriting with API values
- If config change detected: Update all fields with API values (normal behavior)
- During Create/Update: Always use API values (unchanged behavior)

The `io/ioutil` import was also changed because of it being deprecated

### Testing 

- [ ] primary fix: changing a file in the UI and verify no unwanted diffs
- [ ] terraform config changes being updated on the UI if changed locally
- [ ] synthetics tests passed

### Impact 

- one file changed `datadog/resource_datadog_synthetics_test_.go` 
- no impact on Create/Update 
- no data migration and a rollback plan if needed


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -3856,22 +3859,84 @@ func buildDatadogBodyFiles(attr []interface{}) []datadogV1.SyntheticsTestRequest
...
+		} else {
+			// No previous state (new file or import), use API values
+			localFile["name"] = file.GetName()
+			localFile["original_file_name"] = file.GetOriginalFileName()
+			localFile["type"] = file.GetType()
+			localFile["size"] = file.GetSize()
+
+			if bucket_key, ok := file.GetBucketKeyOk(); ok {
+				localFile["bucket_key"] = bucket_key
+			}
```
teodor2312: These 3 else statements are the same right?
Maybe we can put a `continue` before them and collapse them into one instead 🤔 
Wdyt?

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
```
teodor2312: I guess this was not used? :D

mtalec: I had conflicts because of Corentin's [PR]([[datadog_synthetics_test] Rewrite all log.Printf into diags](https://github.com/DataDog/terraform-provider-datadog/pull/3066)) but I'll just remove the logging that's not mandatory

datadog/resource_datadog_synthetics_test_.go
Drarig29: Agreed. Here is the big picture:

```go
if oldLocalBodyFiles[i] {
    if isReadOperation {
        // some logic
        // some variables

        if someConditionThatUsesAboveLogicAndVariables {
            A
        } else {
            B
        }
    } else {
        B
    }
} else {
    B
}
```

You can probably refactor `// some logic` and `// some variables` into a function to use inside the `if` expression:

```go
if isUiFileChangeDetected(file, oldLocalBodyFiles[i]) {
    A
} else {
    B
}
```

mtalec: thanks for the feedback, I changed that, lmk if it works for you :))

Drarig29 (approved):

PR #3142, opened 2025-07-30
[datadog_security_monitoring_default_rule] Make query optional while editing default_rule using TF.

cpatel24:
[SEC-23413] Make query optional while editing default_rule using TF.
- Many teams manage their TF plans. 
- For `resource_datadog_security_monitoring_default_rule` resource, only handful of fields are allowed to change. Some of them are complext/struct fields. If we want to just change/manage selective fields using TF plans, it was not possible. We had to put the entire block (for eg `query {}`). 
- However if original OOTB rule is changed, users have to change other fields in TF config as well, otherwise TF API fails, thinking users are trying to change unmodified fields. 

For eg:
Earlier if we just want to add `custom_query_extension` in query. Our TF has to have entire query block as 
```
query {
    name = "test name"
    custom_query_extension = "test 2"
    trackType = "logs"
    groupByFields = [
        "@github.actor"
     ]
     hasOptionalGroupByFields = false
     distinctFields= []
      ...
  }
```
However only `custom_query_extension` field is allowed to change by TF API.

In this case, we shouldn't need to provide all fields, instead it should just copy from existing rule. Otherwise everything OOTB rule is changed, users have to udpate their TF configs. Which isn't maintainable.

[SEC-23413]: https://datadoghq.atlassian.net/browse/SEC-23413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing
-------

- Tested in staging to update https://mosaic.us1.staging.dog/k9-admin/rule/history/def-000-ouv-2 using 
```
resource "datadog_security_monitoring_default_rule" "possible_brute_force_attempted_against_user" {
  query {
    name = ""
    custom_query_extension = "test 2"
  }
  case {
    status = "medium"
    custom_status = "high"
  }
}
```

- It successfully updated `custom_query_extension = "test 2"`. 
- It ignored already present/same `case`, so no update were applied.


MatthewMSaucedo (approved):
Looks good to me, with just the general nit of preferring. `any` to `interface{}`

datadog/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -541,17 +605,23 @@ func buildSecMonDefaultRuleUpdatePayload(currentState *datadogV2.SecurityMonitor
...
 
-	// Compare options
+	// Handle options - preserve existing options when not specified in config
+	currentOptions := currentState.GetOptions()
+	var payloadOptions *datadogV2.SecurityMonitoringRuleOptions
+
 	if v, ok := d.GetOk("options"); ok {
+		// Options specified in Terraform config
 		tfOptionsList := v.([]interface{})
-		payloadOptions := buildPayloadOptions(tfOptionsList, d.Get("type").(string))
```
fpighi: non-blocking question: what's the intent behind the refactoring here?

cpatel24: You were correct, it wasn't needed. It was added earlier in attempt to fix a condition, but later didn't need it. 

Cleaned up.

PR #3147, opened 2025-08-01
Add incident type tf resource + data source

codhah92:


docs/resources/incident_type.md
```diff
@@ -0,0 +1,48 @@
...
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) Name of the incident type. Must be between 1 and 50 characters.
+
+### Optional
+
+- `description` (String) Description of the incident type. Must be at most 512 characters.
```
maycmlee: Is this saying that the max number of characters is 512?
```suggestion
- `description` (String) Description of the incident type. Must be at most 512 characters.
```

codhah92: yeah I can update the description

maycmlee (approved):

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -0,0 +1,227 @@
...
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: "Name of the incident type. Must be between 1 and 50 characters.",
+				Required:    true,
+			},
+			"description": schema.StringAttribute{
+				Description: "Description of the incident type. The description can have a maximum of 512 characters.",
+				Optional:    true,
+				Computed:    true,
```
cperryk: What does computed mean in this context?

cperryk (approved):

rakshith91 (approved):

datadog/fwprovider/resource_datadog_incident_type.go
codhah92: good catch - Computed indicates that the attributed is calculated by the backend. Optional + Computed indicates that the field can be set by the user, but the backend calculates the value otherwise.

In the case of description, we read and write description as is provided by the end user, so I've omitted Computed

AxelGeist (approved):

PR #3150, opened 2025-08-04
[datadog_observability_pipeline] Amazon S3 Destination

tessneau:
Fix - Including the Amazon S3 destination which was already within the API Spec

API Implementation was in this commit https://github.com/DataDog/datadog-api-spec/pull/3807/commits/161e7ffc1dcbe79f385afa7763ade4c261f03555

### DIFF
--- a/datadog/fwprovider/resource_datadog_observability_pipeline.go
+++ b/datadog/fwprovider/resource_datadog_observability_pipeline.go
@@ -88,10 +88,10 @@ type kafkaSourceSaslModel struct {
 }
 
 type amazonS3SourceModel struct {
-	Id     types.String  `tfsdk:"id"`     // Unique identifier for the component
-	Region types.String  `tfsdk:"region"` // AWS region where the S3 bucket resides
-	Auth   *awsAuthModel `tfsdk:"auth"`   // AWS authentication credentials
-	Tls    *tlsModel     `tfsdk:"tls"`    // TLS encryption configuration
+	Id     types.String                         `tfsdk:"id"`     // Unique identifier for the component
+	Region types.String                         `tfsdk:"region"` // AWS region where the S3 bucket resides
+	Auth   *observability_pipeline.AwsAuthModel `tfsdk:"auth"`   // AWS authentication credentials
+	Tls    *tlsModel                            `tfsdk:"tls"`    // TLS encryption configuration
 }
 
 type tlsModel struct {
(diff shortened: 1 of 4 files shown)


datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -5004,3 +5076,69 @@ func expandDatadogTagsProcessor(ctx context.Context, src *observability_pipeline
...
+	dest.SetRegion(src.Region.ValueString())
+
+	if !src.KeyPrefix.IsNull() {
+		dest.SetKeyPrefix(src.KeyPrefix.ValueString())
+	}
+
+	dest.SetStorageClass(datadogV2.ObservabilityPipelineAmazonS3DestinationStorageClass(src.StorageClass.ValueString()))
+
+	if src.Auth != nil {
+		auth := datadogV2.ObservabilityPipelineAwsAuth{}
```
graphcareful: I think I had generated something like `flatted/expandAwsAuth` that is more in line with the Tls options which already exist. Would be nice to have because the other aws components will have to pretty much do the exact same thing.

datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -3233,3 +3233,113 @@ resource "datadog_observability_pipeline" "custom_processor" {
...
+    destinations {
+      amazon_s3 {
+        id           = "s3-dest-1"
+        inputs       = ["source-1"]
+        bucket       = "my-logs-bucket"
+        region       = "us-east-1"
+        key_prefix   = "logs/"
+        storage_class = "STANDARD"
+
+        		auth {
```
graphcareful: stray tab maybe?

graphcareful (commented):
One suggestion to implement flatten/expand AwsAuth opts, otherwise LGTM

datadog/tests/resource_datadog_observability_pipeline_test.go
tessneau: nice catch https://github.com/DataDog/terraform-provider-datadog/pull/3150/commits/dd53b54b2427f4aaa50ca3b50475a3f254243e94

datadog/fwprovider/resource_datadog_observability_pipeline.go
tessneau: good idea - refactored in https://github.com/DataDog/terraform-provider-datadog/pull/3150/commits/c53fb2733ae2b73af663d3bd06bc5bba110d2cd4 

stored it in a separate file mimicking [what we have for tls](https://github.com/DataDog/terraform-provider-datadog/blob/c53fb2733ae2b73af663d3bd06bc5bba110d2cd4/datadog/fwprovider/observability_pipeline/tls.go)

PR #3152, opened 2025-08-05
on-board dd-octo-sts

ikraemer-dd:


fpighi (approved):

.github/chainguard/self.release.prepare-release.sts.yaml
```diff
@@ -0,0 +1,13 @@
+issuer: https://token.actions.githubusercontent.com
+
+subject: repo:DataDog/terraform-provider-datadog:ref:refs/heads/master
+
+claim_pattern:
+  event_name: workflow_dispatch
+  job_workflow_ref: DataDog/terraform-provider-datadog/\.github/workflows/prepare_release\.yml@.*
+  ref: refs/heads/master
+  ref_protected: "true"
```
xopham, line 9: I think this workflow always runs on `release/*` branches, so all claims will need to be adjust to that. The team should add branch protection to `release/*` branches to restrict access to those branches

xopham: nope, I am wrong. Was looking at "release" workflow and not "prepare release". Sorry, for the confusion

PR #3114, opened 2025-07-11
Josue.agbekodo/op remap vrl

20agbekodo:

PR #3153, opened 2025-08-05
[datadog_observability_pipeline] Amazon Security Lake Destination

graphcareful:
Adding the ASL destination which was introduced in OPW v2.6.0


datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -3343,3 +3343,63 @@ resource "datadog_observability_pipeline" "amazon_s3_basic" {
 		},
 	})
 }
+
+func TestAccDatadogObservabilityPipeline_AmazonSecurityLakeDestination(t *testing.T) {
```
tessneau, line 3347: Run this test with `RECORD=true` (guidance [here](https://datadoghq.atlassian.net/wiki/spaces/OPE/pages/5234361650/Recommended+Workflow+for+Public+API+and+Terraform+Updates#Test-Execution)), it'll generate some cassettes for ASL which you can commit (it will also update a ton of other existing cassettes with a timestamp, you can ignore those)

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -2253,6 +2264,38 @@ func (r *observabilityPipelineResource) Schema(_ context.Context, _ resource.Sch
 							},
 							"socket":    observability_pipeline.SocketDestinationSchema(),
 							"amazon_s3": observability_pipeline.AmazonS3DestinationSchema(),
+							"amazon_security_lake": schema.ListNestedBlock{
```
tessneau: wdyt about extracting the ASL resources - schema, expand/flatten functions, model struct - into a separate file? we've been doing it for the 2.6.0 ones so far, here's an example https://github.com/DataDog/terraform-provider-datadog/pull/3114/files#diff-27764b2b0cac6a7a1401a7663480e52c45a7f0a3214922d428744a1080a5865b

datadog/tests/resource_datadog_observability_pipeline_test.go
tessneau: I actually forgot to run this in my previous S3 destination PR

tessneau (commented):
you can include the `go.mod` and `go.sum` files that you updated locally - we've been doing that in other PRs

tessneau (approved):

PR #3161, opened 2025-08-07
[datadog_monitor] Migrate datadog_monitor to plugin framework

shaneyuandd:
This is the initial effort to migrate the `datadog_monitor` resource from the legacy **SDK v2** to the new **Terraform Plugin Framework**. We are migrating to enable better dev experience on the monitor resource (see [here](https://developer.hashicorp.com/terraform/plugin/framework-benefits)).

- Adds a runtime switch (`TERRAFORM_MONITOR_FRAMEWORK_PROVIDER`) to opt-in to the new framework.
```
TERRAFORM_MONITOR_FRAMEWORK_PROVIDER=true terraform apply
```
- Implements all existing fields on `datadog_monitor`.
- Implements new field `effective_tags` to represent to combination of `default_tags` and resource `tags`.
- Introduces custom types to normalize API quirks, covering data inconsistency.

## Reference ##
- The field-by-field rollout follows HashiCorp’s [official migration guide](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/fields).
- Migration [benefits](https://developer.hashicorp.com/terraform/plugin/framework-benefits).

### DIFF
--- a/datadog/internal/utils/tags.go
+++ b/datadog/internal/utils/tags.go
@@ -1,15 +1,8 @@
 package utils
 
 import (
-	"context"
-	"fmt"
-	"strings"
 	"unicode"
 	"unicode/utf8"
-
-	"github.com/hashicorp/terraform-plugin-framework/attr"
-	"github.com/hashicorp/terraform-plugin-framework/diag"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
 // NormalizeTag takes a string and parses it in accordance to USM Tagging conventions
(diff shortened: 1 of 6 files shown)


fpighi:
Is the plan to make this change in a fully backwards compatible way, so practitioners will not need to change their resources? Also, what's the goal with releasing it in smaller chunks if it's not usable? Would it be ok instead to have a single PR at the end to avoid polluting the changelog with smaller changes?

shaneyuandd:
> Is the plan to make this change in a fully backwards compatible way, so practitioners will not need to change their resources? Also, what's the goal with releasing it in smaller chunks if it's not usable? Would it be ok instead to have a single PR at the end to avoid polluting the changelog with smaller changes?

Yes, our goal is to allow customer migrate their monitors to the new plugin framework without having to modify their existing resource (best effort). 

We don't have to release in smaller chunks, I am posting this PR to get some early feedback, and see if the team like our migration approaches. My plan is to address any comments on this PR, after that we will use this branch as the main branch for the feature. Future PRs will merge into this branch, before merging into `master`.

datadog/internal/utils/tags.go
```diff
@@ -100,3 +107,41 @@ func isValidASCIIStartChar(c byte) bool {
...
+		for _, v := range vals {
+			tag := fmt.Sprintf("%s:%v", k, v)
+			if v == "" {
+				tag = k
+			}
+			resultTags = append(resultTags, tag)
+		}
+	}
+	return types.SetValueFrom(ctx, types.StringType, resultTags)
+}
```
Supam: Could this live in `datadog/internal/fwutils/tags.go` ?
`internal/utils` is where utils for the skdv2 live, if this is to be used in the fwprovider, we would rather it exists in `internal/fwutils`

datadog/internal/utils/types_utils.go
```diff
@@ -0,0 +1,105 @@
+package utils
```
Supam: same comment as for the tags.go utils

Supam (approved):
Looks good !

datadog/internal/utils/tags.go
shaneyuandd: Makes sense, will move these files. Thanks

PR #3164, opened 2025-08-11
[datadog_agentless_scanning_aws_scan_options] Add Terraform provider for AWS scan options

k3nz0:
This pull request introduces a new Terraform resource for managing Datadog Agentless Scanning AWS Scan Options, enabling users to configure scan options for AWS accounts via the provider. 

The changes include the implementation of the resource, supporting CRUD and import operations 

**New Resource Implementation:**

* Added a new resource `datadog_agentless_scanning_aws_scan_options` with support for create, read, update, delete, and import operations, allowing configuration of AWS scan options such as Lambda, sensitive data, and vulnerability scanning for containers and hosts.
* Registered the new resource in the provider's resource list (`fwprovider/framework_provider.go`).

**Testing:**

* Added tests for the new resource, covering creation, update, import, and destruction scenarios.


joepeeples (approved):

fpighi (changes requested):
Please add a line to [CODEOWNERS.md](https://github.com/DataDog/terraform-config/blob/master/.github/CODEOWNERS) with the correct team, otherwise LGTM!

datadog/fwprovider/resource_datadog_agentless_scanning_aws_scan_options.go
```diff
@@ -0,0 +1,225 @@
...
+	var state agentlessScanningAwsScanOptionsResourceModel
+	response.Diagnostics.Append(request.State.Get(ctx, &state)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	accountID := state.ID.ValueString()
+
+	// List all AWS scan options and find the one matching our account ID
+	awsScanOptionsListResponse, _, err := r.Api.ListAwsScanOptions(r.Auth)
```
diogocp, line 127: We need to update the client to support get by ID.

examples/resources/datadog_agentless_scanning_aws_scan_options/resource.tf
```diff
@@ -0,0 +1,8 @@
+# Configure agentless scanning for an AWS account
+resource "datadog_agentless_scanning_aws_scan_options" "example" {
+  id                 = "123456789012" # AWS Account ID
```
diogocp: The presence of the comment suggests that this should be called `account_id` or `aws_account_id` instead.

datadog/fwprovider/resource_datadog_agentless_scanning_aws_scan_options.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		errorMsg := "Error updating AWS scan options"
+		if res != nil {
+			errorMsg += fmt.Sprintf(". API response: %s", res.Body)
+		}
+		response.Diagnostics.AddError(errorMsg, err.Error())
+		return
+	}
+
+	// After update, we need to read the current state since the API doesn't return the updated object
```
diogocp, line 186: We should change this in the API.

diogocp (commented):
LGTM.

It's missing the `Dd-Call-Source` header but I didn't see an obvious way to add it. It might need a change to the client.

The only important change would be to the `id` argument. The rest we can always change later in a backward compatible way.

datadog/fwprovider/resource_datadog_agentless_scanning_aws_scan_options.go
k3nz0: Yep, especially that we already have a route for that one: https://github.com/DataDog/dd-source/blob/f370085d0c3fa979a2e1c004b5c77f86ca3e4d1c/domains/cloud-security-platform/agentless/apps/apis/agentless-api/handler/handlers.go#L183

k3nz0: Created a card in the backlog to handle this: https://datadoghq.atlassian.net/browse/K9VULN-7622

examples/resources/datadog_agentless_scanning_aws_scan_options/resource.tf
k3nz0: Fixed. I just introduced `aws_account_id` as you requested

datadog/fwprovider/resource_datadog_agentless_scanning_aws_scan_options.go
k3nz0: Yes the patch request could return the response in the body (but it's returning 204 today).

PR #3183, opened 2025-08-21
Add remaining monitor parameters to migrated provider

shaneyuandd:
This PR merges into branch (`shaneyuandd/monitor-migration`), which follows https://github.com/DataDog/terraform-provider-datadog/pull/3161. Adds remaining fields to the monitor resource provider (Plugin Framework). 

## TODO ##
- Address Float number difference in `monitor_thresholds`
- Implement `default_tags`
- Add all remaining acceptance tests


datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -487,18 +1340,98 @@ func (r *monitorResource) parseFloat(v types.String) *float64 {
...
+			return types.StringNull()
+		}
+		return r.toTerraformStr(t.Get(), true)
+	case *string:
+		return types.StringValue(*t)
 	}
 	return types.StringNull()
 }
+
+func (r *monitorResource) toTerraformBool(v *bool, ok bool) types.Bool {
```
kevinpombo-datadog: Is this type parsing here and below done per tf integration class? It reads like it could be extracted into a shared class for other types

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -64,6 +177,52 @@ func NewMonitorResource() resource.Resource {
 	return &monitorResource{}
 }
 
+func (r *monitorResource) ConfigValidators(ctx context.Context) []resource.ConfigValidator {
+	return []resource.ConfigValidator{
+		resourcevalidator.Conflicting(
+			frameworkPath.MatchRoot("notify_no_data"),
+			frameworkPath.MatchRoot("on_missing_data"),
```
jasminezh1, line 182: i think there exists some monitors with on_missing_data and notify_no_data=false

shaneyuandd: That is correct, and the only allowed value is `false` in this case. The default value for `notify_no_data ` in terraform is already set to `false`, so there's no need for client to explicitly set this value.

PR #3194, opened 2025-08-29
[resource_datadog_synthetics_test_] Make port in synthetic tests support both int and string

HantingZhang2:


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2782,10 +2782,16 @@ func buildDatadogSyntheticsAPITest(d *schema.ResourceData) (*datadogV1.Synthetic
 		request.SetHost(attr.(string))
 	}
 	if attr, ok := d.GetOk("request_definition.0.port"); ok {
-		port := attr.(string)
-		request.SetPort(datadogV1.SyntheticsTestRequestPort{
-			SyntheticsTestRequestVariablePort: &port,
-		})
+		if val, ok := attr.(int64); ok {
```
thve: nit: We have a lot of duplicates like this: check int or str, then store it. Could we extract a helper to avoid these duplicates, or would that be complicated / not aligned with the code style?

datadog/tests/resource_datadog_synthetics_test_test.go
```diff
@@ -6414,7 +6414,7 @@ func createSyntheticsMultistepAPITestConfigAllStepSubtypes(testName string) stri
 
 				request_definition {
 					host                         = "example.org"
-					port                         = 443
+					port                         = "443"
```
thve, line 6417: Nit: ideally we should check that we accept both str and int format to avoid regression on that part

thve (approved):

datadog/tests/resource_datadog_synthetics_test_test.go
HantingZhang2: The port in the test of other subtypes are ints, now that we are using the same helper one string port is probably enough

fpighi (approved):

PR #3196, opened 2025-08-31
[datadog_role] Add permissions attribute to datadog_role data source

orkhanM:
This PR extends the `datadog_role` data source to include a `permissions` attribute that returns all permissions associated with a role. The permissions are returned as a map (name → ID), consistent with the `datadog_permissions` data source format. This enhancement enables users to programmatically inherit permissions from existing roles when creating custom roles, rather than manually maintaining permission lists.

If preferable this can be made into a different datasource all_together (i.e. `datadog_role_permissions`)

Example:

```hcl
# Get all available permissions to look up IDs by name
data "datadog_permissions" "all" {}

# Get the existing "Datadog Read Only" role and its permissions
data "datadog_role" "readonly" {
  filter = "Datadog Read Only"
}

# Create a custom role that inherits all read permissions plus adds write access for notebooks and dashboards
resource "datadog_role" "custom_analyst" {
  name        = "Custom Analyst"
  description = "Read-only access plus dashboard and notebook write permissions"
  
  default_permissions_opt_out = true

  dynamic "permission" {
    for_each = data.datadog_role.readonly.permissions
    content {
      id = permission.value
    }
  }

  permission {
    id = data.datadog_permissions.all.permissions["dashboards_write"]
  }
  
  permission {
    id = data.datadog_permissions.all.permissions["notebooks_write"]
  }
}
```

Without this feature, users have to manually enumerate every single permission ID from the base role, which is error-prone and difficult to maintain as permissions change over time.

### DIFF
--- a/datadog/data_source_datadog_role.go
+++ b/datadog/data_source_datadog_role.go
@@ -35,6 +35,14 @@ func dataSourceDatadogRole() *schema.Resource {
 					Type:        schema.TypeInt,
 					Computed:    true,
 				},
+				"permissions": {
+					Description: "Map of permissions granted to this role, keyed by permission name and returning the permission ID.",
+					Type:        schema.TypeMap,
+					Computed:    true,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
 			}
 		},
 	}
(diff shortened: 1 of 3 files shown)


joepeeples (approved):

Marcel2603:
Hi,

what is the current state of this PR? I would be actually very handy to have the list of permissions on every role ds. My current workaround is to query the api directly, which is okayisch ... its okay but using data would be so much better

```terraform
data "http" "read_only_permissions" {
  url = "${local.dd_api_url}/api/v2/roles/${data.datadog_role.read_only.id}/permissions"
  request_headers = {
    Accept             = "application/json"
    DD-API-Key         = local.dd_api_key
    DD-APPLICATION-Key = local.dd_app_key
  }
  lifecycle {
    postcondition {
      condition     = 200 == self.status_code
      error_message = "Request failed"
    }
  }
}
```

orkhanM:
@Marcel2603 I'm not sure as I haven't tested since I submitted the PR. It should still work in theory but until it's merged or an equivalent functionality is added to the provider i've been using a workaround similar to yours

PR #3203, opened 2025-09-04
[datadog_security_monitoring_suppression] Add Suppressions Validation during TF Plan

clementgbcn:
## What does this PR do?

Add plan-time validation for datadog_security_monitoring_suppression resource and support for partial updates of suppressions with past dates.

## Motivation

Following the update to datadog-api-client-go that introduced the ValidateSecurityMonitoringSuppression API endpoint, this enhancement enables early detection of configuration errors during terraform plan phase instead of waiting for apply. This provides faster feedback to users and improves the overall developer experience when working with security monitoring suppressions.

Additionally, this enables partial updates of suppressions that have past start dates or expiration dates, which was previously blocked.

## Key Changes

- Added `validate` field to resource schema (defaults to `true`) to control validation behavior
- Implemented `ModifyPlan` method that calls the new ValidateSecurityMonitoringSuppression API endpoint during terraform plan
- Updated datadog-api-client-go dependency to version containing the validation endpoint  
- Fixed data source model separation to prevent struct/schema conflicts between resource and data source
- Added comprehensive test coverage including create/update scenarios with validation enabled
- Maintained full backward compatibility - users can set `validate = false` to skip validation if needed
- Enhanced support for partial updates allowing modification of suppressions with past dates

## Testing

- Added new test case `TestAccSecurityMonitoringSuppression_ValidationEnabled` that verifies validation during plan phase
- Updated existing test cassettes to reflect API changes and validation calls
- All existing acceptance tests pass ensuring backward compatibility
- Test coverage includes scenarios with validation enabled/disabled and partial updates

## Impact

This change improves user experience by catching invalid suppression configurations early in the terraform workflow while maintaining full backward compatibility. Users working with security monitoring suppressions will receive immediate feedback on configuration issues during plan phase rather than discovering them during apply.

The partial update support also resolves limitations around modifying suppressions that have dates in the past, making the resource more flexible for ongoing management.

PR #3204, opened 2025-09-04
Keep Synthetics multistep test steps id in state

romainberger:
Synthetics multistep tests steps have a read-only ID, but currently the provider is not saving them in the state and not sending them when editing tests. This results in the backend re-creating the steps completely when editing, which clutters the version history in the Datadog UI. 
This PR fixes that, saving the IDs in the local state and sending them on edit.


thve (approved):

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/terraform-providers/terraform-provider-datadog
 
 require (
-	github.com/DataDog/datadog-api-client-go/v2 v2.44.1-0.20250825120552-2ee6e2d3e4ed
+	github.com/DataDog/datadog-api-client-go/v2 v2.44.1-0.20250904122701-3794619b8e40
```
Supam, line 4: The go-client was updated earlier today by #3203.
Please make sure that the version you are trying to use is as or more recent then that one.

PR #3206, opened 2025-09-04
CLOUDS-6606: Update GCP STS metric_namespace_configs validation

dtru-ddog:
TBD


dtru-ddog:
Discussed with @tedkahwaji offline, we'll instead go with updating defaults here as a better way to get around this issue due to POST API creation defaults.

PR #3210, opened 2025-09-05
feat: add write-only argument helpers

LiuVII:
[APIR-2172](https://datadoghq.atlassian.net/browse/APIR-2172)

[APIR-2172]: https://datadoghq.atlassian.net/browse/APIR-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary
Add reusable helpers for implementing write-only secret attributes in Terraform Provider Framework resources.

## Changes
- `WriteOnlySecretConfig` - Configuration struct for dual-mode secret support (plaintext + write-only)
- `CreateWriteOnlySecretAttributes()` - Generates schema attributes with `ExactlyOneOf` validators and write-only mode
- `WriteOnlySecretHandler` - Manages secret retrieval with two update patterns:
- **Partial updates** (`SecretRequiredOnUpdate=false`): Only sends secret when version trigger changes
- **Full updates** (`SecretRequiredOnUpdate=true`): Always includes secret in API calls
- `SecretResult` - Return type with `Value`, `ShouldSetValue`, and `Diagnostics`

## Test Plan
  - [x] Table-driven tests for `GetSecretForCreate` and `GetSecretForUpdate`
  - [x] Tests cover both `SecretRequiredOnUpdate` modes
  - [x] Tests verify version trigger change detection

### DIFF
--- a/datadog/internal/fwutils/writeonly_helpers.go
+++ b/datadog/internal/fwutils/writeonly_helpers.go
@@ -1,0 +1,209 @@
+package fwutils
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	frameworkPath "github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// MergeAttributes combines multiple attribute maps into a single map.
+// Later maps take precedence over earlier ones for duplicate keys.
+func MergeAttributes(attributeMaps ...map[string]schema.Attribute) map[string]schema.Attribute {
+	result := make(map[string]schema.Attribute)
+	for _, attrs := range attributeMaps {
+		for key, attr := range attrs {
+			result[key] = attr
+		}
+	}
+	return result
+}
+
+// WriteOnlySecretConfig configures a secret attribute that supports both modes:
+// - Plaintext mode: for Terraform <1.11 or users preferring state storage
+// - Write-only mode: for Terraform 1.11+ with secrets not stored in state
+type WriteOnlySecretConfig struct {
+	OriginalAttr         string // Plaintext attribute (e.g., "secret_key")
+	WriteOnlyAttr        string // Write-only attribute (e.g., "secret_key_wo")
+	TriggerAttr          string // Version trigger (e.g., "secret_key_wo_version")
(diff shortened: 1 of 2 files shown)


datadog/internal/fwutils/writeonly_helpers_test.go
```diff
@@ -0,0 +1,561 @@
...
+		OriginalAttr:  "api_key",
+		WriteOnlyAttr: "api_key_wo",
+		TriggerAttr:   "api_key_wo_version",
+	}
+	handler := WriteOnlySecretHandler{
+		Config:                 config,
+		SecretRequiredOnUpdate: true,
+	}
+
+	testSchema := schema.Schema{
```
fpighi: This test schema, among other things, seems identical across all tests. I suggest extracting the common stuff to make tests a bit more easier to digest.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.37.0
 	github.com/hashicorp/terraform-plugin-testing v1.12.0
 	github.com/jonboulle/clockwork v0.2.2
+	github.com/stretchr/testify v1.8.3
```
fpighi, line 21: We should minimize the number of dependencies unless strictly needed. In this case it's mostly helpers functions, so I believe we can write good tests with just the standard library.

datadog/internal/fwutils/writeonly_helpers.go
```diff
@@ -0,0 +1,193 @@
...
+type SecretResult struct {
+	// Value contains the secret value fetched from the write-only attribute in config.
+	// It is only set when SendWriteOnlySecret is true.
+	Value string
+
+	// SendWriteOnlySecret indicates whether the caller should send the write-only secret to the API.
+	//
+	// Important: when this is false, it does NOT necessarily mean "use plaintext".
+	// It can also mean "write-only secret exists in config, but version trigger didn't change,
+	// so omit the secret for a partial update" (Pattern 1 behavior).
```
fpighi: Is it maybe worth thinking about a more explicit interface here, to control this 
better? For example adding a `SecretResultMode` field could be a way for the resource to have more explicit conditionals.

If I understand this correctly, this would help reducing the risk of currently letting a caller resource code accidentally send plaintext when write‑only was configured but version didn’t change, or when the version is unknown because computed. [Ref (line 442, sorry deep links to line in diffs don't seem to work right now for me)](https://github.com/DataDog/terraform-provider-datadog/pull/3382/changes#diff-8debe959cef44fd2098a8a02e2c6404ad6c81d0feabb4ec67a404435395d2ba7R442).

LiuVII: I've updated to return the value and if it should be set directly from helpers, but I don't think we can help the line 442 mentioned, as it's in updateState, and the helper for this we can do but it's trivial and like very generic, not sure how `Mode` would help here.

fpighi (approved):

PR #3215, opened 2025-09-09
[datadog_csm_threats] Update description to use Workload Protection

QuentinGuillard:
Update TF provider description to use `Workload Protection` product name instead of the old name `CSM Threats`


Supam (approved):
LGTM

PR #3220, opened 2025-09-10
[datadog_metric_tags] fix: data source incorrect metric name

bharling:
This fixes an error using the datadog_metric_tags data source where we incorrectly call `state.Metric.String()` which is intended for human-readable uses, and returns the quoted metric name in a string.

Instead use `.ValueString` which returns a regular string of the metric name.


datadog/fwprovider/data_source_datadog_metric_tags.go
```diff
@@ -69,7 +69,7 @@ func (d *datadogMetricTagsDataSource) Read(ctx context.Context, req datasource.R
 		return
 	}
 
-	ddResp, _, err := d.Api.ListTagsByMetricName(d.Auth, state.Metric.String())
+	ddResp, _, err := d.Api.ListTagsByMetricName(d.Auth, state.Metric.ValueString())
```
mdb, line 72: I believe you may need to generate fresh cassettes in order for the acceptance tests to pass, no? The current cassettes include a `%22`-encapsulated metrics query.

mdb: Thanks @bharling - I see this is now done!

fpighi (approved):
Thanks for the fix!

Issue #2916, opened 2025-03-17
Add support for ephemeral resources


jon-signal:
### What resources or data sources are affected?

- datadog_api_key
- datadog_application_key
- datadog_service_account_application_key

### Feature Request

Terraform recently introduced [ephemeral resources](https://developer.hashicorp.com/terraform/language/resources/ephemeral). In the introduction, they explain:

> Managing infrastructure often requires creating and handling sensitive values that you may not want Terraform to persist outside of the current operation. Terraform provides two tools for resources to manage data you do not want to store in state or plan files: the ephemeral resource block and ephemeral write-only arguments on specific resources.

It would be great to make the various API/application key resources available as ephemeral resources so operators can create or read API keys without storing them in Terraform state (i.e. to add them to vaults/secret management systems).

fpighi:
Thanks for your request @jon-signal.
It's something we want to explore, so we'll keep you posted.

PR #3227, opened 2025-09-12
feat: add framework for ephemeral resources

LiuVII:
This is the initial PR to start implementation of ephemeral resources requested by [Github Issue](https://github.com/DataDog/terraform-provider-datadog/issues/2916).

**Add core infrastructure for ephemeral resources in Terraform Plugin Framework**
- Framework wrapper with schema enrichment
- Private data utilities for state management between Open/Renew/Close
- Test coverage for wrapper functionality

This enables secure, stateless access to sensitive resources without
storing secrets in Terraform state files.

[APIR-2185]: https://datadoghq.atlassian.net/browse/APIR-2185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/datadog/fwprovider/framework_ephemeral_resource_wrapper.go
+++ b/datadog/fwprovider/framework_ephemeral_resource_wrapper.go
@@ -1,0 +1,104 @@
+package fwprovider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral"
+
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/fwutils"
+)
+
+// Interface assertions for FrameworkEphemeralResourceWrapper
+var (
+	_ ephemeral.EphemeralResource                     = &FrameworkEphemeralResourceWrapper{}
+	_ ephemeral.EphemeralResourceWithConfigure        = &FrameworkEphemeralResourceWrapper{}
+	_ ephemeral.EphemeralResourceWithValidateConfig   = &FrameworkEphemeralResourceWrapper{}
+	_ ephemeral.EphemeralResourceWithConfigValidators = &FrameworkEphemeralResourceWrapper{}
+	_ ephemeral.EphemeralResourceWithRenew            = &FrameworkEphemeralResourceWrapper{}
+	_ ephemeral.EphemeralResourceWithClose            = &FrameworkEphemeralResourceWrapper{}
+)
+
+// NewFrameworkEphemeralResourceWrapper creates a new ephemeral resource wrapper following
+// the same pattern as the existing FrameworkResourceWrapper
+func NewFrameworkEphemeralResourceWrapper(i *ephemeral.EphemeralResource) ephemeral.EphemeralResource {
+	return &FrameworkEphemeralResourceWrapper{
+		innerResource: i,
+	}
+}
+
+// FrameworkEphemeralResourceWrapper wraps ephemeral resources to provide consistent behavior
+// across all ephemeral resources, following the existing FrameworkResourceWrapper pattern
+type FrameworkEphemeralResourceWrapper struct {
+	innerResource *ephemeral.EphemeralResource
+}
+
+// Metadata implements the core ephemeral.EphemeralResource interface
+// Adds provider type name prefix to the resource type name, following existing pattern
+func (r *FrameworkEphemeralResourceWrapper) Metadata(ctx context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	(*r.innerResource).Metadata(ctx, req, resp)
+	resp.TypeName = req.ProviderTypeName + resp.TypeName
+}
+
+// Schema implements the core ephemeral.EphemeralResource interface
+// Enriches schema with common framework patterns
+func (r *FrameworkEphemeralResourceWrapper) Schema(ctx context.Context, req ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	(*r.innerResource).Schema(ctx, req, resp)
+	fwutils.EnrichFrameworkEphemeralResourceSchema(&resp.Schema)
+}
+
+// Open implements the core ephemeral.EphemeralResource interface
+// This is where ephemeral resources create/acquire their temporary resources
+func (r *FrameworkEphemeralResourceWrapper) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
+	(*r.innerResource).Open(ctx, req, resp)
+}
+
+// Configure implements the optional ephemeral.EphemeralResourceWithConfigure interface
+// Uses interface detection to only call if the inner resource supports configuration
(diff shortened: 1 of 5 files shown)


datadog/docs/ephemeral-resources/ephemeral_security_guidelines.md
LiuVII, line 1: **ℹ️ info:** remove this file from PR, this is useful mostly as development checklist for LLMs

docs/data-sources/api_key.md
```diff
@@ -3,12 +3,12 @@
 page_title: "datadog_api_key Data Source - terraform-provider-datadog"
 subcategory: ""
 description: |-
-  Use this data source to retrieve information about an existing api key. Deprecated. This will be removed in a future release with prior notice. Securely store your API keys using a secret management system or use the datadog_api_key resource to manage API keys in your Datadog account.
+  Use this data source to retrieve information about an existing API key. Deprecated: This will be removed in a future release with prior notice. For secure access to API key values without storing them in Terraform state, use the ephemeral datadog_api_key resource instead. See the ephemeral resource documentation for examples of secure API key access patterns.
```
domalessi:
```suggestion
  Use this data source to retrieve information about an existing API key. Deprecated: This data source will be removed in a future release with prior notice. To access API key values without storing them in Terraform state, use the ephemeral datadog_api_key resource. See the ephemeral resource documentation for examples of secure API key access patterns.
```

docs/data-sources/api_key.md
```diff
@@ -3,12 +3,12 @@
...
 ---
 
 # datadog_api_key (Data Source)
 
-Use this data source to retrieve information about an existing api key. Deprecated. This will be removed in a future release with prior notice. Securely store your API keys using a secret management system or use the datadog_api_key resource to manage API keys in your Datadog account.
+Use this data source to retrieve information about an existing API key. **Deprecated**: This will be removed in a future release with prior notice. For secure access to API key values without storing them in Terraform state, use the ephemeral `datadog_api_key` resource instead. See the ephemeral resource documentation for examples of secure API key access patterns.
```
domalessi:
```suggestion
Use this data source to retrieve information about an existing API key. **Deprecated**: This data source will be removed in a future release with prior notice. To access API key values without storing them in Terraform state, use the ephemeral `datadog_api_key` resource. See the ephemeral resource documentation for examples of secure API key access patterns.
```

docs/data-sources/api_key.md
```diff
@@ -29,5 +29,5 @@ data "datadog_api_key" "foo" {
 
 ### Read-Only
 
-- `key` (String, Sensitive) The value of the API Key.
+- `key` (String, Sensitive) The value of the API Key. **Security Note**: This field exposes sensitive data in Terraform state. For secure access without state storage, use the ephemeral `datadog_api_key` resource instead.
```
domalessi:
```suggestion
- `key` (String, Sensitive) The value of the API Key. **Security note**: This field exposes sensitive data in Terraform state. For secure access without state storage, use the ephemeral `datadog_api_key` resource.
```

docs/index.md
```diff
@@ -64,6 +64,7 @@ provider "datadog" {
 - `http_client_retry_max_retries` (Number) The HTTP request maximum retry number. Defaults to 3.
 - `http_client_retry_timeout` (Number) The HTTP request retry timeout period. Defaults to 60 seconds.
 - `org_uuid` (String) The organization UUID; used for cloud-provider-based authentication. See the [Datadog API documentation](https://docs.datadoghq.com/api/v1/organizations/) for more information.
+- `store_sensitive_state` (String) Whether to expose API key values in Terraform state. Valid values are [`true`, `false`]. Defaults to `true` for backwards compatibility. When false, API key resources will not include the key value, requiring the use of ephemeral datadog_api_key resources instead.
```
domalessi:
```suggestion
- `store_sensitive_state` (String) Whether to expose API key values in Terraform state. Valid values are [`true`, `false`]. Defaults to `true` for backwards compatibility. When `false`, API key resources do not include the key value; use the ephemeral `datadog_api_key` resource instead.
```

datadog/fwprovider/ephemeral_resource_datadog_api_key.go
```diff
@@ -0,0 +1,137 @@
...
+func (r *EphemeralAPIKeyResource) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
+	// 1. Extract API key ID from config
+	var config EphemeralAPIKeyModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// 2. Fetch API key from Datadog API
+	apiKey, httpResp, err := r.Api.GetAPIKey(r.Auth, config.ID.ValueString())
```
tyjet: question (blocking) - Is there a way to make this configurable such that someone could retrieve their API key from their own Secret Manager instead of relying on the Datadog API?

LiuVII: @tyjet do you mean using some external (non-Datadog) API/provider?
If so, yes, if someone wants to configure their infra to use a secret manager instead of using this ephemeral resource they 💯 can do that, most known secret managers have terraform providers and resources devs can use.

That would mean, that devs won't be using this resource, as this is specifically to make a state-secure way to fetch this data via Datadog API.

I believe this leads to an important question: do we want to allow a way to get API key via our provider or do we want to just cut this API path from our provider entirely and force devs to use external secret managers?
And if we believe restricting this in our Terraform provider (despite still having API support because of reasons) is "the way" to lead devs to better/proper patterns then we don't need an ephemeral resource for `datadog_api_key` at all.

tyjet: Thanks for the explanation, I _think_ I understand much better now.

> do we want to allow a way to get API key via our provider or do we want to just cut this API path from our provider entirely and force devs to use external secret managers?
^^ -- At this point, yes, we want to allow API key access via the provider (and public API). Restricting the terraform provider would be too onerous.

Frogvall:
Hi,

As someone that has been waiting for this for quite some time. What is happening with this pull request? Was it abandoned due to one person not reviewing it? From the outside, that is what it looks like.

fpighi:
> Hi,
> 
> As someone that has been waiting for this for quite some time. What is happening with this pull request? Was it abandoned due to one person not reviewing it? From the outside, that is what it looks like.

Hello @Frogvall ,
It's a problematic feature for us to maintain long term due to some underlying changes to application keys, so we are still discussing but we are leaning towards closing the PR without merging it.
To elaborate a bit more, we could implement an ephemeral resource for API Keys (this PR), but not for App Keys. This is because of a planned change to remove the ability to retrieve application keys after creation. This means ephemeral resources are not a good model that works for both cases, since they wouldn't be working with App Keys.
We are thinking of possible other solutions to this problem, but have no timeline at the moment.

PR #3234, opened 2025-09-18
[resource_datadog_synthetics_test] Fix typo in browser test steps press key modifiers validation

zach-b:
This PR fixes a typo in the Browser Test Step modifier: the value should me "Meta" and not "meta".

"meta" is rejected by the corresponding API endpoints with the following error:

```
{"errors":["Invalid steps data: Step 16 has invalid params: 'meta' is not one of ['Alt', 'Control', 'Meta', 'Shift']"]}
```


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1411,7 +1411,7 @@ func syntheticsBrowserStepParams() schema.Schema {
 					Optional:    true,
 					Elem: &schema.Schema{
 						Type:         schema.TypeString,
-						ValidateFunc: validation.StringInSlice([]string{"Alt", "Control", "meta", "Shift"}, false),
+						ValidateFunc: validation.StringInSlice([]string{"Alt", "Control", "Meta", "Shift"}, false),
```
lesaint, line 1414: I'm not familiar with this code base, but shouldn't there be a test updated/added?

romainberger (approved):

datadog/resource_datadog_synthetics_test_.go
zach-b: Good point, included modifiers in one of our test cases to validate "Meta" is now accepted.

armcburney (approved):

datadog/resource_datadog_synthetics_test_.go
fpighi, line 1414: In your PR description you say:

>This PR fixes a typo in the Browser Test Step modifier: the value should me "meta" and not "Meta".

But here you do the opposite. Am I not reading this correctly?

zach-b: That's right, this was a mistake in the description which is now updated.

fpighi (approved):

PR #3236, opened 2025-09-18
[datadog_tag_pipeline_ruleset] terraform support for tag-pipeline

congx4:
This PR adds support for DataDog Tag Pipeline Ruleset resource and data source to the Terraform provider, enabling users to manage tag pipeline rules for cost management through terraform: 

  What's Included

  New Resources & Data Sources

  - Resource: 
    - datadog_tag_pipeline_ruleset - Manages tag pipeline rulesets
    - resource_datadog_tag_pipeline_ruleset_order - Manages tag pipeline rulesets ordering
  - Data Source: datadog_tag_pipeline_ruleset - Retrieves existing tag pipeline rulesets

  Supported Rule Types

  - Mapping Rules
  - Query Rules
  - Reference Table Rules

Covered test scenarios: 

1. Create new ruleset with 3 different rule types.
2. Update a ruleset
  - Update a field from a ruleset.
  - Add new rules under a ruleset.
  - Reorder rules within a ruleset.
  - Delete rules
3. Delete a ruleset
4. Import existing rulesets
  - Import all 35 rulesets in Frog datadog env.
5. Reorder rulesets

An example test with a HCL file: 
```
resource "datadog_tag_pipeline_ruleset" "my_test_ruleset" {
  name    = "terraform-test-ruleset"
  enabled = true

  rules {
    name    = "environment-standardization"
    enabled = true
    
    mapping {
      destination_key = "env"
      source_keys     = ["environment", "environ", "stage"]
    }
  }

  rules {
    name    = "service-team-tagging"
    enabled = true
    
    query {
      query             = "service:web-api OR service:backend-api"
      case_insensitivity = true  
      if_not_exists     = false
      
      addition {
        key   = "team"
        value = "platform-team"
      }
    }
  }

  rules {
    name    = "service-owner-lookup"
    enabled = true
    
    reference_table {
      table_name  = "service_ownership"
      source_keys = ["service"]
      
      field_pairs {
        input_column = "service"
        output_key   = "owner"
      }
    }
  }
}
```
It can create rules in datadog frog site. 
https://frog.datadoghq.com/account/login?next=%2F


go.mod
```diff
@@ -103,3 +103,4 @@ require (
 )
 
 go 1.23.0
+replace github.com/DataDog/datadog-api-client-go/v2 v2.46.0 => ../datadog-api-client-go
```
congx4: Use the local client version for local development, this will be removed once this the new go client is released.

docs/data-sources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+<a id="nestedblock--rules--reference_table"></a>
+### Nested Schema for `rules.reference_table`
+
+Read-Only:
+
+- `case_insensitivity` (Boolean) Whether the reference table lookup is case insensitive.
+- `field_pairs` (Block List) The field pairs for the reference table. (see [below for nested schema](#nestedblock--rules--reference_table--field_pairs))
```
rtrieu, line 78:
```suggestion
- `field_pairs` (Block List) The field pairs for the reference table. See [below for nested schema](#nestedblock--rules--reference_table--field_pairs).
```

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,102 @@
...
+### Nested Schema for `rules`
+
+Required:
+
+- `enabled` (Boolean) Whether the rule is enabled.
+- `name` (String) The name of the rule.
+
+Optional:
+
+- `mapping` (Block, Optional) The mapping configuration for the rule. (see [below for nested schema](#nestedblock--rules--mapping))
```
rtrieu, line 99:
```suggestion
- `mapping` (Block, Optional) The mapping configuration for the rule. See [below for nested schema](#nestedblock--rules--mapping).
```

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,102 @@
...
+- `query` (Block, Optional) The query configuration for the rule. (see [below for nested schema](#nestedblock--rules--query))
```
rtrieu, line 100:
```suggestion
- `query` (Block, Optional) The query configuration for the rule. See [below for nested schema](#nestedblock--rules--query).
```

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,102 @@
...
+- `reference_table` (Block, Optional) The reference table configuration for the rule. (see [below for nested schema](#nestedblock--rules--reference_table))
```
rtrieu, line 101:
```suggestion
- `reference_table` (Block, Optional) The reference table configuration for the rule. See [below for nested schema](#nestedblock--rules--reference_table).
```

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,102 @@
...
+
+Required:
+
+- `source_keys` (List of String) The source keys for the reference table lookup.
+- `table_name` (String) The name of the reference table.
+
+Optional:
+
+- `case_insensitivity` (Boolean) Whether the reference table lookup is case insensitive.
+- `field_pairs` (Block List) The field pairs for the reference table. (see [below for nested schema](#nestedblock--rules--reference_table--field_pairs))
```
rtrieu, line 143:
```suggestion
- `field_pairs` (Block List) The field pairs for the reference table. See [below for nested schema](#nestedblock--rules--reference_table--field_pairs).
```

docs/data-sources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,87 @@
...
+### Required
+
+- `id` (String) The ID of the ruleset.
+
+### Read-Only
+
+- `enabled` (Boolean) Whether the ruleset is enabled.
+- `name` (String) The name of the ruleset.
+- `position` (Number) The position of the ruleset in the pipeline.
+- `rules` (Block List) The rules in the ruleset. (see [below for nested schema](#nestedblock--rules))
```
rtrieu, line 27:
```suggestion
- `rules` (Block List) The rules in the ruleset. See [below for nested schema](#nestedblock--rules).
```

rtrieu (approved):
Minor suggestions for consistency, but otherwise looks good!

Supam (approved):

docs/data-sources/tag_pipeline_ruleset.md
congx4: I applied your suggestion, but it failed the CI job, I just keep what it is: 

```
 - `case_insensitivity` (Boolean) Whether the reference table lookup is case insensitive.
-- `field_pairs` (Block List) The field pairs for the reference table. See [below for nested schema](#nestedblock--rules--reference_table--field_pairs).
+- `field_pairs` (Block List) The field pairs for the reference table. (see [below for nested schema](#nestedblock--rules--reference_table--field_pairs))
 - `if_not_exists` (Boolean) Whether to apply the reference table only if the key doesn't exist.
 - `source_keys` (List of String) The source keys for the reference table lookup.
 - `table_name` (String) The name of the reference table.
Uncommitted changes were detected in the autogenerated docs folder. Please run 'make docs' to autogenerate the docs, and commit the changes
M docs/data-sources/tag_pipeline_ruleset.md M docs/resources/tag_pipeline_ruleset.md
make: *** [GNUmakefile:83: check-docs] Error 1
Error: Process completed with exit code 2.
```

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+type tagPipelineRulesetModel struct {
+	ID       types.String `tfsdk:"id"`
+	Name     types.String `tfsdk:"name"`
+	Enabled  types.Bool   `tfsdk:"enabled"`
+	Position types.Int64  `tfsdk:"position"`
+	Version  types.Int64  `tfsdk:"version"`
+	Rules    []ruleItem   `tfsdk:"rules"`
+}
+
+type ruleItem struct {
```
alextmeyer, line 41: So talked about this in our 1:1 - wondering if we maybe want a resource for rules

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+				Computed:    true,
+				Description: "The version of the ruleset.",
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"rules": schema.ListNestedBlock{
+				Description: "The rules in the ruleset.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"enabled": schema.BoolAttribute{
```
alextmeyer, line 116: These feel weird being nested so far under the other resource. If an isolated resource a user could have a rules.tf where all separate rules are specified then this would be a list representing their order

alextmeyer: This is a very complicated nested type.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+func (r *tagPipelineRulesetResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan tagPipelineRulesetModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Validate rules configuration
+	for i, rule := range plan.Rules {
+		// Count how many rule types are defined
```
alextmeyer: I think instead we can use terraform resource validators:
```
func (r *myResource) ConfigValidators(_ context.Context) []resource.ConfigValidator {
  return []resource.ConfigValidator{
    resourcevalidator.ExactlyOneOf(
      path.MatchRoot("mapping"),
      path.MatchRoot("query"),
      path.MatchRoot("reference_table"),
    ),
  }
}
```
However you'd have to do that for how nested this is

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+			resp.Diagnostics.AddError(
+				"Multiple rule configurations",
+				fmt.Sprintf("rules[%d] can only define one of: mapping, query, or reference_table", i),
+			)
+			continue
+		}
+
+		// Validate mapping block
+		if rule.Mapping != nil {
+			if rule.Mapping.DestinationKey.IsNull() || rule.Mapping.DestinationKey.ValueString() == "" {
```
alextmeyer: Why not use Optional: true on top level Mapping, Query, and Reference table with the nested required columns (like here) as Required: True

Terraform should do all this schema validation out of the box for you

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+			}
+			if len(rule.Mapping.SourceKeys) == 0 {
+				resp.Diagnostics.AddError(
+					"Missing required attribute",
+					fmt.Sprintf("rules[%d].mapping.source_keys is required when mapping block is used", i),
+				)
+			}
+		}
+
+		// Validate query block
```
alextmeyer: All this validation should overall be unnecessary - terraform should be able to enforce it

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Use the ID and version from the current state, not the plan (needed for the update API)
+	plan.ID = state.ID
+	plan.Version = state.Version
+
+	// Validate rules configuration
+	for i, rule := range plan.Rules {
```
alextmeyer: Dupe validation to above? If we can't get terraform to do all this for us can we at least remove duplication

alextmeyer: This is line for line the same as above - down to line 479

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+	}
+
+	apiReq := buildUpdateRulesetRequestFromModel(plan)
+	apiResp, response, err := r.Api.UpdateRuleset(r.Auth, rulesetId, apiReq)
+	if err != nil {
+		resp.Diagnostics.AddError("Error updating ruleset", fmt.Sprintf("RulesetID: %s, Error: %s", rulesetId, utils.TranslateClientError(err, response, "").Error()))
+		return
+	}
+
+	// Fix for Terraform CLI issue #32460:
```
alextmeyer: https://github.com/hashicorp/terraform/issues/32460?

Was this AI? Or an actual thing we hit? If this is the right issue it is closed

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+	if newState.Position.IsNull() || newState.Position.IsUnknown() {
+		readResp, _, readErr := r.Api.GetRuleset(r.Auth, rulesetId)
+		if readErr == nil {
+			setModelFromRulesetResp(&newState, readResp)
+		}
+	}
+
+	// The key fix: For each rule, ensure that rule type blocks match the config
+	// If config doesn't have a block type, explicitly null it in the new state
+	// This prevents "planned for existence but config wants absence" errors
```
alextmeyer: Something we actually hit?

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
```diff
@@ -0,0 +1,910 @@
...
+		Type:       datadogV2.CREATERULESETREQUESTDATATYPE_CREATE_RULESET,
+	}
+
+	// Build and return the top-level object
+	return datadogV2.CreateRulesetRequest{
+		Data: &data,
+	}
+}
+
+func buildUpdateRulesetRequestFromModel(plan tagPipelineRulesetModel) datadogV2.UpdateRulesetRequest {
```
alextmeyer, line 611: Looks very very similar to above - can we eliminate duplications

datadog/fwprovider/resource_datadog_tag_pipeline_rulesets.go
```diff
@@ -0,0 +1,256 @@
+package fwprovider
```
alextmeyer, line 1: Maybe just name this resource "tag_pipeline_rulesets"? We only allow one per datadog account and it is the set of their rulesets in a specific order

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset_order.go
```diff
@@ -0,0 +1,256 @@
...
+	r.Auth = providerData.Auth
+}
+
+func (r *tagPipelineRulesetOrderResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = "tag_pipeline_ruleset_order"
+}
+
+func (r *tagPipelineRulesetOrderResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "Provides a Datadog Tag Pipeline Ruleset Order API resource. This can be used to manage the order of Datadog Tag Pipeline Rulesets.",
```
alextmeyer:
```suggestion
		Description: "Provides a Datadog Tag Pipeline Ruleset Order resource that can be used to manage the order of Tag Pipeline Rulesets.",
```

datadog/fwprovider/resource_datadog_tag_pipeline_rulesets.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// The tag pipeline API requires ALL existing rulesets to be included in the reorder call
+	// So we always use the comprehensive approach
+	r.updateOrderWithAllRulesets(state, diag, desiredRulesetIDs)
+}
+
+// Fallback method that includes all existing rulesets
+func (r *tagPipelineRulesetOrderResource) updateOrderWithAllRulesets(state *tagPipelineRulesetOrderModel, diag *diag.Diagnostics, desiredOrder []string) {
+	// Get all existing rulesets
```
alextmeyer, line 175: We're pulling all to update all?

We should be able to use ruleset terraform resources to validate existence right?

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset_order.go
```diff
@@ -0,0 +1,256 @@
...
+	finalOrder := make([]string, 0, len(rulesetPositions))
+	desiredIDsMap := make(map[string]bool)
+
+	// Add desired rulesets in specified order
+	for _, id := range desiredOrder {
+		finalOrder = append(finalOrder, id)
+		desiredIDsMap[id] = true
+	}
+
+	// Add remaining rulesets in their current order
```
alextmeyer, line 270: I'd assume we require all rulesets to exist in terraform or they get overwritten. This seems like we'd support reordering by what is referenced here and throwing anything left at the end

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,164 @@
...
+---
+
+# datadog_tag_pipeline_ruleset (Resource)
+
+Provides a Datadog Tag Pipeline Ruleset resource.
+
+## Example Usage
+
+```terraform
+resource "datadog_tag_pipeline_ruleset" "example" {
```
alextmeyer, line 16: Is this correct? How are multiple rules represented here? Isn't rules a list?

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: The customer can CRUD and reorder rules under ruleSets by current design. If we have a new resource for rules, we need a new reorder resource for rules as well.
Also from API perspective, rulesets are CRUD with their rules as a unit, there is no APIs just create rules alone. Unless API supports CRUD rules separately, then it is not a good idea to this in Terraform.

alextmeyer: Yeah unfortunate we have no CRUD for rules. However it would be nice to have a resource type per rule type - keeps changes isolated more.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: This ConfigValidators checks across ALL rules in the list, not within each rule. This can't be applied here. Unless the rule itself is a separate resource.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: No, it couldn't. If I use ConfigValidators along with Required: True, when I just create a reference_table rule, it still needs required fields from query and mapping rule type to be there. Same as above comment, so we have to do some manually check.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: I agree it would be much cleaner to have a separate resource for rules but will that be weird? if you have a separate resource for rule, when you apply it, you actually create a rule and how can it be created without API?

alextmeyer: I think since we don't have it exposed via API for now the current approach is fine. Maybe something to note for future improvements/iterations. Without having real underlying resources you'd have to do a lot of weird state management to change the parent ruleset on a resource change

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: Since we keep rules inside ruleset, we have to have this nested type.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: terraform can't do validation for this nested blocks. It has some limitations, we have to add manual checks.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: This was added by fixing TF cli error, but api reliability team fixed that for me. I can remove this.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset.go
congx4: It was AI, removed this.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset_order.go
congx4: Initially I was thinking not all rules are managed by terraform when they start using terraform. But we should force the customer to use terraform to manage all their rulesets then start using  datadog_tag_pipeline_rulesets resource. We can just return the error since reordering requires all rulesets ID should be there.

docs/resources/tag_pipeline_ruleset.md
congx4: yes, we use ListNestedBlock in the schema and this is the correct HCL syntax for schema.ListNestedBlock in Terraform

datadog/fwprovider/resource_datadog_tag_pipeline_rulesets.go
congx4: Well, customers can sometime use API or UI create new rulesets without using terraform. We have to ensure the terraform manage all rulesets first, then we allow customers to use datadog_tag_pipeline_rulesets resource to reorder rulesets.

datadog/fwprovider/resource_datadog_tag_pipeline_ruleset_order.go
congx4: Now when they create datadog_tag_pipeline_rulesets resources without full managed rulesets from Terraform, we will throw error like this:

datadog/fwprovider/resource_datadog_tag_pipeline_rulesets.go
alextmeyer: Hmm I'm not sure this would be expected behavior. May be worth confirming with the team/product. I would have the expectation that Terraform is the source of truth and should override anything not in terraform.

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -0,0 +1,164 @@
...
+  name    = "Complete Tag Pipeline Example"
+  enabled = true
+
+  rules {
```
alextmeyer, line 20: Makes sense from your answer above. Unfortunately, this is really not obvious that these are listed in order or order here matters - expected behavior though

examples/resources/datadog_tag_pipeline_ruleset_order/import.sh
```diff
@@ -0,0 +1 @@
+terraform import datadog_tag_pipeline_ruleset_order.my_ruleset_order order
```
alextmeyer: How does this work? We only have a single order per customer - is this an id? Or just nothing?

terraform-provider-datadog
alextmeyer, line 1: Remove

terraform-provider-datadog_v999.0.0
alextmeyer, line 1: Remove

alextmeyer (approved):
Fix to examples for tag_pipeline_rulesets instead of datadog_tag_pipeline_ruleset_order and remove 2 BIN files

PR #3238, opened 2025-09-19
Fix type error in building of unitScale

brlee19:


datadog/resource_datadog_dashboard.go
```diff
@@ -11106,8 +11106,7 @@ func buildTerraformNumberFormatFormulaSchema(datadogStyle datadogV1.WidgetNumber
 		m["unit"] = []map[string]interface{}{unit}
 	}
 	if v, ok := datadogStyle.GetUnitScaleOk(); ok {
-		unitScale := map[string]interface{}{}
-		unitScale["unit_name"] = []map[string]interface{}{{"unit_name": v.UnitName}}
+		unitScale := map[string]interface{}{"unit_name": v.UnitName}
 		m["unit_scale"] = []map[string]interface{}{unitScale}
```
brlee19, line 11119: The schema for this is:

```
"unit_scale": &schema.Schema{
			Description: "",
			Type:        schema.TypeList,
			MinItems:    0,
			MaxItems:    1,
			Optional:    true,
			Elem: &schema.Resource{
					"unit_name": &schema.Schema{
						Description: "",
						Required:    true,
					},
				},
			},
		},
```

so `m["unit_scale"]` is supposed to contain a list of `map[string]string` if I am reading this right, not another map containing "unit_name" as a key

brlee19: The old code appears to be the cause of the error per the [CI output](https://gitlab.ddbuild.io/DataDog/dd-source/-/jobs/1132659662):

```level=fatal msg="Execution failed" error="error: error generating plan: exit status 1\n\nError: widget.1.group_definition.0.widget.0.query_value_definition.0.request.0.formula.0.number_format.0.unit_scale.0.unit_name: '' expected type 'string', got unconvertible type '[]map[string]interface {}', value: '[map[unit_name:0xc0009716e0]]'\n\n  with datadog_dashboard.maze_api_dashboard,\n  on dashboard.tf line 1, in resource \"datadog_dashboard\" \"maze_api_dashboard\":\n   1: resource \"datadog_dashboard\" \"maze_api_dashboard\" {\n\n code: INVALID_ARGUMENT origin: terraform-invocation, action: plan, cause: USER, resource: \n"```

aka it is saying that the first element of `unit_scale` had a `[]map[string]interface {}'` at its `unit_name` key and not just a string

Comment (brlee19) [reply datadog/resource_datadog_dashboard.go:11119]: I am not sure why the old code was working until Wednesday though this probably has something to do with the [new release](https://github.com/DataDog/terraform-provider-datadog/pull/3232) that day.

File: datadog/resource_datadog_dashboard.go [hunk 2]
```diff
@@ -11057,7 +11057,7 @@ func getNumberFormatFormulaSchema() map[string]*schema.Schema {
 	}
 }
 
-func buildNumberFormatFormulaSchema(terraformStyle map[string]interface{}) *datadogV1.WidgetNumberFormat {
+func buildDatadogNumberFormatFormulaSchema(terraformStyle map[string]interface{}) *datadogV1.WidgetNumberFormat {
```
brlee19, line 11060: The `buildDatadog` functions are on the write path (turning HCL to Datadog schema), renamed this for clarity.

datadog/resource_datadog_dashboard.go
```diff
@@ -11087,6 +11087,15 @@ func buildNumberFormatFormulaSchema(terraformStyle map[string]interface{}) *data
 			}
 		}
 	}
+	if v, ok := terraformStyle["unit_scale"].([]interface{}); ok && len(v) > 0 {
```
brlee19, line 11090: I wasn't able to actually create a dashboard with this field with `unitScale` until adding this too. Uses the same schema as below.

brlee19:
> this makes sense to me - is there any risk of this affecting existing terraforms? I would assume not since the comment makes it sound like this was working until recently?

I actually don't think this field was ever working with terraform and that this field only made its way into the dashboard because it was added via in the UI (it was not exclusively terraform-managed).

This seems like a relatively simple change though and I don't think it would cause any issues, especially since it is basically just fixing broken code we have already (that implements a field that has been in our public API spec since February) and also adding the fix to the read path, not just the write.

PR #3257, opened 2025-10-01
[datadog_synthetics_test] Support the `blockedRequestPatterns` option for browser tests

SKYWy:
This PR adds support for the `blockedRequestPatterns` option for browser tests.

api-spec PR: [Add blockedRequestPatterns](https://github.com/DataDog/datadog-api-spec/pull/4393)

While updating recording for tests accordingly, we noticed a change in behaviour for RUM settings which was causing the tests to fail. In our public APIs where the fields `application_id` and `client_token_id` are now auto-generated if RUM is enabled and the user does not explicitly set them.
These fields are updated to optional computed fields accordingly.

### DIFF
--- a/datadog/fwprovider/data_source_datadog_custom_allocation_rule.go
+++ b/datadog/fwprovider/data_source_datadog_custom_allocation_rule.go
@@ -269,7 +269,7 @@ func (d *datadogCustomAllocationRuleDataSource) Read(ctx context.Context, reques
 	}
 
 	ruleId := state.RuleId.ValueInt64()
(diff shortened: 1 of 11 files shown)


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -867,6 +867,14 @@ func syntheticsTestOptionsList() *schema.Schema {
 					Type:        schema.TypeInt,
 					Optional:    true,
 				},
+                "blocked_request_patterns": {
+                    Description: "URL patterns to block for browser tests.",
```
zach-b: What about something like "Blocked URL patterns. Requests made to URLs matching any of the patterns listed here will be blocked" ?

PR #3258, opened 2025-10-01
[CCA-1173] Add support for Cloud Cost Google Cloud Usage Cost Config resource and data source

alextmeyer:


docs/data-sources/gcp_uc_config.md
```diff
@@ -0,0 +1,37 @@
...
+- `created_at` (String) The timestamp when the GCP Usage Cost configuration was created.
+- `dataset` (String) The resolved BigQuery dataset name used for the Usage Cost export.
+- `error_messages` (List of String) List of error messages if the GCP Usage Cost configuration encountered any issues during setup or data processing.
+- `export_prefix` (String) The prefix path within the storage bucket where Usage Cost export files are organized.
+- `export_project_name` (String) The Google Cloud Project ID where the Usage Cost export is configured.
+- `id` (String) The ID of this resource.
+- `months` (Number) The number of months of historical cost data available for analysis.
+- `project_id` (String) The resolved Google Cloud Project ID for the Usage Cost export.
+- `service_account` (String) The Google Cloud service account email that Datadog uses to access the Usage Cost export data.
+- `status` (String) The current status of the GCP Usage Cost configuration (e.g., active, archived).
```
buraizu:
```suggestion
- `status` (String) The current status of the GCP Usage Cost configuration (for example, active or archived).
```

alextmeyer: @buraizu So these are just examples of all possible status values - unknown, active, warn, error, disabled, archived

That is the current set. Would it be better to say "active, archived, etc."?

datadog/fwprovider/data_source_datadog_gcp_uc_config.go
```diff
@@ -0,0 +1,169 @@
...
+	BucketName        types.String `tfsdk:"bucket_name"`
+	CreatedAt         types.String `tfsdk:"created_at"`
+	Dataset           types.String `tfsdk:"dataset"`
+	ExportPrefix      types.String `tfsdk:"export_prefix"`
+	ExportProjectName types.String `tfsdk:"export_project_name"`
+	Months            types.Int64  `tfsdk:"months"`
+	ProjectId         types.String `tfsdk:"project_id"`
+	ServiceAccount    types.String `tfsdk:"service_account"`
+	Status            types.String `tfsdk:"status"`
+	StatusUpdatedAt   types.String `tfsdk:"status_updated_at"`
```
gstroz, line 41: Could we convert timestamps to RFC 3339 format? Terraform has a library for timestamps and that's the recommended format: https://github.com/hashicorp/terraform-plugin-framework-timetypes

Caveat that we need timezone information for the timestamp

datadog/fwprovider/data_source_datadog_gcp_uc_config.go
```diff
@@ -0,0 +1,169 @@
...
+			"service_account": schema.StringAttribute{
+				Computed:    true,
+				Description: "The Google Cloud service account email that Datadog uses to access the Usage Cost export data.",
+			},
+			"status": schema.StringAttribute{
+				Computed:    true,
+				Description: "The current status of the GCP Usage Cost configuration (for example, active, error, or archived).",
+			},
+			"status_updated_at": schema.StringAttribute{
+				Computed:    true,
```
gstroz, line 113: If this is converted to RFC 3339 then we'd add:
CustomType: timetypes.NewRFC3339Type()
and it would validate automatically.

datadog/fwprovider/data_source_datadog_gcp_uc_config.go
```diff
@@ -0,0 +1,169 @@
...
+		state.BucketName = types.StringValue(attributes.GetBucketName())
+		state.Dataset = types.StringValue(attributes.GetDataset())
+		state.ExportPrefix = types.StringValue(attributes.GetExportPrefix())
+		state.ExportProjectName = types.StringValue(attributes.GetExportProjectName())
+		state.ServiceAccount = types.StringValue(attributes.GetServiceAccount())
+		state.Status = types.StringValue(attributes.GetStatus())
+		state.CreatedAt = types.StringValue(attributes.GetCreatedAt())
+		state.Months = types.Int64Value(int64(attributes.GetMonths()))
+		state.ProjectId = types.StringValue(attributes.GetProjectId())
+		state.StatusUpdatedAt = types.StringValue(attributes.GetStatusUpdatedAt())
```
gstroz, line 165: Pretty sure this is where the actual conversion would take place from `2025-10-01T18:20:01.684913`. If we're using UTC for timestamps then we'd need to add that as an assumption.

datadog/fwprovider/data_source_datadog_gcp_uc_config.go
alextmeyer: This is only used in one place int he datadog terraform provider with other timestamps across data sources represented as strings. For consistency I will leave this as is for now. Given this is a data source I'd expect users to infrequently use this anyways (referencing an already existing config independent of the actual resource referencing it)

PR #3262, opened 2025-10-03
Add custom allocation rules terraform resource

elihschiff:


git-thuerk-done (approved):
good for docs

datadog/fwprovider/resource_datadog_custom_allocation_rule_order.go
```diff
@@ -0,0 +1,281 @@
+package fwprovider
```
congx4: This file can be renamed to resource_datadog_custom_allocation_rules.go as there will be only one of this resource for a customer. 
This aligns with the name from tag pipeline as well. 

remember to update the folder for examples, docs and tests as well.

datadog/fwprovider/resource_datadog_custom_allocation_rule_order.go
```diff
@@ -0,0 +1,281 @@
...
+func (r *customAllocationRuleOrderResource) Delete(ctx context.Context, request resource.DeleteRequest, response *resource.DeleteResponse) {
+	// No-op: Deleting the order resource doesn't change the actual order of rules
+	// This follows the same pattern as other order resources in the provider
+}
+
+func (r *customAllocationRuleOrderResource) ImportState(ctx context.Context, request resource.ImportStateRequest, response *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, frameworkPath.Root("id"), request, response)
+}
+
+func (r *customAllocationRuleOrderResource) updateOrder(state *customAllocationRuleOrderModel, diag *diag.Diagnostics) {
```
congx4: We want to make sure that all rules from the backend should be the source of truth. If the request from the TF didn't include all rules, this request should be failed. That says: customers should import all rules in TF first, then invoke reorder along with all rules.

elihschiff: Ok that makes sense, I can change it
I thought this approach was more flexible but I would rather be consistent

datadog/fwprovider/data_source_datadog_custom_allocation_rule.go
```diff
@@ -0,0 +1,479 @@
...
+			"id": utils.ResourceIDAttribute(),
+			// Query Parameters
+			"rule_id": schema.Int64Attribute{
+				Optional:    true,
+				Description: "The ID of the custom allocation rule.",
+			},
+			// Computed values
+			"created": schema.StringAttribute{
+				Computed:    true,
+				Description: "The `attributes` `created`.",
```
alextmeyer: Can we improve all these descriptions across schemas - they're used in the auto-generated docs

datadog/fwprovider/resource_datadog_custom_allocation_rule.go
```diff
@@ -0,0 +1,929 @@
...
+		Attributes: map[string]schema.Attribute{
+			"enabled": schema.BoolAttribute{
+				Required:    true,
+				Description: "The `attributes` `enabled`. Whether the rule is enabled.",
+			},
+			"order_id": schema.Int64Attribute{
+				Computed:    true,
+				Description: "The `attributes` `order_id`. This field is read-only and returned by the API. Use the `datadog_custom_allocation_rule_order` resource to manage the order of rules.",
+			},
+			"rejected": schema.BoolAttribute{
```
alextmeyer, line 229: Ensure the way these read-only values are implemented won't make user terraform view itself as having changes to apply - may already be that way but don't want rejected or timestamps to make state seem invalid

examples/resources/datadog_custom_allocation_rule/import.sh
```diff
@@ -0,0 +1 @@
+terraform import datadog_custom_allocation_rule.new_list ""
```
alextmeyer:
```suggestion
terraform import datadog_custom_allocation_rule.new_list <rule_id>
```

docs/data-sources/custom_allocation_rule.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Optional
+
+- `rule_id` (Number) The ID of the custom allocation rule.
+- `strategy` (Block, Optional) (see [below for nested schema](#nestedblock--strategy))
+
+### Read-Only
+
+- `costs_to_allocate` (Block List) (see [below for nested schema](#nestedblock--costs_to_allocate))
+- `created` (String) The `attributes` `created`.
```
alextmeyer: Description improvements in code will make these more meaningful and useful

datadog/fwprovider/resource_datadog_custom_allocation_rule.go
elihschiff: As far as I can tell I have it implemented correctly to do that. I was not able to get it to create any issues in my testing either. So I think it should hopefully be fine 🤞

alextmeyer: So you tried modifying a rule from the UI that was stored in tfstate (updating last modified and updated) and terraform plan didn't say changes to be made?

alextmeyer: Wanted to make sure as I ran into this with some of the config resources and just had things like status exposed via the datasource (not resource) - maybe that was wrong to do if this would work

elihschiff: Correct me if I am wrong but is this video a demo of what you are asking? And if so I think it is working as expected

https://dd.enterprise.slack.com/files/U022FEE80QK/F09LB97FRB2/screen_recording_2025-10-08_at_4.15.34___pm.mov

elihschiff: Confirmed with alex over DM (because the video link here was not working) that my testing looks good

PR #3264, opened 2025-10-06
[synthetics_test] Remove default value for http version for tests type not supporting it

romainberger:
The `http_version` attribute currently has a default value, however this default value is not needed as the field is not required. Having a default value creates an issue where this field is not used for all test types, and creates unnecessary plan diffs.
This PR removes the default value altogether.


HantingZhang2 (approved):

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -2517,17 +2516,6 @@ func updateSyntheticsAPITestLocalState(d *schema.ResourceData, syntheticsTest *d
...
-				if step.SyntheticsAPITestStep.GetSubtype() == "grpc" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "ssl" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "dns" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "websocket" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "tcp" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "udp" ||
-					step.SyntheticsAPITestStep.GetSubtype() == "icmp" {
-					// the schema defines a default value of `http_version` for any kind of step,
-					// but it's not supported for `grpc` - so we save `any` in the local state to avoid diffs
-					localRequest["http_version"] = datadogV1.SYNTHETICSTESTOPTIONSHTTPVERSION_ANY
```
fpighi, line 2529: Since state for customers is now populated with `any` for most tests, I imagine this to cause drifts in every new plan.
I think we should strive not to have those drifts if feasible.
Also, in the case of http tests, it should probably still default to `any` right?

romainberger: Unfortunately yes this will create drift in new plans, but I think this is for the best as the default value was not needed at all. 
Even for http tests, the default value is not required and the backend will accept requests without it.

I understand the issue of drifts in plan, if you really want to avoid this we can close the PR, the original issue comes from editing in the UI a terraformed test, which should (I guess) not happen anyway so the case should be very rare.

fpighi: Since it might generate confusion for customers, I suggest one of these:
1. detect when state contains something and set the request to the same value (if not specified) to avoid showing confusing plan changes
2. accept the plan will show a diff but document it in the release notes explicitly so it's expected
3. make this change in the next major release (4.0) since as you say it should be rare

romainberger: Sounds good, I'll close the PR and we can keep that for the next major release. It should be a rare case so I don't think it's worth confusing users, nor spending too much time on this to avoid drift for a very minor fix.

PR #3274, opened 2025-10-09
[azure_uc_config][aws_cur_config][gcp_uc_config] Update Cloud cost provider resources to support exposing computed fields

alextmeyer:


datadog/fwprovider/data_source_datadog_gcp_uc_config.go
```diff
@@ -35,7 +35,6 @@ type datadogGcpUcConfigDataSourceModel struct {
 	ExportPrefix      types.String `tfsdk:"export_prefix"`
 	ExportProjectName types.String `tfsdk:"export_project_name"`
 	Months            types.Int64  `tfsdk:"months"`
-	ProjectId         types.String `tfsdk:"project_id"`
```
elihschiff, line 38: Is there a reason to get rid of this?

alextmeyer: Dug into it and it is useless and empty. I think a legacy field that was superseded by account id, project name, and billing_account_id

alextmeyer: Verified with @kaankoseler

PR #3275, opened 2025-10-09
[datadog_monitor][feature] Update Monitors terraform to support draft monitors

kevinpombo-datadog:
Updated monitors terraform to add support for setting draft status as an optional value with it defaulting to published when not set - https://datadoghq.atlassian.net/browse/MA-6686

### DIFF
--- a/datadog/fwprovider/resource_datadog_monitor.go
+++ b/datadog/fwprovider/resource_datadog_monitor.go
@@ -438,7 +438,7 @@ func (r *monitorResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				Optional:    true,
 			},
 			"draft_status": schema.StringAttribute{
-				Description:      "Indicates whether the monitor is in a draft or published state. When set to `draft`, the monitor appears as Draft and does not send notifications. When set to `published`, the monitor is active, and it evaluates conditions and sends notifications as configured.",
+				Description: "Indicates whether the monitor is in a draft or published state. When set to `draft`, the monitor appears as Draft and does not send notifications. When set to `published`, the monitor is active, and it evaluates conditions and sends notifications as configured.",
 				Optional:    true,
 				Computed:    true,
 				Default:     stringdefault.StaticString(string(datadogV1.MONITORDRAFTSTATUS_PUBLISHED)),


docs/resources/monitor.md
```diff
@@ -48,6 +48,7 @@ Email notifications can be sent to specific users by using the same `@username`
 
 ### Optional
 
+- `draft_status` (String) Indicates whether the monitor is in a draft or published state. draft: The monitor appears as Draft and does not send notifications. published: The monitor is active and evaluates conditions and notify as configured. Valid values are `draft`, `published`. Defaults to `"published"`.
```
domalessi:
```suggestion
- `draft_status` (String) Indicates whether the monitor is in a draft or published state. When set to `draft`, the monitor appears as Draft and does not send notifications. When set to `published`, the monitor is active, and it evaluates conditions and sends notifications as configured. Valid values are `draft` and `published`. Defaults to `published`.
```

nickpaw:
Are we missing the read in `updateMonitorState`?

datadog/resource_datadog_monitor.go
```diff
@@ -808,13 +815,15 @@ func buildMonitorStruct(d utils.Resource) (*datadogV1.Monitor, *datadogV1.Monito
 	m.SetName(d.Get("name").(string))
 	m.SetMessage(d.Get("message").(string))
 	m.SetOptions(o)
+	m.SetDraftStatus(d.Get("draft_status").(datadogV1.MonitorDraftStatus))
```
nickpaw: Is the assertion correct? Following the monitor type enum, do we need to assert as a string and convert the same way?
`m.SetDraftStatus(datadogV1.MonitorDraftStatus(d.Get("draft_status").(string)))`

Also present below ln 826

kevinpombo-datadog: In this case SetDraftStatus is expecting an instance of datadogV1.MonitorDraftStatus

kevinpombo-datadog:
@nickpaw Updated updateMonitorState for both providers now leaving it as an optional field

datadog/resource_datadog_monitor.go
nickpaw: thanks, right, but does d.Get("draft_status") return an instance of datadogV1.MonitorDraftStatus or a string? Following the other enum example it looked like it returns a string and needed to be converted?

datadog/resource_datadog_monitor.go
```diff
@@ -427,6 +427,12 @@ func resourceDatadogMonitor() *schema.Resource {
 					Optional:         true,
 					ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewMonitorOptionsNotificationPresetsFromValue),
 				},
+				"draft_status": {
+					Description:      "Indicates whether the monitor is in a draft or published state. When set to `draft`, the monitor appears as Draft and does not send notifications. When set to `published`, the monitor is active, and it evaluates conditions and sends notifications as configured.",
```
phillip-dd, line 431: Can we note that the default value if not configured is `published`?

kevinpombo-datadog: Should be updated in the docs now: https://github.com/DataDog/terraform-provider-datadog/pull/3275/files#diff-022eeb2e601163d1dec7b8127c09ca14269934663bff2193a7356c4baa283370R51

datadog/resource_datadog_monitor.go
kevinpombo-datadog: d.Get("draft_status") will return a string that we have to convert to the enum, this should be updated now

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -435,6 +437,15 @@ func (r *monitorResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				Description: "If set to `false`, skip the validation call done during plan.",
 				Optional:    true,
 			},
+			"draft_status": schema.StringAttribute{
+				Description: "The type of the monitor. The mapping from these types to the types found in the Datadog Web UI can be found in the Datadog API [documentation page](https://docs.datadoghq.com/api/v1/monitors/#create-a-monitor). Note: The monitor type cannot be changed after a monitor is created.",
```
nickpaw: Description here looks like it was for monitor.type?

LiuVII:
iiuc the PR name must start with `[datadog_monitor]` for cleaner changelog

LiuVII (approved):
lgtm

PR #3283, opened 2025-10-14
[datadog_custom_allocation_rule] update terraform go api-client

congx4:


congx4:
> Do we need to run `make update-go-client` or does master already have the newest version? i.e. is the provider currently broken?

we need, just add go.mod to the pr.

PR #3286, opened 2025-10-15
[datadog_deployment_gate] Add deployment gates terraform provider

Aaron-9900:
Adds the terraform resource for deployment gates. Includes docs and tests as well.

One point that differ from how the API behaves:
1. Rule names must be unique within a gate. This is because at creation time, rules don't have an ID and we need something to identify rules. This is enforced by the tf client

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -85,4 +85,5 @@ datadog/**/*datadog_azure_uc_config*             @DataDog/api-reliability @DataD
 datadog/**/*datadog_gcp_uc_config*               @DataDog/api-reliability @DataDog/cloud-cost-management
 datadog/**/*datadog_cost_budget*                 @DataDog/api-reliability @DataDog/cloud-cost-management
 datadog/**/*datadog_tag_pipeline_ruleset*        @DataDog/api-reliability @DataDog/cloud-cost-management
+datadog/**/*datadog_deployment_gate*             @DataDog/api-reliability @DataDog/ci-app-backend
 scripts/*cloud-cost-import-existing-resources*   @DataDog/cloud-cost-management


datadog/fwprovider/resource_datadog_deployment_gate.go
```diff
@@ -0,0 +1,959 @@
...
+	response.TypeName = "deployment_gate"
+}
+
+func (r *deploymentGateResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "Provides a Datadog DeploymentGate resource. This can be used to create and manage Datadog deployment_gate.",
+		Attributes: map[string]schema.Attribute{
+			"dry_run": schema.BoolAttribute{
+				Optional:    true,
+				Description: "The `attributes` `dry_run`.",
```
GabrielAnca: Should we add a more descriptive description in these fields?

GabrielAnca (approved):

docs/resources/deployment_gate.md
```diff
@@ -0,0 +1,100 @@
...
+- `query` (String) The query for monitor rules.
+
+## Import
+
+Import is supported using the following syntax:
+
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
+
+```shell
+terraform import datadog_deployment_gate.new_list ""
```
brett0000FF, line 99: It would be better if we can add a placeholder value for the deployment gate ID between these double quotes.
```suggestion
terraform import datadog_deployment_gate.new_list "<deployment-gate-id>"
```

brett0000FF (approved):
Approved with a small suggestion. I know the docs are auto-generated though.

St0rmingBr4in (approved):

PR #3297, opened 2025-10-22
[datadog_integration_gcp_sts] Fix handling prometheus default in mncs

katherinekim-51:
r? @DataDog/gcp-integrations 
r? @ash-ddog 

## Description

This updates the terraform provider to fix handling of any server provided defaults (ex. prometheus): 

- Today, this default is never honored because an explicit [] is always sent to the API server, overriding the default  
- It was gaveled that defaults should only apply when the user omits the field all together 
- This creates 3 main scenarios:
    1. When the user omits the field in create, and also omits in the update --> we should keep the server default in both instances 
    2. When the user omits the field in create, but does not omit in the update --> the default should be gone on update
    3. When the user does not omit the field in create, and omits in the update --> the default should never be used, and **and the inputs on create should be preserved** (an additional change from today's behavior) 
    
## Implementation Details

Primary changes: 
- Updates mncs to be _computed_ (otherwise there will be an "inconsistent plan after apply" if server returns a default)
- Because the field is now _computed_, and pointers cannot represent Unknown/Null, changes the struct type to Types.Set
- To be consistent with monitored_resource_configs, moves/refactors the elementType into a "*spec" 
- Updates the requestBuilder func to omit the field if it is unknown/not set

## Testing Guidelines

Reran cassettes and added test cases for the 3 scenarios above 
## Dependencies (optional)

https://datadoghq.atlassian.net/browse/GCP-2863
## What will it impact? (optional)

See Also: https://github.com/DataDog/devops/wiki/Code-Workflow#prod

### Additional Notes


datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -152,14 +158,11 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 				Computed:    true,
 			},
 			"metric_namespace_configs": schema.SetAttribute{
-				Optional:    true,
+				Optional: true,
```
katherinekim-51, line 160: for additional context: using a default value does not work because that default gets applied to both update and create which we don't want
   --> and you can't use a plan modifier to omit the Update default because plan modifiers are used **after** defaults are populated

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -119,7 +358,7 @@ func TestAccIntegrationGcpStsBasic(t *testing.T) {
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_resource_change_collection_enabled"), knownvalue.Bool(true)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_per_project_quota_enabled"), knownvalue.Bool(true)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("account_tags"), knownvalue.Null()),
-					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("metric_namespace_configs"), knownvalue.Null()),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("metric_namespace_configs"), knownvalue.SetExact(nil)),
```
katherinekim-51: this is because we get rid of the 'if mncs := attributes.GetMetricNamespaceConfigs(); len(mncs) > 0' conditional, which is now necessary for this computed field because the field can't remain unknown; this should be okay because server-wise we established that omission and [] == same behavior

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -166,7 +405,12 @@ func TestAccIntegrationGcpStsDefault(t *testing.T) {
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_resource_change_collection_enabled"), knownvalue.Bool(false)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_per_project_quota_enabled"), knownvalue.Bool(false)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("account_tags"), knownvalue.Null()),
-					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("metric_namespace_configs"), knownvalue.Null()),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("metric_namespace_configs"), knownvalue.SetExact([]knownvalue.Check{
```
katherinekim-51, line 310: this now matches expectations (the default is prometheus disabled)

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -49,20 +55,20 @@ type MonitoredResourceConfigModel struct {
 }
```
ash-ddog: Let's move these up to the `var` declaration above

```
var (
	integrationGcpStsMutex sync.Mutex
	_                      resource.ResourceWithConfigure   = &integrationGcpStsResource{}
	_                      resource.ResourceWithImportState = &integrationGcpStsResource{}

	MetricNamespaceConfigSpec = types.ObjectType{
		AttrTypes: map[string]attr.Type{
			"id":       types.StringType,
			"disabled": types.BoolType,
		},
	}

	MonitoredResourceConfigSpec = types.ObjectType{
		AttrTypes: map[string]attr.Type{
			"type": types.StringType,
			"filters": types.SetType{
				ElemType: types.StringType,
			},
		},
	}
)
```

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -49,20 +55,20 @@ type MonitoredResourceConfigModel struct {
...
+	ClientEmail                       types.String `tfsdk:"client_email"`
+	DelegateAccountEmail              types.String `tfsdk:"delegate_account_email"`
+	HostFilters                       types.Set    `tfsdk:"host_filters"`               // DEPRECATED: use MonitoredResourceConfigs["gce_instance"]
+	CloudRunRevisionFilters           types.Set    `tfsdk:"cloud_run_revision_filters"` // DEPRECATED: use MonitoredResourceConfigs["cloud_run_revision"]
+	IsCspmEnabled                     types.Bool   `tfsdk:"is_cspm_enabled"`
+	IsSecurityCommandCenterEnabled    types.Bool   `tfsdk:"is_security_command_center_enabled"`
+	IsResourceChangeCollectionEnabled types.Bool   `tfsdk:"is_resource_change_collection_enabled"`
+	IsPerProjectQuotaEnabled          types.Bool   `tfsdk:"is_per_project_quota_enabled"`
+	ResourceCollectionEnabled         types.Bool   `tfsdk:"resource_collection_enabled"`
+	MetricNamespaceConfigs            types.Set    `tfsdk:"metric_namespace_configs"`
```
ash-ddog, line 69: 👍

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -266,7 +269,11 @@ func (r *integrationGcpStsResource) Create(ctx context.Context, request resource
 
 func (r *integrationGcpStsResource) Update(ctx context.Context, request resource.UpdateRequest, response *resource.UpdateResponse) {
```
ash-ddog, line 269: So, general comment:

I think it's worth us getting more familiar with the [Resource interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#Resource) we're implementing, and also look at other implementations in this repo.

I say this bc I think we're not using all the info we have available when making decisions. i.e. seems like there's up to 3 versions of the data we're dealing with, depending on the method getting called.
i.e. for `Update`

```
// UpdateRequest represents a request for the provider to update a
// resource. An instance of this request struct is supplied as an argument to
// the resource's Update function.
type UpdateRequest struct {
	// Config is the configuration the user supplied for the resource.
	//
	// This configuration may contain unknown values if a user uses
	// interpolation or other functionality that would prevent Terraform
	// from knowing the value at request time.
	Config tfsdk.Config

	// Plan is the planned state for the resource.
	Plan tfsdk.Plan

	// State is the current state of the resource prior to the Update
	// operation.
	State tfsdk.State
```

and

```
// UpdateResponse represents a response to an UpdateRequest. An
// instance of this response struct is supplied as
// an argument to the resource's Update function, in which the provider
// should set values on the UpdateResponse as appropriate.
type UpdateResponse struct {
	// State is the state of the resource following the Update operation.
	// This field is pre-populated from UpdateResourceRequest.Plan and
	// should be set during the resource's Update operation.
	State tfsdk.State
```

In fact, we actually deserialize `request.State` in our `Update` function and **then don't use it at all**. And we named `request.Plan` as `state` which is not correct in terms of nomenclature.

TLDR; I think moving forward we need to treat our terraform code with more care rather than what we've been doing which is following the patterns in place which may or may not be correct (probably more the latter).

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -401,13 +408,20 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
...
 	}
-	attributes.SetMetricNamespaceConfigs(mncs)
+
+	// only set this field if:
+	// it's for an Update OR
+	// the user explicitly sets the field
+	// otherwise we want to omit it so that the API server can populate defaults when applicable
+	if forUpdate || !state.MetricNamespaceConfigs.IsUnknown() {
+		attributes.SetMetricNamespaceConfigs(mncs)
+	}
```
ash-ddog: I think we remove `forUpdate` and `mncsContainsOnlyPrometheus` and we just need

```
	if cfgs := plan.MetricNamespaceConfigs; !cfgs.IsUnknown() {
		mncs := make([]datadogV2.GCPMetricNamespaceConfig, 0)
		for _, mnc := range tfCollectionToSlice[*MetricNamespaceConfigModel](ctx, diags, cfgs) {
			mncs = append(mncs, datadogV2.GCPMetricNamespaceConfig{
				Id:       mnc.ID.ValueStringPointer(),
				Disabled: mnc.Disabled.ValueBoolPointer(),
			})
		}
		attributes.SetMetricNamespaceConfigs(mncs)
	}
```

At least this is working for me?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -423,6 +437,20 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
...
+	var items []MetricNamespaceConfigModel
+	diags.Append(s.ElementsAs(ctx, &items, false)...)
+	if diags.HasError() || len(items) != 1 {
+		return false
+	}
+
+	return items[0].ID.ValueString() == "prometheus" &&
+		items[0].Disabled.ValueBool()
+}
```
ash-ddog: I'm not quite sure what this is supposed to be doing? Why is having only 1 `MetricNamespaceConfig` that is `id=prometheus,disabled=true` relevant to helping us make a decision about what to do in this provider?

I think between this and
```
	if forUpdate || !state.MetricNamespaceConfigs.IsUnknown() {
		attributes.SetMetricNamespaceConfigs(mncs)
	}
```
the code has become too hard to reason about. This method also feels "hacky" i.e. I don't think Terraform was expecting devs of providers to have to do things like this. It also encodes knowledge of what does defaults are on the server and so we'll prob run into a bug if/when the defaults on the server change.

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,245 @@ import (
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
 )
 
+func TestAccIntegrationGCPMcs(t *testing.T) {
```
ash-ddog: While I'm ok with more tests, I want to make sure we can't just work these test cases into the other ones? i.e. does this really warrant it's own? These tests are end-to-end so the more we add, the more time it takes to run the test suite. That said, if we deem it necessary it's definitely better to be safe.

ash-ddog: I would also user `t.Run` (i.e subtests) with test names rather than these comments. That way the scenario is visible to devs when tests fail.

ash-ddog: I would also trim these tests down to the bear minimum that matters to test the scenarios. i.e. if this test is meant to be very specifically about `MetricNamespaceConfig`s, then I would really only specify those in the HCL and test those attributes in the assertions. IMO that helps communicate to readers what's important.

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,245 @@ import (
...
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_cspm_enabled"), knownvalue.Bool(false)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_security_command_center_enabled"), knownvalue.Bool(false)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("resource_collection_enabled"), knownvalue.Bool(false)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_resource_change_collection_enabled"), knownvalue.Bool(false)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_per_project_quota_enabled"), knownvalue.Bool(false)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("account_tags"), knownvalue.SetExact([]knownvalue.Check{
+						knownvalue.StringExact("acct:one"),
+						knownvalue.StringExact("acct:two"),
+					})),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("metric_namespace_configs"), knownvalue.SetExact(nil)),
```
ash-ddog: So this scenario will fail with the code I'm suggesting.
`// 3. user creates with the field NOT omitted, and updates with empty: prometheus should be absent in both`

I think we can discuss what should happen in this scenario. I actually think what you have here **is** what we discussed offline, but I am not sure if that was the best decision, mostly bc it's inconsistent with how our other state works.

One of the things I learned this week is that most of our fields are marked as `Computed` whether they should be or not. Again, I think this is bc of blindly following existing patterns.

But, from what I can tell, for a `Computed` field, when you remove it from the HCL, Terraform leaves the current state in place bc `Computed` means "server-managed".

So, I think for now it's prob best that `metric_namespace_configs` behave as our other fields do. And we can introduce breaking changes (with a major version bump) in the somewhat near future if we want to change this behavior.

ash-ddog: Image

(I believe we're in the "no-op" scenario, last one)

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
katherinekim-51: i believe if we do this (only look at whether the plan/state is not set) then we'll fail this scenario: 

> When the user omits the field in create, and also omits in the update --> we should keep the prometheus default in both instances (it would be weird/unexpected for the default to be present and then to be cleared on update when the user didn't change anything)

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
katherinekim-51: iirc we did not want to have the case where user omits mncs in both cases (create and update) and that causes funky behavior where from user pov: input = same, but the outcome differs (prometheus is disabled at first and then suddenly not disabled) 

so to preserve this func checks if the previous state _only_ had this entry in mncs, if it did we should still omit the field so on the server prometheus remains disabled and we don't clear it

datadog/tests/resource_datadog_integration_gcp_sts_test.go
katherinekim-51: think i lost you a bit so can discuss offline but just to make sure we're on the same page for context for discussion: 
- it is intentional to set omitted fields to [] to the API because we want TF updates to be a paint over (not the server just TF) 
- the prometheus defaulting in the server is intentionally only on the Create

katherinekim-51:
Spoke with @ash-ddog  --> preference today is to useStateForUnknown/previous state over checking for prometheus 
- Users have alternate way to clear field and this current implementation would only be preferred if we could do it without explicitly checking for "prometheus" 


In other words scenario 3: "When the user does not omit the field, and in the update omits the field --> the prometheus default should never be populated" becomes 
- When the user does not omit the field, and in the update omits the field  --> **user's old input is preserved**

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
katherinekim-51: > In fact, we actually deserialize request.State in our Update function and then don't use it at all. And we named request.Plan as state which is not correct in terms of nomenclature

fyi we were using it for _priorState_ (but this is changing shortly now)

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -353,15 +356,15 @@ func (r *integrationGcpStsResource) updateState(ctx context.Context, state *inte
...
-				Disabled: types.BoolValue(mnc.GetDisabled()),
-			})
-		}
+	mncsapi := attributes.GetMetricNamespaceConfigs()
+	mncs := make([]*MetricNamespaceConfigModel, 0, len(mncsapi))
+	for _, mnc := range mncsapi {
+		mncs = append(mncs, &MetricNamespaceConfigModel{
+			ID:       types.StringValue(mnc.GetId()),
+			Disabled: types.BoolValue(mnc.GetDisabled()),
+		})
```
ash-ddog: Nit: while slightly less performant, would go with this for symmetry with resource configs below

```
	mncs := make([]*MetricNamespaceConfigModel, 0)
	for _, mnc := range attributes.GetMetricNamespaceConfigs() {
		mncs = append(mncs, &MetricNamespaceConfigModel{
			ID:       types.StringValue(mnc.GetId()),
			Disabled: types.BoolValue(mnc.GetDisabled()),
		})
	}
	state.MetricNamespaceConfigs, _ = types.SetValueFrom(ctx, MetricNamespaceConfigSpec, mncs)
```

Alternatively, you could create a helper for these.
```
func setFrom[I, O any](ctx context.Context, ins []I, typ types.ObjectType, conv func(I) O) types.Set {
	outs := make([]O, 0, len(ins))
	for _, in := range ins {
		outs = append(outs, conv(in))
	}
	res, _ := types.SetValueFrom(ctx, typ, outs)
	return res
}
```

ash-ddog: Out of scope, but we may also want to follow up with changing what we do for `accountTags` so we're consistent across all set types.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -401,13 +404,18 @@ func (r *integrationGcpStsResource) buildIntegrationGcpStsRequestBody(ctx contex
 	attributes.SetAccountTags(tfCollectionToSlice[string](ctx, diags, state.AccountTags))
 
 	mncs := make([]datadogV2.GCPMetricNamespaceConfig, 0)
-	for _, mnc := range state.MetricNamespaceConfigs {
+	for _, mnc := range tfCollectionToSlice[*MetricNamespaceConfigModel](ctx, diags, state.MetricNamespaceConfigs) {
 		mncs = append(mncs, datadogV2.GCPMetricNamespaceConfig{
 			Id:       mnc.ID.ValueStringPointer(),
 			Disabled: mnc.Disabled.ValueBoolPointer(),
 		})
 	}
```
ash-ddog: Let's only build if we're gonna use it:

```
	if cfgs := state.MetricNamespaceConfigs; !cfgs.IsUnknown() {
		mncs := make([]datadogV2.GCPMetricNamespaceConfig, 0)
		for _, mnc := range tfCollectionToSlice[*MetricNamespaceConfigModel](ctx, diags, cfgs) {
			mncs = append(mncs, datadogV2.GCPMetricNamespaceConfig{
				Id:       mnc.ID.ValueStringPointer(),
				Disabled: mnc.Disabled.ValueBoolPointer(),
			})
		}
		attributes.SetMetricNamespaceConfigs(mncs)
	}
```

ash-ddog: Out of scope: we should follow up with doing the same for other sets (account tags, cloudRunFilters, hostFilters, monitoredResourceConfigs)

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
 )
 
+func TestAccIntegrationGCPMcs(t *testing.T) {
```
ash-ddog: Can we rename this test to be a little clearer? something like 

`TestGCPSTSIntegration_MetricNamespaceConfigs`

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
...
+	t.Parallel()
+	ctx, providers, accProviders := testAccFrameworkMuxProviders(context.Background(), t)
+	uniq := uniqueEntityName(ctx, t)
+
+	// this test the setting of mcs (metric_namespace_configs), which can be updated to have a prometheus default if the user omits the field
```
ash-ddog: Would remove this comment, shouldn't this be clear from the naming of tests? 🤔

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
...
+	t.Run("1. user creates with the field omitted, and updates with the field omitted: both results should have the prometheus entry", func(t *testing.T) {
```
ash-ddog: Would remove the numbering in these, doesn't matter?

May also reword in accordance with  given/when/then`

maybe something like:
`when metric namespace configs never specified, uses defaults`
`when metric namespace configs specified on create, overrides defaults`
`when metric namespace configs specified on update, starts with defaults then overrides`

Feel free to modify!

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
...
+								"id":       knownvalue.StringExact("prometheus"),
+								"disabled": knownvalue.Bool(true),
+							}),
+						})),
+					},
+				},
+				{
+					Config: fmt.Sprintf(`
+					resource "datadog_integration_gcp_sts" "foo" {
+					  client_email                          = "%s@test-project.iam.gserviceaccount.com"
```
ash-ddog, line 46: Can we add _some_ update here? (we don't have to assert on it) I'm a bit worried this will result in a no-op but the HCL is identical.

ash-ddog (changes requested):
Some small clean up things. I think the only thing actually blocking is https://github.com/DataDog/terraform-provider-datadog/pull/3297/files#r2469956327

Let me know if you'd like to talk about anything! 🙏

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
katherinekim-51: yerr can cut a ticket/do a follow up for the various out-of-scope TF improvements we discussed!

datadog/tests/resource_datadog_integration_gcp_sts_test.go
katherinekim-51: changing to:
`when mrcs never specified in create and update, uses default`
`when mrcs never specified in create, but specified on update, overwrites default`
`when mrcs specified in create, but omitted on update, initial configs kept`

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
 )
 
+func TestAccIntGcpStsMetricNamespaceConfigs(t *testing.T) {
+	t.Parallel()
+	ctx, providers, accProviders := testAccFrameworkMuxProviders(context.Background(), t)
+	uniq := uniqueEntityName(ctx, t)
+
+	t.Run("when mrcs never specified in create and update, uses default", func(t *testing.T) {
```
ash-ddog, line 23: Extreme nit:
`when mrcs not specified in create nor update, uses default`

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -15,6 +15,142 @@ import (
...
+								"disabled": knownvalue.Bool(true),
+							}),
+						})),
+					},
+				},
+			},
+		})
+	})
+
+	t.Run("when mrcs never specified in create, but specified on update, overwrites default", func(t *testing.T) {
```
ash-ddog, line 63: Extreme nit:
`when mrcs not specified in create, but is specified on update, overwrites default`

PR #3299, opened 2025-10-24
[datadog_integration_gcp_sts] Add granular namespace filters to GCP V2 Terraform Provider

ash-ddog:
https://datadoghq.atlassian.net/browse/GCP-2929

Related:
https://github.com/DataDog/datadog-api-spec/pull/4510
https://github.com/DataDog/dd-go/pull/203050

This PR adds support for granular namespace filters to the GCP Integrations V2 terraform provider.

Should be straight-forward, but please take note of this comment, as I could potentially see it as a blocker (although IMO I do not think it's fixable): https://github.com/DataDog/terraform-provider-datadog/pull/3299#discussion_r2478643705
I have marked the PR with label `breaking-change` bc of this

Also please note **we should not merge this** until we fix some of [our broken metric names](https://datadoghq.atlassian.net/browse/GCP-2951) which is slated to occur on Nov 12 @ 1PM

### DIFF
--- a/docs/resources/integration_gcp_sts.md
+++ b/docs/resources/integration_gcp_sts.md
@@ -70,6 +70,7 @@ resource "datadog_integration_gcp_sts" "foo" {
 Optional:
 
 - `disabled` (Boolean)
+- `filters` (Set of String)
 - `id` (String)
 
 


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/terraform-providers/terraform-provider-datadog
 
 require (
-	github.com/DataDog/datadog-api-client-go/v2 v2.47.1-0.20251014135845-22ba517ffd6e
+	github.com/DataDog/datadog-api-client-go/v2 v2.47.1-0.20251022152720-0c92261b1332
```
ash-ddog: This is temporary, will go away when we upgrade to a version containing these changes:

https://github.com/DataDog/datadog-api-spec/pull/4510

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -261,29 +264,26 @@ func (r *integrationGcpStsResource) Create(ctx context.Context, request resource
...
 
 	// Save data into Terraform state
-	response.Diagnostics.Append(response.State.Set(ctx, &state)...)
+	response.Diagnostics.Append(response.State.Set(ctx, &plan)...)
 }
 
 func (r *integrationGcpStsResource) Update(ctx context.Context, request resource.UpdateRequest, response *resource.UpdateResponse) {
-	var state integrationGcpStsModel
-
-	var priorState integrationGcpStsModel
```
ash-ddog: Removed `priorState` since it was unused. Granted it was checking the prior state was deserializable, but still think this is the right move.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -26,6 +26,9 @@ var (
 		AttrTypes: map[string]attr.Type{
 			"id":       types.StringType,
 			"disabled": types.BoolType,
+			"filters": types.SetType{
+				ElemType: types.StringType,
+			},
```
ash-ddog, line 31: Just want to highlight that while this field is actually optional in terms of our API, it's **required** for terraform and I don't think we have a choice here 😞 .

From my understanding/research, this has to do with terraform provider framework v5 vs v6. In v5, you cannot have a definition for nested types, but you can in v6. Unfortunately, Datadog is stuck on v5 until every team updates to v6 and we stop using the terraform "mux" server.

So, ideally, our schema definition would have:
```
			"metric_namespace_configs": schema.SetNestedAttribute{
				Optional:    true,
				Computed:    true,
				Description: "Configurations for GCP metric namespaces.",
				NestedObject: schema.NestedAttributeObject{
						"id": schema.StringAttribute{
							Required:    true,
							Description: "The namespace ID.",
						},
						"disabled": schema.BoolAttribute{
							Optional:    true,
							Description: "Whether the namespace is disabled.",
						},
						"filters": schema.SetAttribute{
							Optional:    true,
							Description: "Filters for the namespace.",
							ElementType: types.StringType,
						},
					},
				},
			},
```

But this fails our tests with a message like "attributes are not allowed in v5" or something of that sort. Note that our code is actually using v6, but the terraform "mux" server verifies our v6 code is backwards-compatible with v5.

Functionally, what this means is that for customers that currently have `metric_namespace_configs` defined, once they update their terraform version, they'll have to manually add `filters = []` to all of their objects.

katherinekim-51: 👍 aligns with https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol#protocol-version-6 and this seems like the best case thinking of other options (ex: deprecate this field and make a new block-based one,  make filters a top-level attribute, etc) 

> Unfortunately, Datadog is stuck on v5 until every team updates to v6 and we stop using the terraform "mux" server

what's the/is there a timeline for this OOC?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -216,12 +222,11 @@ func (r *integrationGcpStsResource) Read(ctx context.Context, request resource.R
 }
 
 func (r *integrationGcpStsResource) Create(ctx context.Context, request resource.CreateRequest, response *resource.CreateResponse) {
-	var state integrationGcpStsModel
-	response.Diagnostics.Append(request.Plan.Get(ctx, &state)...)
+	var plan integrationGcpStsModel
```
katherinekim-51, line 225: 🎉

PR #3317, opened 2025-11-03
[datadog_reference_table] Add terraform support for reference tables

g-brizolier:
## What & Why

Adds Terraform support for Datadog Reference Tables, enabling users to manage Reference Tables as Infrastructure-as-Code.

**Includes:**

- `datadog_reference_table` resource - CRUD operations for Reference Tables
- `datadog_reference_table` data source - Query tables by ID or name
- `datadog_reference_table_rows` data source - Fetch specific rows by ID
- Schema validation via `ModifyPlan` to prevent destructive changes (only additive schema changes allowed)
- Support for cloud storage sources (AWS S3, GCP GCS, Azure Blob)

**Note:** Depends on OpenAPI spec fix ([DataDog/datadog-api-spec#4532](https://github.com/DataDog/datadog-api-spec/pull/4532)) to resolve OneOf unmarshaling for `file_metadata`.

### Run Acceptance Tests locally (they also run in CI)

```bash
cd ~/dd/terraform-provider-datadog && go build ./... && eval $(dd-auth --domain dd.datad0g.com --output) \
&& export DD_TEST_CLIENT_API_KEY="$DD_API_KEY" \
&& export DD_TEST_CLIENT_APP_KEY="$DD_APP_KEY" \
&& export DD_TEST_SITE_URL="https://api.datad0g.com" \
&& export DD_HOST="https://api.datad0g.com" \
&& export DD_TEST_ORG="yB5yjZ" \
&& TF_ACC=1 go test -v ./datadog/tests/... -run "TestAccReferenceTable"
```

Tests cover:

- Resource CRUD (create, update, delete)
- Schema validation (additive vs destructive changes)
- Cloud storage sources (S3, GCS, Azure)
- Data source queries (by ID, by name, rows)
- Import functionality

### Manual Testing (Optional)

1. Build provider: `make`
2. Setup `~/.terraformrc` with dev_overrides pointing to `$GOPATH/bin`
3. Create test HCL in a separate directory
4. Use `dd-auth --domain dd.datad0g.com --api-key-env=TF_VAR_datadog_api_key --app-key-env=TF_VAR_datadog_app_key terraform plan`

## References

- [Terraform Provider Development Docs](https://datadoghq.atlassian.net/wiki/spaces/API/pages/3881239106/Getting+Started+-+Terraform+Provider)
- [dd-auth CLI Documentation](https://datadoghq.atlassian.net/wiki/spaces/COBS/pages/4094591351/Datadog+Authentication+CLI+dd-auth)

### DIFF
--- a/datadog/fwprovider/data_source_datadog_reference_table.go
+++ b/datadog/fwprovider/data_source_datadog_reference_table.go
@@ -141,7 +141,7 @@ func (d *datadogReferenceTableDataSource) Schema(_ context.Context, _ datasource
 									},
 									"file_path": schema.StringAttribute{
 										Computed:    true,
-										Description: "The relative file path from the AWS S3 bucket root.",
+										Description: "The relative file path from the AWS S3 bucket root to the CSV file.",
 									},
 								},
 							},
@@ -158,11 +158,11 @@ func (d *datadogReferenceTableDataSource) Schema(_ context.Context, _ datasource
 									},
 									"file_path": schema.StringAttribute{
 										Computed:    true,
-										Description: "The relative file path from the GCS bucket root.",
+										Description: "The relative file path from the GCS bucket root to the CSV file.",
 									},
 									"gcp_service_account_email": schema.StringAttribute{
 										Computed:    true,
-										Description: "The email of the GCP service account.",
+										Description: "The email of the GCP service account used to access the bucket.",
 									},
 								},
 							},
@@ -175,7 +175,7 @@ func (d *datadogReferenceTableDataSource) Schema(_ context.Context, _ datasource
 									},
 									"azure_client_id": schema.StringAttribute{
 										Computed:    true,
-										Description: "The Azure client ID.",
+										Description: "The Azure client ID (application ID).",
 									},
 									"azure_storage_account_name": schema.StringAttribute{
 										Computed:    true,
@@ -187,7 +187,7 @@ func (d *datadogReferenceTableDataSource) Schema(_ context.Context, _ datasource
 									},
 									"file_path": schema.StringAttribute{
 										Computed:    true,
-										Description: "The relative file path from the Azure container root.",
+										Description: "The relative file path from the Azure container root to the CSV file.",
 									},
 								},
 							},
--- a/datadog/fwprovider/resource_datadog_reference_table.go
+++ b/datadog/fwprovider/resource_datadog_reference_table.go
@@ -152,7 +152,7 @@ func (r *referenceTableResource) Schema(_ context.Context, _ resource.SchemaRequ
 									},
 									"aws_bucket_name": schema.StringAttribute{
 										Optional:    true,
-										Description: "The name of the Amazon S3 bucket.",
+										Description: "The name of the AWS S3 bucket.",
 									},
 									"file_path": schema.StringAttribute{
 										Optional:    true,
--- a/docs/data-sources/reference_table.md
+++ b/docs/data-sources/reference_table.md
@@ -60,7 +60,7 @@ Read-Only:
 
 - `aws_account_id` (String) The ID of the AWS account.
 - `aws_bucket_name` (String) The name of the AWS S3 bucket.
-- `file_path` (String) The relative file path from the AWS S3 bucket root.
+- `file_path` (String) The relative file path from the AWS S3 bucket root to the CSV file.
 
 
 <a id="nestedblock--file_metadata--access_details--azure_detail"></a>
@@ -68,11 +68,11 @@ Read-Only:
 
 Read-Only:
 
-- `azure_client_id` (String) The Azure client ID.
+- `azure_client_id` (String) The Azure client ID (application ID).
 - `azure_container_name` (String) The name of the Azure container.
 - `azure_storage_account_name` (String) The name of the Azure storage account.
 - `azure_tenant_id` (String) The ID of the Azure tenant.
-- `file_path` (String) The relative file path from the Azure container root.
+- `file_path` (String) The relative file path from the Azure container root to the CSV file.
 
 
 <a id="nestedblock--file_metadata--access_details--gcp_detail"></a>
@@ -80,10 +80,10 @@ Read-Only:
 
 Read-Only:
 
-- `file_path` (String) The relative file path from the GCS bucket root.
+- `file_path` (String) The relative file path from the GCS bucket root to the CSV file.
 - `gcp_bucket_name` (String) The name of the GCP bucket.
 - `gcp_project_id` (String) The ID of the GCP project.
-- `gcp_service_account_email` (String) The email of the GCP service account.
+- `gcp_service_account_email` (String) The email of the GCP service account used to access the bucket.
 
 
 
--- a/docs/resources/reference_table.md
+++ b/docs/resources/reference_table.md
@@ -119,7 +119,7 @@ Optional:
 Optional:
 
 - `aws_account_id` (String) The ID of the AWS account.
-- `aws_bucket_name` (String) The name of the Amazon S3 bucket.
+- `aws_bucket_name` (String) The name of the AWS S3 bucket.
 - `file_path` (String) The relative file path from the AWS S3 bucket root to the CSV file.
 
 


g-brizolier:
* **#3317** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3317?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

datadog/tests/cassettes/TestAccDatadogReferenceTableDataSource.freeze
```diff
@@ -0,0 +1 @@
+2025-11-05T18:58:39.163767-05:00
```
g-brizolier: Still need to record these - waiting for a change in the go client. Ignore for now

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,1099 @@
...
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	id := planState.ID.ValueString()
+
+	// If we're updating schema for a cloud file table, ensure the table is ready (not processing)
+	// This prevents race conditions where we try to update while the initial sync is still running
+	isUpdatingSchema := planState.Schema != nil
+	if isUpdatingSchema && currentState.Source.ValueString() != "LOCAL_FILE" {
```
g-brizolier: TODO: figure out if this is truly necessary

datadog/fwprovider/data_source_datadog_reference_table.go
```diff
@@ -0,0 +1,458 @@
...
+				fmt.Sprintf("Found %d reference tables with table_name='%s', expected exactly 1", len(ddResp.Data), tableName),
+			)
+			return
+		}
+
+		// Get full details using the ID from list response
+		tableId := ddResp.Data[0].GetId()
+		fullResp, _, err := d.Api.GetTable(d.Auth, tableId)
+		if err != nil {
+			response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error getting datadog reference table details"))
```
linus-sun-dd: nit: should we make the error messaging the same for getting a reference table via table_id or name? ("error getting datadog reference table" vs "error getting datadog reference table details")

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (d *datadogReferenceTableRowsDataSource) Read(ctx context.Context, request datasource.ReadRequest, response *datasource.ReadResponse) {
+	var state datadogReferenceTableRowsDataSourceModel
+	response.Diagnostics.Append(request.Config.Get(ctx, &state)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	tableId := state.TableId.ValueString()
```
linus-sun-dd, line 97: more broadly, are there any potential concerns with having inconsistent behavior around allowing to query for a table itself via name/uuid but only accepting uuid for rows?

docs/data-sources/reference_table.md
```diff
@@ -0,0 +1,105 @@
...
+- `updated_at` (String) The timestamp of the last update to the reference table in ISO 8601 format.
+
+<a id="nestedblock--file_metadata"></a>
+### Nested Schema for `file_metadata`
+
+Read-Only:
+
+- `access_details` (Block, Read-only) Cloud storage access configuration. Only present for cloud storage sources (S3, GCS, Azure). (see [below for nested schema](#nestedblock--file_metadata--access_details))
+- `error_message` (String) Error message from the last sync attempt, if any.
+- `error_row_count` (Number) The number of rows that failed to sync.
```
linus-sun-dd, line 43: nit: should we clarify error row messaging to include other potential error types? (eg. duplicate primary key, etc.)

linus-sun-dd:
image

i'm seeing `no tests to run` when executing the acceptance tests, is this expected?

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,165 @@
...
+	// Extract row IDs from the list
+	var rowIds []string
+	response.Diagnostics.Append(state.RowIds.ElementsAs(ctx, &rowIds, false)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	// Call API to get rows by ID with retry logic
+	// Rows are written asynchronously, so we need to retry if the table hasn't synced yet
+	// Use a 5-second interval to avoid spamming the API while waiting for sync
```
p-garratt: It looks like there's a util for this retry, is there a library with exponential backoff? Just worried about the traffic from a bunch of workers on a schedule.
Also, how does this fit into the rate limits for an API user? Could they use all their API traffic in terraform accidentally? Or will they likely get a separate API key for TF?

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,165 @@
...
+	var ddResp datadogV2.TableRowResourceArray
+	var httpResp *http.Response
+	var err error
+
+	retryErr := utils.Retry(5*time.Second, 10, func() error {
+		ddResp, httpResp, err = d.Api.GetRowsByID(d.Auth, tableId, rowIds)
+		if err != nil {
+			// If we get a 404, the table might not have synced yet - retry
+			if httpResp != nil && httpResp.StatusCode == 404 {
```
p-garratt, line 128: Optional: Could we fetch the table by id to check if it really exists so we don't spam the API if the table is already deleted or something? Maybe check the status to differentiate if the rows may still be processing or if the ids are just wrong below?

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Call API to get rows by ID with retry logic
+	// Rows are written asynchronously, so we need to retry if the table hasn't synced yet
+	// Use a 5-second interval to avoid spamming the API while waiting for sync
+	var ddResp datadogV2.TableRowResourceArray
+	var httpResp *http.Response
+	var err error
+
+	retryErr := utils.Retry(5*time.Second, 10, func() error {
+		ddResp, httpResp, err = d.Api.GetRowsByID(d.Auth, tableId, rowIds)
```
p-garratt, line 125: This method has a limit of 250, should we enforce that in the TF code somewhere?

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+			// For other errors, don't retry
+			return &utils.FatalError{Prob: fmt.Sprintf("error getting reference table rows: %v", err)}
+		}
+		// Success - check if we got the expected number of rows
+		if len(ddResp.Data) == len(rowIds) {
+			return nil
+		}
+		// If we got some rows but not all, the table might still be syncing - retry
+		if len(ddResp.Data) > 0 && len(ddResp.Data) < len(rowIds) {
```
p-garratt, line 139: This also happens if some of the ids are not found.

datadog/fwprovider/data_source_datadog_reference_table.go
```diff
@@ -0,0 +1,458 @@
...
+		},
+		Blocks: map[string]schema.Block{
+			"file_metadata": schema.SingleNestedBlock{
+				Description: "File metadata for the reference table. Contains sync settings for cloud storage sources.",
+				Attributes: map[string]schema.Attribute{
+					"sync_enabled": schema.BoolAttribute{
+						Computed:    true,
+						Description: "Whether automatic sync is enabled for this table. Only present for cloud storage sources (S3, GCS, Azure).",
+					},
+					"error_message": schema.StringAttribute{
```
p-garratt, line 114: My TF knowledge is falling short here. For these computed fields, what happens if they change between syncs? Like TF is trying to enforce that infra only changes through code, does it store previous runs? This just seems like a field that could easily change between syncs.

I don't have a good mental model of TF here, so might be totally off on whether this matters.

datadog/fwprovider/data_source_datadog_reference_table.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+func (d *datadogReferenceTableDataSource) updateState(ctx context.Context, state *datadogReferenceTableDataSourceModel, referenceTableData *datadogV2.TableResultV2Data) {
+	attributes := referenceTableData.GetAttributes()
+
+	// Set basic attributes
+	state.ID = types.StringValue(referenceTableData.GetId())
+	state.TableName = types.StringValue(attributes.GetTableName())
+
+	if createdBy, ok := attributes.GetCreatedByOk(); ok {
```
p-garratt, line 317: Does having these fields in this method mean they're editable? Because some of them are, but not all of them.

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+					"error_row_count": schema.Int64Attribute{
+						Computed:    true,
+						Description: "The number of rows that failed to sync.",
+					},
+					"error_type": schema.StringAttribute{
+						Computed:    true,
+						Description: "The type of error that occurred during file processing.",
+					},
+				},
+				Blocks: map[string]schema.Block{
```
p-garratt, line 142: Is it worth having any validation on this? For example: that only one type of access_details is specified, that it matches the SOURCE

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+			"id": utils.ResourceIDAttribute(),
+			"table_name": schema.StringAttribute{
+				Required:    true,
+				Description: "The name of the reference table. This must be unique within your organization.",
+			},
+			"source": schema.StringAttribute{
+				Required:    true,
+				Description: "The source type for the reference table. Must be one of: S3, GCS, AZURE.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("S3", "GCS", "AZURE"),
```
p-garratt, line 87: Didn't you have a comment in data_source_datadog_reference_table.go that folks could represent SaaS Ints tables with this framework? Does that need to be reflected here? Or is this only for creating / editing tables?

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+									Description: "The name of the field.",
+									PlanModifiers: []planmodifier.String{
+										stringplanmodifier.UseStateForUnknown(),
+									},
+								},
+								"type": schema.StringAttribute{
+									Optional:    true,
+									Computed:    true,
+									Description: "The data type of the field. Must be one of: STRING, INT32.",
+									Validators: []validator.String{
```
p-garratt, line 255: Is there a way to add an empty check? Maybe here and for primary key too

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+	if httpResp == nil || httpResp.StatusCode != 201 {
+		statusCode := 0
+		if httpResp != nil {
+			statusCode = httpResp.StatusCode
+		}
+		response.Diagnostics.AddError("API Error", fmt.Sprintf("CreateReferenceTable returned unexpected status (HTTP %d).", statusCode))
+		return
+	}
+
+	// List tables with exact name filter to find the created one
```
p-garratt, line 457: Would it make sense for the create API to return the table_id of the table it created?

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+		cloudStorageMetadata := datadogV2.CreateTableRequestDataAttributesFileMetadataCloudStorage{}
+		cloudStorageMetadata.SetSyncEnabled(state.FileMetadata.SyncEnabled.ValueBool())
+
+		if state.FileMetadata.AccessDetails != nil {
+			accessDetails := datadogV2.CreateTableRequestDataAttributesFileMetadataOneOfAccessDetails{}
+
+			// AWS details
+			if state.FileMetadata.AccessDetails.AwsDetail != nil {
+				awsDetail := datadogV2.CreateTableRequestDataAttributesFileMetadataOneOfAccessDetailsAwsDetail{}
+				awsDetail.SetAwsAccountId(state.FileMetadata.AccessDetails.AwsDetail.AwsAccountId.ValueString())
```
p-garratt: Super nit, optional: consider using a local variable for `state.FileMetadata.AccessDetails.AwsDetail` to make this more readable.

Ditto below

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,746 @@
...
+		attributes.SetDescription(planState.Description.ValueString())
+	}
+
+	if !planState.Tags.IsNull() {
+		var tags []string
+		diags.Append(planState.Tags.ElementsAs(ctx, &tags, false)...)
+		attributes.SetTags(tags)
+	}
+
+	// Note: Schema updates are not supported via PATCH - schema is only set on create.
```
p-garratt, line 812: Should TF return an error if it detects schema updates were attempted?

docs/data-sources/reference_table.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+<a id="nestedblock--schema"></a>
+### Nested Schema for `schema`
+
+Read-Only:
+
+- `fields` (Block List) List of fields in the table schema. (see [below for nested schema](#nestedblock--schema--fields))
+- `primary_keys` (List of String) List of field names that serve as primary keys for the table.
```
p-garratt, line 97: Should probably have some comment about how the primary_key must be in fields. It's pretty obvious, but good to be explicit about how these fields relate.

examples/resources/datadog_reference_table/resource.tf
```diff
@@ -0,0 +1,48 @@
...
+      type = "INT32"
+    }
+  }
+
+  tags = [
+    "source:s3",
+    "team:catalog",
+    "env:production"
+  ]
+}
```
p-garratt: nit: no newline here, but there's one in the other files.

examples/resources/datadog_reference_table/azure.tf
```diff
@@ -0,0 +1,47 @@
...
+      type = "INT32"
+    }
+
+    fields {
+      name = "status"
+      type = "STRING"
+    }
+  }
+
+  tags = ["source:azure", "team:warehouse"]
```
p-garratt: Very optional: include an example with no tags?

docs/data-sources/reference_table.md
```diff
@@ -0,0 +1,105 @@
...
+- `error_row_count` (Number) The number of rows that failed to sync.
+- `error_type` (String) The type of error that occurred during file processing. Only present for cloud storage sources.
+- `sync_enabled` (Boolean) Whether automatic sync is enabled for this table. Only present for cloud storage sources (S3, GCS, Azure).
+
+<a id="nestedblock--file_metadata--access_details"></a>
+### Nested Schema for `file_metadata.access_details`
+
+Read-Only:
+
+- `aws_detail` (Block, Read-only) AWS S3 access configuration. (see [below for nested schema](#nestedblock--file_metadata--access_details--aws_detail))
```
p-garratt, line 52: Should these docs link to some shared docs about how to grant read access to us for this bucket?

g-brizolier:
Addressed review comments:
- Unified error messaging for reference table lookups
- Added 250 row limit validation with clear error message
- Added ValidateConfig to enforce: exactly one access_details type, matching source type, non-empty schema fields and primary_keys
- Improved readability with local variables for AWS/GCP/Azure details
- Fixed trailing newlines in examples

datadog/fwprovider/data_source_datadog_reference_table.go
g-brizolier: Fixed - unified to 'error getting reference table'

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Good point. The rows data source requires table_id because the GetRowsByID API endpoint requires a UUID. We could add a lookup-by-name step, but it adds latency. I'll add a note in the docs clarifying this.

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Good point; the GetRowsByID endpoint already does a table existence check and returns a 404 early if it isn't there

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Added validation - now returns clear error if >250 IDs provided.

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Good catch - updated the comment. The retry handles both cases: table still syncing (404) and partial row matches.

datadog/fwprovider/data_source_datadog_reference_table.go
g-brizolier: Correct - for data sources, Computed fields are read-only and reflect API state. TF doesn't enforce drift on data sources; they're always refreshed on each plan/apply. Changes between syncs are expected and reflected in the data source output.

datadog/fwprovider/data_source_datadog_reference_table.go
g-brizolier: This is the data source - all fields are read-only (Computed). The resource (resource_datadog_reference_table.go) has the editable fields marked as Optional.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Added ValidateConfig that enforces: exactly one access_details type specified, and it must match the source type (S3→aws_detail, GCS→gcp_detail, AZURE→azure_detail).

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: The data source supports reading any table type (including SaaS integrations). The resource only supports creating cloud storage tables (S3/GCS/Azure) since those are the only ones that can be created via API.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Added in ValidateConfig - now validates that schema.fields and schema.primary_keys are not empty.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Yes, that would be ideal. Currently the API returns 201 with empty body. I've filed a ticket to return the table_id in the response.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Done - used local variables (aws, gcp, azure) to improve readability.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Schema updates via PATCH aren't supported - schema is only set on create. TF will show a plan diff but the PATCH won't include schema. Added UseStateForUnknown to prevent drift detection on schema fields.

docs/data-sources/reference_table.md
g-brizolier: Good point - will update the description to include common error types (duplicate primary key, invalid format, etc.).

examples/resources/datadog_reference_table/resource.tf
g-brizolier: Fixed - added newline to resource.tf.

docs/data-sources/reference_table.md
g-brizolier: Good idea - I'll add a link to the cloud storage integration docs in a follow-up.

docs/data-sources/reference_table.md
g-brizolier: Good point - the primary_key values must match field names in the fields list. I'll add a note to the docs.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Added ModifyPlan validation - TF now returns a clear error if the user attempts to modify the schema after creation: 'Reference table schema cannot be modified after creation. The schema is derived from the CSV file in cloud storage. To change the schema, update the CSV file and the table will sync automatically if sync_enabled is true.'

docs/data-sources/reference_table.md
g-brizolier: Added! The resource description now links to the official docs: https://docs.datadoghq.com/reference_tables/?tab=cloudstorage#create-a-reference-table which includes setup instructions for granting Datadog read access to cloud storage buckets.

examples/resources/datadog_reference_table/azure.tf
g-brizolier: Done - removed tags from azure.tf to show an example without tags.

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Good catch! Switched from `utils.Retry` (linear retry for tests) to `retry.RetryContext` from the Terraform SDK which implements proper exponential backoff.

Regarding rate limits: The SDK's retry uses the standard provider `http_client_retry_backoff_multiplier` and `http_client_retry_backoff_base` settings. The Datadog API client also has built-in rate limit handling (429 responses trigger backoff). Users typically use a dedicated API key for Terraform, so this shouldn't impact their other usage.

datadog/fwprovider/data_source_datadog_reference_table_rows.go
```diff
@@ -0,0 +1,176 @@
...
+	// Uses the Terraform SDK retry helper which implements exponential backoff
+	var ddResp datadogV2.TableRowResourceArray
+	var httpResp *http.Response
+	var err error
+
+	// Retry for up to 1 minute with exponential backoff (managed by the SDK)
+	retryErr := retry.RetryContext(ctx, 1*time.Minute, func() *retry.RetryError {
+		ddResp, httpResp, err = d.Api.GetRowsByID(d.Auth, tableId, rowIds)
+		if err != nil {
+			// If we get a 404, the table might not have synced yet - retry
```
p-garratt, line 127: No changes needed, but man will it be nice when we have read after write consistency and this is no longer needed.

None of this retry will be needed because we'll get the rows written synchronously too

datadog/fwprovider/data_source_datadog_reference_table_rows.go
p-garratt: The customer behavior on partial row matches will be somewhat frustrating, because they will get an error rather than data and they won't know which row ids are wrong.
Is there a way to return the partial rows if we've exhausted the retries?

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,890 @@
...
+				if state.Schema.Fields[i].Name.ValueString() != plan.Schema.Fields[i].Name.ValueString() ||
+					state.Schema.Fields[i].Type.ValueString() != plan.Schema.Fields[i].Type.ValueString() {
+					schemaChanged = true
+					break
+				}
+			}
+		}
+
+		if schemaChanged {
+			response.Diagnostics.AddError(
```
p-garratt, line 391: Nit: There's two types of changes you're checking here: that the primary keys didn't change and that the fields didn't change. Primary key changes will never be supported, schema modification is supported through the CSV actions described below. The error message for primary key changes should probably be different from general schema changes.

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,890 @@
...
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	// Check if schema is being modified on an existing resource
+	if state.Schema != nil && plan.Schema != nil {
+		schemaChanged := false
+
+		// Compare primary keys
+		var statePKs, planPKs []string
```
p-garratt, line 364: Can you just use slices.Equal()?

datadog/fwprovider/resource_datadog_reference_table.go
```diff
@@ -0,0 +1,890 @@
...
+				response.Diagnostics.AddError(
+					"Invalid access_details for source",
+					"Source 'AZURE' requires azure_detail in access_details.",
+				)
+			}
+		}
+	}
+
+	// Validate schema fields are not empty
+	if config.Schema != nil {
```
p-garratt: Can you use [listvalidator.SizeAtLeast()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.19.0/listvalidator#SizeAtLeast) here?
No strong preference on the impl here, just seems like using the validators that exist is useful

datadog/tests/resource_datadog_reference_table_test.go
p-garratt, line 1: I might of just missed it, but is there a set of tests for the validation logic? Would probably be good to document that.

docs/data-sources/reference_table.md
```diff
@@ -0,0 +1,105 @@
...
+- `gcp_detail` (Block, Read-only) Google Cloud Storage access configuration. (see [below for nested schema](#nestedblock--file_metadata--access_details--gcp_detail))
+
+<a id="nestedblock--file_metadata--access_details--aws_detail"></a>
+### Nested Schema for `file_metadata.access_details.aws_detail`
+
+Read-Only:
+
+- `aws_account_id` (String) The ID of the AWS account.
+- `aws_bucket_name` (String) The name of the AWS S3 bucket.
+- `file_path` (String) The relative file path from the S3 bucket root.
```
p-garratt:
```suggestion
- `file_path` (String) The relative file path from the AWS S3 bucket root.
```
Optional: but the line above says "AWS S3" so probably good to be consistent.

docs/data-sources/reference_table_rows.md
```diff
@@ -0,0 +1,33 @@
...
+
+- `rows` (Block List) List of retrieved rows. Each row contains its ID and field values. (see [below for nested schema](#nestedblock--rows))
+
+<a id="nestedblock--rows"></a>
+### Nested Schema for `rows`
+
+Read-Only:
+
+- `id` (String) The primary key value of the row.
+- `values` (Map of String) Map of field names to values for this row. All values are returned as strings.
```
p-garratt, line 33: > All values are returned as strings.

In a future revision, would it make sense to handle types in rows, once we support more row types?

docs/resources/reference_table.md
```diff
@@ -0,0 +1,214 @@
...
+  ]
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `source` (String) The source type for the reference table. Must be one of: S3, GCS, AZURE. Valid values are `S3`, `GCS`, `AZURE`.
```
p-garratt:
```suggestion
- `source` (String) The source type for the reference table. Valid values are `S3`, `GCS`, `AZURE`.
```

docs/resources/reference_table.md
```diff
@@ -0,0 +1,214 @@
...
+- `description` (String) The description of the reference table.
+- `file_metadata` (Block, Optional) Configuration for cloud storage file access and sync settings. (see [below for nested schema](#nestedblock--file_metadata))
+- `schema` (Block, Optional) The schema definition for the reference table, including field definitions and primary keys. Schema is only set on create; updates are derived from the file asynchronously. (see [below for nested schema](#nestedblock--schema))
+- `tags` (List of String) A list of tags to associate with the reference table.
+
+### Read-Only
+
+- `created_by` (String) UUID of the user who created the reference table.
+- `id` (String) The ID of this resource.
+- `last_updated_by` (String) UUID of the user who last updated the reference table.
```
p-garratt, line 85: Just out of curiosity, when they use the API key via TF, will this field have the service account or just be empty?

docs/data-sources/reference_table.md
p-garratt: If you make this change, ditto in the docs/resources/reference_table.md file too

datadog/fwprovider/data_source_datadog_reference_table_rows.go
g-brizolier: Done - now returns partial results with a warning instead of failing. Users will see which rows were found and a message indicating some IDs may be invalid or not yet synced.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Good point - now returns separate error messages: primary key changes get "Primary key modification not supported" with guidance to recreate the table, while field changes get "Schema field modification not supported" with guidance to update the CSV file.

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Done - switched to slices.Equal().

datadog/fwprovider/resource_datadog_reference_table.go
g-brizolier: Good suggestion - added listvalidator.SizeAtLeast(1) to both primary_keys and fields in the schema definition. Removed the manual validation from ValidateConfig.

datadog/tests/resource_datadog_reference_table_test.go
g-brizolier: Validation is covered by the existing acceptance tests - ValidateConfig and ModifyPlan validations trigger errors that cause plan/apply to fail. The tests verify these errors are raised correctly when invalid configurations are provided. Specific tests: TestAccReferenceTableS3_Basic covers access_details validation, TestAccReferenceTable_SchemaOnCreate covers schema validation.

docs/data-sources/reference_table.md
g-brizolier: Done - updated to "AWS S3 bucket" for consistency.

docs/data-sources/reference_table_rows.md
g-brizolier: Yes, once we support additional row types beyond strings, we should add proper type handling. For now strings cover all current use cases.

docs/resources/reference_table.md
g-brizolier: Fixed - simplified the description. The doc generator auto-appends the valid values from the validator.

docs/resources/reference_table.md
g-brizolier: The created_by field will contain the UUID of the service account associated with the API key. It won't be empty - every API request is attributed to either a user or service account.

docs/data-sources/reference_table.md
g-brizolier: Done - updated both data source and resource docs.

g-brizolier:
@Cursor review

PR #3318, opened 2025-11-04
[datadog_monitor] Make cloud cost query `aggregator` field required

fpighi:
This field is required by the API. This PR updates the schema so it is consistent with that.

Previous behavior:

```
╷
│ Error: error validating monitor from /api/v1/monitor/validate: 400 Bad Request: {"errors":["'aggregator' is a required property"]}
│
│   with datadog_monitor.Cloud_Cost_monitor_terraform_export,
│   on main.tf line 232, in resource "datadog_monitor" "Cloud_Cost_monitor":
│  232: resource "datadog_monitor" "Cloud_Cost_monitor_terraform_export" {
│
╵
```

New behavior:

```
╷
│ Error: Missing required argument
│
│   on main.tf line 238, in resource "datadog_monitor" "Cloud_Cost_monitor":
│  238:     cloud_cost_query {
│
│ The argument "aggregator" is required, but no definition was found.
```

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -581,7 +581,7 @@ func getMonitorFormulaQuerySchema() *schema.Schema {
 							},
 							"aggregator": {
 								Type:             schema.TypeString,
-								Optional:         true,
+								Required:         true,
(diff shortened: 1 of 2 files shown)


rtrieu (approved):

fpighi:
> Looks good! Just want to check that changing from optional to required doesn't break any existing monitors?

It shouldn't because the validation api call would otherwise fail anyway.

PR #3323, opened 2025-11-06
[datadog_user] Add new exclude_service_accounts attribute to the datadog_user ds

tquidetdd:
## Description

This PR adds a new `exclude_service_accounts` attribute to the `datadog_user` data source. 
This boolean option allows users to filter out service accounts from user search results, making it easier to query for regular user accounts without having to manually filter by using the `dataodg_users` data source and then manually exclude service accounts.

**Motivation**: When querying users by email, service accounts with matching attributes cause an error `your query returned more than one result for filter`.
This new attribute provides a clean way to exclude service accounts directly in the data source query.

This is useful in the context of the Datadog On-call product, since we often need to retrieve users to add them to a schedule, and we don’t want to include service accounts in an On-call schedule.

**Changes**:
- Added `exclude_service_accounts` boolean attribute to `datadog_user` data source (defaults to `false` for backward compatibility)
- Updated filtering logic to filter out service accounts when this option is enabled
- Enhanced error messages to indicate when errors occur "after excluding service accounts" for better debugging

## Testing

Added test coverage for the new functionality:

- `TestAccDatadogUserDatasourceWithExcludeServiceAccounts` - Tests successful filtering of service accounts
- `TestAccDatadogUserDatasourceWithExcludeServiceAccountsMultipleUsersWithError` - Tests error handling when multiple non-service-account users remain after filtering
- `TestAccDatadogUserDatasourceWithExcludeServiceAccountsWithError` - Tests error handling when users and service accounts remain

## What will it impact?

This is a new optional attribute with backward compatibility. Existing configurations will be unaffected as the default value is `false`. Users who enable this feature will see filtered results excluding service accounts.

### DIFF
--- a/docs/data-sources/user.md
+++ b/docs/data-sources/user.md
@@ -28,6 +28,7 @@ data "datadog_user" "test" {
 ### Optional
 
 - `exact_match` (Boolean) When true, `filter` string is exact matched against the user's `email`, followed by `name` attribute. Defaults to `false`.
+- `exclude_service_accounts` (Boolean) When true, service accounts are excluded from the result. Defaults to `false`.
 
 ### Read-Only
 


tquidetdd:
> lgtm
> 
> btw, is there some existing Github issue asking for this that we can link or this was requested via some other channels?

I don’t see any issues related to this.
We faced this problem when we started importing On-call team configurations into Terraform, and we decided to update the provider instead of creating a module.

Issue #2651, opened 2024-11-08
widget.toplist_definition.request.query.metric_query do not sort or group by


feoliveira-alelo:
### What resources or data sources are affected?

[widget.toplist_definition.request.query.metric_query](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/powerpack#nested-schema-for-widgettoplist_definitionrequestquerymetric_query)

### Feature Request

widget.toplist_definition.request.query.metric_query do not sort or group by

### References

[widget.toplist_definition.request.query.metric_query](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/powerpack#nested-schema-for-widgettoplist_definitionrequestquerymetric_query)

paleboot:
CMIIW, but I believe OP's issue is: "Unable to sort metric query in toplist definition"

If so, +1, we have several Top List widget in our dashboard and applying via the provider leaves them unsorted

rcasperson-jc:
+1 to this issue

It appears to be a gap in the DataDog Terraform provider that you cannot provide sorting information for a toplist widget like you can if you manually edit the widget.

Let's say I have a toplist widget with two queries, A and B, then a formula that compare A and B to get a percentage. Naturally I want to sort the widget by my formula, displaying the higher percentages at the top.

Via the UI, I can easily do that by changing the `Display only ` section. It appears to be **impossible to do that via terraform**.

When I manually edit via the UI, the JSON section includes something like this inside the `"request"` object:
```
            "sort": {
                "count": 200,
                "order_by": [
                    {
                        "type": "formula",
                        "index": 0,
                        "order": "asc"
                    }
                ]
            }
```

This critical to getting our dashboards working correctly.

dphilipsjc:
+1 to echo @rcasperson-jc 's comment about getting dashboards to work correctly.
Further to elaborate on Ryan's comment, when you open the widget in edit mode, it looks to be sorting as wanted (by the formula), but it is actually sorting by the other value.  This is apparent if you change the sorting to the other value manually, and then change it back to the formula.
So in addition to not allowing the sort specification, the UI is displaying misleading information.

PR #3326, opened 2025-11-12
[datadog_dashboard] Add support for sorting toplist widget.

ric-sigma-computing:
Fixes https://github.com/DataDog/terraform-provider-datadog/issues/2651

Add support for sorting toplist widgets in dashboards

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -8260,8 +8260,59 @@ func getToplistRequestSchema() map[string]*schema.Schema {
 				Schema: getWidgetRequestStyle(),
 			},
 		},
+		"sort": getToplistRequestSortSchema(),
 	}
 }
+
+func getToplistRequestSortSchema() *schema.Schema {
+	return &schema.Schema{
+		Description: "The controls for sorting the widget.",
+		Type:        schema.TypeList,
+		Optional:    true,
+		MaxItems:    1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"count": {
+					Description: "The number of items to limit the widget to.",
+					Type:        schema.TypeInt,
+					Optional:    true,
+				},
+				"order_by": {
+					Description: "The array of items to sort the widget by in order.",
+					Type:        schema.TypeList,
+					Optional:    true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"index": {
+								Description: "The index of the formula to sort by.",
+								Type:        schema.TypeInt,
+								Optional:    true,
+							},
+							"order": {
+								Description:      "Widget sorting methods. Allowed enum values: `asc`, `desc`",
+								Type:             schema.TypeString,
+								Required:         true,
+								ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewWidgetSortFromValue),
+							},
+							"type": {
+								Description:      "Set the sort type to formula. Allowed enum values: `formula`",
+								Type:             schema.TypeString,
+								Required:         true,
+								ValidateDiagFunc: validators.ValidateStringEnumValue(datadogV1.FORMULATYPE_FORMULA, datadogV1.GROUPTYPE_GROUP),
+							},
+							"name": {
+								Description: "The name of the group.",
+								Type:        schema.TypeString,
+								Optional:    true,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
 func buildDatadogToplistRequests(terraformRequests *[]interface{}) *[]datadogV1.ToplistWidgetRequest {
 	datadogRequests := make([]datadogV1.ToplistWidgetRequest, len(*terraformRequests))
 	for i, r := range *terraformRequests {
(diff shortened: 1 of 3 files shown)


estherk15 (approved):

ric-sigma-computing:
@estherk15 Thank you for reviewing - I see I also need a code review from a few other people before merging, is there anything else I need to do from my end? If it's in the review queue that's fine I understand people are busy with the holidays.

estherk15:
> @estherk15 Thank you for reviewing - I see I also need a code review from a few other people before merging, is there anything else I need to do from my end? If it's in the review queue that's fine I understand people are busy with the holidays.

Thanks for the follow up @ric-sigma-computing! Yes, the PR should be on each of the respective teams' queue (I reviewed for documentation). If you don't hear back after the holidays, I'm happy to help nudge the teams!

ric-sigma-computing:
@estherk15 Happy new year! Would you be able to help bump this review with the team? Thanks in advance

estherk15:
@ric-sigma-computing I reached out to the product team and they're reviewing this now!

datadog/resource_datadog_dashboard.go
```diff
@@ -8260,8 +8260,59 @@ func getToplistRequestSchema() map[string]*schema.Schema {
...
 		},
+		"sort": getToplistRequestSortSchema(),
 	}
 }
+
+func getToplistRequestSortSchema() *schema.Schema {
+	return &schema.Schema{
+		Description: "The controls for sorting the widget.",
+		Type:        schema.TypeList,
+		Optional:    true,
```
annagchang, line 8271: ℹ️ sort schema is optional, should be only additional non-breaking change

ric-sigma-computing:
> Reviewed this with @annagchang and while it works we have a few questions.
> 
> We created a dashboard with a top list widget using this PR ([dashboard](https://app.datadoghq.com/dashboard/w47-vza-ab5?fromUser=false&refresh_mode=sliding&tile_focus=1079888866581477&from_ts=1769024662604&to_ts=1769028262604&live=true))
> 
> We specified we wanted the sort to have a count of 5. The JSON looks correct - it has the `sort { count: 5 }` which is what we specified.
> 
> ```
> {
>     "title": "",
>     "type": "toplist",
>     "requests": [
>         {
>             "queries": [
>                 {
>                     "compute": {
>                         "aggregation": "count"
>                     },
>                     "data_source": "logs",
>                     "group_by": [
>                         {
>                             "facet": "service",
>                             "limit": 10,
>                             "sort": {
>                                 "aggregation": "count",
>                                 "metric": "count",
>                                 "order": "desc"
>                             }
>                         }
>                     ],
>                     "indexes": [
>                         "*"
>                     ],
>                     "name": "query1",
>                     "search": {
>                         "query": "env:prod"
>                     }
>                 }
>             ],
>             "response_format": "scalar",
>             "formulas": [
>                 {
>                     "formula": "query1"
>                 }
>             ],
>             "sort": {
>                 "count": 5
>             }
>         }
>     ],
>     "style": {}
> }
> ```
> 
> The widget visualization also looks correct - it only shows the top 5
> 
> Image
> However, the query editor is off:
> 
> Image
> The "display only" says "Top 10" and "count," but in the image, we are specifying the top 5, and it is using the `group by service` field to sort, not the count.
> 
> Is this expected? Also, do we intentionally allow for a different sort/count feature on the widget when the query editor already has one built in on the query?

@bonnierhee @annagchang Thank you for the detailed review and comments. I am able to reproduce the issue described using the following terraform code
```
resource "datadog_dashboard" "dashboard1" {
  title       = "testing"
  description = "testing"
  layout_type = "ordered"
  reflow_type = "auto"

  widget {
    toplist_definition {
      title       = ""
      title_size  = "16"
      title_align = "left"

      request {
        query {
          event_query {
            name        = "query1"
            data_source = "logs"
            indexes     = ["*"]
            search {
              query = "*"
            }
            compute {
              aggregation = "count"
            }
            group_by {
              facet = "service"
              limit = 10
              sort {
                aggregation = "count"
                metric      = "count"
                order       = "desc"
              }
            }
            storage = "hot"
          }
        }

        formula {
          formula_expression = "query1"
        }

        sort {
          count = 5
        }
      }

      style {
        display {
          type = "stacked"
        }
      }
    }
  }
}
```
If I change the `sort` block to include the `order_by` block, then the query editor shows the correct count and sort by values
```
        sort {
          count = 5
          order_by {
            type  = "formula"
            index = 0
            order = "desc"
          }
        }
```
Screenshot 2026-01-21 at 3 13 52 PM

When the `order_by` block is omitted, then the query editor shows the count=5 in the "json" tab, but not in the "edit" tab. However, the dashboard does respect the count and indeed only shows the top 5. 

I believe this may be due to a divergence in the data processing layer within datadog, the dashboard rendering pipeline respects the `count` attribute submitted by terraform, but the query editor tool does not unless both the `count` and `order_by` attributes are present. 

Here are some options I propose:
* I can add a default `order_by` within terraform when none is specified by the user (anti-pattern in terraform, adding unexpected state)
* I can make terraform throw an error if no `order_by` is specified by the user (not datadog api spec compliant)
* Leave as-is (my preference)

The reason for the last option is that this seems to be a larger issue than just the terraform code. Even in the datadog widget query editor, if the `order_by` attribute is removed the same behavior is seen. So the UI would need a fix to default or require the attribute in any case regardless of what is done here.

As for the last question, that may best be answered by datadog product team, but I would guess the query processing layer is separate from the visualization layer and it's a post-processing filter applied on top of the query results hence they are unaware of what sorting/filtering has already taken place.

ric-sigma-computing:
@bonnierhee Hello, just wanted to follow up and see if you had a chance to review my response. Thanks!

PR #3328, opened 2025-11-13
[datadog_agentless_scanning_gcp_scan_options] Better support of resource conflict in agentless GCP scan options TF provider

mohamed-challal:
[K9VULN-8133](https://datadoghq.atlassian.net/browse/K9VULN-8133)

This pull request fixes an edge case in the Datadog Agentless Scanning GCP Scan Options Terraform resource.

How agentless setup works for GCP?
-------------------------------------
To scan a GCP project using agentless scanning, the user needs to:
- Setup [the agentless module](https://github.com/DataDog/terraform-module-datadog-agentless-scanner/tree/main/gcp) (IAM & infra)
- Enable the product in the DD UI for the given project

What the Agentless GCP Scan Options TF provider is used for?
---------------------------------------------------------------
This resource wraps the call to `agentless-api` that automatically enables the product, and removes the need to interact with the UI.

What edge case does this PR fix?
----------------------------------
The agentless api follows [the json api recommendations](https://jsonapi.org/), meaning that if a resource (scan options for a project) already exists, it returns a `CONFLICT` error, in which case the user should use the `Update` route instead.
This behavior is already handled in the TF context by maintaining a state.
The edge case occurs when there are interactions with the UI in parallel to the TF setup, the state is not updated.
The fix of this PR is to check if the scan options exist (GET) before doing any resource creation or update.

Note: there can be a race condition if the resource is updated between the check and the write operation, but this is something we accept for now as the traffic on each resource is extremely low, and re-applying the TF solves this race condition.

Testing
--------
We have tests for the GCP scan options, covering creation, update, import, and destruction scenarios.
These changes have been validated on a dogfood org.

[K9VULN-8133]: https://datadoghq.atlassian.net/browse/K9VULN-8133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


diogocp:
> The edge case occurs when there are interactions with the UI in parallel to the TF setup, the state is not updated.
> The fix of this PR is to check if the scan options exist (GET) before doing any resource creation or update.

Could you expand a bit more on the problem? Terraform should read the remote resource state when planning. If it changes between planning and applying, then failing the apply is appropriate.

k3nz0:
@diogocp I think what he's solving is not covered by Terraform itself. For resources not in the state (i.e., already created from the UI), Terraform calls the provider’s Create implementation without a prior Read. There’s no built-in “does it already exist?” step. 
It's up to the Create implementation of the Terraform provider to handle the conflicts.

PR #3181, opened 2025-08-20
[datadog_security_monitoring_rule] Support Cloud SIEM scheduled rules

pietrodll:
Support `schedulingOptions` and `calculatedFields` for scheduled rules.

PR #3331, opened 2025-11-19
[datadog_security_monitoring_suppression] Fix issue when dealing with empty array of tags in Rules

clementgbcn:
Similar to this PR/change: https://github.com/DataDog/terraform-provider-datadog/pull/3181/files#diff-98e31eb3de7abdef4c78a8f3632d7ccdf52807b0a58945da3a8c610959f394aeR1784, when users update the list for an empty list, the Go client ignores this list.

This PR fixes this issue.

### DIFF
--- a/datadog/resource_datadog_security_monitoring_rule.go
+++ b/datadog/resource_datadog_security_monitoring_rule.go
@@ -1923,6 +1923,9 @@ func buildUpdatePayload(d *schema.ResourceData) (*datadogV2.SecurityMonitoringRu
 			tags[i] = value.(string)
 		}
 		payload.SetTags(tags)
+	} else {
+		// Hack because the Go client does not serialize empty arrays to the JSON payload
+		payload.AdditionalProperties = map[string]any{"tags": []string{}}
 	}
 
 	tfFilters := d.Get("filter")


datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1923,6 +1923,9 @@ func buildUpdatePayload(d *schema.ResourceData) (*datadogV2.SecurityMonitoringRu
 			tags[i] = value.(string)
 		}
 		payload.SetTags(tags)
+	} else {
+		// Hack because the Go client does not serialize empty arrays to the JSON payload
+		payload.AdditionalProperties = map[string]any{"tags": []string{}}
```
LiuVII, line 1928: **❓ question:** ‏can this be `payload.SetTags([]string{})`? iiuc generated MarshalJSON only omits "tags" if it's nil

clementgbcn: We looked at the client code: https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_security_monitoring_rule_update_payload.go#L44 and it omits empty.

LiuVII: gotcha, I'll approve then. Any reason to keep omitempty for this one in a long-term?

clementgbcn: It is the generated code from the Go Client, so I don't know if I can change that.

LiuVII: oh, also should this use merge for additional properties maybe JIC?

clementgbcn: Ah yes, good point for the next steps.
Since there is only one override today, it seems ok.

PR #3334, opened 2025-11-19
[resource_datadog_monitor] Add Support for Saving Assets to Monitors

kevinpombo-datadog:
Adding Support for Saving Monitor Assets alongside monitors to support our new Runbooks integration, includes new test cases that handle saving a monitor with multiple runbooks. 

Note: Needed to update every test that interacts with GET api/v1/monitor/{id} as the enpoint requires you to pass `?with_assets=true` to receive assets. There are two tests that specifically test out the new assets functionality


datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -954,6 +996,12 @@ func (r *monitorResource) buildMonitorStruct(ctx context.Context, state *monitor
 		u.SetRestrictedRolesNil()
 	}
 
+	// Assets
+	if assets := r.buildAssetsStruct(ctx, state.MonitorAssets); len(assets) > 0 {
```
shaneyuandd, line 1000: Question: On API level, if client makes an Update Monitor request with assets not defined, does it removed all assets on the monitor or preserving existing assets?

datadog/resource_datadog_monitor.go
```diff
@@ -1461,3 +1515,54 @@ func getEnv(key, fallback string) string {
...
+	assets := make([]datadogV1.MonitorAsset, 0, len(tfAssets))
+	for _, raw := range tfAssets {
+		aMap, ok := raw.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		categoryStr, _ := aMap["category"].(string)
+		nameStr, _ := aMap["name"].(string)
+		urlStr, _ := aMap["url"].(string)
+		if categoryStr == "" || nameStr == "" || urlStr == "" {
```
shaneyuandd: Curious if this is a possible scenario?

shaneyuandd (approved):
PR still in draft. Overall LGTM, left a couple questions.

datadog/resource_datadog_monitor.go
kevinpombo-datadog: Ended up removing this since we dont really want to skip adding to state

datadog/fwprovider/resource_datadog_monitor.go
kevinpombo-datadog: assets are only update if the assets field is attached on a request, mirroring the patch behavior of our put request 😅

shaneyuandd: i see. Should we remove the `len(assets) > 0` check then? In case user actually want to delete assets.

drichards-87 (approved):

datadog/fwprovider/resource_datadog_monitor.go
kevinpombo-datadog: Tested this out separately and was able to add, remove all assets, and then create new assets on a monitor with the provider with the existing logic

PR #3346, opened 2025-11-24
[datadog_observability_pipeline] Add processor groups instead of standalone processors

vladimir-dd:
#### Replace standalone processors with processor groups. ####

Before:
```
resource "datadog_observability_pipeline" "test" {
  name = "test pipeline"
  config {

    sources {
      kafka {
        id       = "source-1"
        group_id = "my-consumer-group"
        topics   = ["my-topic-1", "my-topic-2"]
      }
    }

    processors {
      parse_json {
        id      = "filter-1"
        include = "service:nginx"
        field   = "message2"
        inputs  = ["source-1"]
      }

      filter {
        id      = "filter-2"
        include = "service:nginx"
        inputs  = ["filter-1"]
      }
    }

    destinations {
      datadog_logs {
        id     = "sink-1"
        inputs = ["filter-3"]
      }
      
      amazon_s3 {
        id = "sink-2"
        inputs = ["filter-3]
        bucket = "my-bucket"
        region = "us-east-1"
      }
    }

  }
}

```

After: 
```
resource "datadog_observability_pipeline" "test" {
  name = "test pipeline"
  config {

    source {
      id       = "source-1"
      kafka = {
        group_id = "my-consumer-group"
        topics   = ["my-topic-1", "my-topic-2"]
      }
    }

    processor_group {
      id = "group-1"
      inputs  = ["source-1"]
      include = "*"
      enabled = true
   
      processor {
        id      = "filter-1"
        include = "service:nginx"
        enabled = true
        parse_json = {
          field   = "message2"
        }
      }
      
      processor {
        id      = "filter-2"
        include = "service:nginx"
        enabled = true
        filter = {
        }
      }
    }

    destination {
       id     = "sink-1"
       inputs = ["filter-3"]
       datadog_logs = {
       }
     }
      
     destination {
       id = "sink-2"
       inputs = ["filter-3]
       amazon_s3 = {
         bucket = "my-bucket"
         region = "us-east-1"
      }
    }
  }
}

```

Important context:
- `processors` is replaced with the list of `processor_group`'s, which have individual `processor` blocks inside. Both `processor_group` and `processor` blocks have `id`, `enabled` and `include`;
- `processor` blocks don't have `inputs` anymore, only processor groups;
- `processor` wrappers are required to maintain the order of processors(TF tracks the order only for the blocks of the same type). See `TestAccDatadogObservabilityPipeline_multipleProcessorGroups` as an example test with reordering.
-  processor blocks(filter, parse_json etc.) must be `ListNestedBlock` with the size validation(<=1), because of the way TF handles optional blocks. If we use `SingleNestedBlock`(which makes more sense) for an optional block, which contains any required field, TF considers this block as required. This is a recommended workaround(see https://github.com/hashicorp/terraform-provider-aws/issues/35813 for example).

#### Replacing sources, destinations with source, destination ####
As in the example above, we have replaced single `sources` and `destinations` blocks with multiple `source` and `destination` blocks. This approach follows the TF convention for list blocks and lets us share common fields(id, inputs) at the component level. 

#### Additional breaking changes combined together to avoid further breaking PRs ####
- all `ListNestedBlock`'s are renamed from plural forms to singular ones: `remaps` -> `remap`, `variables` -> `variable` etc. to follow the TF convention for list blocks;
- all `SingleNestedBlock`s replaced with `ListNestedBlock`s to make the TF schema more robust, future-proof and eliminate potential breaking changes related to required/optional blocks and fields. See https://github.com/hashicorp/terraform-provider-aws/issues/35813 as an example of the same approach.

## Suggested migration flow for existing resources ##

**If you need to make immediate changes, you can safely downgrade your Datadog Terraform provider to `3.83.0` and make updates as before.**

1. Detach the pipeline from the Terraform state (_this does not delete the pipeline!_):
```
terraform state rm datadog_observability_pipeline.test_pipeline
```
2. Open the pipeline in the UI and click Edit/Deploy (no actual changes required) - this will make the necessary config transformations on the API side. **This step will make the pipeline incompatible with the previous Terraform provider versions.**
3. Import the existing pipeline back into the Terraform state:
```
terraform import datadog_observability_pipeline.test_pipeline 4bcf83a6-dcdc-11f0-8624-da7ad0900002
```
4. Export the imported pipeline state into a HCL configuration:
```
terraform state show datadog_observability_pipeline.test_pipeline >> config.tf
```
5. Review the imported configuration and remove the old one.
6. Remove `id` field from the imported configuration.
7. Verify there is no diff for the pipeline:
```
terraform plan --target=datadog_observability_pipeline.test_pipeline
```


datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -2941,66 +2694,338 @@ func expandKafkaSource(src *kafkaSourceModel) datadogV2.ObservabilityPipelineCon
...
+		// Create key from id/enabled/include
+		key := id
+
+		// Get or create processor model for this key
+		procModel, exists := processorsByKey[key]
+		if !exists {
+			procModel = &processorModel{
+				Id:      types.StringValue(id),
+				Enabled: types.BoolValue(enabled),
+				Include: types.StringValue(include),
```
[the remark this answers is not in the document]
vladimir-dd: Nah, I think TF will complain about the "unparsed object" when deserializing an API response.

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -178,10 +206,7 @@ type enrichmentGeoIpModel struct {
 }
 
 type addEnvVarsProcessorModel struct {
-	Id        types.String         `tfsdk:"id"`
-	Include   types.String         `tfsdk:"include"`
-	Inputs    types.List           `tfsdk:"inputs"`
-	Variables []envVarMappingModel `tfsdk:"variables"`
+	Variables []envVarMappingModel `tfsdk:"variable"`
```
20agbekodo, line 210: Could be worth adding a comment here (and in every other field like `merge_strategies`) to explain that the difference in naming between TF and API-shape is normal

datadog/fwprovider/observability_pipeline/custom_processor.go
```diff
@@ -27,39 +23,31 @@ type CustomProcessorRemapModel struct {
...
 	proc := datadogV2.NewObservabilityPipelineCustomProcessorWithDefaults()
-	proc.SetId(src.Id.ValueString())
-
-	var inputs []string
-	src.Inputs.ElementsAs(ctx, &inputs, false)
-	proc.SetInputs(inputs)
 
 	var remaps []datadogV2.ObservabilityPipelineCustomProcessorRemap
 	for _, remap := range src.Remaps {
+		enabled := remap.Enabled.ValueBool()
```
20agbekodo: I'm curious about the reason for this addition

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -2943,66 +2514,338 @@ func expandKafkaSource(src *kafkaSourceModel) datadogV2.ObservabilityPipelineCon
...
+	return *apiGroup
+}
+
+// expandProcessorTypes converts the processor types model to a list of processor items
+// Uses the processor-level id, enabled, and include for all processors in the group
+func expandProcessorTypes(ctx context.Context, processors *processorModel) []datadogV2.ObservabilityPipelineConfigProcessorItem {
+	var items []datadogV2.ObservabilityPipelineConfigProcessorItem
+
+	// Get processor-level id/enabled/include
+	procId := processors.Id.ValueString()
```
20agbekodo: question: why are we sometimes accessing fields like `Id` through `processors.Id.ValueString()`, and sometimes with `processor.GetId()`?

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -628,16 +573,6 @@ type gcpAuthModel struct {
...
 
-type amazonSecurityLakeDestinationModel struct {
-	Id               types.String                         `tfsdk:"id"`
-	Inputs           types.List                           `tfsdk:"inputs"`
-	Bucket           types.String                         `tfsdk:"bucket"`
-	Region           types.String                         `tfsdk:"region"`
-	CustomSourceName types.String                         `tfsdk:"custom_source_name"`
-	Tls              *tlsModel                            `tfsdk:"tls"`
-	Auth             *observability_pipeline.AwsAuthModel `tfsdk:"auth"`
-}
```
20agbekodo, line 639: Was a remain from a previous PR, right?

datadog/fwprovider/observability_pipeline/amazon_s3_destination.go
20agbekodo, line 1: comment(doesn't have to trigger any change in this PR): now that we're having a lot of boilerplate code for each component, I find having separate files for some components and not others very arbitrary. Not a big priority, but putting everything in the main `resources_datadog_observability_pipelines` file would make sense (this file is already very long and full boilerplate, so at least we'd be consistent)

datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -49,51 +49,71 @@ func TestAccDatadogObservabilityPipeline_basic(t *testing.T) {
...
+      processor {
         id      = "parser-1"
+        enabled = true
         include = "service:my-service"
-        field   = "message"
-        inputs  = ["source-1"]
+        
+        parse_json {
+          field = "message"
+        }
```
20agbekodo, line 104: Seeing it like this, I like that syntax where we separate the top-level fields and the component-specific ones

20agbekodo (approved):
Everything looks good to me, especially with the updated tests. 
Still, this is a lot to review (and I'm sure a lot to generate on your side) so it's hard to ensure no regression was introduced.
I could be really worthwhile to update the contribution process doc to reference the new additions/ enrich the LLM prompt used when adding/updating a single component

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: there are quite a few places here and this is a TF idiomatic way of defining lists of blocks, so I'll document it in our internal docs instead.

datadog/fwprovider/observability_pipeline/custom_processor.go
vladimir-dd: Nah, it wasn't necessary - reverted it https://github.com/DataDog/terraform-provider-datadog/pull/3346/commits/9a0076650de2e1570afcc8257af8eb03407a539a

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: I think it depends on which type is it. in the TF model it's `types.String` that has `.ValueString()` and in the API Client it's `string`.
In any case, I noticed that it should be `processor` instead of `processors` here - renamed it.

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: yes, this was unused.

datadog/fwprovider/observability_pipeline/amazon_s3_destination.go
vladimir-dd: Yeah, I think we should actually go to other direction and try to split this monstrous file to dedicated files for each component to be consistent, but I'd like to do it in a separate PR. The recommendation for now is adding new components to separate files.

datadog/fwprovider/observability_pipeline/custom_processor.go
```diff
@@ -82,20 +67,13 @@ func FlattenCustomProcessor(ctx context.Context, src *datadogV2.ObservabilityPip
 func CustomProcessorSchema() schema.ListNestedBlock {
 	return schema.ListNestedBlock{
 		Description: "The `custom_processor` processor transforms events using Vector Remap Language (VRL) scripts with advanced filtering capabilities.",
+		Validators: []validator.List{
+			listvalidator.SizeAtMost(1),
+		},
```
tessneau, line 72: nice catch 👍

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -668,1627 +603,1308 @@ func (r *observabilityPipelineResource) Schema(_ context.Context, _ resource.Sch
...
-									Blocks: map[string]schema.Block{
-										"limit": schema.SingleNestedBlock{
-											Attributes: map[string]schema.Attribute{
-												"enforce": schema.StringAttribute{
-													Required:    true,
-													Description: "Whether to enforce by 'bytes' or 'events'.",
-													Validators: []validator.String{
-														stringvalidator.OneOf("bytes", "events"),
-													},
+										Blocks: map[string]schema.Block{
```
tessneau: not necessarily an immediate blocker but does this currently allow having multiple variants within the `processor` wrapper, something like 
```
[...]
processor {
  id = ""
  enabled = true 
  include = ""
  quota { ... }
  filter { ... }
}
```
where the quota and filter processor would share the same common fields ? 

if so it seems like there's a validator `objectvalidator.ExactlyOneOf` which we may be able to take advantage of to enforce just one of the variants

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -39,11 +39,38 @@ type observabilityPipelineModel struct {
 }
 
 type configModel struct {
-	Sources      sourcesModel      `tfsdk:"sources"`
-	Processors   processorsModel   `tfsdk:"processors"`
-	Destinations destinationsModel `tfsdk:"destinations"`
+	Sources      []*sourceModel         `tfsdk:"source"`
+	Processors   []*processorGroupModel `tfsdk:"processor_group"`
```
tessneau: tioli: should this be updated ?
```suggestion
	ProcessorGroups   []*processorGroupModel `tfsdk:"processor_group"`
```

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -39,11 +39,38 @@ type observabilityPipelineModel struct {
 }
 
 type configModel struct {
-	Sources      sourcesModel      `tfsdk:"sources"`
-	Processors   processorsModel   `tfsdk:"processors"`
-	Destinations destinationsModel `tfsdk:"destinations"`
```
tessneau, line 44: think you forgot to remove the destinationsModel struct definition (below)

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: Yeah, good idea! I'll create a follow-up PR for that.

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: good catch, renamed.

datadog/fwprovider/resource_datadog_observability_pipeline.go
vladimir-dd: thanks, removed.

datadog/fwprovider/observability_pipeline/socket_destination.go
```diff
@@ -117,32 +116,44 @@ func SocketDestinationSchema() schema.ListNestedBlock {
...
-							Attributes: map[string]schema.Attribute{
-								"delimiter": schema.StringAttribute{
-									Optional:    true,
-									Description: "A single ASCII character used as a delimiter.",
+						Blocks: map[string]schema.Block{
+							"character_delimited": schema.ListNestedBlock{
+								Description: "Used when `method` is `character_delimited`. Specifies the delimiter character.",
+								NestedObject: schema.NestedBlockObject{
+									Attributes: map[string]schema.Attribute{
+										"delimiter": schema.StringAttribute{
```
vladimir-dd, line 140: note: delimiter is required and now we can mark it as required with `ListNestedBlock`

datadog/fwprovider/observability_pipeline/compression.go
```diff
@@ -43,18 +45,23 @@ func FlattenCompression(src *datadogV2.ObservabilityPipelineCrowdStrikeNextGenSi
...
-				Optional:    true, // must be optional to make the block optional
-				Description: "Compression algorithm for log events.",
-			},
-			"level": schema.Int64Attribute{
-				Optional:    true,
-				Description: "Compression level.",
+		NestedObject: schema.NestedBlockObject{
+			Attributes: map[string]schema.Attribute{
+				"algorithm": schema.StringAttribute{
+					Required:    true,
```
vladimir-dd, line 54: same here

datadog/fwprovider/observability_pipeline/socket_source.go
```diff
@@ -113,34 +113,45 @@ func SocketSourceSchema() schema.ListNestedBlock {
...
-									"newline_delimited",
-									"bytes",
-									"character_delimited",
-									"octet_counting",
-									"chunked_gelf",
-								),
+					NestedObject: schema.NestedBlockObject{
+						Attributes: map[string]schema.Attribute{
+							"method": schema.StringAttribute{
+								Required:    true,
```
vladimir-dd, line 121: same here

datadog/fwprovider/observability_pipeline/socket_source.go
```diff
@@ -113,34 +113,45 @@ func SocketSourceSchema() schema.ListNestedBlock {
...
-								"delimiter": schema.StringAttribute{
-									Optional:    true, // must be optional to make the block optional
-									Description: "A single ASCII character used as a delimiter.",
+						Blocks: map[string]schema.Block{
+							"character_delimited": schema.ListNestedBlock{
+								Description: "Used when `method` is `character_delimited`. Specifies the delimiter character.",
+								NestedObject: schema.NestedBlockObject{
+									Attributes: map[string]schema.Attribute{
+										"delimiter": schema.StringAttribute{
+											Required:    true,
```
vladimir-dd, line 140: same here

datadog/fwprovider/observability_pipeline/tls.go
```diff
@@ -2,64 +2,74 @@ package observability_pipeline
...
-				Optional:    true,
-				Description: "Path to the Certificate Authority (CA) file used to validate the server's TLS certificate.",
-			},
-			"key_file": schema.StringAttribute{
-				Optional:    true,
-				Description: "Path to the private key file associated with the TLS client certificate. Used for mutual TLS authentication.",
+		NestedObject: schema.NestedBlockObject{
+			Attributes: map[string]schema.Attribute{
+				"crt_file": schema.StringAttribute{
+					Required:    true,
```
vladimir-dd, line 58: same here

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -576,490 +570,477 @@ func (r *observabilityPipelineResource) Schema(_ context.Context, _ resource.Sch
...
-														Optional:    true, // must be optional to make the block optional
-														Description: "SASL mechanism to use (e.g., PLAIN, SCRAM-SHA-256, SCRAM-SHA-512).",
-														Validators: []validator.String{
-															stringvalidator.OneOf("PLAIN", "SCRAM-SHA-256", "SCRAM-SHA-512"),
+												"sasl": schema.ListNestedBlock{
+													Description: "SASL authentication settings.",
+													NestedObject: schema.NestedBlockObject{
+														Attributes: map[string]schema.Attribute{
+															"mechanism": schema.StringAttribute{
+																Required:    true,
```
vladimir-dd, line 637: same here

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -576,490 +570,477 @@ func (r *observabilityPipelineResource) Schema(_ context.Context, _ resource.Sch
...
-													"session_name": schema.StringAttribute{
-														Optional:    true,
-														Description: "A session identifier used for logging and tracing the assumed role session.",
+									},
+									"http_server": schema.ListNestedBlock{
+										Description: "The `http_server` source collects logs over HTTP POST from external services.",
+										NestedObject: schema.NestedBlockObject{
+											Attributes: map[string]schema.Attribute{
+												"auth_strategy": schema.StringAttribute{
+													Required:    true,
```
vladimir-dd, line 674: same here

docs/resources/observability_pipeline.md
```diff
@@ -1154,42 +1058,43 @@ Optional:
...
-### Nested Schema for `config.sources.amazon_data_firehose`
+- `id` (String) The unique identifier for this source.
 
-Required:
+Optional:
+
+- `amazon_data_firehose` (Block List) The `amazon_data_firehose` source ingests logs from AWS Data Firehose. (see [below for nested schema](#nestedblock--config--source--amazon_data_firehose))
+- `amazon_s3` (Block List) The `amazon_s3` source ingests logs from an Amazon S3 bucket. It supports AWS authentication and TLS encryption. (see [below for nested schema](#nestedblock--config--source--amazon_s3))
+- `datadog_agent` (Block List) The `datadog_agent` source collects logs from the Datadog Agent. (see [below for nested schema](#nestedblock--config--source--datadog_agent))
+- `fluent_bit` (Block List) The `fluent` source ingests logs from Fluent Bit. (see [below for nested schema](#nestedblock--config--source--fluent_bit))
```
brett0000FF:
```suggestion
- `fluent_bit` (Block List) The `fluent_bit` source ingests logs from Fluent Bit. (see [below for nested schema](#nestedblock--config--source--fluent_bit))
```

docs/resources/observability_pipeline.md
```diff
@@ -1154,42 +1058,43 @@ Optional:
...
+- `fluentd` (Block List) The `fluent` source ingests logs from a Fluentd-compatible service. (see [below for nested schema](#nestedblock--config--source--fluentd))
```
brett0000FF:
```suggestion
- `fluentd` (Block List) The `fluentd` source ingests logs from a Fluentd-compatible service. (see [below for nested schema](#nestedblock--config--source--fluentd))
```

docs/resources/observability_pipeline.md
```diff
@@ -546,70 +509,82 @@ Optional:
...
+- `ocsf_mapper` (Block List) The `ocsf_mapper` processor transforms logs into the OCSF schema using predefined library mappings. (see [below for nested schema](#nestedblock--config--processor_group--processor--ocsf_mapper))
+- `parse_grok` (Block List) The `parse_grok` processor extracts structured fields from unstructured log messages using Grok patterns. (see [below for nested schema](#nestedblock--config--processor_group--processor--parse_grok))
+- `parse_json` (Block List) The `parse_json` processor extracts JSON from a specified field and flattens it into the event. This is useful when logs contain embedded JSON as a string. (see [below for nested schema](#nestedblock--config--processor_group--processor--parse_json))
+- `quota` (Block List) The `quota` measures logging traffic for logs that match a specified filter. When the configured daily quota is met, the processor can drop or alert. (see [below for nested schema](#nestedblock--config--processor_group--processor--quota))
```
brett0000FF:
```suggestion
- `quota` (Block List) The `quota` processor measures logging traffic for logs that match a specified filter. When the configured daily quota is met, the processor can drop or alert. (see [below for nested schema](#nestedblock--config--processor_group--processor--quota))
```

docs/resources/observability_pipeline.md
vladimir-dd: updated

docs/resources/observability_pipeline.md
vladimir-dd: updated

vladimir-dd:
>  Is it possible to add a test to expect an error when not filling a newly required nested field?

In general yes, but this would require adding quite a few tests to basically double-check whether the field is marked as Required or not... I'm hoping we could find a way to generate different pipeline configs and validating them outside the standard TF testing framework for more thorough testing.

PR #3349, opened 2025-11-26
[CCV-4035] [datadog_cost_budget] Add Validation for terraform plan Flow

zeinageb:
### Problem
The current Terraform Cost Budget resource does not perform sufficient validation during the terraform plan phase. Configuration errors (especially around tag filters and missing budget entries) are only detected during terraform apply, resulting in user confusion and wasted time

### Solution
Implemented `ResourceWithValidateConfig` interface to add client-side validation that mirrors the API's validation logic from dd-source. Validation now runs during terraform plan, catching configuration errors before any API calls are made

### Changes
**1. Client-Side Validation (`resource_datadog_cost_budget.go`)**

Added `ValidateConfig() `method that validates:

- Tags validation:
          - Tags must have 0, 1, or 2 elements
          - Tags must be unique (no duplicate tags in by {tag1,tag2})
- Month validation:
          - start_month > 0 and format YYYYMM
          - end_month > 0 and format YYYYMM
          - end_month >= start_month
- Entry validation:
        - Entry month must be within budget period
        - Entry amount must be >= 0
        - Entry tag_filters count must match tags in metrics_query
        - Each tag_key must be one of the tags from metrics_query
        - Each unique tag combination must have entries for all months in the budget period

**2. Helper Functions**

Added two helper functions copied from dd-source to maintain consistency:

- `extractTagsFromQuery()`: Parses tags from by {tag1,tag2} clause
- `calculateMonthCount()`: Calculates number of months in budget period

### Testing

Created test files to test different scenarios we might have 

 **Valid Configuration (Passes Validation)**

Test case: test_valid_budget/main.tf
Hierarchical budget with 12 months, 2 tags, 1 combinationResult: terraform plan succeeds

image

**Invalid Configurations (Caught at Plan Time)**

**Test case 1: test_invalid_tag_key/main.tf**

metrics_query =` "sum:aws.cost{*} by {account}"tag_filters {  tag_key = "wrong_tag"  # ERROR: Not in metrics_query}`
image

**Test case 2: test_missing_months/main.tf**

start_month = 202501, and end_month = 202503# Only has entry for 202501, missing 202502 & 202503

image

**Test case 3: test_wrong_tag_count/main.tf**

metrics_query = `"sum:aws.cost{*} by {team,account}"  # 2 tagstag_filters {  tag_key = "team"  # ERROR: Missing 'account' tag filter}`

image


docs/resources/cost_budget.md
```diff
@@ -50,3 +169,14 @@ Required:
 
 - `tag_key` (String)
 - `tag_value` (String)
+
+## Import
+
+Import is supported using the following syntax:
+
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
```
iadjivon, line 177:
```suggestion
Import is supported when using the syntax below. This is an example using the [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import):

```

This section is a bit confusing. Is this describing the terraform import command, or is the terraform import command an example of the types of import. 

Would the suggestion made keep the meaning here?

zeinageb: From what I checked, the universal way is using the terraform import command (there's also an import block method, it's in the link that was generated if you want to check it out, but it's less common, [here](https://developer.hashicorp.com/terraform/language/import))

That said, this text is auto-generated when I run `make docs` and is used across all resources in the provider , I only write the example command in `import.sh`

zeinageb:
> Hi there, I added a non-blocking comment and approved, but recommend revising the **Import** section.

Hey! Thanks for the review! I replied to the comment, let me know if you have any other questions or if you're good with keeping the auto-generated text as is

datadog/fwprovider/resource_datadog_cost_budget.go
```diff
@@ -198,6 +208,177 @@ func (r *costBudgetResource) ImportState(ctx context.Context, req resource.Impor
 	resource.ImportStatePassthroughID(ctx, frameworkPath.Root("id"), req, resp)
 }
 
+// ValidateConfig performs client-side validation during terraform plan
+// Mirrors BudgetWithEntries.validate() from dd-source: cost-planning-api/budgets/handler.go
+func (r *costBudgetResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
```
fpighi, line 214: For other complex resources we have a /validate endpoint (e.g. monitors).
I think that's a much better approach for maintainability, since you don't need to duplicate validation logic in two places (terraform provider and backend code).
It would greatly simplify the implementation here too.
Have you considered using that approach instead?

datadog/fwprovider/resource_datadog_cost_budget.go
```diff
@@ -198,6 +208,177 @@ func (r *costBudgetResource) ImportState(ctx context.Context, req resource.Impor
 	resource.ImportStatePassthroughID(ctx, frameworkPath.Root("id"), req, resp)
 }
 
+// ValidateConfig performs client-side validation during terraform plan
+// Mirrors BudgetWithEntries.validate() from dd-source: cost-planning-api/budgets/handler.go
```
fpighi: Not sure if this internal reference was intentional here.

datadog/fwprovider/resource_datadog_cost_budget.go
zeinageb: Thanks for the feedback! Yes, we've considered this approach
This PR is our short-term solution to provide immediate value while we work on the long-term approach. We have an [RFC](https://datadoghq.atlassian.net/wiki/spaces/CCM/pages/5900338014/RFC+Terraform+Cost+Budget+Validation+Improvement+when+terraform+plan) in progress that covers exactly this

datadog/fwprovider/resource_datadog_cost_budget.go
zeinageb: I added it so we know this validation logic is copied from the API and is meant to be temporary. Once we implement API-based validation (dry-run or /validate endpoint, RFC mentioned above in progress), this duplicated code will be removed

zeinageb:
> I also think we should have some tests for the validation if it's implemented in the client like this. If it was an API call you would only need one to validate the api call is executed.

Yes true, I tested the different scenarios manually (see screenshots in the PR description). I can add unit tests for each validation rule if necessary for this short-term solution, before implementing the` dry-run` or `/validate` endpoint

PR #3354, opened 2025-12-03
[datadog_integration_gcp_sts] Add regionFilterConfigs and isGlobalLocationEnabled

katherinekim-51:

## Description

https://datadoghq.atlassian.net/browse/GCP-2892

This updates our terraform provider to support new location filtering fields (regionFilterConfigs and isGlobalLocationEnabled) 

## Implementation Details

Adds 2 new fields to terraform provider 
   - isGlobalLocationEnabled --> optional (to prevent breaking existing users) and computed (because it is true by default) 
   - regionFilterConfigs --> optional (to prevent breaking existing users) 
   
## Testing Guidelines

Amended to existing tests and reran cassettes


datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -174,6 +176,16 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 				Description: "Configurations for GCP monitored resources. Only monitored resources that apply to specified filters are imported into Datadog.",
 				ElementType: MonitoredResourceConfigSpec,
 			},
+			"is_global_location_enabled": schema.BoolAttribute{
+				Description: "When enabled, only collect where location is explicitly stated as “global” or where location information cannot be deduced from GCP labels.",
+				Optional:    true,
+				Computed:    true,
```
ChrisHansenDatadog: For my own understanding: why is this marked computed? Is it because we're going to have the API default to returning `true` for this property and it needs to be marked computed for that to work with the TF SDK?

ChrisHansenDatadog:
I think this makes sense, but probably best to get some eyes from someone who's worked on our TF provider before.

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
ash-ddog, line 183: Is this computed bc the server has a default? I'm assuming the default is `false`?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -174,6 +176,16 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
 				Description: "Configurations for GCP monitored resources. Only monitored resources that apply to specified filters are imported into Datadog.",
 				ElementType: MonitoredResourceConfigSpec,
 			},
+			"is_global_location_enabled": schema.BoolAttribute{
+				Description: "When enabled, only collect where location is explicitly stated as “global” or where location information cannot be deduced from GCP labels.",
```
ash-ddog: I know you got [doc review here](https://github.com/DataDog/datadog-api-spec/pull/4646#discussion_r2586324912) but this does not make sense to me as written. Can we check again?

`When enabled, metric collection is limited to resources that either have no location information or have location explicitly stated as 'global'`

If we fix this here, can we fix it in the API spec as well?

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -174,6 +176,16 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
...
+				Optional:    true,
+				Computed:    true,
+			},
+			"region_filter_configs": schema.SetAttribute{
+				Optional:    true,
+				Description: "Filters to limit metric collection by GCP location, such as region, multi-region, or zone. By default, Datadog collects from all locations.",
```
ash-ddog: Can we use a similar description to other `_configs`? If we update here, can we update API spec as well?

Something like:
`Configurations for GCP region filtering. Only monitored resources that match the specified regions are imported into Datadog."`

datadog/fwprovider/resource_datadog_integration_gcp_sts.go
```diff
@@ -174,6 +176,16 @@ func (r *integrationGcpStsResource) Schema(_ context.Context, _ resource.SchemaR
...
+				ElementType: types.StringType,
```
ash-ddog, line 188: This is fine,but it does mean these configs will not be extensible in the future. I think that's ok if this matches what we decided for the API (I think it does).

datadog/tests/resource_datadog_integration_gcp_sts_test.go
```diff
@@ -350,6 +359,8 @@ func TestAccIntegrationGcpStsDefault(t *testing.T) {
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("cloud_run_revision_filters"), knownvalue.SetExact(nil)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("host_filters"), knownvalue.SetExact(nil)),
 					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("monitored_resource_configs"), knownvalue.SetExact(nil)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("is_global_location_enabled"), knownvalue.Bool(true)),
+					statecheck.ExpectKnownValue("datadog_integration_gcp_sts.foo", tfjsonpath.New("region_filter_configs"), knownvalue.Null()),
```
ash-ddog, line 363: Ah, so, looks like the defaults are
```
is_global_location_enabled = true
region_filter_configs = null
```

How does our crawler handle this? I'm assuming the settings I just posted means "don't do any region filtering". What about?

```
is_global_location_enabled = false
region_filter_configs = null
```
Does this mean only collect metrics where `location is available` and `location != global`?

No action to take here, will ping you offline to understand these settings more.

katherinekim-51: for ^ and other doc related comments, discussed offline + verified with doc teams that we are going with this revision: "When enabled, Datadog collects metrics where location is explicitly stated as "global" or where location information cannot be deduced from GCP" 

will update api spec pr too!

ash-ddog (approved):
Thanks for fixing docs! 🙏

docs/resources/integration_gcp_sts.md
```diff
@@ -52,11 +52,13 @@ resource "datadog_integration_gcp_sts" "foo" {
 - `cloud_run_revision_filters` (Set of String, Deprecated) List of filters to limit the Cloud Run revisions that are pulled into Datadog by using tags. Only Cloud Run revision resources that apply to specified filters are imported into Datadog.
 - `host_filters` (Set of String, Deprecated) List of filters to limit the VM instances that are pulled into Datadog by using tags. Only VM instance resources that apply to specified filters are imported into Datadog.
 - `is_cspm_enabled` (Boolean) Whether Datadog collects cloud security posture management resources from your GCP project. If enabled, requires `resource_collection_enabled` to also be enabled.
+- `is_global_location_enabled` (Boolean) When enabled, Datadog collects metrics where location is explicitly stated as 'global' or where location information cannot be deduced from GCP
```
joepeeples:
```suggestion
- `is_global_location_enabled` (Boolean) When enabled, Datadog collects metrics where location is explicitly stated as 'global' or where location information cannot be deduced from GCP.
```

joepeeples (approved):
Approved with tiny edit, thanks!

PR #3357, opened 2025-12-08
POC: synthetics_global_variable: write only value

bernata:
For the datadog_synthetics_global_variable resource, create value_wo and value_wo_version attributes. value_wo is a wrote only attribute that behaves like the value attribute [except write only]. This will support writing ephemeral resources to a global variable. The reason this is important is ephemeral resources and write only attributes do not get written to the terraform state file. These secrets can be sensitive [like credentials used by a synthetic] and are frequently forbidden by security policy to live in terraform state backends.

### DIFF
--- a/datadog/fwprovider/resource_datadog_synthetics_global_variable.go
+++ b/datadog/fwprovider/resource_datadog_synthetics_global_variable.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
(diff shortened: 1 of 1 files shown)


bernata:
> For the datadog_synthetics_global_variable resource, create value_wo and value_wo_version attributes. value_wo is a wrote only attribute that behaves like the value attribute [except write only]. This will support writing ephemeral resources to a global variable. The reason this is important is ephemeral resources and write only attributes do not get written to the terraform state file. These secrets can be sensitive [like credentials used by a synthetic] and are frequently forbidden by security policy to live in terraform state backends.

@LiuVII I noticed you created a similar [POC](https://github.com/DataDog/terraform-provider-datadog/pull/3210). I was wondering what your thoughts were on continuing this one and if datadog would accept it?

LiuVII:
👋 @bernata
Thank you for taking interest in this and taking a stab at improving security of our TF provider resources 🙇 
We just merged [PR with some helpers](https://github.com/DataDog/terraform-provider-datadog/pull/3210) that will be useful for WO argument implementation.
And here's a [follow-up PR](https://github.com/DataDog/terraform-provider-datadog/pull/3382) that I still need to discuss with codeowners that implements the idea you have here (though unfortunately I couldn't use all helpers we made for it because of `secure` option we have for global vars).
Lmk what you think and if this is in-line with what you imagined for it 🙂

bernata:
> 👋 @bernata Thank you for taking interest in this and taking a stab at improving security of our TF provider resources 🙇 We just merged [PR with some helpers](https://github.com/DataDog/terraform-provider-datadog/pull/3210) that will be useful for WO argument implementation. And here's a [follow-up PR](https://github.com/DataDog/terraform-provider-datadog/pull/3382) that I still need to discuss with codeowners that implements the idea you have here (though unfortunately I couldn't use all helpers we made for it because of `secure` option we have for global vars). Lmk what you think and if this is in-line with what you imagined for it 🙂

Thank you @LiuVII - what a wonderful way to start 2026!  I looked over your [PR](https://github.com/DataDog/terraform-provider-datadog/pull/3382/files#diff-8debe959cef44fd2098a8a02e2c6404ad6c81d0feabb4ec67a404435395d2ba7R167-R190) especially at the schema definition which is very much in-line with what i imagined. Thank you again.

I wanted to understand a bit about this comment you made:
> we can't use CreateWriteOnlySecretAttributes because of secure conditional, which is unique to this resource

Is it because the `secure` property comes from the value which you won't have with a WO attribute.  Since it comes from value, it doesn't seem like a problem that we don't set it in `state` in `updateState` since we don't remember the value_wo anyway. But is your comment more about being unable to use a standard helper function b/c of this unique situation?

LiuVII:
> I looked over your [PR](https://github.com/DataDog/terraform-provider-datadog/pull/3382/files#diff-8debe959cef44fd2098a8a02e2c6404ad6c81d0feabb4ec67a404435395d2ba7R167-R190) especially at the schema definition which is very much in-line with what i imagined. Thank you again.

Happy to hear it! Will try to get reviews soon and should be available then for the next release.

> But is your comment more about being unable to use a standard helper function b/c of this unique situation?

Yeah, this is about the helper.

LiuVII:
The [corresponding PR](https://github.com/DataDog/terraform-provider-datadog/pull/3382) has been merged and should be available in the next release by the end of this month

PR #3365, opened 2025-12-11
[security_monitoring] Fix missing group_signals_by in rules

clementgbcn:
When a customer update the signal group by of a rule, the changes are not propagated.


adriacabeza (approved):

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1939,6 +1939,13 @@ func buildUpdatePayload(d *schema.ResourceData) (*datadogV2.SecurityMonitoringRu
 			payload.SetReferenceTables(make([]datadogV2.SecurityMonitoringReferenceTable, 0))
 		}
 
+		if v, ok := d.GetOk("group_signals_by"); ok {
```
paul-hf, line 1942: Could we add `group_signals_by` to a test to prevent regressions?

clementgbcn: I had to record the test again because they were failing. Therefore, it should prevent any regression from happening.

fpighi (approved):

PR #3375, opened 2025-12-18
[CWS] fix multiple actions support

safchain:
This PR bump `datadog-api-client-go` and fix the Workload Protection actions.

### DIFF
--- a/docs/data-sources/csm_threats_agent_rules.md
+++ b/docs/data-sources/csm_threats_agent_rules.md
@@ -56,6 +56,7 @@ Read-Only:
 
 Read-Only:
 
+- `field` (String)
 
 
 <a id="nestedobjatt--agent_rules--actions--set"></a>
--- a/docs/resources/csm_threats_agent_rule.md
+++ b/docs/resources/csm_threats_agent_rule.md
@@ -63,13 +63,13 @@ Optional:
 <a id="nestedblock--actions--hash"></a>
 ### Nested Schema for `actions.hash`
 
+Optional:
+
+- `field` (String) The field to hash
+
 
 <a id="nestedblock--actions--set"></a>
 ### Nested Schema for `actions.set`
-
-Required:
-
-- `name` (String) The name of the set action
 
 Optional:
 
@@ -78,6 +78,7 @@ Optional:
 - `expression` (String) The expression to use for the set action
 - `field` (String) The field to get the value from
 - `inherited` (Boolean) Whether the set action is inherited
+- `name` (String) The name of the set action
 - `scope` (String) The scope of the set action (process, container, cgroup, or empty)
 - `size` (Number) The maximum size of the set
 - `ttl` (Number) The time to live for the set in nanoseconds


datadog/fwprovider/resource_datadog_csm_threats_agent_rule.go
```diff
@@ -126,7 +126,7 @@ func (r *csmThreatsAgentRuleResource) Schema(_ context.Context, _ resource.Schem
 							Description: "Set action configuration",
 							Attributes: map[string]schema.Attribute{
 								"name": schema.StringAttribute{
-									Required:    true,
+									Optional:    true,
```
QuentinGuillard, line 129: `validateActions()` still checks that it's non-empty. Should we align these? Either keep `Required: true` in the schema or remove the validation check

datadog/tests/data_source_datadog_csm_threats_agent_rules_test.go
```diff
@@ -40,11 +40,15 @@ func TestAccCSMThreatsAgentRulesDataSource(t *testing.T) {
 			actions {
 				set {
 					name   = "test_action"
-					field  = "exec.file.path"
+					field  = "open.file.path"
```
QuentinGuillard, line 43: question: didn't we have issues with `open.file.path` expressions and changed them to `exec.file.path` before?

safchain: Yes it was incorrect and it is not supported now

datadog/fwprovider/resource_datadog_csm_threats_agent_rule.go
safchain: The problem is that the schema applies the constraints for all the types of actions while we would like to apply this only for `set`. I think we can relax it here and let the proper constraint in the validateActions. WDYT ?

datadog/tests/data_source_datadog_csm_threats_agent_rules_test.go
safchain: To elaborate a bit more, we have to use the same event type in the set action than on the rule expression.

datadog/fwprovider/resource_datadog_csm_threats_agent_rule.go
QuentinGuillard: Makes sense! I didn't think about that. I agree this is the right approach then

datadog/tests/data_source_datadog_csm_threats_agent_rules_test.go
QuentinGuillard: so this is not an issue here? or is it only related to the `expression` field and not this specific one?

safchain: The issue is that the backend won't validate the set action if the expression and the field don't use the same event type. As the expression was using `open` and changed the `set` `field` to make the action valid. Before the recent backend fixes it was possible to mix but it was causing agent issues.

PR #3384, opened 2025-12-22
obs-pipelines: remove rate config field in sample processors

clementd-dd:
companion PR to https://github.com/DataDog/datadog-api-client-go/pull/3546

note: the go-client update commit is not included in this PR to avoid merges. Since the initial datadog-api-spec PR has been merged in the op-ga branch, syncing the op-ga datadog branch to the corresponding go client auto-generated PR should work, even if individual commits fail in the meantime


clementd-dd:
tested locally with a go-client update

PR #3385, opened 2025-12-22
Add metric pipelines

20agbekodo:


datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -4469,7 +4468,7 @@ resource "datadog_observability_pipeline" "parse_xml_test" {
     }
 
     destination {
-      id     = "dest-1"
+          id     = "dest-1"
```
taylorchandleryoung: nit: this looks like it could be cleaned up?

PR #3400, opened 2026-01-06
[datadog_security_monitoring_rule] Add instantaneous baseline option to NewValue

hachem-dd:
- Add instantaneous_baseline option for security monitoring new value rules.

[Jira](https://datadoghq.atlassian.net/browse/SEC-26438)

### DIFF
--- a/datadog/resource_datadog_security_monitoring_rule.go
+++ b/datadog/resource_datadog_security_monitoring_rule.go
@@ -233,7 +233,8 @@ func datadogSecurityMonitoringRuleSchema(includeValidate bool) map[string]*schem
 								},
 								"instantaneous_baseline": {
 									Type:        schema.TypeBool,
-									Required:    true,
+									Optional:    true,
+									Default:     false,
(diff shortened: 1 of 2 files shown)


datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -231,6 +231,11 @@ func datadogSecurityMonitoringRuleSchema(includeValidate bool) map[string]*schem
 									Required:         true,
 									Description:      "The duration in days after which a learned value is forgotten.",
 								},
+								"instantaneous_baseline": {
+									Type:        schema.TypeBool,
+									Required:    true,
```
paul-hf: Should this be required? Maybe it should be optional and default to false if not included? Otherwise users of our terraform provider will have to add `instantaneous_baseline = false` to all of their existing rules

rtrieu (approved):

datadog/resource_datadog_security_monitoring_rule.go
hachem-dd: You're right. Oversight from my part 🫡  47d1bd5b62cad04c2087be1ac3ec5922696746e9

fpighi (approved):

PR #3402, opened 2026-01-07
[datadog_sensitive_data_scanner_standard_pattern] Fix filter matching multiple patterns due to substring collision

### LINKED ISSUE #3370: datadog_sensitive_data_scanner_standard_pattern lower case search causes side effects

v3.81.0

### Terraform Version

v1.14.2

- datadog_sensitive_data_scanner_standard_pattern

```hcl
data "datadog_sensitive_data_scanner_standard_pattern" "this" {
  filter = "US Tax Identification Number Scanner"
}
```

_No response_

### Expected Behavior

I want to be able to retrieve that library rule. Matching with contains + lowercase is bound to create side effects, I would at least expect an "exact match" mode.
Or even have the code first perform an exact match on the name and then fallback to a lowercase contains if needed.
Or even have another datasource that can retrieve all the rules & letting the dev handle the match myself.

Since Datadog can rename those rules at any point in time, with only the id being constant, providing more filter options is a must, or at least letting the users handles all the rules & find those they want could also work.

### Actual Behavior

Error : Your query returned more than one result, please try a more specific search criteria

### Steps to Reproduce

terraform plan

### Important Factoids

Image
There is currently one rule named exactly "US Tax Identification Number Scanner" and another named "Cyprus Tax Identification Number Scanner".

### References

_No response_

shinsj4653:
## Description

Fixes an issue where the `filter` parameter would match unintended patterns due to substring collision. For example, searching for "US Tax Identification Number Scanner" would also match "Cyprus Tax Identification Number Scanner" because "Cyprus" ends with "us".

Closes #3370

## Changes

- Modified `dataSourceDatadogSensitiveDataScannerStandardPatternRead()` to prioritize exact match over partial match
- When filter exactly matches a pattern name, that pattern is returned immediately
- Partial match (case-insensitive contains) is used as fallback when no exact match exists
- Added unit tests for the new matching logic
- Added acceptance test case for exact match scenario

## Root Cause

The filter logic only used `strings.Contains()` for matching, which caused false positives when the search term appeared as a substring in other pattern names.

// Before: Only partial match
if strings.Contains(strings.ToLower(name), strings.ToLower(searchedName)) {
    foundStandardPatterns = append(foundStandardPatterns, resource)
}

// After: Exact match takes priority
if name == searchedName {
    exactMatch = &match
    break
}

## Testing
```go
=== RUN   TestExactMatchPriority
=== RUN   TestExactMatchPriority/exact_match_found           ✅ PASS
=== RUN   TestExactMatchPriority/exact_match_priority_over_partial  ✅ PASS
=== RUN   TestExactMatchPriority/partial_match_single        ✅ PASS
=== RUN   TestExactMatchPriority/partial_match_multiple_error ✅ PASS
=== RUN   TestExactMatchPriority/no_match                    ✅ PASS
--- PASS: TestExactMatchPriority
```
- [x] Unit tests for exact match priority logic - all 5 cases pass
- [x] Build verification - success
- [x] Existing functionality preserved (partial match still works as fallback)


apenney-unchained:
Any reason this has been ignored for so long, datadog team?  I just ran into this bug trying to setup SDS and I was surprised to find someone fixed this for you 3 months ago and it's just sat here untouched.

shinsj4653:
Hi team, I've resolved the merge conflicts and updated the branch. Just pinging again for a review. Could someone take a look when you have a chance? Thank you!

PR #3404, opened 2026-01-07
[datadog_observability_pipeline] Add referencetable and datastream

taylorchandleryoung:
Adding reference_table to the enchrichment table processor
Adding datastream option to the Elastic Search destination

### DIFF
--- a/datadog/tests/resource_datadog_observability_pipeline_test.go
+++ b/datadog/tests/resource_datadog_observability_pipeline_test.go
@@ -5659,4 +5659,3 @@ resource "datadog_observability_pipeline" "elasticsearch_datastream" {
 		},
 	})
 }
-


datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -3550,6 +3611,30 @@ func flattenEnrichmentTableProcessor(ctx context.Context, src *datadogV2.Observa
 			},
 		}
 	}
+	if refTableRaw, ok := src.AdditionalProperties["reference_table"]; ok {
```
vladimir-dd: we should be using `ReferenceTable` field(mapped to `reference_table` in the API) field from the updated Go client model instead.

PR #3406, opened 2026-01-07
[datadog_oncall] Add on_call_user_notification_channel resource

kevinconaway:
This commit adds the `on_call_user_notification_channel` resource for allowing users to create On-Call notification channels

It looks like some other docs were out of date when I ran `make docs` as t


datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
```diff
@@ -0,0 +1,322 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"phone": schema.SingleNestedBlock{
+				Attributes: map[string]schema.Attribute{
+					"number": schema.StringAttribute{
+						Optional:    true,
```
daniel-m-dd, line 80: Remind me, it's not optional but we do mark it as optional being part of `map` due to TF plugin restrictions ? (same for email address)

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
```diff
@@ -0,0 +1,322 @@
...
+	if err != nil {
+		response.Diagnostics.AddError("response object was invalid", err.Error())
+	}
+
+	response.Diagnostics.Append(response.State.Set(ctx, &newState)...)
+}
+
+func (c *onCallUserNotificationChannel) Update(_ context.Context, _ resource.UpdateRequest, _ *resource.UpdateResponse) {
+	// This is a no-op as channels cannot currently be updated.
+	// This should not be called as all attributes are marked as require-replace
```
daniel-m-dd, line 180: 👍 could we comment on this on the schema definition? (I was wondering why we do set requiresreplace on all atrributes)

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
bndbsh: If I recall it would make the `phone` block required if the attribute was not optional

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
bndbsh: Could we add a diagnostics error here if it should never be called?

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
kevinconaway: >If I recall it would make the phone block required if the attribute was not optional

I assumed this was the case when I borrowed this pattern from the routing rule actions.

bndbsh: Note I had to manually update docs to reflect this: https://github.com/DataDog/terraform-provider-datadog/pull/3077

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
```diff
@@ -0,0 +1,328 @@
...
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"phone": schema.SingleNestedBlock{
+				Attributes: map[string]schema.Attribute{
+					"number": schema.StringAttribute{
+						Optional:    true,
+						Description: "The E-164 formatted phone number (e.g. +3371234567)",
```
janine-c:
```suggestion
						Description: "The E-164 formatted phone number (for example, `+3371234567`)",
```

PR #3411, opened 2026-01-09
[datadog_dashboard] Support hide_incomplete_cost_data field for widget time spans

yacomink:
### Problem

The Dashboard API supports `hide_incomplete_cost_data` on widget time spans, but the Terraform provider had several issues handling this field:

1. **API Client oneOf Ambiguity**: The `WidgetTime` union type in the API client can fail to deserialize when the API returns the new time format with `hide_incomplete_cost_data`, resulting in the error on `terraform plan` (or any action that reads the dashboard from our API):
   ```
   object contains unparsed element: map[hide_incomplete_cost_data:true type:live unit:hour value:1]
   ```

2. **Missing Schema Support**: The `hide_incomplete_cost_data` field wasn't yet supported in terraform widget schemas.

### Root Cause

The API client's `WidgetTime` is a `oneOf` union that can be:
- `WidgetLegacyLiveSpan`: `{"live_span": "1h", "hide_incomplete_cost_data": true}`
- `WidgetNewLiveSpan`: `{"type": "live", "unit": "hour", "value": 1, "hide_incomplete_cost_data": true}`
- `WidgetNewFixedSpan`: `{"type": "fixed", "from": ..., "to": ...}`

When the API returns the new format with `hide_incomplete_cost_data`, the API client's unmarshaler:
1. Tries `WidgetLegacyLiveSpan` - partial match (has `hide_incomplete_cost_data`, ignores extra fields)
2. Tries `WidgetNewLiveSpan` - full match (all required fields present)
3. Result: Multiple matches → stores entire object in `UnparsedObject` → provider error

### Solution

This PR adds comprehensive support for `hide_incomplete_cost_data` and handles the `WidgetTime` oneOf ambiguity:

### Example Usage

```hcl
resource "datadog_dashboard" "example" {
  title       = "Cost Dashboard"
  layout_type = "ordered"

  widget {
    timeseries_definition {
      live_span                  = "1h"
      hide_incomplete_cost_data = true  # New field!
      title                      = "Cost Metrics"
      
      request {
        query {
          metric_query {
            data_source = "metrics"
            query       = "avg:aws.cost.amortized{*}"
          }
        }
      }
    }
  }
}
```


.github/workflows/sdkv2_deprecation.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
         uses: tj-actions/changed-files@d6e91a2266cdb9d62096cebf1e8546899c6aa18f
         with:
           files: "datadog/*.go"
+          files_ignore: "datadog/*_test.go"
```
brtu, line 23: what's this for?

docs/resources/powerpack.md
```diff
@@ -49,7 +49,7 @@ resource "datadog_powerpack" "foo" {
 
 - `description` (String) The description of the powerpack.
 - `layout` (Block List, Max: 1) The layout of the powerpack on a free-form dashboard. (see [below for nested schema](#nestedblock--layout))
-- `live_span` (String) The timeframe to use when displaying the widget. Valid values are `1m`, `5m`, `10m`, `15m`, `30m`, `1h`, `4h`, `1d`, `2d`, `1w`, `1mo`, `3mo`, `6mo`, `week_to_date`, `month_to_date`, `1y`, `alert`.
+- `live_span` (String) The timeframe to use when displaying the powerpack. Valid values are `1m`, `5m`, `10m`, `15m`, `30m`, `1h`, `4h`, `1d`, `2d`, `1w`, `1mo`, `3mo`, `6mo`, `week_to_date`, `month_to_date`, `1y`, `alert`.
```
brtu: i wonder why it changed from `widget` -> `powerpack`.  this was the output from the `tfdocs` generation tool?

datadog/resource_datadog_dashboard.go
```diff
@@ -313,6 +314,55 @@ func updateDashboardState(d *schema.ResourceData, dashboard *datadogV1.Dashboard
...
+		_, hasValue := m["value"]
+		return hasUnit && hasValue
+	}
+
+	// Check for WidgetNewFixedSpan signature: type="fixed", from, to
+	if typeVal, hasType := m["type"]; hasType && typeVal == "fixed" {
+		_, hasFrom := m["from"]
+		_, hasTo := m["to"]
+		return hasFrom && hasTo
+	}
```
brtu, line 336: i think this can be condensed to a switch case: 
```
  if typeVal, hasType := m["type"]; hasType {
      switch typeVal {
      case "live":
          _, hasUnit := m["unit"]
          _, hasValue := m["value"]
          return hasUnit && hasValue
      case "fixed":
          _, hasFrom := m["from"]
          _, hasTo := m["to"]
          return hasFrom && hasTo
      }
  }
```

datadog/resource_datadog_dashboard.go
```diff
@@ -6395,6 +6426,18 @@ func getTimeseriesDefinitionSchema() map[string]*schema.Schema {
 			},
 		},
 	}
+	schema = mergeSchemas(schema, getCommonSpanSchema())
+	return schema
+}
+
+func mergeSchemas(schemas ...map[string]*schema.Schema) map[string]*schema.Schema {
+	merged := make(map[string]*schema.Schema)
+	for _, schema := range schemas {
```
brtu: rename `schema` to something else (`s`?). to avoid confusion with the `schema` package name on `schema.Schema`?

datadog/resource_datadog_dashboard.go
```diff
@@ -313,6 +314,55 @@ func updateDashboardState(d *schema.ResourceData, dashboard *datadogV1.Dashboard
 	return nil
 }
 
+// isWidgetTimeUnparsedObject checks if an unparsed object is a WidgetTime.
+// WidgetTime can be unparsed due to oneOf ambiguity between legacy and new formats.
```
brtu, line 318: hm do we need to support `CalendarAlignedSpan` here?  or maybe a follow up PR in the future?

datadog/resource_datadog_dashboard.go
```diff
@@ -9732,12 +9768,116 @@ func buildTerraformWidgetEvents(datadogWidgetEvents *[]datadogV1.WidgetEvent) *[
...
+					if hic, ok := legacyLiveSpan.GetHideIncompleteCostDataOk(); ok && *hic {
+						terraformDefinition["hide_incomplete_cost_data"] = true
+					}
+				}
+			}
+		}
+	}
+}
+
+func getCommonSpanSchema() map[string]*schema.Schema {
```
brtu: rename to `getCommonTimeSpanSchema`?

datadog/tests/provider_test.go
```diff
@@ -53,6 +53,7 @@ var isTestOrgC *bool
 var allowedHeaders = map[string]string{"Accept": "", "Content-Type": ""}
 
 var testFiles2EndpointTags = map[string]string{
+	"resource_datadog_dashboard_widget_time_test.go":                                     "dashboard",
```
brtu, line 56: should this be under the `tests` folder?

.github/workflows/sdkv2_deprecation.yaml
yacomink: This is a linting rule that assumes any .go file in the top level directory defines a terraform resource, but I added some unit tests for logic in the dashboard resource, and the test file was triggering the linter.

datadog/tests/provider_test.go
yacomink: I added these unit tests because the logic I added to `resource_datadog_dasboard.go` cannot be tested with a VCR test since they test some round trips that can only happen if there's a manual edit via the web ui.

If the test is in the `tests` directory, it's no longer in the same go module as the dashboard resource, so it cannot access private functions. So to put it in the tests directory I would have to refactor a large chunk of `resource_datadog_dasboard.go` to expose functions outside the module.

go's core testing package excludes _test.go files from regular builds

datadog/tests/resource_datadog_dashboard_time_spans_test.go
```diff
@@ -0,0 +1,196 @@
+package test
```
brtu, line 1: rename file to `resource_datadog_dashboard_time_spans_test`?

datadog/resource_datadog_dashboard.go
yacomink: nice

brtu (approved):
LGTM!

datadog/resource_datadog_dashboard.go
yacomink: I don't think that one is supported in the openapi schema yet https://github.com/DataDog/datadog-api-spec/blob/2f5e1501bbaa395774c7d59a15d586d607f9c231/spec/v1/dashboard.yaml#L4567-L4572

fpighi (approved):

PR #3417, opened 2026-01-10
[datadog_api_key] Add store_key option to prevent storing API key in TF state

LorisFriedel:
### What does this PR do?

Adds a new `store_key` attribute to the `datadog_api_key` resource that controls whether the API key value is stored in Terraform state.

### Motivation

Some organizations have security policies that prohibit storing secrets in Terraform state. This change allows users to provision API keys without persisting the sensitive key value.

### Changes

- Add `store_key` boolean attribute (defaults to `true` for backwards compatibility)
- When `store_key = false`, the `key` attribute is set to null instead of the actual value
- Add acceptance test for `store_key = false` behavior
- Update documentation with example usage

### Usage

```hcl
resource "datadog_api_key" "secure" {
  name      = "my-api-key"
  store_key = false
}
```

### Backwards Compatibility

This is a non-breaking change. The default value of store_key is true, preserving existing behavior.


docs/resources/api_key.md
```diff
@@ -29,6 +29,7 @@ resource "datadog_api_key" "foo" {
 ### Optional
 
 - `remote_config_read_enabled` (Boolean) Whether the API key is used for remote config. Set to true only if remote config is enabled in `/organization-settings/remote-config`.
+- `store_key` (Boolean) Whether to store the API key value in Terraform state. Set to `false` to avoid storing the key in state for security purposes. Defaults to `true` for backwards compatibility. When set to `false`, the `key` attribute will be empty. Defaults to `true`.
```
rtrieu, line 32:
```suggestion
- `store_key` (Boolean) Whether to store the API key value in Terraform state. Set to `false` to avoid storing the key in state for security purposes. Defaults to `true` for backwards compatibility. When set to `false`, the `key` attribute is empty. Defaults to `true`.
```

rtrieu: is the backwards capability necessary information? also, "Defaults to `true`" is mentioned twice here

PR #3424, opened 2026-01-13
[oncall] Add on_call_user_notification_rule resource

kevinconaway:
This commit adds the `on_call_user_notification_rule` resource for allowing users to create On-Call notification rules


docs/resources/on_call_user_notification_rule.md
```diff
@@ -0,0 +1,38 @@
...
+Provides a Datadog On-Call user notification rule resource.
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `category` (String) Notification category to associate the rule with Valid values are `high_urgency`, `low_urgency`.
```
buraizu:
```suggestion
- `category` (String) Notification category to associate the rule with. Valid values are `high_urgency`, `low_urgency`.
```

docs/resources/on_call_user_notification_rule.md
```diff
@@ -0,0 +1,38 @@
...
+- `channel_id` (String) ID of the notification channel to associate the notification rule with.
+- `delay_minutes` (Number) Number of minutes that will elapse before this rule is evaluated.  0 indicates immediate evaluation
```
buraizu:
```suggestion
- `delay_minutes` (Number) Number of minutes to elapse before this rule is evaluated. `0` indicates immediate evaluation.
```

docs/resources/on_call_user_notification_rule.md
```diff
@@ -0,0 +1,38 @@
...
+- `user_id` (String) ID of the user to associate the notification rule with.
+
+### Optional
+
+- `phone` (Block, Optional) (see [below for nested schema](#nestedblock--phone))
```
buraizu, line 27:
```suggestion
- `phone` (Block, Optional) (see [below for nested schema](#nestedblock--phone)).
```

docs/resources/on_call_user_notification_rule.md
```diff
@@ -0,0 +1,38 @@
...
+### Read-Only
+
+- `id` (String) The ID of this resource.
+
+<a id="nestedblock--phone"></a>
+### Nested Schema for `phone`
+
+Required:
+
+- `method` (String) Specifies the method in which a phone is used in a notification rule Valid values are `sms`, `voice`.
```
buraizu:
```suggestion
- `method` (String) Specifies the method in which a phone is used in a notification rule. Valid values are `sms`, `voice`.
```

buraizu (approved):
Flagging some missing and inconsistent punctuation, but otherwise approving

docs/resources/on_call_user_notification_rule.md
kevinconaway: The docs here are generated so I don't think I can modify this one

scripts/generate-docs.sh
```diff
@@ -8,6 +8,7 @@ exclude_files=(
   "docs/resources/on_call_schedule.md"
   "docs/resources/on_call_team_routing_rules.md"
   "docs/resources/on_call_user_notification_channel.md"
+  "docs/resources/on_call_user_notification_rule.md"
```
fpighi, line 11: What's the reason to exclude these from docs generation? I missed a few previous PRs so would appreciate some context.
It would be ideal to have them auto-generating correctly for ease of maintenance.

kevinconaway: https://github.com/DataDog/terraform-provider-datadog/pull/3077 has the original context. 

The short summary is that we use the `oneof` pattern for some configuration resources where the _block_ is optional (in this case phone) but if the block is specific, then the fields are required.  However, we can't mark the field required because then that forces the block to be required as well.

What we're doing is manually validating this in our resources and changing the documentation to reflect that the fields are required, not optional.

fpighi (approved):

PR #3425, opened 2026-01-13
[datadog_api_key] Add encryption helper and use it to encrypt datadog api key in state

LiuVII:
[APIR-2186]

## Summary

Adds encryption support for computed secrets in Terraform state using write-only attributes and ephemeral resources (Terraform 1.11+). This allows API keys and other sensitive values to be encrypted before storing in state, with the encryption key never persisted, enabling secure secret transfer to external secret managers.

## Motivation

- To avoid storing secrets in terraform state and enhance security posture.
- To address customers concerns, as mentioned in this [issue](https://github.com/DataDog/terraform-provider-datadog/issues/2916)

### Encryption Package (`secretbridge`)
- Added `datadog/internal/secretbridge` package providing AES-256-GCM encryption
- Reusable `EncryptionKeyAttribute()` for consistent write-only attribute definitions
- `Encrypt()` and `Decrypt()` functions with proper error handling and diagnostics

### API Key Resource Encryption
- Added `encryption_key_wo` write-only attribute (32-byte key from ephemeral resource)
- Added `encrypted_key` computed attribute containing encrypted API key value
- `key` attribute becomes null when encryption is enabled
- Resource-level `ModifyPlan` handles encryption transition states (adding/removing encryption)
- Graceful handling when encryption is removed with user warning

### Ephemeral Resource
- Added `datadog_secret_decrypt` ephemeral resource to decrypt stored ciphertext
- Takes encrypted value + encryption key, outputs plaintext for secret manager integration
- Provider framework wired up with `EphemeralResources()` method

### Documentation & Examples
- Generated documentation for the new ephemeral resource
- Added usage example showing API key → encryption → AWS Secrets Manager workflow

## Local Test Plan
- [x] Unit and Acceptance tests are passing 
- [x] Verify encryption with: create API key with `encryption_key_wo`, confirm `key` is null and `encrypted_key` is populated
- [x] Verify decryption: use `datadog_secret_decrypt` ephemeral to decrypt the `encrypted_key` value
- [x] Test encryption transitions: add encryption to existing resource, remove encryption

[APIR-2186]: https://datadoghq.atlassian.net/browse/APIR-2186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -21,6 +21,35 @@ Terraform provides helpful [Extending Terraform][1] documentation for best pract
   - One that will have a singular name (ex: `datadog_user`) which returns exactly one objects (and fails if there is 0 or more than 0)
   - One that will have a plural name (ex: `datadog_users`) which returns a list of objects and succeed in all cases (0, 1 or more than 1 objects)
 
+### Ephemeral Resources and Secrets
+
+#### When to Use Ephemerals
+
+Use ephemeral resources when a value should:
+- Never be stored in Terraform state
+- Be available only during apply
+- Be passed to other ephemeral-aware resources
+- Can be returned by the API outside of the initial resource creation (unlike one-time-read secrets)
+
+See [Terraform Ephemeral Resources][17] for concepts.
+
+#### SecretBridge Pattern
+
+For resources that receive sensitive computed values from APIs (like `datadog_api_key`) and can return these values only on creation (one-time-read), use the secretbridge library to encrypt-on-create:
+
+1. Add `encryption_key_wo` attribute using `secretbridge.EncryptionKeyAttribute()`
+2. On Create, encrypt the API response using `secretbridge.Encrypt()`
+3. Store encrypted value in `encrypted_<value_argument_name>`, set plaintext to null
+4. Users decrypt via `datadog_secret_decrypt` ephemeral resource
+
+See [secretbridge/README.md][18] for API details.
+
+#### Important Constraints
+
+- **Transit only**: Encrypted values are for one-time transfer to secret managers
+- **No key rotation**: Encryption algorithm may change without notice
+- **Terraform 1.11+**: Required for write-only attributes
+
 ## Makefile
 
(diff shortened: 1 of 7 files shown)


urnfdog:
Given the usage of encryption, have you had a discussion with the security engineering team?  

AES-256-GCM is fine for compliance but I wonder if the implementation needs a review (nonce from crypto/rand is good and it appears that [random_password "does use a cryptographic random number generator",](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) and log2(62^32) ~= 191 gets you good entropy but I didn't take a deeper look or threat modeling for key interception between secretbridge and ephemeral decrypt helper etc.).

LiuVII:
> Given the usage of encryption, have you had a discussion with the security engineering team?
> 
> AES-256-GCM is fine for compliance but I wonder if the implementation needs a review

I don't believe we necessary need one (and I'll explain the reasoning below) but if you (or anyone else) believes we really should, I'm down for it 🙂 

**Why I think we don't really need security review**
The encryption is for "in transit" to proper secret manager only and no encryption key rotation or permanent state decryption is or going to be supported:
1. If the proper usage of secret manager is done, the encryption key can be thrown away, and whatever encrypted left in a state has zero meaning and can be even "nulled" with `encryption_key_wo` field unset.
2. What I now believe missing here are warning and/or guides (🤔) to suggest that we can change encryption algorithm anytime down the line, so whatever going to be left in state won't be usable on any new release and we won't consider this a "breaking change" because using terraform state as a sort of secret manager is not something we are going to support. 

I'd even explicitly unset the secret after it's decrypted once to avoid reliance on whatever is in state, but this adds complexity and friction if terraform fails midway, so this implementation steers away from this.

LiuVII:
Updated docs and added warnings (cc @urnfdog )

LiuVII:
The security team ruled this approach at this time to be somewhat problematic because of the use of encryption.

PR #3436, opened 2026-01-19
[datadog_observability_pipeline] Validate framing for socket source and destination

vladimir-dd:


datadog/fwprovider/observability_pipeline/socket_framing.go
```diff
@@ -14,3 +18,52 @@ type SocketFramingModel struct {
...
+		return
+	}
+	method := methodAttr.(types.String).ValueString()
+
+	charDelimitedAttr, hasCharDelimitedAttr := attrs["character_delimited"]
+	hasCharacterDelimited := false
+	if hasCharDelimitedAttr && !charDelimitedAttr.IsNull() {
+		if list, ok := charDelimitedAttr.(types.List); ok {
+			hasCharacterDelimited = len(list.Elements()) > 0
+		}
```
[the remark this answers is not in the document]
vladimir-dd: in this case, it does not make sense to define several `character_delimited` blocks, so this use-case is very unrealistic and it does not make sense to complicate the code.

datadog/fwprovider/observability_pipeline/socket_destination.go
```diff
@@ -31,6 +33,11 @@ func ExpandSocketDestination(ctx context.Context, id string, inputs types.List,
 	s.SetMode(datadogV2.ObservabilityPipelineSocketDestinationMode(src.Mode.ValueString()))
 	s.SetEncoding(datadogV2.ObservabilityPipelineSocketDestinationEncoding(src.Encoding.ValueString()))
 
+	if len(src.Framing) == 0 {
```
taylorchandleryoung, line 36: I was expected a `required: true` on the schema. Do we have to do it here like this because its a list?

taylorchandleryoung (approved):

datadog/fwprovider/observability_pipeline/socket_destination.go
clementd-dd: or a sizeAtLeast(1) validator

vladimir-dd: yeah, I think this fix is not ideal, since it'd require similar changes in every other place, where we use `sizeAtLeast(1)`, which turns out is not working when the block is missing. I'll close this PR for now and try to find a more generic solution if possible.

PR #3447, opened 2026-01-22
[datadog_synthetics_suite] Create Synthetics suite resource

romainberger:
Create a new `datadog_synthetics_suite` resource to manage Synthetics Test Suites.

### DIFF
--- a/docs/resources/synthetics_suite.md
+++ b/docs/resources/synthetics_suite.md
@@ -23,7 +23,7 @@ Provides a Datadog Synthetics Suite resource. This can be used to create and man
 
 - `message` (String) Message of the Synthetics suite.
 - `options` (Block List) Options for the Synthetics suite. (see [below for nested schema](#nestedblock--options))
-- `tags` (List of String) A list of tags to associate with your synthetics suite.
+- `tags` (Set of String) A set of tags to associate with your synthetics suite.
 - `tests` (Block List) List of tests in the Synthetics suite. Can be empty but the field is always sent to the API. (see [below for nested schema](#nestedblock--tests))
 
 ### Read-Only
@@ -35,7 +35,7 @@ Provides a Datadog Synthetics Suite resource. This can be used to create and man
 
 Required:
 
-- `alerting_threshold` (Number) Alerting threshold for the suite. Must be between 0 and 1. Value must be between 0.000000 and 1.000000.
+- `alerting_threshold` (Number) Alerting threshold for the suite. Value must be between 0.000000 and 1.000000.
 
 
 <a id="nestedblock--tests"></a>
@@ -47,4 +47,4 @@ Required:
 
 Optional:
 
-- `alerting_criticality` (String) Alerting criticality for the test. Valid values are `ignore`, `critical`. Valid values are `ignore`, `critical`.
+- `alerting_criticality` (String) Alerting criticality for the test. Valid values are `ignore`, `critical`.


datadog/internal/fwutils/fw_enrich_schema.go
```diff
@@ -229,12 +229,24 @@ func buildEnrichedSchemaDescription(rv reflect.Value) {
 				break
 			}
 
-			// Int64 validators
+			// Float64 and Int64 validators (both use "betweenValidator" and "atLeastValidator" names)
+			// Try Float64 first, then fall back to Int64
```
romainberger, line 233: This change is made to allow using a `betweenValidator` on Float, as it currently only works with int. We need [this validator here](https://github.com/DataDog/terraform-provider-datadog/pull/3447/changes#diff-33e2d070befad67fd81e58b1f242a4371ea99af436e375ed93dfdb80bfe3fd98R104)

datadog/fwprovider/resource_datadog_on_call_user_notification_channel.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+
```
romainberger, line 17: I think all changes on the on-call resource have been made when running `make fmt`

datadog/fwprovider/resource_datadog_synthetics_suite.go
```diff
@@ -0,0 +1,384 @@
...
+
+func NewSyntheticsSuiteResource() resource.Resource {
+	return &syntheticsSuiteResource{}
+}
+
+type SyntheticsSuiteModel struct {
+	ID      types.String `tfsdk:"id"`
+	Name    types.String `tfsdk:"name"`
+	Message types.String `tfsdk:"message"`
+	Tags    types.List   `tfsdk:"tags"`
```
LiuVII: **❓ question:** ‏ why not `types.Set`?

datadog/fwprovider/resource_datadog_synthetics_suite.go
```diff
@@ -0,0 +1,384 @@
...
+				Optional:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"options": schema.ListNestedBlock{
+				Description: "Options for the Synthetics suite.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"alerting_threshold": schema.Float64Attribute{
+							Description: "Alerting threshold for the suite. Must be between 0 and 1.",
```
LiuVII:
```suggestion
							Description: "Alerting threshold for the suite.",
```
as validation is added automatically by enrichment

datadog/fwprovider/resource_datadog_synthetics_suite.go
```diff
@@ -0,0 +1,384 @@
...
+			"tests": schema.ListNestedBlock{
+				Description: "List of tests in the Synthetics suite. Can be empty but the field is always sent to the API.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"public_id": schema.StringAttribute{
+							Description: "Public ID of the test.",
+							Required:    true,
+						},
+						"alerting_criticality": schema.StringAttribute{
+							Description: "Alerting criticality for the test. Valid values are `ignore`, `critical`.",
```
LiuVII:
```suggestion
							Description: "Alerting criticality for the test.",
```
as validation is added automatically by enrichment

datadog/fwprovider/resource_datadog_synthetics_suite.go
```diff
@@ -0,0 +1,384 @@
...
+	if attrs, ok := data.GetAttributesOk(); ok {
+		// Update name
+		if name, ok := attrs.GetNameOk(); ok {
+			state.Name = types.StringValue(*name)
+		}
+
+		// Update message
+		if message, ok := attrs.GetMessageOk(); ok {
+			state.Message = types.StringValue(*message)
+		}
```
LiuVII, line 339: **💬 suggestion:** ‏ iiuc if someone removes message via the Datadog UI, a terraform refresh won't reflect that change. Consider adding an else clause:
```suggestion
		if message, ok := attrs.GetMessageOk(); ok {
			state.Message = types.StringValue(*message)
		} else {
			state.Message = types.StringNull()
		}
```

romainberger: The `message` is required in the backend, so at least it will be an empty string. I might be mistaken [but wouldn't that mean `GetMessageOk` should never return `false`](https://github.com/DataDog/datadog-api-client-go/blob/da8bc38cc57cbe6afbfde59abd39599344684b84/api/datadogV2/model_synthetics_suite.go#L71)?

datadog/fwprovider/resource_datadog_synthetics_suite.go
romainberger: No reason, all of our resources are using `types.List` for tags, but it does make sense to use Set, changing that 👍

datadog/fwprovider/resource_datadog_synthetics_suite.go
LiuVII: ah, this makes a good point, I overlooked this, makes sense

PR #3450, opened 2026-01-22
[AWSCORE-663] Remove deprecated resources

dhan0779:
For the 4.0 release, we should remove the following deprecated resources:
- datadog_integration_aws, 
- datadog_integration_aws_tag_filter, 
- datadog_integration_aws_log_collection, 
- datadog_integration_aws_lambda_arn


dhan0779:
Some of the testing is still iffy when switching from the `datadog_integration_aws` resource to the `datadog_integration_aws_account` resource. 

For the EventBridge test, I tested using the sandbox account (694513996623), which works, but I think the cassette tests still verify that the body/params of the request need to be the same. This means, we still need to hardcode that value into the test instead of using the random generated AWS account.

For the logs archive test, I just decided to mark them as skipped since those also require S3 permissions for buckets.

PR #3453, opened 2026-01-23
[datadog_cloud_inventory_sync_config] move to v2 apis from unstable endpoints | STGMON-352

denozor-dog:
move from unstable endpoints to v2 apis


LiuVII:
the cassette wasn't re-recorded, right? IMO this should be checked against real API

PR #3459, opened 2026-01-26
[datadog_observability_pipeline] Adding datastream to opensearch and updating exactlyoneof rules

taylorchandleryoung:
### Motivation

Adding datastream option to Open Search Destination
Added Config validator to enforce exactlyOneOf

### DIFF
--- a/datadog/fwprovider/resource_datadog_observability_pipeline.go
+++ b/datadog/fwprovider/resource_datadog_observability_pipeline.go
@@ -5527,23 +5527,16 @@ func flattenOpenSearchDestination(ctx context.Context, src *datadogV2.Observabil
 		return nil
 	}
 
-	out := &opensearchDestinationModel{}
-	if v, ok := src.GetBulkIndexOk(); ok {
-		out.BulkIndex = types.StringValue(*v)
+	out := &opensearchDestinationModel{
+		BulkIndex: types.StringPointerValue(src.BulkIndex),
 	}
 
 	if ds, ok := src.GetDataStreamOk(); ok && ds != nil {
-		dsModel := opensearchDestinationDataStreamModel{}
-		if v, ok := ds.GetDtypeOk(); ok {
-			dsModel.Dtype = types.StringValue(*v)
-		}
-		if v, ok := ds.GetDatasetOk(); ok {
-			dsModel.Dataset = types.StringValue(*v)
-		}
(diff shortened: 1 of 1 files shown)


datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -23,8 +22,10 @@ import (
 )
 
 var (
-	_ resource.ResourceWithConfigure   = &observabilityPipelineResource{}
-	_ resource.ResourceWithImportState = &observabilityPipelineResource{}
+	_ resource.ResourceWithConfigure        = &observabilityPipelineResource{}
```
vladimir-dd: I think we should use `objectvalidator.ExactlyOneOf()` defined in the component schema.

taylorchandleryoung: thats what I did in the first commit, but it was applying even if the destination didn't exist. It expects all configs to have exactly one of those fields. I may also jsut have been setting it up wrong

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -2185,6 +2188,7 @@ func (r *observabilityPipelineResource) Schema(_ context.Context, _ resource.Sch
 													},
 													Validators: []validator.List{
 														listvalidator.SizeAtMost(1),
+														listvalidator.ConflictsWith(frameworkPath.MatchRelative().AtParent().AtName("bulk_index")),
```
vladimir-dd, line 2191: neat

datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -5489,6 +5532,20 @@ func flattenOpenSearchDestination(ctx context.Context, src *datadogV2.Observabil
...
+		}
+		if v, ok := ds.GetDatasetOk(); ok {
+			dsModel.Dataset = types.StringValue(*v)
+		}
+		if v, ok := ds.GetNamespaceOk(); ok {
+			dsModel.Namespace = types.StringValue(*v)
+		}
+		out.DataStream = []opensearchDestinationDataStreamModel{dsModel}
+	}
+
```
vladimir-dd, line 5541: nit: I think we could just initialize this struct with all the fields - these null checks are unnecessary, because  `StringValue` handles null pointers well.

PR #3471, opened 2026-01-29
fix: prevent deletion of protected logs index (fixes #3470)

### LINKED ISSUE #3470: Prevent Deletion of main datadog_logs_index

datadog_logs_index

### Feature Request

Deleting the main datadog_logs_index is an irreversible action and safeguards should be put in place to prevent it's deletion

### References

_No response_

AndrewCharlesHay:
### Description

This PR adds a new optional argument `prevent_deletion` to the `datadog_logs_index` resource.

When set to `true`, this flag safeguards the index from being deleted by Terraform. Any attempt to destroy the resource will result in a diagnostic error, preventing accidental loss of critical indexes (such as the default `main` index).

Fixes #3470

### Schema Changes

*   **`prevent_deletion`** (Boolean, Optional): If true, the logs index cannot be deleted. Defaults to `false`.

### Use Case

To protect the main index:

```hcl
resource "datadog_logs_index" "main" {
  name             = "main"
  filter           { query = "*" }
  prevent_deletion = true
}
```

If a user attempts `terraform destroy` or removes this block, the provider will return:
`Deletion of logs index 'main' is prevented by 'prevent_deletion' flag`

To allow deletion, the user must explicitly set `prevent_deletion = false` and apply, before destroying.


datadog/resource_datadog_logs_index.go
```diff
@@ -103,6 +103,11 @@ var indexSchema = map[string]*schema.Schema{
 			Schema: exclusionFilterSchema,
 		},
 	},
+	"prevent_deletion": {
+		Description: "If true, the logs index cannot be deleted. Defaults to false.",
+		Type:        schema.TypeBool,
+		Optional:    true,
```
fpighi, line 116: Having an explicit default here would make it consistent and auto-generated docs would benefit from it too, so the "Defaults to false" description could be removed.

PR #3473, opened 2026-01-29
[datadog_observability_pipeline] reshape auth block for Amazon OpenSearch

vladimir-dd:
Refines Amazon OpenSearch auth block shape to use nested auth.aws block.

### DIFF
--- a/datadog/fwprovider/resource_datadog_observability_pipeline.go
+++ b/datadog/fwprovider/resource_datadog_observability_pipeline.go
@@ -5528,18 +5528,10 @@ func expandAmazonOpenSearchDestination(ctx context.Context, dest *destinationMod
 		}
 		if len(authSrc.Aws) > 0 {
 			awsSrc := authSrc.Aws[0]
-			if !awsSrc.AwsRegion.IsNull() {
-				auth.AwsRegion = awsSrc.AwsRegion.ValueStringPointer()
-			}
-			if !awsSrc.AssumeRole.IsNull() {
-				auth.AssumeRole = awsSrc.AssumeRole.ValueStringPointer()
-			}
-			if !awsSrc.ExternalId.IsNull() {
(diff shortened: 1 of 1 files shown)


datadog/fwprovider/resource_datadog_observability_pipeline.go
```diff
@@ -5540,15 +5552,20 @@ func flattenAmazonOpenSearchDestination(src *datadogV2.ObservabilityPipelineAmaz
...
-			Strategy:    types.StringValue(string(src.Auth.Strategy)),
-			AwsRegion:   types.StringPointerValue(src.Auth.AwsRegion),
-			AssumeRole:  types.StringPointerValue(src.Auth.AssumeRole),
-			ExternalId:  types.StringPointerValue(src.Auth.ExternalId),
-			SessionName: types.StringPointerValue(src.Auth.SessionName),
-		},
+	authModel := amazonOpenSearchAuthModel{
+		Strategy: types.StringValue(string(src.Auth.Strategy)),
+	}
+	if string(src.Auth.Strategy) == "aws" {
```
taylorchandleryoung, line 5558: so the original ticket just had region as something that should be marked as required instead of optional. What is the new shape giving us? I would assume we'd mostly want to stay as similarly shaped as the API is possible. Its a pretty direct mapping in the old version but now we have a translation?

vladimir-dd: oh, right, I see what you mean. The idea was that we move all aws-specific fields to a new `aws` block and mark `aws_region` required there. However, your question made me realise that `aws_region` has nothing to do with the aws auth(see the shared aws [auth](https://github.com/DataDog/terraform-provider-datadog/blob/b0de1da07a1bf5d84fbd44185b28d3e05c8a7c93/datadog/fwprovider/observability_pipeline/aws_auth.go) block), so it makes it even more complicated. I'll close this PR to be consistent with the API at least.

Issue #3435, opened 2026-01-19
Cannot replace deprecated host_filters with monitored_resource_configs in datadog_integration_gcp


ivandelcz:
### Datadog Terraform Provider Version

v3.84.0

### Terraform Version

1.13.5

### What resources or data sources are affected?

- datadog_integration_gcp

### Terraform Configuration Files

```hcl
locals {
  integration_gcp_projects = {
    "demo" : {
      "filters" : ["!goog-gke-node"],
    }
    "dev" : {
      "filters" : ["!goog-gke-node"],
    }
  }
}


resource "datadog_integration_gcp" "integrations_gcp" {
  for_each       = local.integration_gcp_projects
  project_id     = "project_id"
  private_key    = "private_key"
  private_key_id = "private_key_id"
  client_email   = "client_email"
  client_id      = "client_id"

  dynamic "monitored_resource_configs" {
    for_each = try(each.value["filters"], null) != null ? [1] : []
    content {
      filters = each.value["filters"]
      type    = "gce_instance"
    }
  }

  is_resource_change_collection_enabled = false
  resource_collection_enabled           = false
}
```

### Expected Behavior

Any change in `host_filters` should not fail, given that the new `monitored_resource_configs` deployed in https://github.com/DataDog/terraform-provider-datadog/pull/3219 are set.

### Actual Behavior

```
Terraform will perform the following actions:

  # datadog_integration_gcp.integrations_gcp["dev"] will be updated in-place
  ~ resource "datadog_integration_gcp" "integrations_gcp" {
      ~ cloud_run_revision_filters            = [] -> (known after apply)
      - host_filters                          = "!goog-gke-node" -> "
        id                                    = "dev"
        # (11 unchanged attributes hidden)
    }

  # datadog_integration_gcp.integrations_gcp["demo"] will be updated in-place
  ~ resource "datadog_integration_gcp" "integrations_gcp" {
      ~ cloud_run_revision_filters            = [] -> (known after apply)
      - host_filters                          = "!goog-gke-node" -> ""
        id                                    = "demo"
        # (11 unchanged attributes hidden)
    }
```

Terraform apply result

```
Error: Provider produced inconsistent result after apply
When applying changes to datadog_integration_gcp.integrations_gcp["dev"], provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced an unexpected new value: .host_filters: was cty.StringVal(""), but now cty.StringVal("!goog-gke-node").

[...]

PR #3489, opened 2026-02-03
[datadog_integration_gcp] Remove default value for HostFilters

katherinekim-51:

## Description

https://datadoghq.atlassian.net/browse/GCP-3282 
Fixes https://github.com/DataDog/terraform-provider-datadog/issues/3435

Right now, if a user tries to "clear" host filters on update (ie. user omits them because it is deprecated and they moved to using mrcs), the terraform provider returns an _inconsistent plan after apply_. This fixes that issue 

## Implementation Details

From a terraform pov, this was happening because we had a default value being set for host filters as "", which defeated the purpose of setting this to computed; so we pass "" in the update, in the server when we merge host_filters with mrcs host_filters becomes computed with "xyz" and to terraform that looks inconsistent. This removes the default value and updates the request body to only be set when the value is known. 

## Testing Guidelines

Added test and ran cassettes


datadog/tests/resource_datadog_integration_gcp_test.go
```diff
@@ -14,6 +14,92 @@ import (
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/fwprovider"
 )
 
+func TestAccDatadogHostFiltersBug(t *testing.T) {
```
ash-ddog: Can you link the GitHub issue here and add to the PR description? (i.e fixes <issue_link>)

ash-ddog: Can you also rename the test to be more explicit about _which_ bug is being fixed?

i.e. `TestLegacy_WhenReplacingHostFiltersWithMRC_Succeeds` or something like this

PR #3498, opened 2026-02-05
[datadog_observability_pipeline] Enforce only one type per processor

taylorchandleryoung:
### Motivation
 The user can currently issue a malformed terraform request that returns in a different shape and terraform responds with an error with the provider. 
 
 We should prevent this terraform shape.
 
 ### Changes
 Using cursor: had it add a validator that checks for all non default processor fields. If we do this based off of processor names, we would need to update this every time we add a new processor.
 
 Alternative considered
 
 `Validators: []validator.Object{
    objectvalidator.ExactlyOneOf(
        path.MatchRelative().AtName("sample"),
        path.MatchRelative().AtName("quota"),
        path.MatchRelative().AtName("filter"),
        // ... all 22+ processor types
    ),
}`

This runs in the same issue mentioned above that it is fragile when we add new processor types


datadog/fwprovider/observability_pipeline/processor_validator.go
tessneau, line 1: tioli: maybe this could fit nicely in `processor_common.go` since its pretty small, to keep all processor things together?

tessneau (approved):
nice ! 🔥

datadog/fwprovider/observability_pipeline/processor_validator.go
taylorchandleryoung: updated!

PR #3510, opened 2026-02-11
[datadog_observability_pipeline] Buffer support on Observability Pipelines

taylorchandleryoung:
### Overview

Updating the terraform provider to support buffer options on Observability pipelines

### DIFF
--- a/docs/resources/observability_pipeline.md
+++ b/docs/resources/observability_pipeline.md
@@ -420,7 +420,36 @@ Optional:
 
 Optional:
 
+- `buffer` (Block List) Configuration for buffer settings on destination components. Exactly one of `disk` or `memory` must be specified. (see [below for nested schema](#nestedblock--config--destination--datadog_logs--buffer))
 - `routes` (Block List) A list of routing rules that forward matching logs to Datadog using dedicated API keys. (see [below for nested schema](#nestedblock--config--destination--datadog_logs--routes))
+
+<a id="nestedblock--config--destination--datadog_logs--buffer"></a>
+### Nested Schema for `config.destination.datadog_logs.buffer`
+
+Optional:
+
+- `disk` (Block List) Options for configuring a disk buffer. Cannot be used with `memory`. (see [below for nested schema](#nestedblock--config--destination--datadog_logs--buffer--disk))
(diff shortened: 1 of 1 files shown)


datadog/fwprovider/observability_pipeline/buffer.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if len(src.MemoryBuffer) > 0 {
+		memBuf := src.MemoryBuffer[0]
+
+		if !memBuf.MaxEvents.IsNull() {
+			buffer := datadogV2.NewObservabilityPipelineMemoryBufferSizeOptionsWithDefaults()
+			buffer.SetType(datadogV2.ObservabilityPipelineBufferOptionsMemoryType("memory"))
+			buffer.SetMaxEvents(memBuf.MaxEvents.ValueInt64())
+			if !memBuf.WhenFull.IsNull() {
+				// buffer.SetWhenFull(datadogV2.ObservabilityPipelineBufferOptionsWhenFull(memBuf.WhenFull.ValueString()))
```
clementd-dd: should this stay commented out?

datadog/fwprovider/observability_pipeline/buffer.go
```diff
@@ -0,0 +1,182 @@
...
+								Validators: []validator.Int64{
+									int64validator.ExactlyOneOf(frameworkPath.MatchRelative().AtParent().AtName("max_events")),
+								},
+							},
+							"max_events": schema.Int64Attribute{
+								Optional:    true,
+								Description: "Maximum events for the memory buffer.",
+							},
+							"when_full": schema.StringAttribute{
+								Optional:    true,
```
clementd-dd, line 170: i think it will require a default value as well, since the api will set it to `block` when it’s not set, which will make terraform report an inconsistency after apply.

clementd-dd (approved):
This generally LGTM, but in my tests, not setting a default value on `when_full` made terraform complain about inconsistencies

datadog/fwprovider/observability_pipeline/buffer.go
taylorchandleryoung: thanks for catching! will remove the comment

datadog/fwprovider/observability_pipeline/buffer.go
taylorchandleryoung: Added default

PR #3519, opened 2026-02-13
[datadog_logs_index] Add support for tags on logs indexes

LorisFriedel:
## Summary
- Add `tags` attribute to the `datadog_logs_index` resource, allowing users to associate `key:value` tags with their logs indexes
- Add `tags` to the `datadog_logs_indexes` data source so tags are exposed when reading index configurations
- Tags are optional, support create/update/delete lifecycle, and are validated server-side (max 20, `key:value` format)

## Changes
- **`datadog/resource_datadog_logs_index.go`**
  - Add `tags` to the resource schema (`TypeSet`, optional)
  - Read tags into state via `updateLogsIndexState`
  - Set tags on create in `buildDatadogIndexCreateRequest`
  - Set or clear tags on update in `buildDatadogIndexUpdateRequest` (explicitly sends empty array when tags are removed)
- **`datadog/data_source_datadog_logs_indexes.go`**
  - Add `tags` to the data source schema (`TypeSet`, computed)
  - Include tags in the read output
- **`docs/resources/logs_index.md`** — Document `tags` attribute and add example usage
- **`docs/data-sources/logs_indexes.md`** — Document `tags` in the data source schema

## Test plan
- [x] `TestAccDatadogLogsIndex_WithTags` — create index with tags, update tags, remove all tags
- [x] Run acceptance tests with `RECORD=none` against live API


mikkoc:
I believe this PR might be the cause of our TF apply failures that we are encountering since updating the provider from 3.88 to 3.91.

```
Error: error updating logs index from /api/v1/logs/config/indexes/development: 400 Bad Request: {"error":{"code":"InvalidArgument","message":"Invalid tag configuration","details":[{"code":"InvalidArgument","message":"Tag format of ManagedBy:terraform is wrong"},{"code":"InvalidArgument","message":"Tag format of Env:development is wrong"}]}}
```

The plan shows:
```
  # module.datadog.datadog_logs_index.this will be updated in-place
  ~ resource "datadog_logs_index" "this" {
        id                                       = "development"
        name                                     = "development"
      ~ tags                                     = [
          + "Env:development",
          + "ManagedBy:terraform",
        ]
        # (5 unchanged attributes hidden)
```

These tags are defined at the provider level. I have tried to override them as an empty list: `tags = []` but it does not work. Any ideas?

LorisFriedel:
Format restrictions on tags for log indexes (and pipelines) are stricter and we're working on better consistency about that. In the meantime, I'd recommend to always use lowercased tags everywhere, as they are normalized anyway within Datadog: this will bring you closer to "what you see is what you get"! If you'd prefer not to, there is unfortunately no way to disable provider-level tag concatenation on specific resources.

PR #3525, opened 2026-02-18
[datadog_sensitive_data_scanner_rule] Fix constant diff on should_save_match

vinckama:
## Summary
- Fix perpetual plan diff on `text_replacement.should_save_match` field
- The API client (v2.55) now has a proper `ShouldSaveMatch` struct field, but the provider was still using `AdditionalProperties` to read/write it — causing the read path to never find the value
- Switch to `GetShouldSaveMatchOk()` / `SetShouldSaveMatch()` API client methods
- re-generate test cases as well in a [second commit](https://github.com/DataDog/terraform-provider-datadog/pull/3525/commits/8d949743d3a1098eb18466f0637d9e027af1ae14)

Fixes: https://datadoghq.atlassian.net/browse/SDS-2250


fbryden (approved):
LGTM!

Re-recording cassettes is not too fun 😓

fpighi (approved):

PR #3528, opened 2026-02-19
[datadog_observability_pipeline] Adding support for OCSF Mapper custom rules in oberservability pipelines provider

taylorchandleryoung:
### Changes 

Adding support for custom rules instead of just library rules to the OCSF Mapper

Schema, flattten, and unflatten added
new test


datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
```diff
@@ -0,0 +1,158 @@
...
+					Description: "List of OCSF mapping entries. Each entry uses either a library mapping or a custom mapping.",
+					NestedObject: schema.NestedBlockObject{
+						Attributes: map[string]schema.Attribute{
+							"include": schema.StringAttribute{
+								Required:    true,
+								Description: "Search query for selecting which logs the mapping applies to.",
+							},
+							"library_mapping": schema.StringAttribute{
+								Optional:    true,
+								Description: "Predefined library mapping for log transformation. Use this or custom_mapping, not both.",
```
clementd-dd, line 33: can this be encoded in a validator?

taylorchandleryoung: let me see if I can use conflictsWith

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
```diff
@@ -0,0 +1,158 @@
...
+			Blocks: map[string]schema.Block{
+				"mapping": schema.ListNestedBlock{
+					Description: "List of OCSF mapping entries. Each entry uses either a library mapping or a custom mapping.",
+					NestedObject: schema.NestedBlockObject{
+						Attributes: map[string]schema.Attribute{
+							"include": schema.StringAttribute{
+								Required:    true,
+								Description: "Search query for selecting which logs the mapping applies to.",
+							},
+							"library_mapping": schema.StringAttribute{
```
20agbekodo, line 31: The naming is `mapping` regardless of if it's library or custom

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
```diff
@@ -0,0 +1,158 @@
...
+								Optional:    true,
+								Description: "Predefined library mapping for log transformation. Use this or custom_mapping, not both.",
+							},
+						},
+						Blocks: map[string]schema.Block{
+							"custom_mapping": schema.ListNestedBlock{
```
20agbekodo, line 40: ditto

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
```diff
@@ -0,0 +1,158 @@
...
+													},
+													"profiles": schema.ListAttribute{
+														ElementType: types.StringType,
+														Optional:    true,
+														Description: "A list of OCSF profiles to apply.",
+													},
+												},
+											},
+										},
+										"field_mapping": schema.ListNestedBlock{
```
20agbekodo: same here, the name is `mapping`

20agbekodo: Might be worthwhile asking cursor to make sure that all the field namings are respected between TF and API Spec

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
taylorchandleryoung: It already was library mapping before my changes. also due to the rather large change between the library mapping shape and the custom mapping shape, I don't think the same name is wise to play well with  terraform

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
taylorchandleryoung: This one I probably can update to mapping easily, but I think its bad practice to have so many things at different levels and shapes all called just 'mapping'

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
taylorchandleryoung: added conflicts with

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
taylorchandleryoung: went ahead and updated this one so I could get tests and docs updated

datadog/fwprovider/observability_pipeline/ocsf_mapper_processor.go
20agbekodo: I see..I guess the only thing bothering me is that the TF naming is now not in sync with the Public API naming, which is a new behaviour. I'm concerned about introducing a shift between TF configs and API payloads.
That being said, OCSF is rarely used and differentiating the naming makes it easier to validate that users choose only one, so I'm okay to go with it.

PR #3534, opened 2026-02-23
[datadog_security_monitoring_rule] Add instantaneous_baseline to anomaly detection

hachem-dd:
- Add instantaneous_baseline option for security monitoring anomaly detection rules.

[Jira](https://datadoghq.atlassian.net/browse/SEC-27123)

### DIFF
--- a/datadog/resource_datadog_security_monitoring_rule.go
+++ b/datadog/resource_datadog_security_monitoring_rule.go
@@ -10,6 +10,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/validators"
 
 	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV2"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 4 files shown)


docs/data-sources/security_monitoring_rules.md
```diff
@@ -130,7 +130,7 @@ Read-Only:
 - `detection_tolerance` (Number)
 - `learning_duration` (Number)
 - `learning_period_baseline` (Number)
-
+- `instantaneous_baseline` (Boolean)
```
brett0000FF: Likely auto-generated, but flagging that this breaks alphabetical order in the list (nit).

hachem-dd: Fixed :)

datadog/resource_datadog_security_monitoring_rule.go
```diff
@@ -1169,10 +1207,22 @@ func buildPayloadAnomalyDetectionOptions(tfOptionsList []interface{}) (*datadogV
...
+	instantaneous := false
+	if v, ok := tfOptions["instantaneous_baseline"]; ok {
 		hasPayload = true
-		learningPeriodBaseline := int64(v.(int))
-		options.LearningPeriodBaseline = &learningPeriodBaseline
+		instantaneous = v.(bool)
+		options.SetInstantaneousBaseline(instantaneous)
+	}
+
+	// Only set learning_period_baseline when not using instantaneous baseline.
```
paul-hf: IMO this validation should be performed by our API rather than trying to do it here. Terraform has basic validation for enums, positive values for numeric fields, etc. but more complex logic that involves more than one field should be handled by our API. A good way forward here might be to use the rawConfig to make sure learning_period_baseline is only set when the user has actually included the field, then let the API validate whatever the user sends. Thoughts?

hachem-dd: This is the hack I had to add to avoid it sending `0` further down the line even if learning_period_baseline is not set.

I fully agree though

hachem-dd: Edit: here is an attempt with `GetRawConfigAt` instead of `GetRawConfig` which might be nicer: 1ec4f7520fe110023682c658aca1a6d7aea03468

paul-hf (approved):
Thanks!

PR #3551, opened 2026-02-27
Add PR-scoped integration tests with flaky test skip list

fpighi:
## Summary

- Add a new workflow (`test_pr_integration.yml`) that runs integration tests (live API, `RECORD=none`) on every PR, scoped to only the resources/datasources changed in that PR
- Introduce a flaky test skip list (`flaky_tests.yaml`) so known-flaky tests don't drown out signal in PR runs
- Apply the skip list to label-triggered (`ci/integrations`) runs in the existing `test_integration.yml` workflow; scheduled 12h runs continue executing everything

### How test selection works

`scripts/select_pr_tests.py` takes changed files from `git diff` and:
1. Identifies resource/datasource files (both SDKv2 and Framework)
2. Extracts the resource type name (e.g., `datadog_monitor`)
3. Searches all test files for references to that resource type (HCL blocks + test check references), catching cross-resource dependencies
4. Directly changed test files are always included
5. Shared infrastructure changes (e.g., `datadog/internal/`) are ignored — those rely on scheduled runs + `ci/integrations` label

The `--escape-for-make` flag doubles `$` anchors to `$$` so Make's recursive variable expansion doesn't corrupt the regex.

## Test plan

- [x] Unit tests for `select_pr_tests.py` (10 cases) and `build_skip_regex.py` (7 cases) — all passing
- [x] Open a PR touching a single resource, verify workflow triggers and runs only relevant tests
- [x] Open a PR not touching any resource, verify workflow skips (can be checked with current PR)

## How it was tested

Validated end-to-end via test PR #3553 which touched `resource_datadog_synthetics_concurrency_cap.go` (a resource with a single test).

### Workflow run: https://github.com/DataDog/terraform-provider-datadog/actions/runs/22568243705

**determine-tests** correctly detected the changed resource and called the Python scripts:
```
Changed files:
datadog/fwprovider/resource_datadog_synthetics_concurrency_cap.go
scripts/build_skip_regex.py
scripts/select_pr_tests.py
scripts/test_build_skip_regex.py
scripts/test_select_pr_tests.py

RUN_REGEX=$(python3 scripts/select_pr_tests.py --escape-for-make < /tmp/changed_files.txt)
SKIP_REGEX=$(python3 scripts/build_skip_regex.py --escape-for-make)
Test args: -run '^TestAccDatadogSyntheticsConcurrencyCap$$'
```

**integration-tests** ran only the scoped test and passed:
```
PASS datadog/tests.TestAccDatadogSyntheticsConcurrencyCap (1.91s)
PASS Package datadog/tests (1.944s)
```

APIR-2647


.github/workflows/test_pr_integration.yml
```diff
@@ -0,0 +1,106 @@
...
+  pull_request:
+    branches: [master, 4.x]
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}-pr-integration
+  cancel-in-progress: true
+
+env:
+  TERRAFORM_VERSION: "1.13.1"
```
LiuVII, line 16: **💬 suggestion:** ‏maybe we should track latest, or is there some idea of how we will bump these versions periodically?

scripts/build_skip_regex.py
LiuVII, line 1: **💭 thought:** ‏I wonder if this file could be replaced by smth like `grep -oP '^\s*-?\s*test:\s*\K\S+' flaky_tests.yaml | sed 's/.*/^\&$$/' | paste -sd'|'` though your approach is more readable and easier to reason about, maybe ok to have an overkill like this given how cheap it's to support smth like this

fpighi: I started with pure shell scripts, I went back to Python for ease of maintaining/reading this and ability to write reasonable tests...

.github/workflows/test_pr_integration.yml
fpighi: I think this was based on most used version, which is something we should re-assess every now and then I agree!

PR #3561, opened 2026-03-03
[datadog_dashboard] Add DD-Terraform-Resource header opt-in framework

yacomink:
## Context

We plan to introduce a new `datadog_dashboard_v2` resource to the terraform provider, and we'd like to be able to differentiate which requests come for the `datadog_dashboard` vs `datadog_dashboard_v2` resource types.

## Summary

- Adds a platform-level opt-in framework for annotating API requests with a `DD-Terraform-Resource` header, so backends can identify which Terraform resource type originated a request
- Implements as an `http.RoundTripper` wrapper installed at provider init that reads a resource name from the request context — resources opt in with a single `utils.WithTerraformResource(auth, "resource_name")` call
- `datadog_dashboard` is the initial adopter (all CRUD functions)

## Details

**New files:**
- `datadog/internal/utils/transport.go` — context key, RoundTripper, and `WrapTransportWithResourceHeader` helper
- `datadog/internal/utils/transport_test.go` — unit tests (header set, header absent, no mutation of original request)

**Modified files:**
- `datadog/internal/utils/utils.go` — `NewHTTPClient()` wraps transport so both SDKv2 and framework providers get the RoundTripper
- `datadog/resource_datadog_dashboard.go` — CRUD functions use `utils.WithTerraformResource()` to annotate auth context

Cassettes are unaffected — `filterHeaders` strips all headers except `Accept` and `Content-Type`.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./datadog/internal/utils/... ./datadog/fwprovider/... ./datadog/...` passes
- [x] `go test ./datadog/internal/utils/...` passes (3 new tests)
- [x] `make check-docs` passes

### DIFF
--- a/datadog/internal/utils/transport_test.go
+++ b/datadog/internal/utils/transport_test.go
@@ -9,10 +9,10 @@ import (
 
 func TestResourceHeaderTransport(t *testing.T) {
 	tests := []struct {
-		name           string
-		resourceName   string // empty means no context value set
-		wantHeader     string
-		checkNoMutate  bool
+		name          string
+		resourceName  string // empty means no context value set
+		wantHeader    string
+		checkNoMutate bool
 	}{
 		{
 			name:         "sets header from context",


datadog/resource_datadog_dashboard.go
```diff
@@ -154,7 +154,7 @@ func resourceDatadogDashboard() *schema.Resource {
 func resourceDatadogDashboardCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	providerConf := meta.(*ProviderConfiguration)
 	apiInstances := providerConf.DatadogApiInstances
-	auth := providerConf.Auth
+	auth := utils.WithTerraformResource(providerConf.Auth, "datadog_dashboard")
```
brtu: should `datadog_dashboard` be extracted to a constant?

datadog/internal/utils/transport_test.go
```diff
@@ -0,0 +1,80 @@
+package utils
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestResourceHeaderTransport_SetsHeader(t *testing.T) {
```
brtu: these tests look pretty similar with just minor differences, can they be consolidated?

PR #3576, opened 2026-03-05
[datadog_observability_pipeline] Adds tests for upcoming Opentelemetry support for metrics pipelines

apatruddog:
Adds tests for the upcoming Opentelemetry source support for metrics pipelines.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/terraform-providers/terraform-provider-datadog
 
 require (
-	github.com/DataDog/datadog-api-client-go/v2 v2.56.1-0.20260313211508-8a1e1dceb6e8
+	github.com/DataDog/datadog-api-client-go/v2 v2.56.1-0.20260317083541-751df6b23c73
 	github.com/DataDog/dd-sdk-go-testing v0.0.0-20211116174033-1cd082e322ad
 	github.com/Masterminds/semver/v3 v3.3.1
 	github.com/google/go-cmp v0.7.0
--- a/go.sum
+++ b/go.sum
@@ -2,8 +2,8 @@ cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMT
 dario.cat/mergo v1.0.0 h1:AGCNq9Evsj31mOgNPcLyXc+4PNABt905YmuqPYYpBWk=
 dario.cat/mergo v1.0.0/go.mod h1:uNxQE+84aUszobStD9th8a29P2fMDhsBdgRYvZOxGmk=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/DataDog/datadog-api-client-go/v2 v2.56.1-0.20260313211508-8a1e1dceb6e8 h1:81Mwk6e1BWzPn6fu5y5AmVaunrpLAnKFikH9PCMCWxM=
-github.com/DataDog/datadog-api-client-go/v2 v2.56.1-0.20260313211508-8a1e1dceb6e8/go.mod h1:d3tOEgUd2kfsr9uuHQdY+nXrWp4uikgTgVCPdKNK30U=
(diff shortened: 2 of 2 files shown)


datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -4236,6 +4236,191 @@ resource "datadog_observability_pipeline" "opentelemetry_update" {
 	})
 }
```
taylorchandleryoung, line 4237: just so its tracked here too, make sure to generate the casettes for the test as well and include in the PR. (Sorry for double ping)

apatruddog: Ok good call – I thought they only needed to be [created here](https://github.com/DataDog/datadog-api-spec/pull/5154/changes#diff-8c9fbd38cdd29953ca5c049f65e1de0fe5a837b1451cfcec2c6cca4f988a9700)

datadog/tests/resource_datadog_observability_pipeline_test.go
```diff
@@ -4236,6 +4236,191 @@ resource "datadog_observability_pipeline" "opentelemetry_update" {
...
+	resourceName := "datadog_observability_pipeline.opentelemetry_metrics"
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: accProviders,
+		CheckDestroy:             testAccCheckDatadogPipelinesDestroy(providers.frameworkProvider),
+		Steps: []resource.TestStep{
+			{
+				Config: `
+resource "datadog_observability_pipeline" "opentelemetry_metrics" {
+  name = "opentelemetry-metrics-pipeline"
```
taylorchandleryoung, line 4251: So tests in this repo are here to confirm the terraform provider works and test logic inside of it. Were there changes made earlier to support this feature, or did this work without needing updates? Or are we missing the go client bump still?

May not need these tests if its just testing the client being generated correctly, that seems more schema related

apatruddog: Good questions.

So:

> Were there changes made earlier to support this feature, or did this work without needing updates?

So in fact the only change that needed to land in prod to get these tests passing was [the dd-go change](https://github.com/DataDog/dd-go/pull/225519/changes) that now allows (behind a feature flag) the `(otel, metrics pipeline)` combo.

I'm pretty certain that about 1 week ago when I started this PR, these tests were failing. They must have started passing once the dd-go code went to prod.

> are we missing the go client bump still?

We were not, but I have bumped it now anyway. That's probably good practice.

This project I'm working on, which is about allowing the `(otel, metrics)` combo, doesn't actually change the API structure in any way.

So after the dd-go changes went in, we could have really used any Go client that was not too old, because it would have presented the same API contract that terraform could have called. So that's why the terraform tests succeeded here – the API contract was the same between the old and new Go client versions.

In fact, the changes we make to `datadog-api-spec` e.g. [the one here](https://github.com/DataDog/datadog-api-spec/pull/5154) are mostly cosmetic – they affect the docs and code examples presented online but they don't contain actual validation logic. (see attached screenshot actually taken from https://docs.datadoghq.com/ a few minutes ago)

To summarize: tests started passing after the `dd-go` PR even without `go.mod` changes because the API contract did not change. Therefore, the code that converts TF code to API calls was able to call the same API, just this time with `(otel, metrics)`, and the code in `dd-go` allowed it.

taylorchandleryoung: In that case, we probably don't need these tests? They should probably be written at a dd-go level? But since we've already got them probably not bad to just check them in for more coverage

apatruddog: I guess it's better to have them than not, since it is testing a path that didn't exist before – otel source in metrics pipeline.

Either way, I still can't merge this PR because I can't update to the latest api-spec Go library without introducing the breaking changes. I need to [wait for another PR](https://dd.slack.com/archives/C013K2M4KH6/p1773251468276289) to be merged before I can proceed with this one.

Otherwise I would have to merge this PR without a Go library update, but I feel like that's ... not best practice.

tessneau: since your change doesn't depend on the go client changing, I feel like it's ok to omit that for your PR, up to you though, definitely appreciate helping us with PRs who depend on it getting unblocked 🙇

fpighi (approved):

PR #3588, opened 2026-03-09
[CCR-3007] AWS action connection for Workflow Automation

zeinageb:
**[DO NOT MERGE, FOR DEMO AND REFERENCE PURPOSES]**

Adds a minimal example for creating a Datadog AWS action connection (Workflow Automation / EBS workflows) via Terraform

**How to use it**

**Credentials via env**

```
   export TF_VAR_datadog_api_key="your_api_key"
   export TF_VAR_datadog_app_key="your_registered_app_key"
``` 

The app key must be registered for the Actions API 

**Apply:**

```
   cd examples/resources/datadog_action_connection_aws
   terraform init
   terraform plan
   terraform apply
```

IAM: Use terraform output `external_id` in the IAM role trust policy for the role you specified, then use the connection in Workflow Automation 

See `README.md` in the example directory for full steps

### DIFF
--- a/examples/resources/datadog_action_connection_aws/.terraform.lock.hcl
+++ b/examples/resources/datadog_action_connection_aws/.terraform.lock.hcl
@@ -1,0 +1,24 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/datadog/datadog" {
+  version = "4.0.0"
+  hashes = [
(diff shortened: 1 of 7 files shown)


dave-handy:
This looks right to me, although the use of vars should be pretty straight forward for anyone that approaches doing this in terraform, and that causes a fair amount of indirection. Also the "outputs" is only relevant if you package this up as a module iirc, which I don't think is really needed either

PR #3608, opened 2026-03-16
[ACCESS-3859] Include permission name aliases in permissions data source

retsguj:
https://datadoghq.atlassian.net/browse/ACCESS-3859

I updated the go client to the most recent commit.

This allows terraform users to refer to permissions by any of their names, canonical or otherwise, in their configurations (for roles, mostly).

To test, you can create a role in terraform and give it the infrastructure_read permission, and plan/apply will work.


amarshall-dd (approved):
Only reviewed go.{mod,sum} change

PR #3490, opened 2026-02-03
[datadog_team_permission_setting] Add import support

grv231:
# Summary

- Add ImportState method to allow importing existing team permission settings
- Import format: \`team_id:action\`
- Add import example for documentation

## Test Plan
- [x] Existing test `TestAccTeamPermissionSettingBasic` passes
- [x] `make fmtcheck` passes
- [x] `make vet` passes
- [x] `make docs && make check-docs` passes

Fixes #3488

Issue #3488, opened 2026-02-02
team_permission_setting resource doesnt support imports using datadog terraform provider


grv231:
### What resources or data sources are affected?

- Resource name: `team_permission_setting`
- Resource link: https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/team_permission_setting

### Feature Request

I was trying to import existing Datadog teams and associated resources recently and found that the datadog resource: `team_permission_setting` doesn't support imports right now. Current version of provider I am using: `3.86.0`

PR attached

PR #3628, opened 2026-03-20
[datadog_team_permission_setting] Fix panic in Read after import

### LINKED ISSUE #3629: [datadog_team_permission_setting] Read panics with SIGSEGV after import (regression from #3490)
### Terraform Version, Provider Version and Datadog Environment

- Terraform: v1.14.6
- Provider: v3.90.0 and v3.91.0 (both affected)
- Datadog Environment: datadoghq.com

### Description

`datadog_team_permission_setting` panics with a nil pointer dereference (SIGSEGV) whenever `ImportState` is used. This is a regression introduced in #3490 (v3.90.0).

### Steps to reproduce

1. Create a terraform config with `datadog_team_permission_setting` import blocks:
```hcl
import {
  to = datadog_team_permission_setting.example
  id = "<team_id>:manage_membership"
}
```

2. Run `terraform plan` or `terraform apply`

3. Provider crashes:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x18ef8ce]

goroutine 688 [running]:
github.com/.../datadog/internal/utils.FrameworkErrorDiag({0x0?, 0x0?}, ...)
    utils.go:140 +0x32e
github.com/.../datadog/fwprovider.(*teamPermissionSettingResource).Read(...)
    resource_datadog_team_permission_setting.go:123 +0x465
```

### Root cause

Two bugs in the `Read` method:

1. `ImportState` sets `team_id` and `action` but not `id`. The `Read` loop only matches by `id` (empty string post-import), so no permission is ever found.

2. When no match is found, `err` (nil since the API call succeeded) is passed to `FrameworkErrorDiag`, which dereferences it → SIGSEGV.

### Expected behavior

`terraform import` should complete without panic, populating the full resource state from the API.

### Fix

PR #3628 addresses both issues.

grv231:
### Description

Fixes #3629 — a panic (nil pointer dereference / SIGSEGV) in `teamPermissionSettingResource.Read()` that occurs after every `ImportState` call. This is a regression introduced in #3490 (v3.90.0) and still present in v3.91.0+.

### How the panic occurs — step by step

**Step 1:** `ImportState` correctly sets `team_id` and `action`, but NOT `id`:
```go
// ImportState parses "team_id:action" and sets both
response.State.SetAttribute(ctx, path.Root("team_id"), result[0])
response.State.SetAttribute(ctx, path.Root("action"), result[1])
// id is NOT set — it remains "" (empty string)
```
This is correct — the internal resource ID (e.g., `TeamPermission-{team_id}-{action}`) is not known at import time.

**Step 2:** Terraform automatically calls `Read` to populate the full state. The API call at line 104 **succeeds** (returns all permission settings for the team), so `err` is `nil`. The loop at line 116 tries to find the matching permission:
```go
for _, permission := range permissions.Data {
    if permission.Id == state.ID.ValueString() {  // compares against "" — never matches
```
`state.ID.ValueString()` is `""` after import. The API returns IDs like `TeamPermission-{team_id}-edit`. Empty string never matches → `found` stays `false`.

**Step 3:** The not-found path passes `nil` err to `FrameworkErrorDiag`:
```go
if !found {
    response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error getting..."))
}
```
`err` is `nil` (the API call succeeded). `FrameworkErrorDiag` at `utils.go:137` performs:
```go
switch v := err.(type) {
    ...
    default:
        summary = v.Error()  // nil pointer dereference → SIGSEGV
```

**Stack trace:**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x18e7b8e]

goroutine 688 [running]:
github.com/.../datadog/internal/utils.FrameworkErrorDiag({0x0?, 0x0?}, ...)
    utils.go:140 +0x32e
github.com/.../datadog/fwprovider.(*teamPermissionSettingResource).Read(...)
    resource_datadog_team_permission_setting.go:123 +0x465
```

### What was missed in #3490 and why

During development of #3490, this risk was identified:

> *Note: May need to modify Read to also match by `action` field during import (since `ID` won't be set until after first read).*

This was noted but not implemented. The `ImportState` method was modeled after `team_membership`, which handles import correctly. However, the key difference was missed — `team_membership`'s `Read` at line 136 **already has a dual-match fallback**:
```go
// team_membership Read — correctly handles import with dual match
if userTeam.GetId() == state.ID.ValueString() ||
    state.UserId.ValueString() == userTeam.Relationships.User.Data.GetId() {
```

`team_permission_setting`'s `Read` only matches by `ID` — the equivalent fallback for `action` was never added.

Additionally, the import test cassette (`TestAccTeamPermissionSettingImport.yaml`) was created but left empty (`interactions: []`), so the import code path was never actually exercised in CI before the PR was merged.

### The fix — two targeted changes in `Read`

**Bug 1: Matching logic (line 116)** — Add fallback to match by `action` when `id` is empty (the import case), mirroring `team_membership`:
```go
// Before (only matches by ID — fails after import when ID is "")
if permission.Id == state.ID.ValueString() {

// After (also matches by action for import reads)
if permission.Id == state.ID.ValueString() ||
    (state.ID.ValueString() == "" && string(permission.Attributes.GetAction()) == state.Action.ValueString()) {
```

**Bug 2: Nil error handling (line 123)** — `FrameworkErrorDiag` performs `switch v := err.(type)` which panics on nil input. `err` is nil by design here (the API call succeeded; the "not found" is a logic condition, not an API error). Replace with `AddError()` for nil-safe error reporting:
```go
// Before (passes nil err → panic)
response.Diagnostics.Append(utils.FrameworkErrorDiag(err, ...))

// After (safe error reporting)
response.Diagnostics.AddError(
    fmt.Sprintf("error getting team permission setting with id %s", state.ID.ValueString()),
    "team permission setting not found",
)
return
```

Also adds `break` after finding a match (minor optimization — no need to continue iterating).

### Evidence

- CI failures with full stack traces: both v3.90.0 and v3.91.0 crash identically at `resource_datadog_team_permission_setting.go:123`
- Tested locally with `dev_overrides`: multiple team permission setting imports complete successfully with this fix (0 panics, plan completes as expected)

### Changes

Single file: `datadog/fwprovider/resource_datadog_team_permission_setting.go` (+8/-2)

### Related

- Bug issue: #3629
- Original import support PR that introduced the regression: #3490
- Original issue (closed): #3488

### DIFF
--- a/datadog/fwprovider/resource_datadog_team_permission_setting.go
+++ b/datadog/fwprovider/resource_datadog_team_permission_setting.go
@@ -113,14 +113,20 @@ func (r *teamPermissionSettingResource) Read(ctx context.Context, request resour
 
 	found := false
 	for _, permission := range permissions.Data {
-		if permission.Id == state.ID.ValueString() {
+		if permission.Id == state.ID.ValueString() ||
+			(state.ID.ValueString() == "" && string(permission.Attributes.GetAction()) == state.Action.ValueString()) {
 			r.updateState(ctx, &state, &permission)
 			found = true
+			break
 		}
 	}
 
 	if !found {
-		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, fmt.Sprintf("error getting team permission setting with id %s", state.ID.ValueString())))
+		response.Diagnostics.AddError(
+			fmt.Sprintf("error getting team permission setting with id %s", state.ID.ValueString()),
+			"team permission setting not found",
+		)
+		return
 	}
 
 	// Save data into Terraform state


grv231:
@botz-syd confirming if there are any steps required on my end to be done? I see some pending checks, hence not sure what are the next steps here?

grv231:
@dshivara is there something that needs to be done for this? Trying to estimate when this can be merged so that I can use it. I see a new provider version was put out 5 hours ago and this was not cherry picked/merged to main for the new version. 

If there is something I need to do, please let me know.

PR #3640, opened 2026-03-26
[datadog_role] Deprecate and stop setting user_count to fix unstoppable drift

### LINKED ISSUE #3468: datadog_role resource unstoppable drift

v3.85.0

### Terraform Version

v1.11.4

datadog_role

```hcl
resource "datadog_role" "this" {
  name = var.name

  dynamic "permission" {
    for_each = toset(var.permissions)
    content {
      id = permission.value
    }
  }
}
```

```
  # module.roles["read_only_non_phi"].datadog_role.this has changed
  ~ resource "datadog_role" "this" {
        id         = "fee4688e-93e5-11f0-9017-da7ad0900002"
        name       = "read_only_non_phi"
      ~ user_count = 143 -> 145

        # (71 unchanged blocks hidden)
    }
```

### Expected Behavior

Terraform should not show a change when the inputs for the resource change.

### Actual Behavior

The read-only attribute `user_count` automatically increases as SCIM users are automatically added to the role. Because it is a read_only attribute `lifecycle` policies cannot be used to ignore the change. This causes an unstoppable drift.  This is an anti-pattern considering that datasources can be used for retrieving read-only data that changes dynamically. It also causes problems with DevOps workflows that perform drift detection since it will create non-stop diffs in busy orgs.

### Steps to Reproduce

1. Create a role in TF
2. Attach that role to a SCIM group mapping
3. Add user to SCIM group
4. Run new plan or apply

### Important Factoids

_No response_

### References

_No response_

fpighi:
## Summary
- Deprecate the `user_count` attribute on `datadog_role` resource and stop refreshing it from the API
- If anyone needs to rely on the user count, it can be done using the `datadog_role` [datasource](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/data-sources/role) instead.

## Motivation
- The read-only `user_count` field was updated on every Read, causing perpetual state drift when users were dynamically added to a role (e.g. via SCIM group mappings)
- Since the attribute is Computed-only, `lifecycle { ignore_changes }` cannot suppress the diff
- Fixes #3468

## Test plan
- [x] `make fmtcheck` passes
- [x] `make test` passes (1012 tests, 0 failures)
- [x] Acceptance tests: `RECORD=false TESTARGS="-run TestAccDatadogRole" make testacc` (9 tests, 0 failures)

### DIFF
--- a/docs/resources/role.md
+++ b/docs/resources/role.md
@@ -44,7 +44,7 @@ resource "datadog_role" "api_key_manager" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
-- `user_count` (Number) Number of users that have this role.
+- `user_count` (Number, Deprecated) Number of users that have this role. **Deprecated.** This field is no longer set and will be removed in a future release.
 
 <a id="nestedblock--permission"></a>
 ### Nested Schema for `permission`
--- a/datadog/resource_datadog_role.go
+++ b/datadog/resource_datadog_role.go
@@ -56,6 +56,7 @@ func resourceDatadogRole() *schema.Resource {
 					Type:        schema.TypeInt,
 					Computed:    true,
 					Description: "Number of users that have this role.",
+					Deprecated:  "This field is no longer set and will be removed in a future release.",
 				},
 				"validate": {
 					Description: "If set to `false`, skip the validation call done during plan.",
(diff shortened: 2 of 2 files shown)


fpighi:
Currently failing test in CI is [failing in master](https://github.com/DataDog/terraform-provider-datadog/actions/runs/23571103850/job/68633627011#step:9:771) too. Working on a [fix](https://github.com/DataDog/terraform-provider-datadog/pull/3643) separately.

docs/resources/role.md
```diff
@@ -44,7 +44,7 @@ resource "datadog_role" "api_key_manager" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
-- `user_count` (Number) Number of users that have this role.
+- `user_count` (Number, Deprecated) Number of users that have this role. **Deprecated.** This field is no longer set and will be removed in a future release.
```
maycmlee, line 47:
```suggestion
- `user_count` (Number, Deprecated) Number of users that have this role. **Deprecated**: This field is no longer set and will be removed in a future release.
```

fpighi: This is auto-generated and can't be changed unfortunately.

PR #3642, opened 2026-03-26
Fix make docs failing in git worktrees

fpighi:
## Summary
- `tfplugindocs` infers the provider name from the directory name by stripping the `terraform-provider-` prefix. In a worktree (e.g. `terraform-provider-datadog-APIR-2568-foo`), it derives the wrong name, causing doc generation to fail.
- Explicitly pass `--provider-name datadog` and `--rendered-provider-name "terraform-provider-datadog"` to decouple doc generation from the directory name.
- This unblocks developers and coding agents that use `git worktree` for branch isolation.

## Test plan
- [x] Ran `make docs` from a worktree (`terraform-provider-datadog-APIR-2743-fix-make-docs-worktree`) — succeeds, produces identical docs to master


templates/resources/synthetics_test.md.tmpl
```diff
@@ -1,7 +1,7 @@
 ---
 # Schema inspired by https://github.com/hashicorp/terraform-plugin-docs/blob/b7abf704f51e7356251d59bceb27d4973d78bee4/internal/provider/template.go#L130
 # generated by https://github.com/hashicorp/terraform-plugin-docs
-page_title: "{{.Type}} {{.Name}} - {{.ProviderName}}"
+page_title: "{{.Type}} {{.Name}} - {{.RenderedProviderName}}"
```
fpighi, line 4: This was identical previously, but with the change we have to explicitly use the rendered name.

PR #3643, opened 2026-03-26
[datadog_monitor] Remove deprecated restricted_roles checks from restriction policy tests

fpighi:
## Summary

- Remove `verifyRestrictedRolesSize` checks from `TestAccMonitor_Fwprovider_WithRestrictionPolicy` and `TestAccDatadogMonitor_WithRestrictionPolicy`
- Remove `restricted_roles.#` state assertions and the step that tested restriction policy destruction clearing restricted_roles
- Delete the now-unused `verifyRestrictedRolesSize` function and `testAccCheckDatadogMonitorWithRestrictionPolicyDestroyed` config helper
- Re-record cassettes for both tests

The `restricted_roles` field was deprecated in v4.0 in favor of `datadog_restriction_policy`. These tests were verifying that creating a restriction policy automatically populates the monitor's `restricted_roles` — this API side-effect is no longer guaranteed. Tests now only verify the restriction policy resource itself works correctly alongside a monitor.

## Test plan

- [x] `RECORD=true` — both tests pass against the live API
- [x] `RECORD=false` — both tests pass in replay mode with new cassettes


fpighi:
Superseded by https://github.com/DataDog/terraform-provider-datadog/pull/3645

PR #3652, opened 2026-03-30
[datadog_dataset] Expand `datadog_dataset` resource explanation

jyanezt:
Expand `datadog_dataset` resource explanation to clarify that `metrics` product refers to "Custom Metrics" and to point out that datasets are also known as "restricted datasets" and that are part of Datadog's Data Access Control


fpighi:
The docs are generated using `make docs` from the respective resource source code.

PR #3655, opened 2026-03-31
[sensitive_data_scanner] Allow filtering rules by standard pattern ID

fbryden:
The current way of only being able to select SDS OOTB rules in terraform based on rule name is limiting in two ways:
* Impossible to select rules whose name is a subset of a another rule (e.g. `US Tax Identification Number Scanner` being a subset of `Cyprus Tax Identification Number Scanner`)
* TF plan/apply being broken when a Library Rule gets renamed

This PR introduces a new way of filtering on the Library Rule ID.

### DIFF
--- a/datadog/tests/data_source_datadog_sensitive_data_scanner_standard_pattern_test.go
+++ b/datadog/tests/data_source_datadog_sensitive_data_scanner_standard_pattern_test.go
@@ -100,6 +100,26 @@ func TestAccDatadogStandardPatternDatasourceIDFilter(t *testing.T) {
 	})
 }
 
+func TestAccDatadogStandardPatternDatasourceIDFilterErrorNotFound(t *testing.T) {
+	t.Parallel()
+	if isRecording() || isReplaying() {
+		t.Skip("This test doesn't support recording or replaying")
+	}
+
+	_, accProviders := testAccProviders(context.Background(), t)
+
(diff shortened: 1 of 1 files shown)


LiuVII:
The new `TestAccDatadogStandardPatternDatasourceIDFilter` only covers the happy path. The name filter has three tests (happy path, multiple results, not found) — worth adding a matching not-found case for the ID path:

```go
func TestAccDatadogStandardPatternDatasourceIDFilterErrorNotFound(t *testing.T) {
	t.Parallel()
	if isRecording() || isReplaying() {
		t.Skip("This test doesn't support recording or replaying")
	}

	_, accProviders := testAccProviders(context.Background(), t)

	resource.Test(t, resource.TestCase{
		PreCheck:          func() { testAccPreCheck(t) },
		ProviderFactories: accProviders,
		Steps: []resource.TestStep{
			{
				Config:      testAccDatasourceStandardPatternConfigByID("nonexistent-id"),
				ExpectError: regexp.MustCompile("Couldn't find the standard pattern with id nonexistent-id"),
			},
		},
	})
}
```

Otherwise the change looks good — `ExactlyOneOf` wiring is correct, the ID lookup uses exact match (right call), and the change is fully backward-compatible.

LiuVII (approved):
lgtm
left a small comment above on test if it makes sense

PR #3686, opened 2026-04-09
[datadog_org_group] Terraform support for org groups and policies

k-phan:
## Summary

Adds Terraform support for managing Datadog organization groups — the mechanism customers use to share policy configuration across orgs in their organization hierarchy.

Resources cover group identity, org-to-group membership, policies defined at the group level, and per-organization policy overrides. Matching data sources expose each of these for discovery and for adopting existing server-side configuration via `for_each`.

## Test plan

- [x] Acceptance tests for each resource covering create, update, import, and destroy
- [x] Acceptance tests for each data source covering filters and result shape
- [x] Recorded cassettes included for CI replay

## For Reviewers

**Note [1]** I don't expect the **live** integration tests to pass (recorded acceptance tests should still pass). Org Group tests require a pretty unique setup that isn't friendly with the existing one.

**Note [2]** I decided to keep all of the resources/data sources in one PR to avoid multiple round-trips of disjoint context switching. My rationale was that all of these are new resources and are part of a single new feature in preview, and that it would make sense to look at them holistically in one PR. A large portion of the diff is cassettes (filter out `.yaml` files!).

If we feel strongly that the PR should be split up into smaller buckets, happy to do so.

### Structure

```
datadog/
├── fwprovider/
│   ├── resource_datadog_org_group.go                    ← review in depth
│   ├── resource_datadog_org_group_membership.go         ← review in depth
│   ├── resource_datadog_org_group_policy.go             ← review in depth
│   ├── resource_datadog_org_group_policy_override.go    ← review in depth
│   ├── data_source_datadog_org_groups.go                ← review in depth
│   ├── data_source_datadog_org_group_memberships.go     ← review in depth
│   ├── data_source_datadog_org_group_policies.go        ← review in depth
│   └── data_source_datadog_org_group_policy_overrides.go ← review in depth
└── tests/
    ├── resource_datadog_org_group_*_test.go             ← review in depth
    ├── data_source_datadog_org_group_*_test.go          ← review in depth
    └── cassettes/                                       ← recorded VCR; skim only

templates/resources/org_group_*.md.tmpl                  ← drives the "Behavior" sections in rendered docs
docs/resources/org_group_*.md                            ← GENERATED; preview via links below
docs/data-sources/org_group_*.md                         ← GENERATED; preview via links below
examples/{resources,data-sources}/datadog_org_group_*    ← HCL snippets embedded in Registry docs
```

### Where to focus

- **Provider code** (`datadog/fwprovider/*.go`) — the actual implementation.
- **Tests** (`datadog/tests/resource_*`, `datadog/tests/data_source_*`) — acceptance-style, replay against cassettes.
- **Templates** (`templates/resources/*.md.tmpl`) — source of the "Behavior" prose in rendered docs.
- **Cassettes** (`datadog/tests/cassettes/*.{yaml,freeze}`) — skim only; they're recorded VCR data that replays in CI.

### Preview the generated docs

Easier to read rendered than as raw markdown:

**Resources:**
- [`datadog_org_group`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/resources/org_group.md)
- [`datadog_org_group_membership`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/resources/org_group_membership.md)
- [`datadog_org_group_policy`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/resources/org_group_policy.md)
- [`datadog_org_group_policy_override`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/resources/org_group_policy_override.md)

**Data sources:**
- [`datadog_org_groups`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/data-sources/org_groups.md)
- [`datadog_org_group_memberships`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/data-sources/org_group_memberships.md)
- [`datadog_org_group_policies`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/data-sources/org_group_policies.md)
- [`datadog_org_group_policy_overrides`](https://github.com/DataDog/terraform-provider-datadog/blob/khai.phan/org-group-resource/docs/data-sources/org_group_policy_overrides.md)


k-phan:
* **#3686** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

joepeeples:
Opened [DOCS-14170](https://datadoghq.atlassian.net/browse/DOCS-14170) to assign a Docs writer and follow up with editorial review. 

[DOCS-14170]: https://datadoghq.atlassian.net/browse/DOCS-14170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

datadog/fwprovider/data_source_datadog_org_group_policies.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, p := range policies {
+		attrs := p.GetAttributes()
+		item := &OrgGroupPolicyItemModel{
+			ID:              types.StringValue(p.GetId().String()),
+			PolicyName:      types.StringValue(attrs.GetPolicyName()),
+			EnforcementTier: types.StringValue(string(attrs.GetEnforcementTier())),
+			PolicyType:      types.StringValue(string(attrs.GetPolicyType())),
+		}
+
+		contentBytes, err := json.Marshal(attrs.GetContent())
```
madie-dd, line 136: in theory this field is a map - it shouldn't matter for org configs - but do we need to worry about future-casing this for potentially more complicated "contents" down the road that might have more than 1 value? 

I'm thinking that if/when we have more complicated contents, this will become a non-deterministic "map ordering -> string" issue. 

Should we add some ordering/sorting now to be safe?

datadog/fwprovider/data_source_datadog_org_group_policy_overrides.go
```diff
@@ -0,0 +1,220 @@
...
+
+		if attrs.HasContent() {
+			bytes, err := json.Marshal(attrs.GetContent())
+			if err != nil {
+				response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "error marshaling override content"))
+				return
+			}
+			item.Content = types.StringValue(string(bytes))
+		} else {
+			item.Content = types.StringValue("{}")
```
madie-dd, line 171: Is there any "meaningful" API difference between isEmpty and "has no content" here? 

wondering if there's a semantic difference (we want?) to define -- "no overrides" vs "never has overrides" if the field is omitted entirely? 

once again - a more hypothetical long term qq about resource that can't be overridden (e.g. roles?) not a blocker for org configs.

datadog/fwprovider/data_source_datadog_org_group_memberships.go
```diff
@@ -0,0 +1,176 @@
...
+	d.Auth = providerData.Auth
+}
+
+func (d *datadogOrgGroupMembershipsDataSource) Metadata(_ context.Context, _ datasource.MetadataRequest, response *datasource.MetadataResponse) {
+	response.TypeName = "org_group_memberships"
+}
+
+func (d *datadogOrgGroupMembershipsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, response *datasource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "Use this data source to retrieve org group memberships, optionally filtered by org group or organization.",
```
madie-dd: Is it "optionally by either" or "at least one"? 

iirc we don't support looking up "get all memberships" which this seems to imply?

datadog/fwprovider/resource_datadog_org_group_membership.go
```diff
@@ -0,0 +1,273 @@
...
+		return
+	}
+
+	orgUuid, err := uuid.Parse(state.OrgUuid.ValueString())
+	if err != nil {
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "org_uuid must be a valid UUID"))
+		return
+	}
+
+	params := datadogV2.NewListOrgGroupMembershipsOptionalParameters().WithFilterOrgUuid(orgUuid)
```
madie-dd, line 108: other filters included datacenter... is this a miss in oui-public? or in this terraform?

we should be consistent if datacenter is required across all "get" requests.

madie-dd (changes requested):

datadog/fwprovider/data_source_datadog_org_group_memberships.go
k-phan: It's at least one, we should add `(at least one is required)` in the description for clarity.

datadog/fwprovider/data_source_datadog_org_group_policies.go
k-phan: I think we should be ok! `json.Marshal` sorts:

> Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement [encoding.TextMarshaler](https://pkg.go.dev/encoding#TextMarshaler). The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:

https://pkg.go.dev/encoding/json#Marshal

Also verified it through Claude as a potentially hallucinating second check 😅

datadog/fwprovider/data_source_datadog_org_group_policy_overrides.go
k-phan: Not one right now! I think we can embed those semantics inside the content itself. I don't feel strongly either way, though. Let me know if you feel strongly!

datadog/fwprovider/resource_datadog_org_group_membership.go
k-phan: This is smoothed over by the API to assume current site (site of the manager)! We'll keep the behavior going forward by introducing the optional site argument. Agree this is a bit inconsistent but needs to be improved upstream instead

datadog/fwprovider/data_source_datadog_org_group_policies.go
madie-dd: Sounds good! I have been burned by this enough times with proto - I wasn't sure that json.Marshal did this for us! Thanks for looking into it!

madie-dd (approved):

datadog/fwprovider/data_source_datadog_org_group_policies.go
k-phan: Thanks for calling it out! Good to confirm the behavior to make sure :)

docs/data-sources/org_group_memberships.md
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `org_group_id` (String) Filter memberships to those within the given org group. At least one of `org_group_id` or `org_uuid` is required. Must be a valid UUID.
+- `org_uuid` (String) Filter memberships to those for the given organization. At least one of `org_group_id` or `org_uuid` is required. Must be a valid UUID.
```
OliviaShoup:
```suggestion
- `org_group_id` (String) Filter memberships to those within the given org group. At least one filter (`org_group_id` or `org_uuid`) is required. Must be a valid UUID.
- `org_uuid` (String) Filter memberships to those for the given organization. At least one filter (`org_group_id` or `org_uuid`) is required. Must be a valid UUID.
```

docs/data-sources/org_group_memberships.md
```diff
@@ -0,0 +1,49 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_org_group_memberships Data Source - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Use this data source to retrieve org group memberships filtered by org group or organization. At least one of org_group_id or org_uuid is required.
+---
+
+# datadog_org_group_memberships (Data Source)
+
+Use this data source to retrieve org group memberships filtered by org group or organization. At least one of `org_group_id` or `org_uuid` is required.
```
OliviaShoup:
```suggestion
Use this data source to retrieve org group memberships. At least one filter (`org_group_id` or `org_uuid`) is required.
```

docs/data-sources/org_group_memberships.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Optional
+
+- `org_group_id` (String) Filter memberships to those within the given org group. At least one of `org_group_id` or `org_uuid` is required. Must be a valid UUID.
+- `org_uuid` (String) Filter memberships to those for the given organization. At least one of `org_group_id` or `org_uuid` is required. Must be a valid UUID.
+
+### Read-Only
+
+- `id` (String) The ID of this resource.
+- `memberships` (List of Object) The list of org group memberships. (see [below for nested schema](#nestedatt--memberships))
```
OliviaShoup, line 38:
```suggestion
- `memberships` (List of Object) The list of org group memberships. See [below for nested schema](#nestedatt--memberships).
```

docs/data-sources/org_group_policies.md
```diff
@@ -0,0 +1,54 @@
...
+- `org_group_id` (String) The UUID of the org group whose policies to list. Must be a valid UUID.
+
+### Optional
+
+- `policy_name` (String) Filter policies by name.
+
+### Read-Only
+
+- `id` (String) The ID of this resource.
+- `policies` (List of Object) The list of policies. (see [below for nested schema](#nestedatt--policies))
```
OliviaShoup, line 42:
```suggestion
- `policies` (List of Object) The list of policies. See [below for nested schema](#nestedatt--policies).
```

docs/data-sources/org_group_policy_overrides.md
```diff
@@ -0,0 +1,58 @@
...
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `org_group_id` (String) The UUID of the org group whose overrides to list. Must be a valid UUID.
+
+### Optional
+
+- `org_uuid` (String) Filter overrides to those for the given organization. Applied client-side after the List call since the API does not accept an org_uuid filter on this endpoint. Must be a valid UUID.
```
OliviaShoup:
```suggestion
- `org_uuid` (String) Filter overrides to those for the given organization. Applied client-side after the List call because the API does not accept an `org_uuid` filter on this endpoint. Must be a valid UUID.
```

docs/data-sources/org_group_policy_overrides.md
```diff
@@ -0,0 +1,58 @@
...
+- `policy_id` (String) Filter overrides to those on the given policy. Must be a valid UUID.
+
+### Read-Only
+
+- `id` (String) The ID of this resource.
+- `overrides` (List of Object) The list of policy overrides. (see [below for nested schema](#nestedatt--overrides))
```
OliviaShoup, line 46:
```suggestion
- `overrides` (List of Object) The list of policy overrides. See [below for nested schema](#nestedatt--overrides).
```

docs/data-sources/org_groups.md
```diff
@@ -0,0 +1,36 @@
...
+# List every org group in the organization.
+data "datadog_org_groups" "all" {}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Read-Only
+
+- `groups` (List of Object) The list of org groups. (see [below for nested schema](#nestedatt--groups))
```
OliviaShoup, line 25:
```suggestion
- `groups` (List of Object) The list of org groups. See [below for nested schema](#nestedatt--groups).
```

docs/resources/org_group_membership.md
```diff
@@ -0,0 +1,58 @@
...
+  org_group_id = datadog_org_group.prod.id
+  org_uuid     = "ff4a8255-6931-58d1-add0-a6b3602d5421"
+}
+```
+
+## Behavior notes
+
+### In-place moves vs. replacement
+
+Changing `org_group_id` moves the organization between groups in place via the server's membership-update endpoint — the resource is **not** re-created. Changing `org_uuid` triggers replacement since the resource represents a specific org's membership.
```
OliviaShoup:
```suggestion
Changing `org_group_id` moves the organization between groups in place via the server's membership-update endpoint. The resource is **not** re-created. Changing `org_uuid` triggers replacement because the resource represents a specific org's membership.
```

docs/resources/org_group_membership.md
```diff
@@ -0,0 +1,58 @@
...
+
+### State-only destroy
+
+Destroying this resource is state-only: Terraform removes the resource from its state without calling an API, because the server has no endpoint to "delete" a membership outright. The organization stays in its current group until another `datadog_org_group_membership` resource (or an out-of-band API call) moves it.
+
+Because of this, destroying a [`datadog_org_group`](./org_group) requires that it has **zero memberships pointing at it**. Before destroying a group, move every member org to another group first — either by updating `org_group_id` on each membership resource (in-place reassignment) or by removing the member-org configs so Terraform reassigns them elsewhere.
```
OliviaShoup:
```suggestion
Because of this, destroying a [`datadog_org_group`](./org_group) requires that it has **zero memberships pointing at it**. Before destroying a group, move every member org to another group first, either by updating `org_group_id` on each membership resource (in-place reassignment) or by removing the member-org configs so Terraform reassigns them elsewhere.
```

docs/resources/org_group_policy.md
```diff
@@ -0,0 +1,69 @@
...
+## Behavior notes
+
+### Side effects on member orgs
+
+Creating or updating a policy with a non-`ENFORCE` tier (`DEFAULT` or `DELEGATE`) triggers propagation across every member org in the group. For each member, the server compares the org's current config value to the policy value:
+
+- **Match** — no action.
+- **Mismatch** — the server auto-creates a [`datadog_org_group_policy_override`](./org_group_policy_override) for that org, recording its existing value so the org is exempted from the new policy.
+
+Auto-created overrides show up in the [`datadog_org_group_policy_overrides`](../data-sources/org_group_policy_overrides) data source and can be adopted into Terraform via import. See the override resource documentation for the full lifecycle.
```
OliviaShoup: just double-checking: should there be two periods in `(../data-sources/org_group_policy_overrides)` ?

docs/resources/org_group_policy.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Transitioning to ENFORCE
+
+Changing `enforcement_tier` to `"ENFORCE"` atomically deletes every override associated with this policy server-side. Any `datadog_org_group_policy_override` resources pointing at this policy must be removed from configuration in the same commit — otherwise Terraform's next apply will try to recreate the server-deleted overrides and fail with a `FailedPrecondition` error.
```
OliviaShoup:
```suggestion
Changing `enforcement_tier` to `"ENFORCE"` automatically deletes every override associated with this policy server-side. Any `datadog_org_group_policy_override` resources pointing at this policy must be removed from configuration in the same commit. Otherwise, Terraform's next apply will try to recreate the server-deleted overrides and fail with a `FailedPrecondition` error.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+## Behavior notes
+
+### Server-side auto-creation
+
+Overrides are also created automatically by the server when an org's existing config differs from a policy's value. Two triggers:
+
+1. **Membership change** — when an org is moved into an org group, the server compares that org's current config to each non-`ENFORCE` policy in the group. Any mismatch produces an auto-created override recording the org's existing value.
```
OliviaShoup:
```suggestion
1. **Membership change:** When an org is moved into an org group, the server compares that org's current config to each non-`ENFORCE` policy in the group. Any mismatch produces an auto-created override that records the org's existing value.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+2. **Policy create/update** with a non-`ENFORCE` tier — the server performs the same comparison across every member org at policy apply time, creating overrides for orgs whose config doesn't match the new policy value.
```
OliviaShoup:
```suggestion
2. **Policy create/update** with a non-`ENFORCE` tier: The server performs the same comparison across every member org at policy apply time, creating overrides for orgs whose config doesn't match the new policy value.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+
+Auto-created and user-declared overrides are indistinguishable at the API level. They can be adopted into Terraform via `terraform import` or by iterating the [`datadog_org_group_policy_overrides`](../data-sources/org_group_policy_overrides) data source with `for_each` + `import` blocks.
```
OliviaShoup: same question about the double periods here

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Delete behavior
+
+Removing an override does **not** just remove the exemption marker — it resets the target org's config value to match the parent policy's current value. The server treats override deletion as "re-apply the policy to this org" and propagates the policy value accordingly.
```
OliviaShoup:
```suggestion
Removing an override does **not** just remove the exemption marker. The target org's config value is also reset to match the parent policy's current value. The server treats override deletion as "re-apply the policy to this org" and propagates the policy value accordingly.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+
+If you want to stop managing an override without changing the org's value, use `terraform state rm` instead of removing the resource block.
+
+### ENFORCE tier cascade
+
+When the parent [`datadog_org_group_policy`](./org_group_policy) transitions to `enforcement_tier = "ENFORCE"`, the server atomically deletes every override for that policy. Creating an override against an already-`ENFORCE`-tier policy also fails with a `FailedPrecondition` error.
+
+If you plan to flip a policy to `ENFORCE`, remove the `datadog_org_group_policy_override` resource blocks for that policy in the same commit — otherwise Terraform's next apply will try to re-create the server-deleted overrides and fail.
```
OliviaShoup:
```suggestion
If you plan to flip a policy to `ENFORCE`, remove the `datadog_org_group_policy_override` resource blocks for that policy in the same commit. Otherwise, Terraform's next apply will try to re-create the server-deleted overrides and fail.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `org_group_id` (String) The UUID of the org group that owns the policy. Must be a valid UUID.
+- `org_site` (String) The short site name of the organization (e.g. `us1`, `eu1`, `us1-fed`). Part of the override's server-side identity; changing it replaces the resource. String length must be at least 1.
```
OliviaShoup:
```suggestion
- `org_site` (String) The short site name of the organization (e.g., `us1`, `eu1`, `us1-fed`). Part of the override's server-side identity; changing it replaces the resource. String length must be at least 1.
```

docs/resources/org_group_policy_override.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Delete behavior
+
+Removing an override does **not** just remove the exemption marker — it resets the target org's config value to match the parent policy's current value. The server treats override deletion as "re-apply the policy to this org" and propagates the policy value accordingly.
+
+If you want to stop managing an override without changing the org's value, use `terraform state rm` instead of removing the resource block.
+
+### ENFORCE tier cascade
+
+When the parent [`datadog_org_group_policy`](./org_group_policy) transitions to `enforcement_tier = "ENFORCE"`, the server atomically deletes every override for that policy. Creating an override against an already-`ENFORCE`-tier policy also fails with a `FailedPrecondition` error.
```
OliviaShoup: just double-checking: this is supposed to say "atomically" and not "automatically" right?

docs/data-sources/org_group_memberships.md
```diff
@@ -0,0 +1,49 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_org_group_memberships Data Source - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Use this data source to retrieve org group memberships filtered by org group or organization. At least one of org_group_id or org_uuid is required.
```
OliviaShoup:
```suggestion
Use this data source to retrieve org group memberships. At least one filter (`org_group_id` or `org_uuid`) is required.
```

docs/resources/org_group_policy.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## Behavior notes
+
+### Side effects on member orgs
+
+Creating or updating a policy with a non-`ENFORCE` tier (`DEFAULT` or `DELEGATE`) triggers propagation across every member org in the group. For each member, the server compares the org's current config value to the policy value:
+
+- **Match** — no action.
+- **Mismatch** — the server auto-creates a [`datadog_org_group_policy_override`](./org_group_policy_override) for that org, recording its existing value so the org is exempted from the new policy.
```
OliviaShoup:
```suggestion
- **Match:** No action.
- **Mismatch:** The server auto-creates a [`datadog_org_group_policy_override`](./org_group_policy_override) for that org, recording its existing value so the org is exempted from the new policy.
```

OliviaShoup (changes requested):
this looks great! left some small comments. please let me know if you have any questions!

docs/resources/org_group_policy_override.md
k-phan: Yes! But I wonder if that is too much detail. I'll change the wording to "automatically" since that was already implied and we don't need to convey the atomic nature of the deletion!

docs/resources/org_group_policy.md
k-phan: The double period is correct but other docs use the absolute path so I'll do that instead!

docs/data-sources/org_groups.md
k-phan: This is unfortunately auto-generated by the plugin so this change (along with the three similar suggestions) would be a big lift. On the plus side, it'll be consistent with other docs 😄

OliviaShoup: haha that makes sense! thanks for letting me know. no big deal then :)

OliviaShoup (approved):
thanks for addressing my comments so fast + explaining so well! looks great 🚀

k-phan:
> this looks great! left some small comments. please let me know if you have any questions!

@OliviaShoup Thank you!! I accepted all the changes except the following:

- atomically --> automatically
- `(...)` --> Sentence (this is automatically generated so I left as is)
- `../` references -> I switched to absolute links like [here](https://github.com/DataDog/terraform-provider-datadog/pull/3686/changes#diff-01e79b2539331c957bd559545b40670b7dab6ed2f9497884be5da8dd7f09d2b2R52) 

Let me know if you have any concerns!

Supam (approved):

PR #3691, opened 2026-04-10
[datadog_metric_metadata] Fix non-empty plan — guard statsd_interval in Read

fpighi:
## Summary

- The Read function unconditionally set `statsd_interval` using `GetStatsdInterval()` which returns 0 for unset values, causing state drift
- Now uses `GetStatsdIntervalOk()` to only set the field when actually present in the API response

## Test plan

- [x] `TestAccDatadogMetricMetadata_Updated` passes with `RECORD=none` (plan is now empty after apply)
- [x] CI integration tests pass


Supam (approved):

St0rmingBr4in (approved):

ClementTsang (approved):
Didn't know we were partial owners of this but this seems reasonable.

PR #3712, opened 2026-04-16
[datadog_security_monitoring_rule] Expand new value rule learningDuration and forgetAfter from enum to int range (0-30 and 1-30)

clementgbcn:
## What does this PR do?

Fixes the build break caused by DataDog/datadog-api-spec#5468 which converts `SecurityMonitoringRuleNewValueOptionsLearningDuration` and `SecurityMonitoringRuleNewValueOptionsForgetAfter` from fixed enums to integer ranges.

## Who will it impact?

* [slack/#k9-ask-cloud-siem](https://dd.enterprise.slack.com/archives/C02PKN3LV96)

## Motivation

The API spec PR (DataDog/datadog-api-spec#5468) removes the named enum constants in favor of `minimum`/`maximum` integer bounds (0-30 for learningDuration, 1-30 for forgetAfter). This aligns the Terraform provider with the new client types and expands the accepted range beyond the previous fixed values.

## Testing Guidelines

No new test scenarios — schema-only change replacing enum validation with `validation.IntBetween` bounds.


docs/resources/security_monitoring_rule.md
```diff
@@ -175,12 +175,12 @@ Optional:
 
 Required:
 
-- `forget_after` (Number) The duration in days after which a learned value is forgotten. Valid values are `1`, `2`, `7`, `14`, `21`, `28`.
+- `forget_after` (Number) The duration in days after which a learned value is forgotten.
```
brett0000FF: It would be good to include the new accepted ranges for this and learning_duration.

brett0000FF (approved):

docs/resources/security_monitoring_rule.md
clementgbcn: I updated the documentation

alessandro-siragusa-datadog (approved):

fpighi:
I don't think this is a breaking change, because it's more permissive than before from a TF practitioner standpoint. I'll remove the breaking-change label.

clementgbcn:
@fpighi thank for the insights.
I propagated the breaking change label from the Datadog API Client which is indeed a breaking change since we moved from en enum to a int.
I agree that in fact it should be transparent since it is an integer in the terraform and not an enum.

fpighi (approved):

PR #3716, opened 2026-04-17
[datadog_security_monitoring_rule] Migrate to plugin framework

alvizi-dd:
## What does this PR do?

Migrates `datadog_security_monitoring_rule` from `terraform-plugin-sdk/v2` to [terraform-plugin-framework](https://developer.hashicorp.com/terraform/plugin/framework).

The resource now lives in `datadog/fwprovider/resource_datadog_security_monitoring_rule.go` and is registered via `fwprovider.NewSecurityMonitoringRuleResource`. The SDKv2 entry is removed from `datadog/provider.go`; shared helpers (`buildPayloadFilters`, `buildPayloadReferenceTables`, `buildPayloadCaseActions`, schema, etc.) stay in the SDKv2 file because `data_source_datadog_security_monitoring_rules` and `datadog_cloud_configuration_rule` still consume them.

### `tags` / `effective_tags` split (one-time upgrade diff)

Under SDKv2, `tags` was Optional+Computed and the provider merged `default_tags` into it server-side. The framework enforces stricter config/state consistency on Optional attributes, so that merge is no longer viable. This PR adopts the same split `datadog_monitor` uses:

- `tags` is now user-input only (Optional, no longer Computed).
- `effective_tags` is a new Computed `SetAttribute` containing the merged set (user `tags` ∪ provider `default_tags`).
- Create/Update/Validate payloads send `effective_tags`.

References to `datadog_security_monitoring_rule.<name>.tags` in outputs, locals, or other resources will no longer include tags inherited from provider `default_tags`. Switch those references to `effective_tags` if the merged set is needed.

### One-time diffs after upgrade or import

The first plan after upgrade or import may show a converging diff in these cases.
Each settles after one apply.

- **`default_tags`**: default-tag entries move out of `tags` into `effective_tags`. Matches `datadog_monitor`; no `UpgradeState` upgrader is provided.
- **`validate = true|false`**: produces one no-op API update. SDKv2's `DiffSuppressFunc` for write-only flags has no faithful Framework equivalent (Terraform core enforces `plan == config` for known config values).
- **`terraform import` of a config with explicit empty list values** (`distinct_fields = []`, `group_by_fields = []`, `notifications = []`): Read returns `null` on cold import for Optional list fields; the explicit empty list in HCL drives a one-time apply to converge state.
- **`terraform import` of a third-party rule mixing `case`/`query` blocks with `third_party_case`**: SDKv2-parity behavior — Read does not auto-derive the legacy `case`/`query` blocks from `third_party_case` on cold import.  

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt`, `make fmtcheck`, `make docs && make check-docs` clean.
- [x] Full acceptance suite re-recorded and passing under `RECORD=false`: all 18 `TestAccDatadogSecurityMonitoringRule_*` + `TestAccDatadogSecurityMonitoringRuleJSON_Basic` + the data source tests.
- [x] `default_tags` (`TestAccDatadogSecurityMonitoringRule_DefaultTags`) passing with the new `tags` / `effective_tags` split.
- [x] State migration tests added (`resource_datadog_security_monitoring_rule_migration_test.go`) covering basic, signal_correlation, CWS, third_party, new_value, impossible_travel, and default_tags.

### Migration tests — RECORD-only, manual gate

The migration tests use `ExternalProviders` to spin up provider v4.5.0 as a subprocess with its own HTTP client, bypassing the in-process VCR recorder. They `t.Skip` unless `RECORD=true`, so **CI does not execute them** — they are a manual pre-release gate, not automated regression coverage.

Runner shape (`runSecurityMonitoringRuleMigrationTest`):

1. v4.5.0 Create
2. FW Apply + per-shape `Check` asserting every tracked attribute round-trips through Read
3. PlanOnly empty
4. FW Update (message tweak forces a PUT, asserts `message` round-trips)
5. PlanOnly empty

`DefaultTags` has a bespoke case verifying the documented one-time `effective_tags` diff after upgrading.


AndrewCharlesHay:
Plugin Framework migrations are always a good time to tighten up validation — a few things to verify given this is a security-monitoring-rule resource where misconfiguration directly translates to detection gaps:

1. **Schema completeness parity**: the SDKv2 version of this resource accumulated a fair bit of API-driven validation over its lifetime. Worth a diff-check that every enum, regex, and cross-field validator present in the old schema made it into the Plugin Framework schema. The classic migration bug is "old schema validated `type` against 4 enum values, new schema is a free-form string" — at which point a typo like `type: "loss-signal"` vs `"log_signal"` creates a rule that silently never fires. If there's a golden test set for the old resource, running it against the new one would catch most of this.

2. **`Validate types.Bool` field**: I see `Validate` in the model. If this maps to the API's dry-run validation flag, a user who accidentally leaves `validate = true` will get confusing behavior ("my rule doesn't seem to detect anything"). Worth confirming this field's semantics match the API and is surfaced clearly in docs / warnings.

3. **Plan modifiers on sensitive fields**: `Enabled` defaults to... what? For a security detection resource, the safest default is usually "fail closed" — i.e., `Enabled` default `false` so operators have to explicitly opt in. Using `booldefault.StaticBool(...)` gives this for free but I can't tell from the snippet which way it goes. Please confirm the default is the conservative choice.

4. **Import handling**: Plugin Framework import paths are different from SDKv2. A user running `terraform import datadog_security_monitoring_rule.foo <id>` against a state file from the old resource should not silently succeed but end up with a partially-populated state. Worth a test that imports an SDKv2-created rule into the new resource and asserts no drift on next plan.

5. **Parallel resource coexistence**: `NewSecurityMonitoringRuleResource` is added alongside the existing `NewSecurityMonitoringRuleJSONResource`. A user can technically create the same rule with both resource types. Consider whether a `ConflictsWith`-equivalent at the provider level (documented) is warranted, or at minimum a release note "do not mix `datadog_security_monitoring_rule` and `datadog_security_monitoring_rule_json` pointed at the same rule ID".

6. **Deprecation window for SDKv2 resource**: is the plan to keep both indefinitely or deprecate the SDKv2 version? If deprecation is coming, signaling it *now* via docs so users can plan the migration reduces the scramble later. Also: what's the migration path — is there a state migration or do users have to `terraform state rm` + `terraform import`?

7. **Validation of `Cases[].Condition`**: these conditions are evaluated server-side. No injection risk (API is authoritative), but worth confirming the plan-time validator rejects obviously-malformed conditions rather than letting a typo through to a silently-broken rule at apply time.

A 1500-line migration is hard to fully review on GitHub — if there's a summary of "what changed beyond translating SDKv2 → Plugin Framework" (e.g., new features, schema changes, validator additions), that'd make reviewing the security-relevant deltas much faster.

drichards-87 (approved):

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_DefaultTags.yaml
```diff
@@ -34,20 +34,20 @@ interactions:
...
...on":"third \u003e 9000","name":"low case","notifications":[],"status":"low"}],"hasExtendedTitle":true,"isEnabled":false,"message":"acceptance rule triggered","name":"tf-TestAccDatadogSecurityMonitoringRule_DefaultTags-local-1778008986","options":{"decreaseCriticalityBasedOnEnv":true,"detectionMethod":"threshold","evaluationWindow":300,"keepAlive":600,"maxSignalDuration":900},"queries":[{"aggregation":"count","dataSource":"logs","groupByFields":["host"],"hasOptionalGroupByFields":false,"name":"first","query":"does not really match much"},{"aggregation":"cardinality","dataSource":"logs","distinctFields":["@orgId"],"groupByFields":["host"],"hasOptionalGroupByFields":false,"name":"second","query":"does not really match much either"},{"aggregation":"sum","dataSource":"logs","groupByFields":["host"],"hasOptionalGroupByFields":false,"metric":"@network.bytes_read","metrics":["@network.bytes_read"],"name":"third","query":"does not really match much either"}],"referenceTables":[{"checkPresence":true,"columnName":"column1","logFieldPath":"@testattribute","ruleQueryName":"first","tableName":"table1"}],"tags":["baz","default_key:default_value","foo:bar","foo:double_bar"],"type":"log_detection"}
```
hachem-dd, line 50: `distinctFields` changed between these 2 versions - the empty ones dissapeared. Why? Is this ok?

docs/resources/security_monitoring_rule.md
```diff
@@ -100,8 +101,8 @@ Required:
 Optional:
 
 - `action` (Block List) Action to perform when the case trigger (see [below for nested schema](#nestedblock--case--action))
-- `condition` (String) A rule case contains logical operations (`>`,`>=`, `&&`, `||`) to determine if a signal should be generated based on the event counts in the previously defined queries.
-- `name` (String) Name of the case.
+- `condition` (String) A rule case contains logical operations (`>`,`>=`, `&&`, `||`) to determine if a signal should be generated based on the event counts in the previously defined queries. Defaults to `""`.
```
hachem-dd, line 104: Need to double check if `""` is allowed for a condition

docs/resources/security_monitoring_rule.md
```diff
@@ -139,16 +140,16 @@ Required:
...
-- `keep_alive` (Number) Once a signal is generated, the signal will remain “open” if a case is matched at least once within this keep alive window (in seconds). Valid values are `0`, `60`, `300`, `600`, `900`, `1800`, `3600`, `7200`, `10800`, `21600`, `43200`, `86400`.
-- `max_signal_duration` (Number) A signal will “close” regardless of the query being matched once the time exceeds the maximum duration (in seconds). This time is calculated from the first seen timestamp. Valid values are `0`, `60`, `300`, `600`, `900`, `1800`, `3600`, `7200`, `10800`, `21600`, `43200`, `86400`.
-- `new_value_options` (Block List, Max: 1) New value rules specific options. (see [below for nested schema](#nestedblock--options--new_value_options))
-- `sequence_detection_options` (Block List, Max: 1) Options for rules using the sequence detection method. (see [below for nested schema](#nestedblock--options--sequence_detection_options))
-- `third_party_rule_options` (Block List, Max: 1) Options for rules using the third-party detection method. (see [below for nested schema](#nestedblock--options--third_party_rule_options))
```
hachem-dd, line 151: The lack of `Max` is due to the new provider. This is still properly enforced in the code iinm.

docs/resources/security_monitoring_rule.md
```diff
@@ -159,7 +160,7 @@ Optional:
 - `detection_tolerance` (Number) An optional parameter that sets how permissive anomaly detection is. Higher values require higher deviations before triggering a signal. Valid values are 1, 2, 3, 4, 5. Valid values are `1`, `2`, `3`, `4`, `5`.
 - `instantaneous_baseline` (Boolean) When set to true, Datadog uses previous values that fall within the defined learning window to construct the baseline, enabling the system to establish an accurate baseline more rapidly rather than relying solely on gradual learning over time. Defaults to `false`.
 - `learning_duration` (Number) Learning duration in hours. Anomaly detection waits for at least this amount of historical data before it starts evaluating. Valid values are 1, 6, 12, 24, 48, 168, 336. Valid values are `1`, `6`, `12`, `24`, `48`, `168`, `336`.
-- `learning_period_baseline` (Number) An optional override baseline to apply while the rule is in the learning period. Must be greater than or equal to 0.
+- `learning_period_baseline` (Number) An optional override baseline to apply while the rule is in the learning period. Must be greater than or equal to 0. Value must be at least 0.
```
hachem-dd: The docs is a little wonky - it is optional meaning it can be not set at all. And it says `Must be greater than or equal to 0. Value must be at least 0.` which is slightly a double-saying.

docs/resources/security_monitoring_rule.md
```diff
@@ -256,14 +257,14 @@ Required:
...
 - `data_source` (String) Source of events. Valid values are `logs`, `audit`, `app_sec_spans`, `spans`, `security_runtime`, `network`, `events`, `security_signals`. Defaults to `"logs"`.
 - `distinct_fields` (List of String) Field for which the cardinality is measured. Sent as an array.
 - `group_by_fields` (List of String) Fields to group by.
 - `has_optional_group_by_fields` (Boolean) When false, events without a group-by value are ignored by the rule. When true, events with missing group-by fields are processed with `N/A`, replacing the missing values. Defaults to `false`.
 - `indexes` (List of String) List of indexes to run the query on when the data source is `logs`. Supports only one element. Used only for scheduled rules (in other words, when `scheduling_options` is defined).
-- `metric` (String, Deprecated) The target field to aggregate over when using the `sum`, `max`, or `geo_data` aggregations. **Deprecated.** Configure `metrics` instead. This attribute will be removed in the next major version of the provider.
+- `metric` (String, Deprecated) The target field to aggregate over when using the `sum`, `max`, or `geo_data` aggregations.
```
hachem-dd, line 267: This is **_still_** deprecated, and _metrics_ should be used instead, still.

docs/resources/security_monitoring_rule.md
```diff
@@ -256,14 +257,14 @@ Required:
 
 Optional:
 
-- `agent_rule` (Block List, Deprecated) **Deprecated**. It won't be applied anymore. **Deprecated.** `agent_rule` has been deprecated in favor of new Agent Rule resource. (see [below for nested schema](#nestedblock--query--agent_rule))
+- `agent_rule` (Block List, Deprecated) **Deprecated**. It won't be applied anymore. (see [below for nested schema](#nestedblock--query--agent_rule))
```
hachem-dd, line 260: Same thing - still Deprecated.

datadog/tests/data_source_datadog_security_monitoring_rules_test.go
hachem-dd, line 1: 👍

datadog/fwprovider/resource_datadog_security_monitoring_rule.go
```diff
@@ -0,0 +1,2648 @@
...
+	// Mirror into the local plan so validation (which reads `plan`, not
+	// `response.Plan`) sees the same tags create/update will send.
+	plan.EffectiveTags = combinedTags
+
+	normalizeDeprecatedMetricPlan(ctx, &plan, &config, response)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	r.resourceDatadogSecurityMonitoringRuleCustomizeDiff(ctx, &plan, &state, response)
```
hachem-dd, line 2572: This function says `error validating security monitoring rule, retrying` on an error path - it does not do any retrying.

Did we want to do retry?

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_RemoveReferenceTables.yaml
```diff
@@ -305,20 +305,20 @@ interactions:
...
+            {"cases":[{"condition":"first_updated \u003e 3","name":"high case (updated)","notifications":["@user"],"status":"medium"},{"condition":"first_updated \u003e 0","name":"warning case (updated)","notifications":[],"status":"high"}],"hasExtendedTitle":false,"isEnabled":true,"message":"acceptance rule triggered (updated)","name":"tf-TestAccDatadogSecurityMonitoringRule_RemoveReferenceTables-local-1778008975 - updated","options":{"decreaseCriticalityBasedOnEnv":false,"detectionMethod":"threshold","evaluationWindow":60,"keepAlive":300,"maxSignalDuration":600},"queries":[{"aggregation":"cardinality","dataSource":"logs","distinctFields":["@orgId"],"groupByFields":["service"],"hasOptionalGroupByFields":false,"name":"first_updated","query":"does not really match much (updated)"}],"tags":["i:tomato","u:tomato"],"type":"log_detection"}
```
hachem-dd, line 321: `metric` and `metrics` were removed here as they wre empty, from the request body. 

Is the behavior still identical API-wise? ( I think so but double checking again )

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRuleJSON_Basic.yaml
hachem-dd, line 1: 👍

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_ThirdParty.yaml
```diff
@@ -6,14 +6,14 @@ interactions:
...
+            {"cases":null,"hasExtendedTitle":false,"isEnabled":true,"message":"third party rule triggered","name":"tf-TestAccDatadogSecurityMonitoringRule_ThirdParty-local-1778008986","options":{"decreaseCriticalityBasedOnEnv":false,"detectionMethod":"third_party","keepAlive":0,"maxSignalDuration":900,"thirdPartyRuleOptions":{"defaultStatus":"info","rootQueries":[{"groupByFields":["instance-id"],"query":"source:guardduty @data.resourceType:*EC2*"},{"groupByFields":[],"query":"source:guardduty"}]}},"queries":null,"tags":[],"thirdPartyCases":[{"name":"High severity alert","notifications":["@slack-channel"],"query":"@alert.severity:[5 TO 10]","status":"high"},{"name":"Low severity alert","notifications":[],"query":"@alert.severity:[1 TO 4]","status":"low"}],"type":"log_detection"}
```
hachem-dd, line 16: Tags used to be omitted. Double-checking again: is the behavior identical when we don't provide the tags field at all than when we do provide it but have it as an empty list?

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_ThirdParty.yaml
```diff
@@ -131,26 +131,26 @@ interactions:
...
+            {"cases":null,"hasExtendedTitle":false,"isEnabled":true,"message":"third party rule triggered","name":"tf-TestAccDatadogSecurityMonitoringRule_ThirdParty-local-1778008986","options":{"decreaseCriticalityBasedOnEnv":false,"detectionMethod":"third_party","evaluationWindow":0,"keepAlive":0,"maxSignalDuration":900,"thirdPartyRuleOptions":{"defaultStatus":"info","rootQueries":[{"groupByFields":["instance-id"],"query":"source:guardduty @data.resourceType:*EC2*"},{"groupByFields":[],"query":"source:guardduty"}]}},"queries":null,"tags":[],"thirdPartyCases":[{"name":"High severity alert","notifications":["@slack-channel"],"query":"@alert.severity:[5 TO 10]","status":"high"},{"name":"Low severity alert","notifications":[],"query":"@alert.severity:[1 TO 4]","status":"low"}],"type":"log_detection"}
```
hachem-dd, line 153: Why did we remove signalTitleTemplate and defaultNotifications here?

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_SignalCorrelation.yaml
hachem-dd, line 1: 👍

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_SequenceDetection.yaml
hachem-dd, line 1: 👍

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_ScheduledRule.yaml
hachem-dd, line 1: 👍

docs/resources/security_monitoring_rule.md
alvizi-dd: Correct — listvalidator.SizeAtMost(1) is still applied to all four *_options blocks (new_value_options, impossible_travel_options, third_party_rule_options, sequence_detection_options). It's just a cosmetic difference in how tfplugindocs renders FW list-size validators vs SDKv2 MaxItems. The runtime constraint is unchanged.

docs/resources/security_monitoring_rule.md
alvizi-dd: Confirmed — "" is accepted by the API (e.g., a single-case rule doesn't need a comparison expression). The Default: "" was added intentionally to preserve SDKv2 migration parity: in SDKv2 an unset Optional string serialized to "" automatically, so without the explicit default we'd see spurious diffs when migrating existing state.

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_DefaultTags.yaml
alvizi-dd: Yes, this is expected. The SDKv2 schema serialized every unset Optional field as its zero value, so the request always carried "distinctFields":[], "metric":"", "metrics":[] for queries that didn't use them. The framework provider correctly omits unset Optional fields from the payload, so only the cardinality query (which actually sets distinct_fields) sends the field now. The API treats omitted and [] identically, so behavior is unchanged — the new payloads are just cleaner.

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_RemoveReferenceTables.yaml
alvizi-dd: Same as `distinctFields` above

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_ThirdParty.yaml
alvizi-dd: Same as distinctFields above

hachem-dd:
Only this 
image

And I think the rest can be resolved for later versions

alvizi-dd:
> Only this image
> 
> And I think the rest can be resolved for later versions

@hachem-dd In practice, this is the same behavior of the SDKv2

datadog/fwprovider/resource_datadog_security_monitoring_rule.go
alvizi-dd: Nope, nice catch - Updated in [[SEC-28844] Address PR review: extractor + validate cleanup](https://github.com/DataDog/terraform-provider-datadog/pull/3716/commits/b5b7891e4bca680863d1d6cb33ba2db00a0fe748)

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringRule_ThirdParty.yaml
alvizi-dd: Yes — for the security monitoring rules API, tags omitted and tags:[] are treated identically (both mean "no user-defined tags"). The reason FW now sends tags:[] here is that we always compute effective_tags during plan (fwutils.CombineTags(plan.Tags, defaultTags)), so it ends up as a known empty set rather than null even when neither user tags nor provider default tags are set, and we serialize the resolved value.

docs/resources/security_monitoring_rule.md
alvizi-dd: Fixed in [[SEC-28844] Address PR review: extractor + validate cleanup](https://github.com/DataDog/terraform-provider-datadog/pull/3716/commits/b5b7891e4bca680863d1d6cb33ba2db00a0fe748)

docs/resources/security_monitoring_rule.md
alvizi-dd: This is auto generated, the deprecation message is set in our side, but the docs weren't generated accordingly. I could find at least one other fw resource with the same mismatch

docs/resources/security_monitoring_rule.md
alvizi-dd: Same as above

datadog/fwprovider/resource_datadog_security_monitoring_rule.go
hachem-dd, line 1037: Can be looked at in a later commit - logic here could be mergeable for simpler code.

Not a blocker.

datadog/fwprovider/resource_datadog_security_monitoring_rule.go
```diff
@@ -1171,8 +1168,11 @@ func extractSignalRuleQueries(ctx context.Context, responseRuleQueries []datadog
 			ruleQuery.CorrelatedQueryIndex = types.StringValue("")
 		}
 
-		// Optional-only — only set when API returns a meaningful value
-		if name, ok := responseRuleQuery.GetNameOk(); ok && *name != "" {
+		// Optional-only — pass the API value through. The API returns "" even
+		// when the user didn't author `name`, so reconcileEmptySignalQueryFields
+		// reconciles empty strings against the prior plan/state to keep
+		// post-apply state equal to plan.
+		if name, ok := responseRuleQuery.GetNameOk(); ok {
```
hachem-dd: Can you explain this change?

hachem-dd (approved):
I'm tempted to write "YOLO" but I did spend a _good good chunk_ reviewing this, heh. And from my tests, catch the small diffs you did mention in the PR description regarding tags and validate, it seems to be in parity and fixes the previous errors of "null == 0".

In either case if we missed any edge case we can fix in a followup PR.

datadog/fwprovider/resource_datadog_security_monitoring_rule.go
alvizi-dd: The API returns `""` for `name` even when the user didn't write one. Without the reconciler, state would store `""` while the plan has `null`, producing a forever-diff.

Example: user config has query `{ query = "service:foo" }` (no name). Plan: `name = null`. API response: `name = ""`. Reconciler: sees empty-known string in API state with prior `null` → restores `null`. State == plan.

Supam (approved):

PR #3744, opened 2026-04-30
[datadog_synthetics_test] Add drag and drop browser step support

dd-victor-cumer:
Adds drag and drop as browser test step types in datadog_synthetics_test, closing the gap opened by the API-side work in SYNTH-23610.

What's new:
- New **drag_drop_options** param block (delay, offset.x/y) shared by both step types
- Plan-time validation that a drag step is always immediately followed by a drop step
- Acceptance tests for the happy path and both ordering failure modes
- Docs and example updated

Note: there is an ongoing `synthetics-drag-and-drop` FF that was GA to all orgs and all envs earlier this week.

References:
- Jira: SYNTH-25730
- API work: SYNTH-23610


datadog/resource_datadog_synthetics_test_.go
```diff
@@ -5750,6 +5844,25 @@ func convertStepParamsValueForState(key string, value interface{}) (interface{},
...
+		if delay, ok := optsMap["delay"]; ok {
+			result["delay"] = int(delay.(float64))
+		}
+		if offsetRaw, ok := optsMap["offset"].(map[string]interface{}); ok {
+			result["offset"] = []interface{}{
+				map[string]interface{}{
+					"x": int(offsetRaw["x"].(float64)),
+					"y": int(offsetRaw["y"].(float64)),
+				},
+			}
```
Supam, line 5872: The type assertions on the payload can panic if the fields are missing or null 
- int(delay.(float64)) panics if delay is nil (JSON null)
- int(offsetRaw["x"].(float64)) and int(offsetRaw["y"].(float64)) panic if either is missing or null.
This can cause a crash of the provider instead of a graceful error.

A quick fix could be to call the `.(float64)` in the if statement directly.
```go
if d, ok := optsMap["delay"].(float64); ok {
    result["delay"] = int(d)
}
if x, ok := offsetRaw["x"].(float64); ok {
    // …
}
```

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1980,6 +2022,39 @@ func resourceDatadogSyntheticsTestCustomizeDiff(ctx context.Context, diff *schem
 	return nil
 }
 
+func validateDragDropStepOrdering(diff *schema.ResourceDiff) error {
```
Supam, line 2027: NIT: This validator returns on the first ordering error. With multiple drag/drop pairs, users might fix one, re-plan, and still see an error. 

We can use `errors.Join` to surface all errors at once, avoiding unpleasant fix->plan->new error loops.

Supam (commented):
One nit, one comment

estherk15 (approved):

datadog/resource_datadog_synthetics_test_.go
dd-victor-cumer: Yes indeed! Thank you @Supam I'll push some changes 👍

datadog/resource_datadog_synthetics_test_.go
dd-victor-cumer: done

datadog/resource_datadog_synthetics_test_.go
```diff
@@ -1582,6 +1583,41 @@ func syntheticsBrowserStepParams() schema.Schema {
...
 				},
+				"drag_drop_options": {
+					Description: `Options for a "drag" or "drop" step.`,
+					Type:        schema.TypeList,
+					MaxItems:    1,
+					Optional:    true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"delay": {
+								Description: "Delay in milliseconds before performing the action (0–9999).",
```
romainberger: nit: we could use a `ValidateFunc` here to have a faster validation loop for users.

romainberger (approved):

datadog/resource_datadog_synthetics_test_.go
dd-victor-cumer: @Supam good for you?

datadog/resource_datadog_synthetics_test_.go
dd-victor-cumer: You mean instead of using the validateDragDropStepOrdering afterward?

dd-victor-cumer: Ah no for the allowed value range for delay ? Let me add that 👍

dd-victor-cumer: done

Supam (approved):

PR #3757, opened 2026-05-05
[datadog_provider] Add native PAT (Bearer) auth support

luczhou:
## Description

CRED-2570: PATs (`ddpat_*`) currently only work in the Datadog Terraform provider via a workaround — stuff the PAT into `DD_APP_KEY` and set `validate = false`. This PR adds a first-class `bearer_token` attribute that plumbs the token through the SDK's `ContextAccessToken` so requests emit `Authorization: Bearer <token>` natively. The same attribute also accepts service-account access tokens (`ddsat_*`), and the name was chosen to be forward-compatible with OAuth tokens / WIF / future identity-issued bearer credentials (per the [#aaa-credential-mgmt naming thread](https://dd.slack.com/archives/C09JJ76EE0P/p1779811175996209), final ack from AAA + AuthN was `DD_BEARER_TOKEN`).

JIRA: https://datadoghq.atlassian.net/browse/CRED-2570

## Changes

- New `bearer_token` provider attribute (Optional, Sensitive) on both the SDKv2 and framework provider schemas, with `DD_BEARER_TOKEN` / `DATADOG_BEARER_TOKEN` env-var fallback.
- Auth context now sets `datadog.ContextAccessToken` when `bearer_token` is configured. `bearer_token` takes precedence over `api_key`/`app_key` when both are set — a configured bearer token is an explicit signal to use Bearer auth.
- Startup validation for bearer-token-only configurations calls `KeyManagementApi.ValidateAPIKey` (`GET /api/v2/validate_keys`) with the token injected into `ContextAPIKeys["appKeyAuth"]` — the authn-validation backend accepts PATs/SATs in the `DD-APPLICATION-KEY` slot. API/APP key configs still hit `/api/v1/validate`.
- Bumps `datadog-api-client-go` to `v2.59.1-0.20260512205120-0d69104e7dd3` to pick up the regenerated `Validate API key` operation (datadog-api-spec [#5605](https://github.com/DataDog/datadog-api-spec/pull/5605) / [#5638](https://github.com/DataDog/datadog-api-spec/pull/5638)).
- Configure-time gate relaxed: `api_key + app_key` is no longer required when a `bearer_token` or `cloud_provider_type` is provided.
- Provider docs (`docs/index.md`) list `bearer_token` alongside `api_key` / `app_key`.

**Out of scope** (per the ticket): the legacy `zorkian/go-datadog-api` community client still uses API/APP keys only.

## Security review follow-up

Review surfaced that `datadog-api-client-go`'s debug-dump path redacts `ContextAPIKeys` values but not `ContextAccessToken`, so `TF_LOG=DEBUG` with a bearer token would leak the credential to stderr/CI/log shippers. Fix is being made upstream rather than worked around locally, tracked under [CRED-2625](https://datadoghq.atlassian.net/browse/CRED-2625):

- **[datadog-api-client-go#4098](https://github.com/DataDog/datadog-api-client-go/pull/4098)** — adds `ContextAccessToken` redaction to the SDK's `CallAPI` debug dump. **Once this merges, this PR bumps `go.mod` to a commit that includes it and the security finding is closed.**
- [datadog-api-client-typescript#4168](https://github.com/DataDog/datadog-api-client-typescript/pull/4168) — same fix shape (cross-language parity).
- [datadog-api-client-ruby#3340](https://github.com/DataDog/datadog-api-client-ruby/pull/3340) — same fix shape (cross-language parity).
- Java leaks all credentials in debug mode (Jersey `LoggingFeature.PAYLOAD_ANY` with no header filter); fix shape being coordinated with api-platform separately (tracked in CRED-2625, no public issue).

## Testing

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `gofmt` clean on touched files
- [x] Framework provider configure unit tests pass: `TestFrameworkProviderConfigure_BearerTokenOnly`, `_BearerTokenEnvVar`, `_BearerTokenWinsOverAPIKey`, `_NoCredentialsErrors`
- [x] SDKv2 provider configure unit tests pass: `TestProviderConfigure_BearerTokenOnly`, `_BearerTokenEnvVar`, `_BearerTokenWinsOverAPIKey`, `_NoCredentialsErrors`
- [x] Staging smoke test (`api.datad0g.com`, real PAT): `terraform plan` succeeded; `ValidateAPIKey` returned 200 with the token in `DD-APPLICATION-KEY` and empty `DD-API-KEY`. *(Run pre-rename — wire format and validation path are unchanged; only the attribute/env-var name changed.)*

[CRED-2625]: https://datadoghq.atlassian.net/browse/CRED-2625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joepeeples (approved):

tausman (approved):
only thing to consider is pat vs sat naming? we've been renaming a lot of things & not sure if this should be included

datadog/fwprovider/framework_provider.go
```diff
@@ -584,6 +598,10 @@ func defaultConfigureFunc(p *FrameworkProvider, request *provider.ConfigureReque
 			diags.AddError("cloud_provider_type must be set to a valid value unless validate = false", "")
 			return diags
 		}
+	} else if bearerToken != "" {
+		// bearer_token takes precedence over api_key/app_key when both are set:
+		// a configured bearer token is an explicit signal to use Bearer auth.
```
tausman, line 605: [Question] Is there a benefit of giving bearer token preference? Should we just make it so that this fails fast? I don't think we should ever have both set & seems like a footgun to even allow customers to set both.

If this makes implementation significantly easier or something, I think it's okay to keep, but I would consider the fail fast option.

tausman (approved):
LGTM on the rename 👍 
Up to you on whether we go this route, but I think we can consider making API+App key & Bearer Token illegal. But don't know the level of effort there.

fpighi (approved):

PR #3759, opened 2026-05-06
[datadog_azure_integration] Introduce Preview Secretless Auth Flag 

benjjs:
### Motivation
We are starting a closed beta of a new azure authentication mechanism which does not use a client secret. Some of the beta customers use terraform, so we'd like them to be able to provision azure integrations without client secrets.

### Summary
Make the client secret field optional and add a new `secretless_auth_enabled` field. I considered simply using a null client secret to indicate the intent to use secretless auth. This would prevent awkward states where, e.g., no client secret is provided but secretless auth is not enabled, but would also make it very easy to accidentally enable secretless auth by simply omitting the client secret field.

This is problematic because certain things need to be true on the azure side for secretless auth to work. We can't really validate that except at runtime, but we can at least force them to explicitly flip the flag, where we have documented that they will need a federated credential on the app registration for it to work. Plus, the feature is currently in beta so we don't want this to be in any sense "default" behavior.

Note that I had to bump the api spec version to get the new field.

### Testing
Tested locally in the method described in `DEVELOPMENT.md` with 3 cases: null client secret + flag enabled, client secret + flag disabled, null client secret + flag disabled.  Go the expected results each time with the first two passing `terraform validate` and `terraform plan` but the third failing.

Added a new validation test specifically for the no secret + no secretless case mentioned above.


hestonhoffman (approved):

amarshall-dd (approved):
Only high-level review from api-reliability

parsons90 (approved):

PR #3761, opened 2026-05-06
[datadog_service_access_token] CRED-2421 Add service_access_token resource for managing SATs

tausman:
## Summary

Adds a new managed Terraform resource `datadog_service_access_token` for full lifecycle management of Datadog **Service Access Tokens (SATs)** — access tokens scoped to a service account.

The resource interface mirrors `datadog_service_account_application_key` (RequiresReplace on `service_account_id`, sensitive `key` only available on creation, composite-id import that drops the secret with a warning).

By design the resource is limited to service-account-owned tokens; the user/me PAT path is intentionally not exposed.

## Schema

| Attribute | Required | Optional | Computed | Notes |
|---|---|---|---|---|
| `service_account_id` | yes | | | RequiresReplace |
| `name` | yes | | | Mutable. Min length 1 (`stringvalidator.LengthAtLeast(1)`). |
| `scopes` | yes | | | At least 1 element (`setvalidator.SizeAtLeast(1)`). Mutable. |
| `expires_at` | | yes | | RFC3339 string. **Immutable after creation** — adding, changing, or removing it is rejected at plan time with a clear error. To rotate the expiration, destroy + re-create the resource. The Datadog API caps expirations at 365 days from creation. |
| `key` | | | yes | Sensitive, returned only on creation, `UseStateForUnknown` |
| `created_at` | | | yes | RFC3339, `UseStateForUnknown` (immutable) |
| `last_used_at` | | | yes | RFC3339, empty if never used |
| `public_portion` | | | yes | Identifies the token; matches the API field name. `UseStateForUnknown`. |
| `id` | | | yes | |

The immutability of `expires_at` is enforced by a reusable `planmodifiers.ImmutableString` plan modifier rather than `RequiresReplace`. Rationale: silently destroying and re-creating a credential when a user types a new date is a footgun; failing loudly at plan time forces an explicit destroy/re-create. The modifier lives in `datadog/internal/planmodifiers/types_string_immutable.go` so future resources can reuse it.

## Backing API

Bumps `datadog-api-client-go` to `v2.60.1-0.20260529072814-6fdb70862f77` to pick up the rename from `personal_access_tokens` to `service_access_tokens` (regenerated from datadog-api-spec PR #5744). The resource uses these methods on `ServiceAccountsApi`:

- `CreateServiceAccountAccessToken` → returns `ServiceAccessTokenCreateResponse`
- `GetServiceAccountAccessToken` → returns `ServiceAccessTokenResponse`
- `UpdateServiceAccountAccessToken` → takes `ServiceAccountAccessTokenUpdateRequest`
- `RevokeServiceAccountAccessToken`

No more `PersonalAccessToken*` references in our resource code.

## Files

New:
- `datadog/fwprovider/resource_datadog_service_access_token.go`
- `datadog/internal/planmodifiers/types_string_immutable.go` — reusable `ImmutableString` plan modifier
- `datadog/tests/resource_datadog_service_access_token_test.go`
- `examples/resources/datadog_service_access_token/{resource.tf,import.sh}`
- `docs/resources/service_access_token.md` (generated by `tfplugindocs`)

Edited:
- `datadog/fwprovider/framework_provider.go` — register `NewServiceAccessTokenResource`
- `datadog/tests/provider_test.go` — register the test file in `testFiles2EndpointTags`
- `.github/CODEOWNERS` — three entries for `datadog_service_access_token` files, the docs file, and the examples directory, all owned by `@DataDog/credential-management` (overrides the broader `*service_account*` rule routing to `team-aaa`)
- `go.mod`, `go.sum`

## Test plan

- [x] `make fmtcheck` passes
- [x] `make test` passes (replay mode)
- [x] `make vet` passes
- [x] `make docs` produces clean output
- [x] `golangci-lint run --new-from-rev=origin/master` — 0 issues
- [x] All 6 acceptance tests pass live against staging (`RECORD=none make testacc TESTARGS='-run TestAccServiceAccessToken'`):
  - `TestAccServiceAccessTokenBasic` — create + rename + scope update (multi-step in-place)
  - `TestAccServiceAccessToken_WithExpiry` — create with expiry → try changing → try dropping (both expect immutability error)
  - `TestAccServiceAccessToken_AddExpiryFails` — create without expiry → try adding (expects immutability error)
  - `TestAccServiceAccessToken_Error` — empty `scopes` (client validator) + invalid scope name (server)
  - `TestAccServiceAccessToken_EmptyName` — empty `name` (client validator)
  - `TestAccServiceAccessToken_Import` — import via composite `<service_account_id>:<token_id>`

## Notes for reviewers

- `scopes` is required (the SAT API requires it), which differs from `datadog_application_key` where it's optional.
- `expires_at` is intentionally immutable rather than RequiresReplace. To change an expiration, the user must explicitly destroy and re-add the resource (e.g. `terraform destroy -target=...` then re-apply).
- The Datadog API caps `expires_at` at 365 days from creation; values beyond that are rejected server-side.
- Closes CRED-2421


cswatt (approved):

examples/resources/datadog_service_account_token/import.sh
```diff
@@ -0,0 +1,2 @@
+# Importing a service account token cannot import the value of the key.
```
Supam: Could you add this file the the CODEOWNERS as being owned by your team ?

examples/resources/datadog_service_account_token/resource.tf
```diff
@@ -0,0 +1,24 @@
+# Source the permissions for scoped tokens
+data "datadog_permissions" "dd_perms" {}
```
Supam: Could you add this file the the CODEOWNERS as being owned by your team ?

Supam (approved):
Two CODEOWNER requests.
Otherwise looks good to me

emubello:
Can we call these `service_access_token` to match the public apis? Those changes are here https://github.com/DataDog/dd-source/pull/428383

datadog/fwprovider/resource_datadog_service_account_token.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func (r *serviceAccountTokenResource) Configure(_ context.Context, request resource.ConfigureRequest, response *resource.ConfigureResponse) {
+	providerData, _ := request.ProviderData.(*FrameworkProvider)
+	r.Api = providerData.DatadogApiInstances.GetServiceAccountsApiV2()
+	r.Auth = providerData.Auth
+}
+
+func (r *serviceAccountTokenResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = "service_account_token"
```
emubello: the response type is `service_access_tokens`

datadog/fwprovider/resource_datadog_service_access_token.go
```diff
@@ -0,0 +1,363 @@
...
+	req := datadogV2.NewServiceAccountAccessTokenCreateRequestWithDefaults()
+	req.SetData(*data)
+
+	return req, diags
+}
+
+func (r *serviceAccountTokenResource) buildServiceAccountTokenUpdateRequestBody(_ context.Context, state *serviceAccountTokenModel) (*datadogV2.PersonalAccessTokenUpdateRequest, diag.Diagnostics) {
+	diags := diag.Diagnostics{}
+
+	attributes := datadogV2.NewPersonalAccessTokenUpdateAttributesWithDefaults()
```
emubello: Why is this personal instead of service here?

examples/resources/datadog_service_account_token/import.sh
tausman: Done — added `examples/resources/datadog_service_account_token/` to CODEOWNERS owned by `@DataDog/credential-management`.

🤖 Assisted with AI

examples/resources/datadog_service_account_token/resource.tf
tausman: Done — see the new CODEOWNERS entry for `examples/resources/datadog_service_account_token/`.

🤖 Assisted with AI

datadog/fwprovider/resource_datadog_service_account_token.go
tausman: Good catch — the wire-level type field is `personal_access_tokens` in the current api-client because the SAT endpoints currently reuse the PAT types in the OpenAPI spec. That value is set by `PERSONALACCESSTOKENSTYPE_PERSONAL_ACCESS_TOKENS` in [`model_service_account_access_token_create_data.go`](https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_service_account_access_token_create_data.go#L40) — we can't override it from here.

Once dd-source#428383 lands and the api-client regenerates with a dedicated `ServiceAccessTokensType` enum, we can swap in `service_access_tokens`. I've cleaned up our user-facing description and Go identifiers to use "service account token" terminology and added a code comment pointing at the upcoming rename for the spots that still reference the PAT-named types.

🤖 Assisted with AI

datadog/fwprovider/resource_datadog_service_access_token.go
tausman: Same root cause as the line-57 comment — the update request type is `datadogV2.PersonalAccessTokenUpdateRequest` because the SAT update endpoint reuses the PAT update types in the OpenAPI spec. There's no separate `ServiceAccessTokenUpdateRequest` in the api-client yet. Our wrapper is named `buildServiceAccountTokenUpdateRequestBody`, and I've added a comment above the state-update helpers explaining the upcoming rename to be tracked alongside dd-source#428383.

🤖 Assisted with AI

datadog/fwprovider/resource_datadog_service_account_token.go
tausman: Okay, taking a look at this manually now, we will need to update the open api spec. Can you handle that as part of the rename work & then I can update this? @emubello

datadog/fwprovider/resource_datadog_service_access_token.go
tausman: ^^yeah, looks like the open api spec needs to be updated for the recent service account token changes

datadog/fwprovider/resource_datadog_service_account_token.go
emubello: Sure, I can update it. But I was also suggesting that we rename the entire terraform resource to be service _access_ token for consistency.

tausman: ah, I see. my bad. I didn't realize the names were different. agreed that we should stay consistent w/ the public apis. fixing now!

emubello: Thanks!

emubello:
The description prob needs to be updated too, but this looks good!

emubello (approved):

PR #3770, opened 2026-05-11
[datadog_monitor] Add support for metrics augmented queries

kevinpombo-datadog:
This Work supports the new augmented query syntax in terraform today to support saving f&f Metric monitors. This was previously introduced in [kima](https://github.com/DataDog/dogweb/pull/159684) and [api spec](https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/monitor.yaml#L768) services

Example Query:
```
resource "datadog_monitor" "r" {
  name    = "%s"
  type    = "query alert"
  message = "aggregate augmented fw test"
  query   = "formula(\"query1\").rollup(\"sum\").last(\"5m\") > 124"

  monitor_thresholds {
    critical = "124"
  }

  variables {
    aggregate_augmented_query {
      name        = "query1"
      data_source = "aggregate_augmented_query"

      augment_reference_table {
        name         = "filter_query"
        data_source  = "reference_table"
        table_name   = "test_table"
        columns {
          name = "org_id"
        }
        columns {
          name = "name"
        }
      }

      base_metrics_query {
        data_source = "metrics"
        name        = "query1"
        query       = "avg:dd{*} by {org_id}.as_count()"
      }

      join_condition {
        augment_attribute = "org_id"
        base_attribute    = "org_id"
        join_type         = "inner"
      }

      compute {
        name        = "compute_result"
        aggregation = "max"
      }

      group_by {
        facet  = "org_id"
        source = "filter_query"
      }
      group_by {
        facet  = "name"
        source = "filter_query"
      }
    }
  }
}

### CODE REVIEW

**Review by hestonhoffman — APPROVED** [review]

File: datadog/fwprovider/resource_datadog_monitor.go [hunk 1]
```diff
@@ -138,9 +138,10 @@ type Recurrence struct {
 }
 
 type Variable struct {
-	EventQuery       []EventQuery       `tfsdk:"event_query"`
-	CloudCostQuery   []CloudCostQuery   `tfsdk:"cloud_cost_query"`
-	DataQualityQuery []DataQualityQuery `tfsdk:"data_quality_query"`
+	EventQuery              []EventQuery              `tfsdk:"event_query"`
+	CloudCostQuery          []CloudCostQuery          `tfsdk:"cloud_cost_query"`
+	DataQualityQuery        []DataQualityQuery        `tfsdk:"data_quality_query"`
+	AggregateAugmentedQuery []AggregateAugmentedQuery `tfsdk:"aggregate_augmented_query"`
```
unclebconnor, line 144: Bummer this isn't available as a shared struct somewhere (owned by metrics folks).  I'm assuming these are also usable in dashboards, etc. Is it worth asking them to add and manage that or is it better for us to keep these separate?

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -263,6 +309,128 @@ func (r *monitorResource) Metadata(_ context.Context, request resource.MetadataR
...
+					Description: "Group by options.",
+					NestedObject: schema.NestedBlockObject{
+						Attributes: map[string]schema.Attribute{
+							"facet": schema.StringAttribute{
+								Required:    true,
+								Description: "The event facet.",
+							},
+							"source": schema.StringAttribute{
+								Optional:    true,
+								Description: "For composite aggregate-augmented queries, identifies which sub-query this group-by facet refers to (for example `filter_query`).",
```
unclebconnor: I think there are only 2 options for this.  Worth making it an enum?

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -875,6 +934,214 @@ func (r *monitorResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+																Optional:    true,
+																Description: "Optional alias for the column.",
+															},
+														},
+													},
+												},
+											},
+										},
+									},
+									"augment_event_query": r.variablesEventQueryListBlock("Events augment query. Conflicts with `augment_reference_table` in configuration."),
```
shaneyuandd: nit: in SDKv2, a `MaxItem: 1` is set, but same behavior is not in fwprovider. Wonder if we want to have some validation here as well. 
Same applies to `base_event_query`.

datadog/resource_datadog_monitor.go
```diff
@@ -466,6 +466,352 @@ func resourceDatadogMonitor() *schema.Resource {
...
+			Type:        schema.TypeList,
+			Optional:    true,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "An array of index names to query in the stream.",
+		},
+		"compute": {
+			Type:        schema.TypeList,
+			Required:    true,
+			Description: "The compute options.",
+			Elem: &schema.Resource{
```
shaneyuandd, line 505: can set a minItem: 1 here like API spec.

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -1258,6 +1529,12 @@ func (r *monitorResource) buildEventQueryStruct(ctx context.Context, eventQs []E
 			for _, groupBy := range groupBys {
 				groupByReq := datadogV1.MonitorFormulaAndFunctionEventQueryGroupBy{}
 				fwutils.SetOptString(groupBy.Facet, groupByReq.SetFacet)
+				if !groupBy.Source.IsNull() && groupBy.Source.ValueString() != "" {
+					if groupByReq.AdditionalProperties == nil {
+						groupByReq.AdditionalProperties = map[string]interface{}{}
+					}
+					groupByReq.AdditionalProperties["source"] = groupBy.Source.ValueString()
```
shaneyuandd: I think this is a miss in the original API spec PR. Is it possible that we add `source` there and not define `AdditionalProperties`?

datadog/fwprovider/resource_datadog_monitor.go
shaneyuandd, line 144: In the API spec PR, we also introduce `MonitorFormulaAndFunctionAggregateFilteredQueryDefinition`, wonder if that's something we also want to cover to reach feature parity.

kevinpombo-datadog: Will create a seperate pr for this since its >1000 line change: https://github.com/DataDog/terraform-provider-datadog/pull/3796

kevinpombo-datadog: This should be fine for the currently pr, this can be swapped later on if desired from the metrics team with the same schema

shaneyuandd (approved):

amarshall-dd (approved):
Only reviewed api-reliability files (docs)

PR #3773, opened 2026-05-11
[datadog_security_monitoring_filter] Migrate to plugin framework

alvizi-dd:
## Summary

Migrates `datadog_security_monitoring_filter` from `terraform-plugin-sdk/v2` to [terraform-plugin-framework](https://developer.hashicorp.com/terraform/plugin/framework).

The SDKv2 file `datadog/resource_datadog_security_monitoring_filters.go` is deleted; the two helpers the data source still consumed (`securityMonitoringFilterSchema` and an SDKv2-typed `extractExclusionFiltersTF`) moved into `data_source_datadog_security_monitoring_filters.go`.

Schema shape is unchanged — same attributes (`name`, `query`, `is_enabled`, `version`, `filtered_data_type`, `exclusion_filter`), same validators, same defaults.

## Notable behaviors

- **`Delete` now treats 404 as success.** SDKv2 surfaced 404 on delete as an error diagnostic; the framework version treats it as success (resource is already gone — Terraform's standard convention, matches `datadog_security_monitoring_suppression`).
- **In-process write mutex.**  Added a `sync.Mutex` around Create/Update/Delete (mirrors `datadog_security_monitoring_suppression`) to prevent 409s. Note this only serializes writes within a single provider instance — out-of-process callers can still hit the 409.

## Test plan

- [x] `go build ./...`, `make fmtcheck`, `make docs && make check-docs` clean.
- [x] `TestAccDatadogSecurityMonitoringFilter` (Create + Update) passing under `RECORD=false`.
- [x] `TestAccDatadogSecurityMonitoringFilter_Import` (cold import + `ImportStateVerify`) passing.
- [x] `TestAccDatadogSecurityMonitoringFiltersDataSource` passing (data source unaffected by the resource migration).
- [x] Four migration tests added in `resource_datadog_security_monitoring_filter_migration_test.go`:
  - `_Basic` — typical config with two `exclusion_filter` blocks
  - `_NoExclusionFilters` — empty-list round-trip (SDKv2 always sent `[]` on PUT)
  - `_ExplicitFilteredDataType` — `filtered_data_type = "logs"` set explicitly (Optional+Computed user-equals-default drift)
  - `_DefaultFilteredDataType` — `filtered_data_type` omitted (framework `Default` materialization)
- [x] Manually tested in staging
 
### Migration tests — RECORD-only, manual gate

The migration tests use `ExternalProviders` to spin up provider v4.7.0 as a subprocess with its own HTTP client, bypassing the in-process VCR recorder. They `t.Skip` unless `RECORD=true`, so **CI does not execute them** — they are a manual pre-release gate. Step 1 (v4.7.0 Create) → Step 2 (FW Apply, `ExpectEmptyPlan`) is the empirical proof of HashiCorp's no-drift guarantee.

The migration tests intentionally do **not** call `t.Parallel()` — the v4.7.0 subprocess doesn't share the in-process mutex, so parallel runs hit the API's 409 lock.


datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	filterCreate := buildSecMonFilterCreatePayload(&plan)
+
+	filterWriteMutex.Lock()
+	defer filterWriteMutex.Unlock()
+
+	filterResponse, httpResponse, err := r.api.CreateSecurityFilter(r.auth, *filterCreate)
+	if err != nil {
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(utils.TranslateClientError(err, httpResponse, "error creating security monitoring filter"), ""))
```
hachem-dd, line 131: The header error is empty `""` - maybe we can add more information to it :)

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+		return
+	}
+
+	filterResponse, httpResponse, err := r.api.GetSecurityFilter(r.auth, state.ID.ValueString())
+	if err != nil {
+		if httpResponse != nil && httpResponse.StatusCode == 404 {
+			response.State.RemoveResource(ctx)
+			return
+		}
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(utils.TranslateClientError(err, httpResponse, "error fetching security monitoring filter"), ""))
```
hachem-dd, line 156: Same as above

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	filterUpdate := buildSecMonFilterUpdatePayload(&plan, int32(state.Version.ValueInt64())) //nolint:gosec // API-assigned monotonic counter, fits int32
+
+	filterWriteMutex.Lock()
+	defer filterWriteMutex.Unlock()
+
+	filterResponse, httpResponse, err := r.api.UpdateSecurityFilter(r.auth, state.ID.ValueString(), *filterUpdate)
+	if err != nil {
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(utils.TranslateClientError(err, httpResponse, "error updating security monitoring filter"), ""))
```
hachem-dd, line 197: Same as above

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+
+	filterWriteMutex.Lock()
+	defer filterWriteMutex.Unlock()
+
+	httpResponse, err := r.api.DeleteSecurityFilter(r.auth, state.ID.ValueString())
+	if err != nil {
+		if httpResponse != nil && httpResponse.StatusCode == 404 {
+			return
+		}
+		response.Diagnostics.Append(utils.FrameworkErrorDiag(utils.TranslateClientError(err, httpResponse, "error deleting security monitoring filter"), ""))
```
hachem-dd, line 224: Same as above

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+func (r *securityMonitoringFilterResource) Schema(_ context.Context, _ resource.SchemaRequest, response *resource.SchemaResponse) {
+	response.Schema = schema.Schema{
+		Description: "Provides a Datadog Security Monitoring Rule API resource for security filters.",
+		Attributes: map[string]schema.Attribute{
+			"id": utils.ResourceIDAttribute(),
+			"name": schema.StringAttribute{
+				Required:    true,
+				Description: "The name of the security filter.",
+			},
+			"version": schema.Int64Attribute{
```
hachem-dd, line 71:
```
version is Computed: true with no plan modifier. Every terraform plan that touches any field shows:
  ~ version = 1 -> (known after apply)

  SDKv2 didn't exhibit this because it used state values for Computed fields implicitly. With the Framework, Computed-only fields show as (known after apply) unless UseStateForUnknown() is applied. The version
  value is known before apply — it's the current state value — so displaying it as unknown is incorrect and confusing.

  Fix: Add PlanModifiers: []planmodifier.Int64{int64planmodifier.UseStateForUnknown()} to the version attribute.
```

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
```diff
@@ -0,0 +1,300 @@
...
+				Computed:    true,
+				Default:     stringdefault.StaticString("logs"),
+				Description: "The filtered data type.",
+				Validators: []validator.String{
+					validators.NewEnumValidator[validator.String](datadogV2.NewSecurityFilterFilteredDataTypeFromValue),
+				},
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"exclusion_filter": schema.ListNestedBlock{
```
hachem-dd, line 94: Previously this behaved without any sorting, so like a SetNestedBlock I think?

Actually nevermind, the API allows more than 1 exclusion_filter with the exact same format
```
  exclusion_filter { name = "same-name", query = "env:staging" }
  exclusion_filter { name = "same-name", query = "env:staging" }   # identical pair
```

The difference here is that now they are sorted. I think this should be mentioned in the release note - a diff is possible on the order of exclusion_filters

```
~ exclusion_filter {
      ~ name  = "exclude staging" -> "exclude canary"   ← only order changed
      ~ query = "env:staging"     -> "env:canary"
    }
  ~ exclusion_filter {
      ~ name  = "exclude canary"  -> "exclude staging"
      ~ query = "env:canary"      -> "env:staging"
    }
```

datadog/resource_datadog_security_monitoring_filters.go
```diff
@@ -1,253 +0,0 @@
...
-			StateContext: schema.ImportStatePassthroughContext,
-		},
-
-		SchemaFunc: func() map[string]*schema.Schema {
-			return securityMonitoringFilterSchema()
-		},
-	}
-}
-
-func securityMonitoringFilterSchema() map[string]*schema.Schema {
```
hachem-dd, line 35: This was moved to datadog/data_source_datadog_security_monitoring_filters.go 👍

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
alvizi-dd: Good point, but that's the standard around this repo when we use both `utils.FrameworkErrorDiag()` and `utils.TranslateClientError()`. We either set the `msg` parameter in the first or in the second method.

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
alvizi-dd: We actually had `[]planmodifier.Int64{int64planmodifier.UseStateForUnknown()}` in an earlier iteration [here](https://github.com/DataDog/terraform-provider-datadog/pull/3773/changes/6942dd750b35a9a77c5225a1157a503ed134687f#diff-a56e072da17da80088b37913297cda1cd96ff107037362895b333b6e70834142L73), but we dropped it since the the security filter API increments version on every write (both Create and Update). With `UseStateForUnknown()`, the plan would copy version = 1 from state, but UpdateSecurityFilter returns version = 2 — causing Terraform to error with "Provider produced inconsistent result after apply" because the planned value didn't match the returned value.

The (known after apply) display is intentional: the version truly is unknown before apply because the API will return a different value after any mutation. This is the correct pattern for API-managed monotonic counters.

datadog/fwprovider/resource_datadog_security_monitoring_filter.go
alvizi-dd: On the ordering concern: ListNestedBlock has identical list semantics to SDKv2's TypeList — no sorting is introduced. Both implementations store exclusion_filters in the order the API returns them. Our migration test (ExpectEmptyPlan with ExternalProviders: "4.7.0" → new provider) confirms that. Also, I tried to reproduce this scenario manually, and I didn't get any diffs

hachem-dd (approved):
Checked some examples manually

fpighi (approved):

PR #3780, opened 2026-05-13
[datadog_security_monitoring_default_rule] Migrate to plugin framework

alvizi-dd:
## Summary

Migrates `datadog_security_monitoring_default_rule` from `terraform-plugin-sdk/v2` to [terraform-plugin-framework](https://developer.hashicorp.com/terraform/plugin/framework).

The SDKv2 file `datadog/resource_datadog_security_monitoring_default_rule.go` is deleted. Schema shape (HCL attributes/blocks) is unchanged — same `case`, `query`, `filter`, `options` blocks; same `enabled`, `custom_message`, `custom_name`, `custom_tags`, `type` attributes; same validators and defaults.

## One-time migration UX

Users with `case`, `query`, or `options` blocks in 4.9.0 HCL will see a one-time `+ block` diff on first plan after upgrade. Apply reconciles state and it is non-destructive, subsequent plans are empty.

```
terraform plan    # may show + block diffs — verify values
terraform apply   # no-op against the API; state aligns
terraform plan    # empty
```

The same applies after `terraform import`: the first plan shows `+ block` diffs for any blocks declared in config, apply reconciles, and subsequent plans are empty.

Adding or removing blocks in ongoing use produces the expected `+` / `-` diffs and applies normally.

## Hard FW constraint that shaped the design

`ListNestedBlock` cannot be `Computed` in the framework. SDKv2 4.9.0 declared `case`, `query`, `options` as `Optional+Computed` and silently absorbed API-populated values into state. FW enforces strict consistency — non-Computed blocks must have plan/state counts matching the user's config, or Terraform Core emits `Provider produced inconsistent result after apply: plan (N) disagrees with count in config (M)`.

Keeping the existing HCL surface (no `block { … }` → `attr = [{ … }]` break) required:

1. **`UpgradeState` (schema v0→v1) strips `Cases`, `Queries`, `Options`** from migrated SDKv2 state.
2. **`Read` keeps in state only the blocks the user is managing.** The API always returns all blocks; we filter them down to what was already in state, so unmanaged blocks never appear in plans.
3. **`Update` sets state from plan for blocks, then patches in API-owned scalars.** Required by FW's strict post-apply consistency check.
4. **`extractDefaultRuleCases` notifications carry-forward**: when the API returns `[]` and prior was null, restore null to avoid `null ↔ []` drift.

## Behavior change vs SDKv2: un-declared `case` blocks no longer mutate the API

**Before:** Removing a `case` block from HCL and running `terraform apply` cleared that case's notifications via the API.

**After:** Un-declared cases are left untouched. Fixes a footgun where a minimal-config import + apply could wipe notifications set elsewhere (UI, other team members).

### Clearing notifications now requires declaring the case

The block is the "I manage this case" toggle. Within a declared case, omitting fields follows the same rule as top-level scalars (omit = clear):

| HCL | Effect on `high` case notifications |
|---|---|
| (no `case` block for `high`) | preserved — `high` is not managed |
| `case { status = "high" }` | cleared (declared but no `notifications`) |
| `case { status = "high" notifications = [] }` | cleared (explicit) |
| `case { status = "high" notifications = ["@x"] }` | set to `["@x"]` |

Consistent with `custom_message`, `custom_name`, `custom_tags` — omitting from config clears.

## Other behavior change: `options` block no longer auto-appears in state

**Before:** Minimal configs ended up with `options { decrease_criticality_based_on_env = false }` quietly in state for `log_detection` rules.

**After:** State only contains an `options` block when the user declared one. Same uniform pattern for `case` and `query`.

## Test plan

- [x] `go build ./...`, `make fmtcheck`, `make docs && make check-docs` clean.
- [x] `TestAccDatadogSecurityMonitoringDefaultRule_Basic` and `_DeprecationWarning` passing under `RECORD=false`.
- [x] `TestAccDatadogSecurityMonitoringDefaultRule_Migration_*` migration tests (v4.7.0 ExternalProviders → FW with `ExpectEmptyPlan`).
- [x] Manually tested in staging — minimal-config import, UI-export import, case notifications update, custom_message `""` sentinel, decrease_criticality toggle, custom_query_extension toggle.

### Migration test — RECORD-only, manual gate

Uses `ExternalProviders` to spin up provider v4.7.0 as a subprocess with its own HTTP client, bypassing the in-process VCR recorder. `t.Skip`s unless `RECORD=true` — manual pre-release gate. Step 1 (v4.7.0 import + apply) → Step 2 (FW apply, `ExpectEmptyPlan`) is the empirical no-drift proof.


joepeeples (approved):

datadog/internal/validators/security_monitoring_data_source_warning_validator.go
```diff
@@ -0,0 +1,41 @@
+package validators
```
Supam: Could you add this file to the CODEOWNERs as being owned by your team ?

Supam (approved):

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"custom_tags": schema.SetAttribute{
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Custom tags for generated signals.",
+			},
+		},
+		Blocks: map[string]schema.Block{
```
hachem-dd, line 137: We discussed internally. The drift is unavoidable - either drift happens here on the 1-time plan for existing resources, fixed once and never again bothers the user, or the drift is shifted elsewhere.

The reason being that Blocks cannot be `Computed`.
I suggest a warning message for users upgrading if it's possible


```go
// UpgradeState migrates state written by SDKv2 (schema version 0) to version 1.
func (r *securityMonitoringDefaultRuleResource) UpgradeState(_ context.Context) map[int64]resource.StateUpgrader {
    v0 := securityMonitoringDefaultRuleSchema(0)
    return map[int64]resource.StateUpgrader{
        0: {
            PriorSchema: &v0,
            StateUpgrader: func(ctx context.Context, req resource.UpgradeStateRequest, resp *resource.UpgradeStateResponse) {
                var state securityMonitoringDefaultRuleResourceModel
                resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
                if resp.Diagnostics.HasError() {
                    return
                }

                hadCases   := len(state.Cases) > 0
                hadOptions := len(state.Options) > 0
                hadQueries := len(state.Queries) > 0

                state.Cases   = nil
                state.Options = nil
                state.Queries = nil
                resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)

                // Emit a one-time warning only when there were blocks that
                // needed to be stripped. This tells users what to expect.
                if hadCases || hadOptions || hadQueries {
                    resp.Diagnostics.AddWarning(
                        "One-time plan diff expected after provider upgrade",
                        "This resource had `case`, `query`, or `options` blocks in the "+
                            "previous provider version. The new provider framework cannot "+
                            "carry these blocks forward automatically during state migration. "+
                            "If your configuration declares any of these blocks, the next "+
                            "`terraform plan` will show them as additions. The diff is "+
                            "cosmetic: the API already holds those values. Running "+
                            "`terraform apply` once will re-align state and subsequent "+
                            "plans will be clean.",
                    )
                }
            },
        },
    }
}
```

down below.

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+	}
+}
+
+// UpgradeState migrates state written by SDKv2 (schema version 0) to version 1.
+// SDKv2 declared case/query/options as Optional+Computed and auto-populated all
+// API-returned blocks into state even when the user didn't declare them. FW blocks
+// cannot be Computed, so those extra rows must be stripped here. After the upgrade,
+// Read will repopulate only the blocks whose statuses/indices are present in the
+// user's actual config (via the prior-state filter in Read).
+func (r *securityMonitoringDefaultRuleResource) UpgradeState(_ context.Context) map[int64]resource.StateUpgrader {
```
hachem-dd, line 309: This is where I mentioned that maybe we can include a warning message for users who will get a diff. It will help them understand it.

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+		},
+		Blocks: map[string]schema.Block{
+			"case": schema.ListNestedBlock{
+				Description: "Cases of the rule, this is used to update notifications.",
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(10),
+				},
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"status": schema.StringAttribute{
```
hachem-dd, line 145: Given this is aiming a minor terraform provider update, we can probably skip it here.

But for a major upgrade, it would make much more sense to remove all the fields that are not modifiable by the user from this list. If I'm not mistaken, only these are modifiable for default rules: custom_query_extension, decrease_criticality_based_on_env, notifications and custom_status

If we go this route, can we add a deprecation message to them?

```
    DeprecationMessage: "This field is read-only for default rules and may be removed from the provider in a future version.",
```

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Description: "Notification targets for each rule case.",
+						},
+					},
+				},
+			},
+			"query": schema.ListNestedBlock{
+				Description: "Queries for selecting logs which are part of the rule.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"aggregation": schema.StringAttribute{
```
hachem-dd, line 171: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							Description: "The aggregation type.",
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"distinct_fields": schema.ListAttribute{
```
hachem-dd, line 178: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							ElementType: types.StringType,
+							Description: "Field for which the cardinality is measured. Sent as an array.",
+							PlanModifiers: []planmodifier.List{
+								listplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"group_by_fields": schema.ListAttribute{
```
hachem-dd, line 186: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							ElementType: types.StringType,
+							Description: "Fields to group by.",
+							PlanModifiers: []planmodifier.List{
+								listplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"has_optional_group_by_fields": schema.BoolAttribute{
```
hachem-dd, line 194: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							Description: "When false, events without a group-by value are ignored by the rule. When true, events with missing group-by fields are processed with `N/A`, replacing the missing values.",
+							PlanModifiers: []planmodifier.Bool{
+								boolplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"data_source": schema.StringAttribute{
```
hachem-dd, line 201: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							Description: "Source of events.",
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"metric": schema.StringAttribute{
```
hachem-dd, line 208: s a above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:           true,
+							Description:        "The target field to aggregate over when using the `sum`, `max`, or `geo_data` aggregations.",
+							DeprecationMessage: "Configure `metrics` instead. This attribute will be removed in the next major version of the provider.",
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"metrics": schema.ListAttribute{
```
hachem-dd, line 216: s a above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							ElementType: types.StringType,
+							Description: "Group of target fields to aggregate over when using the `sum`, `max`, `geo_data`, or `new_value` aggregations. The `sum`, `max`, and `geo_data` aggregations only accept one value in this list, whereas the `new_value` aggregation accepts up to five values.",
+							PlanModifiers: []planmodifier.List{
+								listplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"name": schema.StringAttribute{
```
hachem-dd, line 224: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+							Computed:    true,
+							Description: "Name of the query. Not compatible with `new_value` aggregations.",
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"query": schema.StringAttribute{
```
hachem-dd, line 231: same as above

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
```diff
@@ -0,0 +1,1029 @@
...
+										Required:    true,
+										Description: "**Deprecated**. It won't be applied anymore.",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			"filter": schema.ListNestedBlock{
```
hachem-dd, line 263: Same here - if we can't modify it via the provider, maybe plan on removing it for the next major release

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringDefaultRule_AddTag.freeze
hachem-dd, line 1: The cassettes in their current state are pretty much unusable. They need to be revamped to allow actual humans to see how changes happen.

Here they serve no purpose if they're in the megabytes of size and no git diff tool is happy to show them.

hachem-dd: Otherwise, for the brave amongst the public: `git diff HEAD..master -- datadog/tests/cassettes/TestAccDatadogSecurityMonitoringDefaultRule_AddTag.yaml`

docs/resources/security_monitoring_default_rule.md
```diff
@@ -78,14 +78,17 @@ Optional:
...
+
+Read-Only:
+
+- `aggregation` (String) The aggregation type.
+- `data_source` (String) Source of events.
 - `distinct_fields` (List of String) Field for which the cardinality is measured. Sent as an array.
 - `group_by_fields` (List of String) Fields to group by.
 - `has_optional_group_by_fields` (Boolean) When false, events without a group-by value are ignored by the rule. When true, events with missing group-by fields are processed with `N/A`, replacing the missing values.
-- `metric` (String, Deprecated) The target field to aggregate over when using the `sum`, `max`, or `geo_data` aggregations. **Deprecated.** Configure `metrics` instead. This attribute will be removed in the next major version of the provider.
+- `metric` (String, Deprecated) The target field to aggregate over when using the `sum`, `max`, or `geo_data` aggregations.
```
hachem-dd, line 91: This is **still** deprecated!

docs/resources/security_monitoring_default_rule.md
```diff
@@ -29,13 +29,13 @@ resource "datadog_security_monitoring_default_rule" "adefaultrule" {
...
 
-- `case` (Block List, Max: 10) Cases of the rule, this is used to update notifications. (see [below for nested schema](#nestedblock--case))
+- `case` (Block List) Cases of the rule, this is used to update notifications. (see [below for nested schema](#nestedblock--case))
 - `custom_message` (String) Custom Message (will override default message) for generated signals.
 - `custom_name` (String) The name (will override default name) of the rule.
 - `custom_tags` (Set of String) Custom tags for generated signals.
 - `enabled` (Boolean) Enable the rule. Defaults to `true`.
 - `filter` (Block List) Additional queries to filter matched events before they are processed. (see [below for nested schema](#nestedblock--filter))
-- `options` (Block List, Max: 1) Options on default rules. Note that only a subset of fields can be updated on default rule options. (see [below for nested schema](#nestedblock--options))
+- `options` (Block List) Options on default rules. Note that only a subset of fields can be updated on default rule options. (see [below for nested schema](#nestedblock--options))
```
hachem-dd, line 38: There is max 1 block still

docs/resources/security_monitoring_default_rule.md
```diff
@@ -29,13 +29,13 @@ resource "datadog_security_monitoring_default_rule" "adefaultrule" {
 
 ### Optional
 
-- `case` (Block List, Max: 10) Cases of the rule, this is used to update notifications. (see [below for nested schema](#nestedblock--case))
+- `case` (Block List) Cases of the rule, this is used to update notifications. (see [below for nested schema](#nestedblock--case))
```
hachem-dd, line 32: IIRC we cannot add more cases to default rules, so it doesn't matter if we mention that them max is 10 or not

docs/resources/security_monitoring_default_rule.md
alvizi-dd: Yes, it's a problem with `tfplugindocs` auto generating these docs. It's still marked as deprecated, but it loses the DeprecationMessage.

docs/resources/security_monitoring_default_rule.md
alvizi-dd: `options` blocks cannot be added either, like `case` blocks as you said below. This was also auto generated.

datadog/tests/cassettes/TestAccDatadogSecurityMonitoringDefaultRule_AddTag.freeze
alvizi-dd: They are still useful for test replay

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
alvizi-dd: I think it’s a valid point, yes. But I’d rather have this discussion internally with the other teams first.

datadog/internal/validators/security_monitoring_data_source_warning_validator.go
alvizi-dd: Decided to put this validator alongside other validators, just like its SDKv2 version, if it's ok. https://github.com/DataDog/terraform-provider-datadog/pull/3780/changes/e060ed48059b26bab112df67d21609e3ef4ffd77

Supam: That is perfectly fine as well 👍

datadog/fwprovider/resource_datadog_security_monitoring_default_rule.go
alvizi-dd: Good idea! I added the warning message in [[SEC-30101] Add UpgradeState one-time warning when blocks are stripped](https://github.com/DataDog/terraform-provider-datadog/pull/3780/commits/f9940d989ca57ee4fc0ae736cc2356689ecb0def)

hachem-dd (approved):

PR #3783, opened 2026-05-13
feat[datadog_metric_tag_configuration]: add data source

mdb:
Addresses https://github.com/DataDog/terraform-provider-datadog/issues/3790 to implement a `datadog_metric_tag_configuration ` by using the `ListTagConfigurationByName` endpoint
(`GET /api/v2/metrics/{metric_name}/tags`) to surface "does a tag configuration exist for this metric?"

None of the existing metric data sources offer this insight:

- datadog_metric_tags returns ingested tag values, unrelated to the managed tag configuration.
- datadog_metric_metadata returns metric metadata, also unrelated.
- datadog_metric_active_tags_and_aggregations returns active ingested data, also unrelated.

The data source returns successfully on 404 with exists = false rather than emitting a diagnostic error. This is a deliberate deviation from the convention used by e.g. data_source_datadog_incident_type, becaus one purpose served by this data source is existence checking — a data source that errors when its target doesn't exist cannot be used inside a `for_each` filter to discriminate.

## Example motivating use case (other examples exist; this is just one):

Callers that dynamically discover metric names via `datadog_metrics` and want to `import {}` the subset that already have a tag configuration. Today, that import fails when any discovered metric lacks a tag configuration; this data source lets callers filter on `.exists` before populating the `for_each` set.


cswatt (approved):

mdb:
@cswatt Thanks for the review! What are next steps for stewarding this along towards being merged/released? Anything I can help with?

PR #3802, opened 2026-05-19
[datadog_security_monitoring_rule] Prevent API-echoed metric field from causing apply inconsistency

alvizi-dd:
## Summary

- Extends `reconcileEmptyQueryFields` to handle the deprecated `metric` field
- The Datadog API derives and echoes back `metric` from the `metrics` list even when the user never configured `metric`. Since `metric` is `Optional` (not `Computed`), the framework plans it as `null` when not configured. The API echo caused a "provider produced inconsistent result" error after apply.
- Fix: when the prior plan/state had `metric = null`, restore it to null after reading the API response, regardless of the echoed value.

## Test plan

- [x] Verify rules using only `metrics` (not `metric`) apply without inconsistency error
- [x] Verify rules using the deprecated `metric` field still work correctly
- [x] Verify import of existing rules works (defaults to null for unconfigured `metric`)


clementgbcn:
Added an acceptance test for the `metrics`-only path (`TestAccDatadogSecurityMonitoringRule_ImpossibleTravelRuleMetricsField`): it drives the full geo_data create + update + update lifecycle authored with `metrics = ["@usr.handle"]` and no `metric`. The recorded API responses echo `metric` back (as prod does), so reverting the `reconcileEmptyQueryFields` change makes it fail with exactly `Provider produced inconsistent result after apply … .query[0].metric: was null, but now …`; with the fix it passes and `metric` reconciles to null. Full `TestAccDatadogSecurityMonitoringRule*` suite green in replay (20/20).

Closing my duplicate #4018 in favor of this — your read-side reconcile is the cleaner fix (keeps `metric` Optional and out of state).

clementgbcn (approved):

PR #3805, opened 2026-05-20
[datadog_observability_pipeline] add tls.verify_certificate field to multiple server based sources

tessneau:
for several server-based sources this replaces the Tls struct with a new Mtls one that simply introduces the field `tls.verify_certificate`

public api spec PR: https://github.com/DataDog/datadog-api-spec/pull/5712


amarshall-dd (approved):
Only reviewed go.mod+go.sum

taylorchandleryoung (approved):

PR #3830, opened 2026-05-28
[datadog_on_call_team_routing_rules] Add customizable_content to send_slack_message

### LINKED ISSUE #3740: datadog_on_call_team_routing_rules: send_slack_message missing customizable_content (causes data loss on apply)
## Resources affected

`datadog_on_call_team_routing_rules`

## Problem

The Slack action `customizable_content` field (with `include_description` and `include_source` booleans) on `send_slack_message` is not exposed by the provider, and an apply silently wipes any value set via the UI. Without it, the Slack post is title-only and lacks alert context. They're toggle-checkboxes in the UI's Slack action editor.

The field can be set in the DD UI and persists there until `terraform apply` touches the rule — at which point the PUT omits it and DD resets it to default. This means anyone using the provider for routing rules has to re-flip these UI toggles after every apply that mutates the rule, which is operationally untenable for IaC.

## Underlying API

`customizable_content` is already in the GET response from `/api/v2/on-call/teams/{team_id}/routing-rules?include=rules` when set via UI:

```json
{
  "actions": [{
    "type": "send_slack_message",
    "channel": "C012ABCDE",
    "workspace": "T012ABCDE",
    "customizable_content": {
      "include_description": true,
      "include_source": true
    }
  }]
}
```

So adding it to the provider's `send_slack_message` schema is straightforward — the field already round-trips through the API. Note: the field is not yet in the generated `datadog-api-client-go` `SendSlackMessageAction` struct, so an interim implementation can round-trip via `AdditionalProperties` until the API spec / client is regenerated.

## Current workaround

Either:
1. Set the toggle manually in the DD UI after every apply that touches routing rules.
2. Use `lifecycle { ignore_changes = [rule] }` on the resource so TF stops reconciling rule changes after initial create — UI changes then persist, but you lose the ability to update rules via TF without temporarily removing the lifecycle directive.

Neither is great.

## Request

Add `customizable_content { include_description, include_source }` to the `send_slack_message` block in the rule.action schema. The API already supports it.

PR opened: tylergohl/terraform-provider-datadog → DataDog/terraform-provider-datadog (link to follow once filed).

## Datadog Terraform Provider Version

Latest (v4.6.x at time of filing).

---

## ~~Hold outside support hours~~ (covered by #3788)

~~The "Outside support hours, hold escalation policy notifications until the next window" toggle on a rule, paired with `time_restrictions`. Required for daytime-only paging to actually hold pages overnight rather than falling through to the next rule.~~

This is being addressed by Datadog's own [PR #3788](https://github.com/DataDog/terraform-provider-datadog/pull/3788), which adds a new `action { escalation_policy { policy_id, ack_timeout_minutes, urgency, support_hours { ... } } }` block. The `support_hours` sub-block is mutually exclusive with rule-level `time_restrictions` (per that PR's validators) but provides the same operational capability — pages held outside the configured window are queued until the window opens rather than falling through to a 24/7 catch-all rule.

tylergohl:
## Summary

Adds `customizable_content { include_description, include_source }` to the `send_slack_message` action on `datadog_on_call_team_routing_rules`.

Closes #3740.

## Background

The On-Call routing-rules API supports a `customizable_content` object on `send_slack_message` actions, controlling whether the alert description and source are rendered in the Slack message. The DD UI exposes this as the "Include description" / "Include source" checkboxes when editing a Slack action. The field is round-tripped through `GET/PUT /api/v2/on-call/teams/{team_id}/routing-rules`, but the provider has not yet exposed it — so applying the resource silently wipes whatever the user set in the UI.

## Changes

- `datadog/fwprovider/resource_datadog_on_call_team_routing_rules.go` — new `customizable_content` sub-block under `send_slack_message`, with `include_description` and `include_source` bool attributes; wired through `stateFromResponse` and `teamRoutingRulesRequestFromModel`.
- `examples/resources/datadog_on_call_team_routing_rules/resource.tf` — example updated to demonstrate the new field.
- `docs/resources/on_call_team_routing_rules.md` — manually updated (this doc is in the `generate-docs.sh` exclude list).

## Note on the API client

`customizable_content` is not yet a typed field on `datadogV2.SendSlackMessageAction` in `datadog-api-client-go`. As an interim measure, this change round-trips it via the SDK's `AdditionalProperties` map (the SDK's `MarshalJSON` folds `AdditionalProperties` into the wire payload, and unknown JSON keys land back in `AdditionalProperties` on read — verified end-to-end). The two relevant call sites have comments calling out that when the API spec is updated and the client is regenerated, the implementation should swap to the typed accessors.

If you'd prefer to land the spec/client regeneration first, I'm happy to revise.

## Testing

- `make fmtcheck` ✓
- `make docs && make check-docs` ✓ (the manually maintained `on_call_team_routing_rules.md` is excluded from regeneration; I updated it by hand)
- `go vet ./...` ✓
- `go build ./...` ✓
- Built provider binary + `terraform validate` on a scratch config that exercises both the new `customizable_content` block and a Slack action without it (backward compatibility) — both pass.

I did not add a new acceptance test step or record a cassette — that needs `DD_TEST_CLIENT_API_KEY` sandbox credentials, which I don't have. Happy to extend the existing `TestAccOnCallTeamRoutingRulesCreateAndUpdate` with a `customizable_content` config and re-record the cassette if a maintainer can hand off creds, or leave the test addition to you.

## Test plan

- [ ] Maintainer extends `TestAccOnCallTeamRoutingRulesCreateAndUpdate` with a step that sets `customizable_content` and re-records the cassette under sandbox.
- [ ] Maintainer verifies the round-trip in `RECORD=none` mode against a live test team to confirm the UI reflects `include_description` / `include_source` after apply.

### DIFF
--- a/datadog/fwprovider/resource_datadog_on_call_team_routing_rules.go
+++ b/datadog/fwprovider/resource_datadog_on_call_team_routing_rules.go
@@ -52,8 +52,14 @@ type teamRuleActionModel struct {
 }
 
 type slackMessageModel struct {
-	Workspace types.String `tfsdk:"workspace"`
-	Channel   types.String `tfsdk:"channel"`
+	Workspace           types.String                   `tfsdk:"workspace"`
+	Channel             types.String                   `tfsdk:"channel"`
+	CustomizableContent *slackCustomizableContentModel `tfsdk:"customizable_content"`
+}
+
+type slackCustomizableContentModel struct {
+	IncludeDescription types.Bool `tfsdk:"include_description"`
+	IncludeSource      types.Bool `tfsdk:"include_source"`
 }
 
 type teamsMessageModel struct {
@@ -219,6 +225,21 @@ func (r *onCallTeamRoutingRulesResource) Schema(_ context.Context, _ resource.Sc
 											"workspace": schema.StringAttribute{
 												Optional:    true,
 												Description: "Slack workspace ID.",
+											},
+										},
+										Blocks: map[string]schema.Block{
+											"customizable_content": schema.SingleNestedBlock{
+												Description: "Controls what alert context is included in the Slack message. If omitted, the API default (title only) is used.",
+												Attributes: map[string]schema.Attribute{
+													"include_description": schema.BoolAttribute{
+														Optional:    true,
+														Description: "Whether to include the alert description in the Slack message.",
+													},
+													"include_source": schema.BoolAttribute{
+														Optional:    true,
+														Description: "Whether to include the alert source in the Slack message.",
+													},
+												},
 											},
 										},
 									},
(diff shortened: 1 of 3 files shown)


tylergohl:
Closing temporarily to iterate on review/simplify in our fork. Will reopen once cleaned up.

PR #3834, opened 2026-05-29
[datadog_dashboard] Add default_timeframe support

Gurik-M:
## Summary

Resolves [DASHB-970](https://datadoghq.atlassian.net/browse/DASHB-970) (Update the terraform resource to respect `default_time`)
- Adds an optional `default_timeframe` block to `datadog_dashboard` and `datadog_dashboard_v2` for configuring the dashboard's default time span on load
- Supports `live` (`type`, `unit`, `value`) and `fixed` (`type`, `from`, `to`) variants, matching the dashboard API schema
- Allows `default_timeframe = null` on update to explicitly disable after it has been configured

## Test plan

- [x] Unit tests for build/flatten/null handling (`datadog/dashboardmapping/dashboard_default_timeframe_test.go`)
- [x] Acceptance tests (cassette replay, `RECORD=false`):

  ```bash
  TF_ACC_TERRAFORM_VERSION=1.13.1 \
  RECORD=false \
  TESTARGS='-run TestAccDatadogDashboardDefaultTimeframe' \
  make testacc
  ```

  (`TF_ACC_TERRAFORM_VERSION` pins/downloads a Terraform CLI for the harness; optional if `terraform version` already works on your `PATH`.)

  **Result (local):**

  ```
  PASS datadog/tests.TestAccDatadogDashboardDefaultTimeframe (5.65s)
  PASS datadog/tests.TestAccDatadogDashboardDefaultTimeframe_import (6.23s)
  PASS datadog/tests (cached)
  ```

- [ ] `make docs && make check-docs` after merge prep

Made with [Cursor](https://cursor.com)

[DASHB-970]: https://datadoghq.atlassian.net/browse/DASHB-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hestonhoffman (approved):

yacomink (changes requested):
Can you please load this claude skill: https://github.com/DataDog/claude-marketplace/tree/main/dashboard-terraform-openapi-sync and go over this change with it?

The claude skill has some conventions and pointers for making changes to the datadog_dashboard_v2 spec. One of those is preferring expanding the engine over adding custom build/flatten code for individual fields. With that guidance in place the skill gave me this suggestion:

> The custom build code is all a consequence of one missing engine capability: **no way to emit JSON `null` for a field that was cleared in HCL**.

> The engine already has two states for an `OmitEmpty: true` TypeBlock — include it or skip it. A third state (`nil` in the data map → serialize as JSON `null`) would cover this case generically. The abstraction would be a `NullOnClear: bool` flag on `FieldSpec`:

```go
// In BuildEngineJSONFromMap, TypeBlock case:
rawVal := data[f.HCLKey]
if rawVal == nil && f.NullOnClear {
    result[f.effectiveJSONKey()] = nil  // explicit JSON null
    continue
}
// ... existing block handling unchanged ...
```

I'd also suggest making your OpenAPI change PR at the same time. The skill works best if you make the two changes in one session.

datadog/dashboardmapping/engine.go
```diff
@@ -1519,10 +1526,26 @@ func getIntFromMap(data map[string]interface{}, key string) int {
 		switch iv := v.(type) {
 		case int:
 			return iv
+		case int32:
```
yacomink, line 1544: I don't think the three added case statements (int32, json.Number, string) here are needed for your PR. Can you remove them unless they are?

datadog/dashboardmapping/engine.go
```diff
@@ -2586,7 +2614,8 @@ func buildTabsJSONFromMap(data map[string]interface{}, builtWidgets []interface{
 // Parallel to MarshalDashboardJSON in engine.go.
 // The trailing newline matches behavior of json.NewEncoder used when cassettes were recorded.
 func MarshalDashboardJSONFromMap(data map[string]interface{}, id string) (string, error) {
-	body, err := json.Marshal(BuildDashboardEngineJSONFromMap(data, id))
+	bodyMap := BuildDashboardEngineJSONFromMap(data, id)
```
yacomink: There's no functional change here, can you revert to keep the diff clean?

datadog/resource_datadog_dashboard_v2.go
```diff
@@ -308,6 +314,17 @@ func setDashboardStateSDKv2(d *schema.ResourceData, resp map[string]interface{})
 		if err := d.Set("tab", flatTabs); err != nil {
 			diags = append(diags, diag.FromErr(err)...)
 		}
+	} else if err := d.Set("tab", []interface{}{}); err != nil {
```
yacomink: Was this change to the "tab" field handling intentional?

datadog/dashboardmapping/dashboard_default_timeframe.go
```diff
@@ -0,0 +1,100 @@
...
+		Type:        TypeBlock,
+		OmitEmpty:   true,
+		Description: "The default timeframe applied when opening the dashboard. Set to `null` to disable after it has been configured.",
+		Children:    dashboardDefaultTimeframeFields,
+	})
+}
+
+// BuildDefaultTimeframeJSONFromMap converts a Terraform default_timeframe block to API JSON.
+// Used by the v1 datadog_dashboard resource.
+func BuildDefaultTimeframeJSONFromMap(block map[string]interface{}) (map[string]interface{}, error) {
```
yacomink: When you land the oneOf change for the open-api spec you should be able to do something similar in the field mapping and then you won't need these two functions.

yacomink (changes requested):

datadog/dashboardmapping/engine.go
Gurik-M: Removed these extra cases. They weren’t needed for this PR.

datadog/resource_datadog_dashboard_v2.go
Gurik-M: No, that was unrelated to this PR. I reverted it.

datadog/dashboardmapping/engine.go
Gurik-M: 👍 Reverted this to keep the diff clean

datadog/dashboardmapping/dashboard_default_timeframe.go
Gurik-M: I kept the current flat Terraform shape for this PR to avoid a user-facing schema change, but pulled the shared FieldSpec into DashboardDefaultTimeframeField() so this can move cleanly to the oneOf field mapping once the OpenAPI change lands

yacomink: This PR is introducing the default_timeframe field for the first time, so there is no existing user-facing schema for the field to preserve.

Gurik-M: Updated the schema so that `default_timeframe` in the v2 resource now uses `TypeOneOf`:
default_timeframe {
live {
unit = "week"
value = 1
}
} instead of the flat type = "live" / unit / value shape that the v1 resource uses.

Also, all the ubuntu version errors are fixed now! A client version bump introduced `UnmarshalUseNumber`, causing `additionalProperties` to deserialize as `json.Number` instead of `float64` which `getIntFromMap` didn't handle and returned 0. **Fix**: Hardened `getIntFromMap` to handle `json.Number`

yacomink (approved):

michlee1337 (approved):

amarshall-dd (approved):
Only high-level review of api-reliability owned files.

PR #3845, opened 2026-06-01
[datadog_custom_allocation_rule] Fix "inconsistent result after apply" on in-place update

### LINKED ISSUE #3829: datadog_custom_allocation_rule: "Provider produced inconsistent result after apply" on `updated` timestamp during update
### Terraform / provider versions
- Terraform: 1.x
- `DataDog/datadog` provider: 4.10.0

### What happened

Updating an existing `datadog_custom_allocation_rule` (changing the `costs_to_allocate` filter) fails at apply with:

```

When applying changes to module.this.datadog_custom_allocation_rule.<name>,
provider produced an unexpected new value: .updated: was
cty.StringVal("2026-05-28T19:11:04.654224Z"), but now
cty.StringVal("2026-05-28T19:45:20.015381Z").
```

The update *does* take effect server-side — the rule is modified correctly — but Terraform errors on the post-apply consistency check, leaving the apply marked as failed.

### Root cause

In `datadog/fwprovider/resource_datadog_custom_allocation_rule.go`, the computed `updated` attribute uses the `UseStateForUnknown()` plan modifier:

```go
"updated": schema.StringAttribute{
    Computed:    true,
    Description: "The timestamp (in ISO 8601 format) when the rule was last updated.",
    PlanModifiers: []planmodifier.String{
        stringplanmodifier.UseStateForUnknown(),
    },
},
```

`UseStateForUnknown()` pins the planned value to the prior state during an update. But `updated` is a last-modified timestamp the API changes on every update, so the post-apply value differs from the pinned plan value, violating the plan/apply consistency contract.

The `version` attribute (which increments on each update) has the same latent issue. `created` is fine — it doesn't change after creation.

### Expected behavior

Updating the rule should succeed without an inconsistency error.

### Steps to reproduce

1. Create a `datadog_custom_allocation_rule` and apply.
2. Change its `costs_to_allocate` filter.
3. `terraform apply` → inconsistent-result error on `.updated`.

billykern:
Fixes https://github.com/DataDog/terraform-provider-datadog/issues/3829

`updated`, `version`, and `last_modified_user_uuid` are computed attributes the API rewrites on every update, but they used `UseStateForUnknown()`, which pins the planned value to prior state. On an in-place update the post-apply value differs from the pinned plan value, so Terraform fails the consistency check even though the update succeeded server-side. `created` keeps the modifier since it never changes after creation.

The existing `...Update` acceptance test missed this because it also changes `rule_name` (a `RequiresReplace` field), which destroys and recreates the rule rather than updating it in place. Added `...InPlaceUpdate`, which changes only `costs_to_allocate`, exercising the PATCH path that triggers the bug.


fpighi:
We have an issue with integration tests coming from forks right now, so it's ok to ignore that failure.

ashleyyip (approved):

PR #3850, opened 2026-06-02
[TFgen] Added a identifier translator to match the go SDK

Supam:
The translator will transform the OAS snake_case names into the
equivalent PascalCase that are used in the go SDK.
Added tests for this translation function.

Removed internal ticket names.

PR #3849, opened 2026-06-02
[TFgen] Added types and enums for parsing the OAS

Supam:
Added enums :
- ArtifactKind
- SchemaKind
- IdStrategy

Added structs:
- Spec
- Operation
- Schema
- Artifact
- TrackingFieldMetaData
- various other supporting structs

PR #3842, opened 2026-06-01
[TFgen] Phase 1 Setup CLI

j10czar:
[APIR-2491](https://datadoghq.atlassian.net/browse/APIR-2491)
 ## What this does

  Phase 1 scaffolding for the new generator (tfgen): build, CI, and test-fixture
  plumbing so later phases have a consistent place to land. No generator logic
  yet. Targets the `generator-v2` branch, which already has the stubs and deps.

  ## Changes

  - CI workflow (`.github/workflows/tfgen.yml`): runs `go build` and `go vet`
    against the generator on any PR that touches `.generator-v2/internal/**`,
    `.generator-v2/cmd/tfgen/**`, or the workflow file. Build and vet only for
    now. Unit tests, e2e, coverage, and the `generate --check` / `verify --strict`
    gates come in later phases.

  - Make targets: `make tfgen-build` builds the binary to `bin/tfgen`,
    `make tfgen-test` runs `go test ./internal/... ./cmd/tfgen/...`. Both run
    inside the `.generator-v2` module and honor a `GO=` override.

  - Fixture catalogue (`.generator-v2/internal/testdata/fixtures/README.md`):
    documents the per-fixture layout (input `openapi.yaml`, committed golden
    `out/`, optional README) so e2e fixtures get added consistently. No fixtures
    included here.

  ## Testing

  - `make tfgen-build` produces `bin/tfgen`.
  - `make tfgen-test` exits 0 (no test files yet).
  - Workflow runs build + vet clean.

[APIR-2491]: https://datadoghq.atlassian.net/browse/APIR-2491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3840, opened 2026-06-01
[TFgen] Adding a new terraform generator

Supam:
## Summary

Introduces `.generator-v2/`, a new standalone Go module that will replace the hand-written Terraform resource/data source authoring workflow. The generator reads the Datadog OpenAPI Specification, respects a `x-datadog-tf-generator` tracking field, and emits Go source code using the Datadog go-sdk.

This PR adds the module skeleton only:
- `cmd/tfgen/main.go` entry point with Cobra root command
- Global flags (`--spec`, `--output-root`, `--hooks-root`, `--tracking-field`, `--max-depth`, `--report`, `--quiet`)
- `tfgen generate` and `tfgen verify` subcommands (no-op stubs)
- Internal package layout (`parser`, `model`, `emit`, `hooks`, `report`)
- CI step: `cd .generator-v2 && go test ./...`

## Test Plan

```bash
cd .generator-v2
go build ./cmd/tfgen
./tfgen --help
./tfgen generate --help
./tfgen verify --help
go test ./...

PR #3853, opened 2026-06-02
[TFgen] Implement tracking-field decoder + duplicate artifact name detection

j10czar:
## Changes

Decodes the `x-datadog-tf-generator` vendor extension into typed metadata on
`model.Operation.Tracking` during `LoadSpec`, and enforces global uniqueness of the resulting
`artifact_name`. Both run inside a single `LoadSpec` call, which now returns a fully-populated
`*model.Spec` or an actionable error.

- **`parser.DecodeTracking(op, path, method, extensionName) (*model.TrackingFieldMetadata, error)`** — parses the extension into a typed `*model.TrackingFieldMetadata` (`artifact_kind`,
`artifact_name`, `group`, `id_strategy`, `sensitive`, `skip`) for downstream codegen,
applying the `id_strategy` default (`data.id`). Before populating the struct it validates the
raw value at runtime against the committed JSON Schema
(`internal/contracts/tracking-field.schema.json`, embedded via a new `contracts` package), so
callers only ever get well-formed metadata back. Returns `(nil, nil)` when the operation is
out of scope (no extension, or `skip: true`), and a typed `*TrackingError` naming the OpenAPI
location (method + path + operationId) when the extension is malformed (unknown
`artifact_kind`, missing required field, unknown property, bad name pattern, …).
- **`parser.CheckDuplicateArtifactNames(spec)`** — walks every `Operation.Tracking` and fails
with a single aggregated `*DuplicateArtifactNameError` listing each `artifact_name` collision
and **all** of its source locations, sorted by name then (path, method) for deterministic
output.
- **Pipeline wiring** — decode happens per-operation in the enumeration loop (storing the
metadata on `Operation.Tracking`, fail-fast on a malformed extension); the duplicate check
runs after the deterministic sort.
- **`--tracking-field`** **seam** — adds a `WithTrackingFieldName` `LoadSpec` option and threads
`flags.trackingField` through the `generate` command, mirroring the existing `--max-depth`
wiring.
- **New dependency** — `github.com/santhosh-tekuri/jsonschema/v6` (a JSON-Schema validator,
_not_ a second OAS parser).

> Note: the model carries `json:` tags, so once the value is validated `DecodeTracking`
> bridges YAML→JSON (`json.Marshal`/`Unmarshal`) to fill the typed struct, rather than decoding
> the `yaml.Node` directly.

## Test Plan

New coverage — **decode (asserts the returned metadata):** valid resource / valid data source
(kind, name, group, id_strategy populated), absent extension, `skip:true`, missing required
fields, unknown `artifact_kind`, unknown property (`additionalProperties:false`), bad
`artifact_name` pattern, applied `id_strategy` default, error-carries-OpenAPI-path, custom
extension name. **Duplicates:** none / single collision / all-sources-listed /
multiple-collisions-sorted / ignores-untracked / deterministic-across-input-order. **LoadSpec**
**wiring fixtures:** tracking populated (resource + data source + unflagged op), duplicate →
typed error naming both ops, malformed → typed error carrying the real OpenAPI path.

```sh
❯ make tfgen-test
cd .generator-v2 && go test ./internal/... ./cmd/tfgen/...
ok    generator/internal/cli        0.505s
?     generator/internal/contracts  [no test files]
ok    generator/internal/model      0.495s
ok    generator/internal/parser     0.561s
      generator/cmd/tfgen           [no test files]
tfgen tests passed
```

## Follow-ups / action items

- [ ] **Remove the parser-side default extension name once the global CLI flags are wired into**
**the cobra root (APIR-2923).** `DefaultTrackingFieldName` and the empty-name fallback are a
temporary stopgap so `LoadSpec` is usable standalone — the real default
(`x-datadog-tf-generator`) already lives on the `--tracking-field` flag in `root.go`. Once the
wiring guarantees the name is always supplied, delete:
    - the `DefaultTrackingFieldName` const — `internal/parser/tracking.go`
    - the `if name != ""` fallback in `WithTrackingFieldName` **and** the `trackingFieldName: DefaultTrackingFieldName` seed in `LoadSpec` — `internal/parser/parser.go`

    …so the extension name has a single source of truth (the CLI flag), instead of a default
defined in two places.

-  [ ] **Update OAS extension contract**

> ⚠️ **Note:** artifact-name uniqueness is enforced **per `artifact_kind`**, not globally — a
  `resource` and a `data_source` may share a name (matches Terraform's separate namespaces).
  This intentionally refines the literal AC; the schema is updated to match, but the ticket AC +
  APIR-2492 discovery doc still say "globally unique" and need an owner's confirmation.

### DIFF
--- a/.generator-v2/internal/contracts/tracking-field.schema.json
+++ b/.generator-v2/internal/contracts/tracking-field.schema.json
@@ -17,7 +17,7 @@
         "pattern": "^[a-z][a-z0-9_]*$",
         "minLength": 1,
         "maxLength": 64,
-        "description": "Terraform-facing artifact name without the 'datadog_' prefix. Must be lowercase snake_case. Must be unique across the entire spec."
+        "description": "Terraform-facing artifact name without the 'datadog_' prefix. Must be lowercase snake_case. Must be unique per artifact_kind — resources and data sources are separate Terraform namespaces, so a resource and a data source may share a name."
       },
       "group": {
         "type": "object",
--- a/.generator-v2/internal/contracts/tracking-field.schema.json
+++ b/.generator-v2/internal/contracts/tracking-field.schema.json
@@ -1,0 +1,72 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://datadog.github.io/terraform-provider-datadog/contracts/tracking-field.schema.json",
+    "title": "x-datadog-tf-generator",
+    "description": "OpenAPI 3.0 vendor extension that opts an operation in to Datadog Terraform provider generation.",
+    "type": "object",
+    "additionalProperties": false,
+    "required": ["artifact_kind", "artifact_name"],
+    "properties": {
+      "artifact_kind": {
+        "type": "string",
+        "enum": ["resource", "data_source"],
+        "description": "Whether this operation should be exposed as a Terraform resource (with full CRUD lifecycle) or a Terraform data source (read-only)."
+      },
+      "artifact_name": {
+        "type": "string",
+        "pattern": "^[a-z][a-z0-9_]*$",
+        "minLength": 1,
+        "maxLength": 64,
+        "description": "Terraform-facing artifact name without the 'datadog_' prefix. Must be lowercase snake_case. Must be unique across the entire spec."
+      },
+      "group": {
+        "type": "object",
+        "description": "Declares which OpenAPI operations form the C/R/U/D quadruple. May reference operations by their operationId.",
+        "additionalProperties": false,
+        "properties": {
+          "create": { "type": "string", "description": "operationId of the Create endpoint." },
+          "read":   { "type": "string", "description": "operationId of the Read endpoint." },
+          "update": { "type": "string", "description": "operationId of the Update endpoint. May be omitted; the generator will then mark all attributes ForceNew per the missing-CRUD edge case in the spec." },
+          "delete": { "type": "string", "description": "operationId of the Delete endpoint." }
+        },
+        "required": ["read"]
+      },
+      "id_strategy": {
+        "type": "string",
+        "enum": ["data.id", "data.attributes.id", "data.attributes.uuid", "header.location"],
+        "default": "data.id",
+        "description": "How to derive the Terraform resource ID from the API response on Create/Read."
+      },
+      "sensitive": {
+        "type": "boolean",
+        "default": false,
+        "description": "When attached to a Schema Object, marks the attribute as Terraform-sensitive (suppressed in plan output)."
+      },
+      "skip": {
+        "type": "boolean",
+        "default": false,
+        "description": "Explicitly disable generation for this operation while keeping the annotation in place. Equivalent to removing the extension, but documented in-spec so reviewers see the choice."
+      }
+    },
+    "examples": [
+      {
+        "artifact_kind": "data_source",
+        "artifact_name": "team",
+        "group":{
+          "read": "GetTeam"
+        }
+      },
+      {
+        "artifact_kind": "resource",
+        "artifact_name": "incident_type",
+        "group": {
+          "create": "CreateIncidentType",
+          "read": "GetIncidentType",
+          "update": "UpdateIncidentType",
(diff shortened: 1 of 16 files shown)


j10czar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3853?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3853** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3840** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3840?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 16 other dependent PRs ([#3842](https://github.com/DataDog/terraform-provider-datadog/pull/3842) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3842?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3849](https://github.com/DataDog/terraform-provider-datadog/pull/3849) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3849?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3850](https://github.com/DataDog/terraform-provider-datadog/pull/3850) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3850?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 13 others)
* `master`


<h2></h2>

.generator-v2/internal/parser/tracking.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, wrap(fmt.Errorf("re-encoding extension: %w", err))
+	}
+	var meta model.TrackingFieldMetadata
+	if err := json.Unmarshal(jsonBytes, &meta); err != nil {
+		return nil, wrap(fmt.Errorf("decoding extension into metadata: %w", err))
+	}
+	if meta.Skip {
+		return nil, nil
+	}
```
Supam: Is is possible to check before validating that the user added the skip boolean ?

If there is a validation issue but the user has specified that this resources should be skipped, we do not expect for a validation error to pop up.

j10czar:
Need to add coverage to ignore collisions when both the resource and datasource have the same artifact name. Right now the `CheckDuplicateArtifactNames` function flags it. There are many instances where something that is a datasource and something that is a resource has the same artifact name ex. dashboard, monitors, slos etc.). Are we are adding some kind of uniqueness to artifact name in the future @Supam ?

.generator-v2/internal/contracts/tracking-field.schema.json
```diff
@@ -17,7 +17,7 @@
         "pattern": "^[a-z][a-z0-9_]*$",
         "minLength": 1,
         "maxLength": 64,
-        "description": "Terraform-facing artifact name without the 'datadog_' prefix. Must be lowercase snake_case. Must be unique across the entire spec."
+        "description": "Terraform-facing artifact name without the 'datadog_' prefix. Must be lowercase snake_case. Must be unique per artifact_kind — resources and data sources are separate Terraform namespaces, so a resource and a data source may share a name."
```
j10czar, line 20: @Supam is this a valid statement? Should the OAS extension contract be updated?

fpighi: This is correct.

Supam:
There *technically* is already some differentiation for both datasources and resources, it is just not obvious at the OAS level.

In the OAS we will be asking for the artefact name, this name will be used in both datasources and resources.
Once we run the generator we will create datasources as `datasource_<artefact-name>` and resources as `resource_<artefact-name>`. This is how we differentiate between both.

Supam (approved):

PR #3870, opened 2026-06-05
Run integration tests for PRs from forks

fpighi:
## What

The integration-test workflows now mint short-lived Datadog credentials at job time over GitHub OIDC (via `DataDog/dd-sts-action`) instead of reading long-lived API/APP keys from repository secrets. PR runs use `pull_request_target`, and PRs from forks run in a protected `forks-prs` environment.

## Why

PRs opened from forks don't have access to repository secrets, so the integration test jobs could not authenticate and had to be skipped and merged manually. This lets fork PRs run integration tests while removing long-lived credentials from the repository.

## Changes

- `pull_request` → `pull_request_target` on both integration-test workflows.
- Credentials are minted per run via `DataDog/dd-sts-action` (short-lived, test-org scoped); `DD_TEST_CLIENT_*` now come from the action outputs and the static secret references are removed.
- `id-token: write` is granted only on the credentialed jobs (the unprivileged jobs cannot mint tokens).
- Fork PRs run in the `forks-prs` environment; same-repo PRs and the scheduled run are unaffected.
- The PR test-selection job no longer executes untrusted PR code: it checks out the trusted base, fetches the PR head, and diffs by filename only.

## Required to land / follow-ups

- Depends on the server-side OIDC trust configuration being deployed first.
- Requires a `forks-prs` environment configured in repository settings.
- The static `DD_CLIENT_API_KEY` / `DD_CLIENT_APP_KEY` secrets can be removed once this is validated.

## Security note

`pull_request_target` runs the base branch's workflow definition but the credentialed job checks out and runs the PR's code. Fork PRs run in the protected `forks-prs` environment, and minted credentials are short-lived and scoped to a dedicated test org.

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -3,8 +3,11 @@ name: Run Integration Tests
 permissions:
   contents: read
 
+# pull_request_target so that PRs from forks can mint Datadog credentials via
+# OIDC. The credentialed job checks out the PR head and, for forks, runs in the
+# forks-prs environment.
 on: # yamllint disable-line rule:truthy
-  pull_request:
+  pull_request_target:
     types:
       - labeled
       - unlabeled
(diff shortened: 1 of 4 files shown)


Supam (approved):

fpighi:
@codex review

PR #3889, opened 2026-06-12
[datadog_monitor] Add ignore_tag_keys support

j10czar:
## Summary

Adds an optional `ignore_tag_keys` mechanism that lets users suppress Terraform drift on tag keys managed outside of Terraform — for example, tags applied by the Datadog UI, a tagging service, or a deployment pipeline — without affecting how other tags are managed. It is exposed at **two levels**:

- **Provider level** — `ignore_tag_keys` on the `datadog` provider block, applied across all resources that support it (currently `datadog_monitor` and `datadog_service_level_objective`), wired similarly to `default_tags`.
- **Resource level** — `ignore_tag_keys` on `datadog_monitor` (SDKv2 + Framework) and `datadog_service_level_objective` (SDKv2).

**Precedence:** the resource-level list is **unioned** with the provider-level list; the effective ignore set for a resource is the union of both. There is no per-resource opt-out, and a resource opts in simply by declaring the attribute.

### Core logic (`internal/utils/tags.go`)
- **`StripIgnoredTags(planTags, stateTags, ignoreKeys []string) []string`** — SDK-agnostic helper. Drops plan-side tags whose key is in `ignoreKeys` and re-injects the matching state value(s) instead, so drift is neither reported nor stripped on apply. Multi-valued keys are preserved. Keys are normalized via `tagKey`/`NormalizeTag` so SDKv2 callers (which apply the `NormalizeTag` state func) and Framework callers (which don't always) resolve to the same canonical form. Output is sorted for stable plan output; returns `planTags` unchanged when `ignoreKeys` or `stateTags` is empty.
- **`tagKey`** — unexported helper that isolates and normalizes the key portion of a `key:value` tag.
- Unit-tested in `tags_test.go`.

### Framework bridge (`internal/fwutils/tags_utils.go`)
- **`ApplyIgnoreTagKeys`** — wraps `StripIgnoredTags` for use in Framework `ModifyPlan`. Operates on `effective_tags` (the Computed field that mirrors the API) rather than the user-owned `tags`, satisfying the Framework rule that a computed attribute's planned value must equal its config or its own prior state. Short-circuits on a null/unknown state `Set` (the Create path) to avoid an `ElementsAs` panic.
- Unit-tested in `tags_utils_test.go`.

### Provider-level wiring
- **SDKv2 (`provider.go`)** — adds the `ignore_tag_keys` `TypeSet` provider-schema attribute, reads it into `ProviderConfiguration.IgnoreTagKeys` in `providerConfigure`, and adds a new **`ignoreTagKeysDiff`** customdiff that unions provider- and resource-level keys, calls `StripIgnoredTags`, and writes the filtered set back via `d.SetNew("tags")`. It runs **before** `tagDiff` so `default_tags` merges over the ignore-filtered set.
- **Framework (`fwprovider/framework_provider.go`)** — adds the matching `ignore_tag_keys` provider attribute (byte-identical description), `ProviderSchema.IgnoreTagKeys`, and `FrameworkProvider.IgnoreTagKeys`, populated in `defaultConfigureFunc`.

### SDKv2 resources (`resource_datadog_monitor.go`, `resource_datadog_service_level_objective.go`)
- Added `ignore_tag_keys` `TypeSet` attribute with a `NormalizeTag` `StateFunc`.
- Added `ignoreTagKeysDiff` to each resource's `customdiff.All(...)` chain, ahead of `tagDiff`.

### Framework resource (`fwprovider/resource_datadog_monitor.go`)
- Added the `ignore_tag_keys` schema attribute and `monitorResourceModel` field, plus `IgnoreTagKeys` on `monitorResource` (populated in `Configure`).
- `ModifyPlan` now always runs `CombineTags` (producing `effective_tags`), unions provider- and resource-level keys, calls `ApplyIgnoreTagKeys`, and sets the result on `resp.Plan` only. The local `plan` is deliberately left untouched so the `validate` request body — and the pre-recorded monitor cassettes — are unchanged.

### Documentation
- `docs/index.md` — provider-level `ignore_tag_keys` entry.
- `docs/resources/monitor.md`, `docs/resources/service_level_objective.md` — resource-level `ignore_tag_keys` entries.

### Test harness
- **`withIgnoreTagKeys`** (`provider_test.go`) and **`withIgnoreTagKeysFw`** (`framework_provider_test.go`) — wrap the SDKv2 and muxed Framework providers respectively to set provider-level `ignore_tag_keys`, mirroring `withDefaultTags`/`withDefaultTagsFw`.

### Acceptance tests
| Test | Resource | Stack | Level |
|------|----------|-------|-------|
| `TestAccDatadogMonitor_IgnoreTagKeys` | `datadog_monitor` | SDKv2 | resource |
| `TestAccDatadogMonitor_ProviderIgnoreTagKeys` | `datadog_monitor` | SDKv2 | provider |
| `TestAccDatadogMonitor_ProviderIgnoreTagKeysOverride` | `datadog_monitor` | SDKv2 | provider + resource union |
| `TestAccMonitor_Fwprovider_IgnoreTagKeys` | `datadog_monitor` | Framework | resource |
| `TestAccMonitor_Fwprovider_ProviderIgnoreTagKeys` | `datadog_monitor` | Framework | provider |
| `TestAccMonitor_Fwprovider_ProviderIgnoreTagKeysOverride` | `datadog_monitor` | Framework | provider + resource union |
| `TestAccDatadogServiceLevelObjective_IgnoreTagKeys` | `datadog_slo` | SDKv2 | resource |
| `TestAccDatadogServiceLevelObjective_ProviderIgnoreTagKeys` | `datadog_slo` | SDKv2 | provider |

Cassettes recorded against the `frog.datadoghq.com` test org. Unit tests for `StripIgnoredTags` and `ApplyIgnoreTagKeys` run via `make test`.


OliviaShoup (approved):

datadog/resource_datadog_service_level_objective.go
```diff
@@ -92,7 +92,7 @@ func resourceDatadogServiceLevelObjective() *schema.Resource {
 		ReadContext:   resourceDatadogServiceLevelObjectiveRead,
 		UpdateContext: resourceDatadogServiceLevelObjectiveUpdate,
 		DeleteContext: resourceDatadogServiceLevelObjectiveDelete,
-		CustomizeDiff: customdiff.All(resourceDatadogServiceLevelObjectiveCustomizeDiff, tagDiff),
+		CustomizeDiff: customdiff.All(resourceDatadogServiceLevelObjectiveCustomizeDiff, ignoreTagKeysDiff, tagDiff),
```
Supam: Does the ordering of `tagdiff` & `ignoreTagKeysDiff` pose a problem ?
I see that both monitors and SLO have a different ordering. 

Intuitively I feel like there might be some wierd interactions between both that might not behave the same way depending on the ordering of both `CustomizeDiff`s.

Supam (approved):
Approved with one comment

datadog/resource_datadog_service_level_objective.go
j10czar: It does not because of how SLOs are validated. Fixed anyway because its more consistent with the monitor implementation.

shaneyuandd (approved):

ely-dd (approved):

PR #3894, opened 2026-06-16
feat(auth): support TFE dynamic provider credentials via web identity token fallback

wynbennett:
## Summary

- Guards \`ContextAWSVariables\` injection in both the SDK and framework providers so it only fires when static AWS credentials are actually present. Previously the provider always injected an empty map, which caused \`AWSAuth.GetCredentials\` to short-circuit before reaching the web identity exchange path — breaking Terraform Enterprise dynamic provider credentials and IRSA.
- Bumps \`datadog-api-client-go\` to [\`3a2c1b0\`](https://github.com/DataDog/datadog-api-client-go/pull/4231) (\`wyn.bennett/web-identity-credential-chain\`), which adds \`assumeRoleWithWebIdentity()\` as a fallback in \`GetCredentials\` when no static credentials are present, plus the new \`ContextAWSWebIdentityVariables\` / \`AWSWebIdentityVariables\` API.
- Adds \`aws_web_identity_token_file\`, \`aws_role_arn\`, and \`aws_role_session_name\` provider attributes so callers can configure web identity token exchange explicitly (rather than relying solely on \`AWS_WEB_IDENTITY_TOKEN_FILE\` / \`AWS_ROLE_ARN\` env var autodiscovery).
- Adds unit tests covering: no static creds → \`ContextAWSVariables\` absent; static creds present → \`ContextAWSVariables\` injected.

## Background

Terraform Enterprise uses \`TFC_AWS_PROVIDER_AUTH\` (dynamic provider credentials): TFE's internal SDK assumes the target role and exposes only \`AWS_WEB_IDENTITY_TOKEN_FILE\` + \`AWS_ROLE_ARN\` in the run environment — never \`AWS_ACCESS_KEY_ID\`/\`AWS_SECRET_ACCESS_KEY\`/\`AWS_SESSION_TOKEN\`. With the old code, the provider injected \`ContextAWSVariables\` with all three values empty, causing \`GetCredentials\` to return empty creds and \`GenerateAwsAuthData\` to fail with \`"missing AWS credentials"\` before the new web identity path could run.

### TFE (env var autodiscovery, no explicit config)

1. Provider reads empty \`aws_access_key_id\`, empty \`aws_web_identity_token_file\` → injects neither context key
2. \`AWSAuth.GetCredentials\`: no context → static env vars empty → \`assumeRoleWithWebIdentity(AWS_WEB_IDENTITY_TOKEN_FILE, AWS_ROLE_ARN)\` → STS → temp creds ✓

### Explicit IRSA / cross-account config

1. Provider reads \`aws_web_identity_token_file\` + \`aws_role_arn\` → injects \`ContextAWSWebIdentityVariables\`
2. \`AWSAuth.GetCredentials\`: context key present → \`assumeRoleWithWebIdentity(tokenFile, roleARN)\` → STS → temp creds ✓
3. If STS fails, auth fails immediately — no silent fallback to ambient env-var creds

### Static credentials (existing path, unchanged)

1. Provider reads non-empty \`aws_access_key_id\` → injects \`ContextAWSVariables\`
2. \`AWSAuth.GetCredentials\`: step 1 short-circuits, returns static creds directly ✓

## \`GetCredentials\` priority after this PR

| Priority | Source |
|----------|--------|
| 1 | \`ContextAWSVariables\` in context (static, explicit) |
| 2 | \`ContextAWSWebIdentityVariables\` in context (web identity, explicit — authoritative, no fallthrough) |
| 3 | \`AWS_ACCESS_KEY_ID\` / \`AWS_SECRET_ACCESS_KEY\` / \`AWS_SESSION_TOKEN\` env vars |
| 4 | \`AWS_WEB_IDENTITY_TOKEN_FILE\` + \`AWS_ROLE_ARN\` env vars (TFE autodiscovery) |

## Dependency note

The \`datadog-api-client-go\` bump points to the open PR branch (\`wyn.bennett/web-identity-credential-chain\`). This should be updated to a released tag once that PR merges.

## Test plan

- [x] \`go test ./datadog/tests/ -run "TestProviderConfigure|TestFrameworkProviderConfigure"\` — all 24 tests pass
- [ ] Manual end-to-end: TFE workspace with \`cloud_provider_type = "aws"\` + \`org_uuid\`, dynamic provider credentials enabled, confirm delegated token is obtained
- [ ] Manual: explicit \`aws_web_identity_token_file\` + \`aws_role_arn\` in provider block, confirm token exchange succeeds
- [ ] Update \`go.mod\` to a tagged release of \`datadog-api-client-go\` once PR #4231 merges

### DIFF
--- a/datadog/provider.go
+++ b/datadog/provider.go
@@ -431,15 +431,21 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 		)
 		switch cloudProviderType {
 		case "aws":
-			auth = context.WithValue(
-				auth,
-				datadog.ContextAWSVariables,
-				map[string]string{
-					datadog.AWSAccessKeyIdName:     awsAccessKeyId,
-					datadog.AWSSecretAccessKeyName: awsSecretAccessKey,
-					datadog.AWSSessionTokenName:    awsSessionToken,
-				},
-			)
+			// Only inject static credentials when they are actually present.
+			// When absent (e.g. TFE dynamic provider credentials / IRSA), omitting
+			// ContextAWSVariables lets AWSAuth.GetCredentials fall through to the
+			// AWS_WEB_IDENTITY_TOKEN_FILE + AWS_ROLE_ARN exchange path.
+			if awsAccessKeyId != "" {
+				auth = context.WithValue(
+					auth,
+					datadog.ContextAWSVariables,
+					map[string]string{
+						datadog.AWSAccessKeyIdName:     awsAccessKeyId,
+						datadog.AWSSecretAccessKeyName: awsSecretAccessKey,
+						datadog.AWSSessionTokenName:    awsSessionToken,
+					},
+				)
+			}
 		default:
 			return nil, diag.FromErr(errors.New("cloud_provider_type must be set to a valid value unless validate = false"))
 		}
--- a/datadog/provider.go
+++ b/datadog/provider.go
@@ -135,6 +135,21 @@ func Provider() *schema.Provider {
 				Sensitive:   true,
 				Description: "The AWS session token; used for cloud-provider-based authentication. This can also be set using the `AWS_SESSION_TOKEN` environment variable. Required when using `cloud_provider_type` set to `aws` and using temporary credentials.",
 			},
+			"aws_web_identity_token_file": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Path to the OIDC web identity token file; used for web identity token exchange (e.g. explicit IRSA configuration). This can also be set using the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable. Takes precedence over the environment variable when both are set.",
+			},
+			"aws_role_arn": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The ARN of the IAM role to assume via web identity token exchange. This can also be set using the `AWS_ROLE_ARN` environment variable. Takes precedence over the environment variable when both are set.",
+			},
+			"aws_role_session_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The session name to use when assuming the IAM role via web identity token exchange. Appears in CloudTrail and is useful for multi-workspace attribution. This can also be set using the `AWS_ROLE_SESSION_NAME` environment variable. Defaults to `datadog-api-client`.",
+			},
 			"http_client_retry_enabled": {
 				Type:         schema.TypeString,
 				Optional:     true,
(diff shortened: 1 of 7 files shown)


datadog/provider.go
```diff
@@ -431,15 +431,21 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
-				map[string]string{
-					datadog.AWSAccessKeyIdName:     awsAccessKeyId,
-					datadog.AWSSecretAccessKeyName: awsSecretAccessKey,
-					datadog.AWSSessionTokenName:    awsSessionToken,
-				},
-			)
+			// Only inject static credentials when they are actually present.
+			// When absent (e.g. TFE dynamic provider credentials / IRSA), omitting
+			// ContextAWSVariables lets AWSAuth.GetCredentials fall through to the
+			// AWS_WEB_IDENTITY_TOKEN_FILE + AWS_ROLE_ARN exchange path.
```
wynbennett: Guards `ContextAWSVariables` injection so it only fires when static AWS credentials are actually present (`aws_access_key_id` populated from provider config or `AWS_ACCESS_KEY_ID` env var).

**Why this matters for TFE:** Terraform Enterprise dynamic provider credentials only expose `AWS_WEB_IDENTITY_TOKEN_FILE` + `AWS_ROLE_ARN` — never the static `AWS_ACCESS_KEY_ID` etc. The old unconditional injection put an empty map in the context, causing `AWSAuth.GetCredentials` to short-circuit and return empty creds before the web identity fallback (added in the `datadog-api-client-go` bump) could run. With the guard absent/empty, `GetCredentials` falls through to `assumeRoleWithWebIdentity`, which exchanges the TFE-provided token for STS creds.

The same change is made to `datadog/fwprovider/framework_provider.go` for the framework provider.

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/terraform-providers/terraform-provider-datadog
 
 require (
-	github.com/DataDog/datadog-api-client-go/v2 v2.60.1-0.20260522183654-aa58b6b70588
+	github.com/DataDog/datadog-api-client-go/v2 v2.60.1-0.20260609164814-849e9030d419
```
wynbennett: Bumps `datadog-api-client-go` from `v2.60.1-0.20260522183654-aa58b6b70588` to `v2.60.1-0.20260609164814-849e9030d419` (commit `849e9030d` on branch `wyn.bennett/web-identity-credential-chain`, open as PR #4231).

The new commit adds `assumeRoleWithWebIdentity()` as a third tier in `AWSAuth.GetCredentials`: when `ContextAWSVariables` is absent and no static `AWS_*` env vars are set, it reads `AWS_WEB_IDENTITY_TOKEN_FILE` + `AWS_ROLE_ARN`, calls STS `AssumeRoleWithWebIdentity`, and returns temporary credentials with a session token.

**Action needed:** once PR #4231 merges and a release tag is cut, update this pseudo-version to the tagged release.

datadog/provider.go
```diff
@@ -431,15 +458,34 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
+						datadog.AWSSessionTokenName:    awsSessionToken,
+					},
+				)
+			} else if awsWebIdentityTokenFile != "" || awsRoleARN != "" {
+				if awsWebIdentityTokenFile == "" || awsRoleARN == "" {
+					return nil, diag.FromErr(errors.New("aws_web_identity_token_file and aws_role_arn must both be set when using web identity token exchange"))
+				}
+				auth = context.WithValue(
+					auth,
+					datadog.ContextAWSWebIdentityVariables,
```
wynbennett, line 481: When `aws_web_identity_token_file` or `aws_role_arn` is set (and no static `aws_access_key_id`), injects `ContextAWSWebIdentityVariables` so `AWSAuth.GetCredentials` uses the explicit web identity exchange path rather than env var autodiscovery.

Key design points:
- Both-or-neither validation fires before any network call — partial config (e.g. token file without role ARN) gives a clear error naming the two provider attributes to set together.
- **Explicit config is authoritative**: if `ContextAWSWebIdentityVariables` is set and STS fails, `GetCredentials` returns `nil` immediately rather than silently falling through to ambient env-var credentials. This prevents the wrong-identity scenario where an IRSA misconfiguration silently authenticates as a different static-key identity.
- `aws_role_session_name` defaults to `datadog-api-client` (matching the library default) but can be set explicitly for CloudTrail attribution in multi-workspace setups or IAM policies that condition on `sts:RoleSessionName`.

The same changes are mirrored in `datadog/fwprovider/framework_provider.go`.

PR #3905, opened 2026-06-22
[Tfgen] Datasource coverage increase

j10czar:
## Motivation

The v2 code generator (.generator-v2/) could only emit a narrow slice of data sources. This PR builds out the emit pipeline so the generator produces working Terraform data sources across a larger slice of Datadog's v2 API surface. Singular, plural, and searchable variants of datasource, including nested objects, arrays, enums, oneOf, and datetime fields can now be generated.
This also wires the generated data sources into the framework provider.

### Emit pipeline
- New emit/builder.go : artifact building with OpenAPI->Terraform type conversion, SDK binding resolution, response-envelope flattening, and state mapping.
- New emit/registration.go : provider constructor wiring for generated data sources.
- Templates extended for plural and singular data sources, with support for nested objects, hoisted arrays/objects, and searchable singular data sources.
- Added extension-level TF description support.

### Model / parser
- New model/artifact.go and model/report.go (report moved out of the deleted report/ package).
- Parser schema normalization improvements; enum, datetime, oneOf, and snake-case naming fixes; api_key divergence fix.
- Removed the now-unused parser/representability.go.

### CLI
- generate command wired end-to-end; now exits non-zero with a failed-artifact count instead of silently swallowing model/emit/write errors.

### Provider integration
- datadog/fwprovider/datasources_generated.go registers the generated data sources in the framework provider.

### Test corpus
- Added minimal per-data-source OpenAPI slices (testdata/mini-oas/) plus the scripts that generate/validate/annotate them (documented in the README), used as generation fixtures.

## Tests

- New unit tests for builder, registration, artifact, and updated parser/model/template tests.
- Golden files added/updated for singular, plural, nested, object, list, search, and no-params variants.
- Restored determinism and writer-error tests for RunReport.Write.


j10czar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3905?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3905** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3905?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3905?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3840** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3840?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 16 other dependent PRs ([#3842](https://github.com/DataDog/terraform-provider-datadog/pull/3842) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3842?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3849](https://github.com/DataDog/terraform-provider-datadog/pull/3849) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3849?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3850](https://github.com/DataDog/terraform-provider-datadog/pull/3850) <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/3850?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 13 others)
* `master`


<h2></h2>

.generator-v2/internal/cli/generate.go
```diff
@@ -136,6 +136,9 @@ func generateArtifact(op *model.Operation, outputRoot string, check bool) model.
 	}
 	artifact.SourceFile = filepath.Join(outputRoot, "data_source_datadog_"+artifact.Name+".go")
 	entry.Path = artifact.SourceFile
+	// Non-fatal notes (e.g. query params dropped from a plural filter set) ride
+	// along on a successful entry; failEntry below overrides them on failure.
+	entry.Diagnostics = append(entry.Diagnostics, artifact.Diagnostics...)
```
j10czar, line 160: Looking for some feedback on if this was needed or not

.generator-v2/internal/contracts/tracking-field.schema.json
```diff
@@ -24,6 +24,12 @@
         "minLength": 1,
         "description": "Human-readable doc string for the generated artifact's top-level Terraform schema (the data source / resource description shown in `terraform docs`). Optional; when omitted the generated Description is left empty. Convention: \"Use this data source to retrieve information about an existing <thing>.\""
       },
+      "cardinality": {
+        "type": "string",
+        "enum": ["singular", "plural"],
```
j10czar, line 29: should this be a multi select instead or should it remain an enum?

Supam: This should remain an enum. A datasource cannot be both a singular and a plural source.

.generator-v2/internal/emit/builder_test.go
```diff
@@ -77,31 +102,191 @@ var _ = Describe("BuildDataSourceView", func() {
...
+				}),
+			},
+		}),
+	}
+}
+
+// datastoreBothOperation is an id-optional singular data source: the tracked op is
+// the by-id GET, and SearchOp points at the list GET (no query params, matching the
+// real ListDatastores). Its element mirrors data_source_datadog_datastore.go.
+func datastoreBothOperation() *model.Operation {
```
Supam, line 262: `datastoreBothOperation` has no query params on its list op, so the "both" shape never exercises filter partitioning.
`buildSingularFilters` is only called through the search-only fixture, could you add a variant where the lis op has a scalar filter (e.g. `filter[keyword]`) and check that it appears as an Optional attribute as well as a FilterParamView in `view.Search.Filters` ?

.generator-v2/internal/emit/registration.go
```diff
@@ -0,0 +1,148 @@
...
+// datasources_generated.go holds nothing else that fits the pattern, so it
+// safely recovers the already-registered set from the file's current contents.
+var datasourceConstructorRe = regexp.MustCompile(`New[A-Za-z0-9_]+DataSource`)
+
+// generatedDatasourcesHeader is everything in datasources_generated.go up to and
+// including the slice literal's opening brace. SyncGeneratedDatasources appends
+// the sorted constructors and the closing brace, then gofmt canonicalizes it.
+const generatedDatasourcesHeader = `package fwprovider
+
+import "github.com/hashicorp/terraform-plugin-framework/datasource"
```
Supam, line 42: Why use inline text and regexes here ?
Is there a specific reason why we are not using a template ?

.generator-v2/internal/emit/templates/data_source_singular.go.tmpl
```diff
@@ -70,15 +91,83 @@ func (d *{{ .GoName }}DataSource) Read(ctx context.Context, request datasource.R
...
+	defer cancel()
+	for paginationResult := range result {
+		if paginationResult.Error != nil {
+			response.Diagnostics.Append(utils.FrameworkErrorDiag(paginationResult.Error, "Error when calling `{{ .Search.Method }}WithPagination`"))
+			return
+		}
+		items = append(items, paginationResult.Item)
+	}
+{{- else }}
+	resp, _, err := d.Api.{{ .Search.Method }}(d.Auth{{ if .Search.OptionalParamsType }}, optionalParams{{ end }})
```
Supam: AFAIU, in the non-paginated searchFetch branch, the list response is bound to resp:

```
resp, _, err := d.Api.{{ .Search.Method }}(...)
items = append(items, resp.Data...)
```

In the "both" shape's by-id branch, the outer scope also declares resp. They're in separate if/else branches so it compiles. However this is a tad confusing. Future extension of the by-id branch (e.g. inspecting resp after the else) would silently shadow. Could we rename the search local to `listResp` to make the two responses unambiguous.

.generator-v2/internal/model/artifact.go
```diff
@@ -7,37 +7,319 @@ import (
...
+	searchOp := op.SearchOp
+	if searchOp == nil {
+		return nil, fmt.Errorf("model: data source %q declares group.search %q but no such operation exists",
+			op.Tracking.ArtifactName, op.Tracking.Group.Search)
+	}
+
+	// When the by-id record shape diverges from the list element shape (e.g.
+	// api_key: FullAPIKey vs PartialAPIKey), one state mapper cannot serve both.
+	// Degrade to by-id-only (the full shape, id required) and record why.
+	if op.ResponseDataRefName != "" && searchOp.ItemRefName != "" &&
```
Supam: This check triggers when both `ResponseDataRefName` and `searchOp.ItemRefName` are non-empty. When the by-id response uses an inline schema (no $ref on its data property), `ResponseDataRefName` is empty, the check is skipped, and the data source stays in "both" mode even if the shapes actually differ. This will produces a state mapper that reads the wrong fields with no error or diagnostic.

Could we flip this check ? degrade unless you can positively confirm the shapes match. 
```
// Only stay "both" when both ref names are known and agree.
if op.ResponseDataRefName == "" || searchOp.ItemRefName == "" ||
    op.ResponseDataRefName != searchOp.ItemRefName {
    // degrade to by-id-only
}
```

This should be a bit safer.

.generator-v2/internal/emit/registration.go
j10czar: Registration is inserting into the existing file so using a template would render things forward instead of reading the existing file. For the inline text, there is no data-driven structure or control flow for a template to earn its keep on, and gofmt fixes all the spacing anyway. The regex is needed to see if something already exists in either file before deciding what to do. It can only be an issue if someone uses a constructor name in a comment for example.

Supam (approved):

PR #3912, opened 2026-06-23
[datadog_tag_pipeline_ruleset] Enforce canonical query form at plan time for tag pipeline resources

dd-CyrilleChev:
-  [Ticket here](https://datadoghq.atlassian.net/browse/CCA-1536)

!!! heads-up !!! most changes are auto-generated recorded cassettes for testing, the only 'real' changes are in datadog/fwprovider & docs/resources only

## Motivation

One can set up tag pipelines in Cloud Cost feature, using either the UI, the API or by submitting a terraform file. The problem is, when one uses terraform and do 'terraform apply' (or when the auto deployment does) , it actually : 
- calls the API to create the given resource (a tag pipeline)
- the API parses and canonicalizes the query (e.g. rewrites it), creates the tag pipeline and returns this new rewritten query to terraform
- terraform then performs a sanity check 

-> error : query from resource from .tf file is different from query given back by API -> drift !!!

## Solution chosen

- Enforce a canonicalized query in config by : 
1. Call the [ValidateQuery API](https://docs.datadoghq.com/api/latest/cloud-cost-management/validate-query/) on the given config's tag pipeline query at plan time (more specs in [datadog-api-client-go](https://github.com/DataDog/datadog-api-client-go/tree/master/api/datadogV2) -> model_rules_validate_query_8.go + api_cloud_cost_management.go)
2. If the API-returned canonicalized query is different from the one in config : error in plan, and the error message contains the right canonicalized query to replace the config's with.
3. If needed, the user has now to replace himself his bad query with the right ones from error logs in 2.

## How to reproduce the error + tests

All in the google doc linked in the ticket above !


datadog/internal/utils/tag_pipeline_query_formatting.go
```diff
@@ -0,0 +1,435 @@
...
+	p.pos++
+	var parts []string
+	var b strings.Builder
+	for p.pos < len(p.in) {
+		c := p.in[p.pos]
+		switch {
+		case c == '\\' && p.pos+1 < len(p.in) && isEscapable(p.in[p.pos+1]):
+			b.WriteByte(p.in[p.pos+1])
+			p.pos += 2
+		case c == '\'' || c == '"':
```
ashleyyip: 🔴 **Blocker (worth confirming)** — `readArbitrary` doesn't track which quote opened the string, so this terminator treats `'` and `"` as interchangeable. A double-quoted value with an apostrophe closes early, the parse fails, and the query is returned unchanged. But the API canonicalizes to the escaped form (`query.Parse(...).Canonical()` escapes `'`→`\'`), so `"Tom's Account"` stays `"Tom's Account"` here while the API returns `"Tom\'s Account"` → perpetual diff.

Could we capture `q := p.in[p.pos]` before `p.pos++` and only close on that byte? A `k:"a'b"` → `k:"a\'b"` case would cover it.

docs/resources/tag_pipeline_ruleset.md
```diff
@@ -124,7 +124,7 @@ Optional:
 - `case_insensitivity` (Boolean) Whether the query matching is case insensitive.
 - `if_not_exists` (Boolean, Deprecated) Whether to apply the query only if the key doesn't exist.
 - `if_tag_exists` (String) Behavior when the tag already exists. Valid values: `append` (append to the existing tag value), `replace` (replace existing tag value), `do_not_apply` (never apply if tag already exists). Valid values are `append`, `replace`, `do_not_apply`.
-- `query` (String) The query string.
+- `query` (String) The query string. Must be in canonical form. Datadog normalizes queries (operator casing, spacing, redundant parentheses, quoting, and similar), and the provider validates the configured value against this canonical form during planning. If they differ, the plan fails with an error showing the canonical query to use in the configuration.
```
rtrieu:
```suggestion
- `query` (String) The query string. Datadog normalizes queries to a canonical form (operator casing, spacing, redundant parentheses, quoting, and similar formatting differences). The provider validates the configured value against this canonical form during planning. If the value doesn't match, the plan fails with an error showing the canonical query to use in the configuration.
```

rtrieu (approved):
minor suggestions for clarity and conciseness

ashleyyip (approved):

PR #3926, opened 2026-07-01
[datadog_logs_archive] Support access_key_id S3 destination

glecetre:
## Motivation

[LOST-66](https://datadoghq.atlassian.net/browse/LOST-66)

Allow authenticating a `datadog_logs_archive` S3 destination with an AWS **access key id**, as an alternative to the existing IAM-role auth (`account_id` + `role_name`). The API models this as a `oneOf`: `access_key_id` alone, OR both `account_id` and `role_name` together.

## Changes

- Add `access_key_id` (optional) to the `s3_archive` block; `account_id` and `role_name` are now optional and required together, all three mutually exclusive via `ConflictsWith`/`RequiredWith`.
- `buildS3Destination`: build the auth as the new `LogsArchiveIntegrationS3` oneOf union (access-key variant vs role variant).
- `buildS3Map`: flatten via a type switch on the union's actual instance.
- Bump `datadog-api-client-go` to the build containing the S3 access-key union (pseudo-version, pinned to the merged client commit — swap to a tagged release once cut).
- Add a (skipped) acceptance test mirroring the existing S3 archive tests, which can't be recorded via VCR (they depend on a live `datadog_integration_aws_account`).
- Regenerate docs.

## QA

1. `terraform apply` an `s3_archive` with `access_key_id` set → archive created with access-key auth.
2. `terraform apply` an `s3_archive` with `account_id` + `role_name` (no `access_key_id`) → still works, no plan diff on existing archives (regression check).
3. Set `access_key_id` together with `account_id`/`role_name` → plan errors with a mutual-exclusivity validation message.

## Blast radius

- Changes the schema and API payload of the `datadog_logs_archive` resource's `s3_archive` block.
- Bumps the `datadog-api-client-go` dependency to a pseudo-version.

[LOST-66]: https://datadoghq.atlassian.net/browse/LOST-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


datadog/resource_datadog_logs_archive.go
```diff
@@ -38,12 +38,26 @@ func resourceDatadogLogsArchive() *schema.Resource {
 							"bucket": {Description: "Name of your s3 bucket.", Type: schema.TypeString, Required: true},
 							"path":   {Description: "Path where the archive is stored.", Type: schema.TypeString, Optional: true},
 							"account_id": {
-								Description:      "Your AWS account id.",
+								Description:      "Your AWS account id. Required with `role_name`; mutually exclusive with `access_key_id`.",
 								Type:             schema.TypeString,
-								Required:         true,
+								Optional:         true,
```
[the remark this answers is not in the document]
glecetre: Fixed in b9e4987b

St0rmingBr4in (approved):

datadog/tests/resource_datadog_logs_archive_test.go
```diff
@@ -378,6 +378,64 @@ func TestAccDatadogLogsArchiveS3Update_basic(t *testing.T) {
 	})
 }
 
+// create then update: Ok s3 with access_key_id (mutually exclusive with account_id/role_name)
```
bdelamarre, line 382: I think we can add some more test cases to check that we cannot specify both acces_key_id and role_name at the same time and that we also cannot specify neither of them

jeff-morgan-dd (approved):

datadog/tests/resource_datadog_logs_archive_test.go
glecetre: Good point, added!

bdelamarre (approved):

PR #3928, opened 2026-07-02
[datadog_incident_user_defined_field] allow CRUD operations on Incident management's property fields

rakshith91:
## Summary
- Adds a new terraform-plugin-framework resource `datadog_incident_user_defined_field` for managing custom fields on Datadog incidents.
- Full CRUD wired to the generated `datadogV2.IncidentsApi` (`Create/Get/Update/Delete IncidentUserDefinedField`).
- Registers the resource, adds acceptance tests (Basic / Updated / Import), an example, and generated docs.

**Source API:** `/api/v2/incidents/config/user-defined-fields` (rapid-api, fully migrated, JSON:API `type: user_defined_field`)
**api-documentability:** ~93/100 — documentable.
**datadog-api-client-go:** relies on types already present in the pinned pseudo-version `v2.61.1-0.20260625212749-6aa2781f8c13` — **no SDK bump required**.

Schema was generated from the rapid-api model (`domains/incidents/libs/models/incident/user_defined_field.go`) and the published `datadog-api-spec`.

## Derived schema
| Attribute | Type | R/O/C | Modifiers |
|---|---|---|---|
| name | String | Required | ForceNew |
| type | String (enum) | Required | ForceNew |
| incident_type | String | Required | ForceNew |
| display_name | String | Opt+Computed | UseStateForUnknown |
| category / collected / default_value | String | Optional | — |
| ordinal | String | Opt+Computed | UseStateForUnknown |
| required | Bool | Opt+Computed | default false |
| tag_key | String | Optional | ForceNew |
| valid_values | Block List | Optional | — |
| id / attached_to / created / modified / deleted / reserved / prerequisite / table_id / metadata | — | Computed | — |

`type` is exposed as a validated string enum — one of `dropdown`, `multiselect`, `textbox`, `textarray`, `metrictag`, `autocomplete`, `number`, `datetime` — and translated to/from the API's numeric field type internally, so configs avoid magic numbers.

## Usage
```hcl
resource "datadog_incident_type" "example" {
  name        = "Security Incident"
  description = "Security-related incidents requiring immediate attention"
}

resource "datadog_incident_user_defined_field" "example" {
  name          = "root_cause"
  display_name  = "Root Cause"
  type          = "dropdown"
  category      = "what_happened"
  default_value = "service_bug"
  incident_type = datadog_incident_type.example.id

  valid_values {
    display_name = "Service Bug"
    value        = "service_bug"
    description  = "A bug in the service code."
  }

  valid_values {
    display_name = "Human Error"
    value        = "human_error"
  }
}
```

## Notes
- `name`/`type`/`tag_key`/`incident_type` are create-only in the API, so they're marked ForceNew.
- `default_value` cannot be cleared via PATCH (server ignores null) — noted in code.
- `metadata` is a computed, server-populated nested attribute (autocomplete-only).

## Test plan
- [ ] `make testall TESTARGS='-run TestAccDatadogIncidentUserDefinedField'` (acceptance, records/replays cassettes)
- [ ] `RECORD=true make testall TESTARGS='-run TestAccDatadogIncidentUserDefinedField'` against a real org to (re)record cassettes
- [ ] `go build ./...` / `go vet ./datadog/fwprovider/` (verified locally)
- [ ] `make docs` produces no unexpected diff (verified locally)

```release-note:feature
datadog_incident_user_defined_field: Add new resource for managing incident user-defined fields.
```


cperryk:
I know the underlying attribute is `valid_values` but in this context IMO it would make sense if the nested blocks are titled `valid_value` to make it clear that each block is one singular valid value

cperryk:
I think perhaps we should also scrap `collected` from this — I think we want to deprecate that feature totally now that we have transition forms

rakshith91:
> I know the underlying attribute is `valid_values` but in this context IMO it would make sense if the nested blocks are titled `valid_value` to make it clear that each block is one singular valid value

@cperryk fair call - lemme address it

rakshith91:
Addressed both in `ef0bb0b`:
- Renamed the nested block `valid_values` → `valid_value` (singular), so each block reads as one value. Config is now `valid_value { ... }`.
- Removed the `collected` attribute since we're deprecating that in favor of transition forms.

@cperryk PTAL.

docs/resources/incident_user_defined_field.md
```diff
@@ -0,0 +1,108 @@
...
+- `search_params` (Map of String) Additional query parameters to include in the search URL.
+- `search_query_param` (String) The query parameter used to pass typed input to the search URL.
+- `search_result_path` (String) The JSON path to the results in the response body.
+- `search_url` (String) The URL used to populate autocomplete options.
+
+## Import
+
+Import is supported using the following syntax:
+
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
```
jeff-morgan-dd, line 101: One of these lines (102 or 104) could probably be removed while keeping the other? With the colon leading into the colon, it seems a bit redundant. Non-blocking.

jeff-morgan-dd (approved):

cperryk:
@rakshith91 Sorry last thing — scrap `attached_to` (this is always `incidents`). Probably need to remove that attribute from the API at some point.

🤔 Also `table_id`

rakshith91:
Done in `a99f1e3` — dropped both `attached_to` (always `incidents`) and `table_id` (always 0) from the resource surface.

(Left `prerequisite` for now since it wasn't called out, but happy to drop it too — it's the same always-null situation.)

@jeff-morgan-dd re: the docs nit — that's the shared `tfplugindocs` Import boilerplate, identical across every resource in the repo, so I left it as-is to stay consistent. Thanks for the review!

rakshith91:
Went ahead and dropped `prerequisite` too in `f5fe846` — confirmed it's a vestigial always-null column (nothing in rapid or dogweb ever sets it, and it's not used by web-ui), so it's the same case as `attached_to`/`table_id`.

cperryk (approved):

fpighi (approved):

PR #3929, opened 2026-07-02
Adds make dev-build to create custom tf binary that we can use in an override config

apatruddog:
Adds new `make dev-build` command to create a custom .tfrc file that allows terraform to use a custom TF provider when using this command:

```
TF_CLI_CONFIG_FILE=~/.terraform.d/dev/datadog.tfrc terraform plan
TF_CLI_CONFIG_FILE=~/.terraform.d/dev/datadog.tfrc terraform apply
```

.tfrc file looks like this:
```
provider_installation {
  dev_overrides {
    "DataDog/datadog" = "/Users/andrei.patru/.terraform.d/dev/datadog"
  }
  direct {}
}
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -3,6 +3,8 @@ RECORD?=false
 GOIMPORTS_FILES?=$$(find . -name '*.go')
 PKG_NAME=datadog
 DIR=~/.terraform.d/plugins
+DEV_PLUGIN_DIR=$(HOME)/.terraform.d/dev/datadog
+DEV_TFRC=$(HOME)/.terraform.d/dev/datadog.tfrc
 ZORKIAN_VERSION?=master
 API_CLIENT_VERSION?=master
 LOCAL_PACKAGE="github.com/terraform-providers/terraform-provider-datadog"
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -19,6 +21,26 @@ install: fmtcheck
 uninstall:
 	@rm -vf $(DIR)/terraform-provider-datadog
 
+# Build the provider from the current branch and generate a Terraform CLI config
+# that points the datadog provider at this local build via dev_overrides. This lets
+# you exercise in-development changes with the regular `terraform` CLI.
+# Re-run after any code change. Usage:
+#   make dev-build
+#   TF_CLI_CONFIG_FILE=$(DEV_TFRC) terraform apply   # do NOT run `terraform init`
+dev-build:
```
fpighi, line 30: instead of a new directory, can we simply automate the manual steps instead?

fpighi, line 30: Also, probably worth extending `build` or `install` targets instead of a totally fresh target?
Your new target doesn't run fmtcheck for example.

fpighi (changes requested):

GNUmakefile
apatruddog: I don't understand what you mean here. The tfrc file needs to live somewhere – in a directory...

apatruddog: Can you give me more info about what needs to happen here?

Otherwise, can your team take this on after we merge? I'm not as familiar with this repo – I just know we really need something like this for our own use case.

fpighi: My idea was to reuse the same output dir as `make build`, which is `$GOBIN` or `$GOPATH/bin`. But no big deal so I'm fine with this in the end, we might change it in another PR.

fpighi (approved):

PR #3936, opened 2026-07-03
[datadog_sensitive_data_scanner_group] omitted samplings are the same as 100%

vhourdel:
### Summary

Fixes configuration drifts on `samplings` in a `datadog_sensitive_data_scanner_group` resource. 

### Details

The `samplings` of a group determine what proportion of a particular product to scan. Not setting any sampling for a product is equivalent to having 100% sampling (i.e. scan everything) on it.

When the API returns a 100% sampling for a product, for which the Terraform configuration did not specify any sampling, Terraform would try to remove this sampling over and over.

We now consider that an omitted sampling in the Terraform configuration is the same as a 100% sampling returned by the API, removing this drift.


datadog/resource_datadog_sensitive_data_scanner_group.go
```diff
@@ -208,24 +208,46 @@ func buildTerraformSamplings(ddSamplings []datadogV2.SensitiveDataScannerSamplin
...
-func alignTerraformSamplingsOrder(apiSamplings []interface{}, preferredOrder []interface{}) []interface{} {
-	if !samplingsEqualOrderInsensitive(apiSamplings, preferredOrder) {
-		return apiSamplings
-	}
-
-	byProduct := make(map[string]map[string]interface{}, len(apiSamplings))
-	for _, sampling := range apiSamplings {
-		samplingMap := sampling.(map[string]interface{})
-		byProduct[samplingMap["product"].(string)] = samplingMap
+func reconcileTerraformSamplings(apiSamplings, referenceSamplings []interface{}) []interface{} {
```
isabella-garza-datadog: small nit: could we rename `referenceSamplings` to something like `tfConfiguredSamplings` ? I initially wasn’t sure what “reference” meant here but it seems to represent the Terraform config we want to reconcile against

datadog/resource_datadog_sensitive_data_scanner_group.go
```diff
@@ -208,24 +208,46 @@ func buildTerraformSamplings(ddSamplings []datadogV2.SensitiveDataScannerSamplin
...
+	// must be surfaced to keep config and state in sync.
+	for _, ref := range referenceSamplings {
+		product := ref.(map[string]interface{})["product"].(string)
+		referenceProducts[product] = true
+		if apiEntry, ok := apiByProduct[product]; ok {
+			result = append(result, apiEntry)
+		} else {
+			result = append(result, map[string]interface{}{
+				"product": product,
+				"rate":    float64(100),
```
isabella-garza-datadog: nit: can we create a const for `100`? That way we could also use it below at [line 245](https://github.com/DataDog/terraform-provider-datadog/pull/3936/changes#diff-4096a74723aab5faf6d62a24180c3b2675f5765364da045b46b6a8a8f2ba3613R245)

datadog/resource_datadog_sensitive_data_scanner_group.go
```diff
@@ -208,24 +208,46 @@ func buildTerraformSamplings(ddSamplings []datadogV2.SensitiveDataScannerSamplin
...
+			})
+		}
 	}
 
-	aligned := make([]interface{}, 0, len(preferredOrder))
-	for _, sampling := range preferredOrder {
-		samplingMap := sampling.(map[string]interface{})
-		aligned = append(aligned, byProduct[samplingMap["product"].(string)])
+	// A missing sampling is equivalent to a 100% sampling: drop implicit 100%
+	// entries for unconfigured products, but surface genuine non-default drift.
```
isabella-garza-datadog: tiny nit: would you mind also updating this comment to explain a bit more what this loop is doing? I was a bit confused at first how this one differed than the above loop. For example,
```
// Next, process products returned by the API that are not configured in
// Terraform. Ignore implicit default (100%) samplings, but surface any
// non-default rates as drift.
```

isabella-garza-datadog (approved):
lgtm overall! left some small nit comments

datadog/resource_datadog_sensitive_data_scanner_group.go
vhourdel: Agreed this could be confusing. Addressed in ce6baa09

datadog/resource_datadog_sensitive_data_scanner_group.go
vhourdel: Yes, definitely seems like a good idea. Addressed in ce6baa09

datadog/resource_datadog_sensitive_data_scanner_group.go
vhourdel: Updated wording in ce6baa09

PR #3942, opened 2026-07-07
[datadog_incident_type] Add configuration attribute

rakshith91:
## Summary

- Adds a `configuration` block to the `datadog_incident_type` resource and data source, exposing the incident type's behavior settings.
- All fields are optional; any field you don't set takes its server-side default.

Parent ticket: https://datadoghq.atlassian.net/browse/INL-245

## Usage

```hcl
resource "datadog_incident_type" "example" {
  name        = "Customer Impacting"
  description = "Incidents that impact customers"

  configuration = {
    private_incidents                          = false
    private_incidents_by_default               = false
    allow_workflows                            = true
    allow_incident_deletion                    = false
    editable_timestamps                        = false
    test_incidents                             = true
    create_message                             = ""
    disable_out_of_the_box_postmortem_template = false
    slug_source                                = "default"
  }
}
```

| Field | Type | Default |
|---|---|---|
| `private_incidents` | bool | `false` |
| `private_incidents_by_default` | bool | `false` |
| `allow_workflows` | bool | `true` |
| `allow_incident_deletion` | bool | `false` |
| `editable_timestamps` | bool | `false` |
| `test_incidents` | bool | `true` |
| `create_message` | string | `""` |
| `disable_out_of_the_box_postmortem_template` | bool | `false` |
| `slug_source` | string (`default` \| `servicenow`) | `default` |

`slug_source = "servicenow"` displays the ServiceNow record ID instead of the public ID on incidents of this type (falls back to the public ID when no ServiceNow integration exists).

## Test plan

- [ ] Acceptance tests pass (`TestAccDatadogIncidentType_*`)
- [ ] Docs regenerated (`make docs`)


datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -64,9 +96,75 @@ func (r *incidentTypeResource) Schema(_ context.Context, _ resource.SchemaReques
...
+				},
+				Attributes: map[string]schema.Attribute{
+					"private_incidents": schema.BoolAttribute{
+						Description:   "Whether responders can create private incidents of this type. Defaults to `false`.",
+						Optional:      true,
+						Computed:      true,
+						PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
+					},
+					"private_incidents_by_default": schema.BoolAttribute{
+						Description:   "Whether incidents of this type are created as private by default. Defaults to `false`.",
```
codhah92: i think we should update this description because this setting only actually handles whether the create modal toggles the private setting on by default

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -64,9 +96,75 @@ func (r *incidentTypeResource) Schema(_ context.Context, _ resource.SchemaReques
...
+						Optional:      true,
+						Computed:      true,
+						PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
+					},
+					"allow_workflows": schema.BoolAttribute{
+						Description:   "Whether automation workflows can be triggered for incidents of this type. Defaults to `true`.",
```
codhah92: Image

 maybe update this description too? this setting is specifically for manually running a workflow from an incident

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -64,9 +96,75 @@ func (r *incidentTypeResource) Schema(_ context.Context, _ resource.SchemaReques
...
+						PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
+					},
+					"editable_timestamps": schema.BoolAttribute{
+						Description:   "Whether responders can edit incident timestamps for incidents of this type. Defaults to `false`.",
+						Optional:      true,
+						Computed:      true,
+						PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
+					},
+					"test_incidents": schema.BoolAttribute{
+						Description:   "Whether incidents of this type are treated as test incidents. Defaults to `true`.",
```
codhah92: same here - this toggle just handles whether test incidents of this type can be created

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -64,9 +96,75 @@ func (r *incidentTypeResource) Schema(_ context.Context, _ resource.SchemaReques
...
+						Optional:      true,
+						Computed:      true,
+						PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
+					},
+					"create_message": schema.StringAttribute{
+						Description:   "An optional message shown to users when they declare an incident of this type.",
```
codhah92: Image

probably needs an update as well

docs/data-sources/incident_type.md
```diff
@@ -27,6 +27,22 @@ data "datadog_incident_type" "example" {
...
+Read-Only:
+
+- `allow_incident_deletion` (Boolean) Whether incidents of this type can be deleted.
+- `allow_workflows` (Boolean) Whether automation workflows can be triggered for incidents of this type.
+- `create_message` (String) An optional message shown to users when they declare an incident of this type.
+- `disable_out_of_the_box_postmortem_template` (Boolean) Whether the out-of-the-box postmortem template is disabled for incidents of this type.
+- `editable_timestamps` (Boolean) Whether responders can edit incident timestamps for incidents of this type.
+- `private_incidents` (Boolean) Whether responders can create private incidents of this type.
+- `private_incidents_by_default` (Boolean) Whether incidents of this type are created as private by default.
+- `slug_source` (String) The source used to derive the incident slug (`default` or `servicenow`).
```
buraizu:
```suggestion
- `slug_source` (String) The source used to derive the incident slug. When set to `servicenow`, incidents display the ServiceNow record ID instead of the public ID. If no ServiceNow integration exists, the public ID is displayed.
```

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -64,9 +96,75 @@ func (r *incidentTypeResource) Schema(_ context.Context, _ resource.SchemaReques
...
+						Optional:      true,
+						Computed:      true,
+						PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+					},
+					"disable_out_of_the_box_postmortem_template": schema.BoolAttribute{
```
codhah92: Image

i think the naming of this setting is confusing, but if that is difficult to change then maybe just updating the description will suffice here

docs/resources/incident_type.md
```diff
@@ -30,13 +49,29 @@ resource "datadog_incident_type" "example" {
 
 ### Optional
 
+- `configuration` (Attributes) The incident type's behavior settings. Any field left unset is managed by the server and returned as its server-side value. This block is applied after creation via a separate update call, since the create endpoint does not accept configuration. (see [below for nested schema](#nestedatt--configuration))
```
buraizu:
```suggestion
- `configuration` (Attributes) The incident type's behavior settings. Any field left unset takes its server-side default. This block is applied in a separate call after the incident type is created. (see [below for nested schema](#nestedatt--configuration))
```

buraizu: Suggesting to clarify and trim this down to just what a user needs to know.

Source: `resource_datadog_incident_type.go:105`

docs/resources/incident_type.md
```diff
@@ -30,13 +49,29 @@ resource "datadog_incident_type" "example" {
...
 - `id` (String) The ID of the incident type.
 
+<a id="nestedatt--configuration"></a>
+### Nested Schema for `configuration`
+
+Optional:
+
+- `allow_incident_deletion` (Boolean) Whether incidents of this type can be deleted. Defaults to `false`.
+- `allow_workflows` (Boolean) Whether automation workflows can be triggered for incidents of this type. Defaults to `true`.
+- `create_message` (String) An optional message shown to users when they declare an incident of this type.
```
buraizu:
```suggestion
- `create_message` (String) An optional message shown to users when they declare an incident of this type. Defaults to an empty string.
```

buraizu: Suggesting to add the default here for consistency.

Source: `resource_datadog_incident_type.go:149`

docs/data-sources/incident_type.md
buraizu: Suggestion for consistency with the resource schema's version of the same field (`docs/resources/incident_type.md`).

Source: `data_source_datadog_incident_type.go:76`

buraizu (commented):
Thanks for the PR, I've left some suggestions along with their associated lines in the `.go` files (since the docs team doesn't own those files).

datadog/fwprovider/resource_datadog_incident_type.go
```diff
@@ -116,7 +220,44 @@ func (r *incidentTypeResource) Create(ctx context.Context, request resource.Crea
...
+			Data: datadogV2.IncidentTypePatchData{
+				Type: datadogV2.INCIDENTTYPETYPE_INCIDENT_TYPES,
+				Id:   resp.Data.GetId(),
+				Attributes: datadogV2.IncidentTypeUpdateAttributes{
+					Configuration: cfg,
+				},
+			},
+		}
+		patchResp, patchHTTPResp, err := r.Api.UpdateIncidentType(r.Auth, resp.Data.GetId(), patchBody)
+		if err != nil {
```
codhah92, line 233: when creating a new incident type, should we be cleaning up the **newly created incident type** if there is some kind of error during the patch? 

it's a pretty unlikely scenario, but wouldn't this now be susceptible to creating orphaned incident types?

datadog/fwprovider/resource_datadog_incident_type.go
cperryk: @codhah92 I think you're referring to `allow_postmortem_gen_on_all_states`

@rakshith91 Can you double check that `disable_out_of_the_box_postmortem_template` (1) actually does something and (2) is released globally? I'm not sure it should be here.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: Reworded to "Whether the private toggle is enabled by default in the incident creation modal for this type." in f8a9a3c.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: Reworded to "Whether users can manually run a workflow from an incident of this type." in f8a9a3c.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: Reworded to "Whether test incidents of this type can be created." in f8a9a3c.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: Reworded to "An optional message shown to users when they declare an incident of this type. Defaults to an empty string." in f8a9a3c.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: The name comes straight from the API so I can't rename it, but I clarified the description to "When enabled, incidents of this type do not use Datadog's out-of-the-box postmortem template." in f8a9a3c.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: Good catch. Added a best-effort cleanup in f8a9a3c: if the follow-up configuration PATCH fails during Create, we delete the just-created incident type (since it never gets written to state) so it isn't orphaned. If that cleanup delete also fails, we surface a warning telling the user to remove it manually. Applied to both the error and non-200 PATCH paths.

docs/data-sources/incident_type.md
rakshith91: Applied — updated the data source slug_source description to this wording (and the resource's to match) in f8a9a3c; docs regenerated.

docs/resources/incident_type.md
rakshith91: Applied this wording to the configuration block description in f8a9a3c; docs regenerated.

docs/resources/incident_type.md
rakshith91: Applied — added "Defaults to an empty string." to create_message in f8a9a3c; docs regenerated.

datadog/fwprovider/resource_datadog_incident_type.go
rakshith91: this was claude^ i guess we could rename the API, but lemme check if it's doing something first

rakshith91: yeah - verified `disable_out_of_the_box_postmortem_template` is being used in postmortem templates -  i'll remove this from here - they belong to postmortem templates (potentially)

codhah92 (approved):

buraizu (approved):

armcburney (approved):
LGTM

amarshall-dd (approved):
Only high-level review of APIR-owned docs+examples.

PR #3949, opened 2026-07-08
[datadog_software_catalog] Add includeDiscovered support

### LINKED ISSUE #3948: `datadog_software_catalog` never finds auto-discovered entities that haven't been written to via the API yet

v4.14.0

### Terraform Version

N/A (provider-level, independent of Terraform CLI version)

`datadog_software_catalog` (data source and resource)

```hcl
data "datadog_software_catalog" "test" {
  filter_ref = "datastore:default/peer.db.name:my_service,peer.db.system:postgresql"
}
```

### Expected Behavior

A Software Catalog entity Datadog has auto-discovered (visible in the Datadog UI) should be findable by `filter_ref`, even if nothing has ever written to it via the API before.

### Actual Behavior

`entities` comes back empty. `GET /api/v2/catalog/entity` only returns entities that already have an `ingestionSource` unless the request also passes `includeDiscovered=true` — which `datadog-api-client-go`'s `ListCatalogEntityOptionalParameters` already supports, but this provider never passes.

### Steps to Reproduce

1. Find a real, auto-discovered entity nothing has written to via the API yet (`GET /api/v2/catalog/entity?filter[kind]=datastore&includeDiscovered=true` shows it with `ingestionSource: null`).
2. Query it with `filter_ref` in the data source above → empty result, despite it existing.
3. Compare directly against the API: the same query with `includeDiscovered=true` added returns it.

### References

`datadog-api-client-go`: [`ListCatalogEntityOptionalParameters.IncludeDiscovered`](https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/api_software_catalog.go)

AshTharappel:
## Summary

This PR adds `includeDiscovered` support to `datadog_software_catalog`, so entities Datadog's APM/infra integrations have auto-discovered but that have never been written to via the API before can actually be found.
**Fixes #3948**

## Problem

`GET /api/v2/catalog/entity` only returns entities that already have an `ingestionSource` set unless the request also passes `includeDiscovered=true`. `datadog-api-client-go`'s `ListCatalogEntityOptionalParameters` already supports this, but neither the `datadog_software_catalog` data source nor the resource's `Read` (used by refresh and `terraform import`) ever pass it — so a real, freshly-discovered entity can never be found or enriched for the first time via Terraform.

### `data_source_datadog_software_catalog.go`
- Added a new optional `include_discovered` (Boolean) schema attribute, wired to `ListCatalogEntityOptionalParameters.WithIncludeDiscovered`.
- The `id` hash only factors this attribute in when it's actually set, so existing configurations that don't use it get an unchanged `id`.

### `resource_datadog_software_catalog.go`
- `Read` now retries with `includeDiscovered=true` only when the plain `filter[ref]` lookup finds nothing — the common case (an entity already tracked/created by this resource) sends the exact same request as before, so no existing recorded acceptance-test cassette needed changes. Confirmed `TestAccDatadogSoftwareCatalogEntity_Basic`, `TestAccDatadogCatalogEntity_Order`, and `TestAccDatadogSoftwareCatalogEntity_Datasource` all still pass unmodified under `RECORD=false`.

### `docs/data-sources/software_catalog.md`
- Regenerated via `make docs` for the new attribute.

## Testing

- [x] `gofmt`, `go build`, `go vet` clean
- [x] `make docs` — exactly the expected one-line diff
- [x] Ran the existing `datadog_software_catalog` acceptance tests against their recorded cassettes: `RECORD=false TF_ACC=1 go test ./datadog/tests/... -run 'SoftwareCatalog|CatalogEntity'` — all pass unmodified.
- [ ] Didn't add a new cassette-based acceptance test for the `includeDiscovered=true` retry path itself — no sandbox credentials to record one. Verified that specific behavior directly against the live API instead: for a known real, previously API-untouched entity, `filter[ref]` alone returns nothing, and the same query with `includeDiscovered=true` added returns it.

## Related

- `datadog-api-client-go`'s `ListCatalogEntityOptionalParameters.IncludeDiscovered` already exists — this only wires the provider to use it.
- Datadog API docs: https://docs.datadoghq.com/api/latest/software-catalog/#get-a-list-of-entities
- Per `DEVELOPMENT.md`, this should get a `changelog/bugfix` label — I don't have permission to add it myself as an external contributor.

### DIFF
--- a/datadog/fwprovider/data_source_datadog_software_catalog.go
+++ b/datadog/fwprovider/data_source_datadog_software_catalog.go
@@ -220,10 +220,15 @@ func (d *datadogSoftwareCatalogDataSource) updateState(ctx context.Context, resp
 		return softwareEntities[i].Name.String() < softwareEntities[j].Name.String()
 	})
 
+	includeDiscoveredStr := ""
+	if !state.IncludeDiscovered.IsNull() {
+		includeDiscoveredStr = strconv.FormatBool(state.IncludeDiscovered.ValueBool())
+	}
+
 	idHash := fmt.Sprintf("%x", sha256.Sum256([]byte(
 		state.FilterID.ValueString()+state.FilterName.ValueString()+state.FilterExcludeSnapshot.ValueString()+
 			state.FilterKind.ValueString()+state.FilterOwner.ValueString()+state.FilterRelationType.ValueString()+
-			state.FilterRef.ValueString()+strconv.FormatBool(state.IncludeDiscovered.ValueBool()),
+			state.FilterRef.ValueString()+includeDiscoveredStr,
 	)))
 
 	state.ID = types.StringValue(idHash)
--- a/datadog/fwprovider/data_source_datadog_software_catalog.go
+++ b/datadog/fwprovider/data_source_datadog_software_catalog.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV2"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
@@ -41,6 +42,7 @@ type datadogSoftwareCatalogDataSourceModel struct {
 	FilterKind            types.String `tfsdk:"filter_kind"`
 	FilterOwner           types.String `tfsdk:"filter_owner"`
 	FilterRelationType    types.String `tfsdk:"filter_relation_type"`
+	IncludeDiscovered     types.Bool   `tfsdk:"include_discovered"`
 
 	// Results
 	Entities []*CatalogEntityModel `tfsdk:"entities"`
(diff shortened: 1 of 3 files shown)


AshTharappel:
Pushed an update: the resource's `Read` now only retries with `includeDiscovered=true` when the plain lookup finds nothing, instead of always adding it — the original version changed the request URL unconditionally, which broke this repo's own existing acceptance-test cassettes (`TestAccDatadogSoftwareCatalogEntity_Basic`, `TestAccDatadogCatalogEntity_Order`, `TestAccDatadogSoftwareCatalogEntity_Datasource`) since they match requests by exact URL. Verified all three now pass unmodified against their recorded cassettes. Also fixed the data source's `id` hash to not change for existing users who don't set `include_discovered`.

buraizu (approved):

fpighi (approved):

christinexlee (approved):

fpighi:
Hey @AshTharappel we require signed commits in this repo. Can you please amend and sign them?

AshTharappel:
@fpighi 

ah my bad.
i have amended and pushed up signed commits.
Thank you for your help

AshTharappel:
@fpighi Hey just checking in, 
Do you have an idea on when i should expect this to be merged/released?

PR #3970, opened 2026-07-13
[generated] Add datadog_gcp_uc_configs data source

j10czar:
> ℹ️ **This PR is part of a new project that auto-generates Terraform provider data
> sources to increase coverage.** The data source code is generated **deterministically
> by tfgen from an annotated OpenAPI spec, without the use of LLMs**. This PR is
> **reviewed via AI, though human review is still necessary**. If you decide to use
> this generated data source, please **review it thoroughly and test it** before relying
> on it — see the verification note and testing guide below.

> ⚠️ **This PR contains auto-generated code and must be verified before merging.** Do not
> merge until the acceptance test has been recorded and replays green against the Frog org
> (see "How to test"). A clean build and a green generator report do **not** prove runtime
> correctness.

> ⚠️ **Merge risks flagged — read before approving.** Optional attributes (project_id, months, status_updated_at) are mapped with unguarded getters, so an absent field collapses to "" or 0 rather than null.
> Details under "Reviewer notes / risks" below.

> Please add manually add acceptance tests. The generator is unable to emit them. Boilerplate acceptance tests have been generated.
## gcp_uc_configs data source (generator-v2)

Generated by tfgen from an annotated slice of the Datadog v2 OpenAPI spec
(operations: `ListCostGCPUsageCostConfigs`; cardinality: `plural`).

### Generated
- `datadog/fwprovider/data_source_datadog_gcp_uc_configs.go`
- `datadog/tests/data_source_datadog_gcp_uc_configs_test.go`
- `datadog/fwprovider/framework_provider.go` (updated) — constructor hand-wired into `Datasources` (branch mitigation; see above)
- `docs/data-sources/gcp_uc_configs.md` (created via `make docs`)

### Cardinality
plural

### Test / cassette
- Acceptance test: `TestAccDatadogGcpUcConfigsDatasource` in `datadog/tests/data_source_datadog_gcp_uc_configs_test.go`
- Cassette: scaffold, not yet recorded

### Reviewer notes / risks
- Schema contains enums — a new/unknown enum value can trip the SDK silent-empty trap; require cassette verification.
- Plural/list-all: no server-side filter (id is computed) — narrow client-side; test with set-membership, not a fixed index or count.
- Read-after-write lag: create-then-list in one apply can transiently return 0 rows — require cassette replay before claiming it works.
- project_id, months, and status_updated_at are NOT in the attributes `required` list in the spec, yet updateState uses unguarded GetProjectId()/GetMonths()/GetStatusUpdatedAt() (item.Attributes.GetX()) instead of the GetXOk-guarded pattern; per the plural idiom this becomes "" or 0 when absent instead of null, which is ambiguous for `months` (deprecated int, 0 is also a valid value) and for `status_updated_at` (empty string vs. "never updated").
- Mapping of the remaining attributes (account_id, bucket_name, dataset, export_prefix, export_project_name, service_account, status, error_messages) is complete and matches spec required-ness/types; created_at and updated_at are deliberately dropped per the generator diagnostics, and the `type` discriminator field is omitted consistent with generator convention — no evidence of silent field loss elsewhere.
- List endpoint has no pagination parameters in the spec and the code calls the non-paginated ListCostGCPUsageCostConfigs directly (not a WithPagination variant), which is correct for this operation but should be confirmed against the live API in case pagination is added later without a spec update.
- generator info: dropped "response.data[].attributes.created_at": server-managed audit field
- generator info: dropped "response.data[].attributes.updated_at": server-managed audit field

### How to test
```bash
eval "$(dd-auth --domain frog.datadoghq.com --force-app-key --no-cache --output)"
export DD_TEST_CLIENT_API_KEY="$DD_API_KEY"
export DD_TEST_CLIENT_APP_KEY="$DD_APP_KEY"
# Do not set DD_TEST_SITE_URL — keys route themselves.
```

From the repo root, record the cassette against the live Frog test org:

```bash
make testacc RECORD=true TESTARGS='-run TestAccDatadogGcpUcConfigsDatasource'
```

Commit both files this writes under `datadog/tests/cassettes/`:

- `TestAccDatadogGcpUcConfigsDatasource.yaml`
- `TestAccDatadogGcpUcConfigsDatasource.freeze`

Then replay offline — this is the only step that proves the test actually works, and it's what CI enforces:

```bash
make testacc RECORD=false TESTARGS='-run TestAccDatadogGcpUcConfigsDatasource'
```

Only a green replay counts as verification of this data source; recording alone does not.

This is a plural (list) data source, so before recording, confirm `TestAccDatadogGcpUcConfigsDatasource` in `datadog/tests/data_source_datadog_gcp_uc_configs_test.go`:

- Depends on its seed resource so the seed is created before the list read.
- Asserts membership with `TestCheckTypeSetElemNestedAttrs` against the collection (e.g. `gcp_uc_configs.*`), not a fixed index or hard-coded count — the test org holds other records in unstable order.

---
> 🚧 **This Terraform data source generation is still under development.** Reach out to
> **#api-platform** with any questions.


docs/data-sources/gcp_uc_configs.md
```diff
@@ -0,0 +1,39 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_gcp_uc_configs Data Source - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Use this data source to retrieve information about existing Datadog gcp uc configs.
+---
+
+# datadog_gcp_uc_configs (Data Source)
+
+Use this data source to retrieve information about existing Datadog gcp uc configs.
```
maycmlee, line 11:
```suggestion
Use this data source to retrieve information about existing Datadog GCP Usage Cost configs.
```

maycmlee (commented):
Just a small suggestion

Issue #1649, opened 2022-12-01
Setting 'saml_autocreate_access_role' to "ERROR" is ignored once manually set in web interface


jorneilander:
### Terraform Version

`Terraform v1.2.5` but also applies to `v1.3.5`

### Affected Resource(s)

-   `datadog_organization_settings`

### Terraform Configuration Files

```hcl
resource "datadog_organization_settings" "suborganization" {
  name  = lower(local.customer)

  settings {
    saml_autocreate_access_role = "ERROR"
    saml_autocreate_users_domains {
      domains = ["example.com"]
      enabled = true
    }
    saml_idp_initiated_login {
      enabled = true
    }
    saml {
      enabled = true
    }
    saml_strict_mode {
      enabled = false
    }
  }
}
```

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

The value for `Choose a role for provisioning new Just-in-Time Users` in the web interface to either be blank or set to `ERROR`

### Actual Behavior

Value remains as-is when a default role has been selected in the past.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Not to my knowledge.

### References

N/A

### Other

Tried looking at the Python library for these settings but it found the following:

[OrganizationSettings.saml_autocreate_access_role()](https://datadoghq.dev/datadog-api-client-python/datadog_api_client.v1.model.html#module-datadog_api_client.v1.model.organization_settings_saml_strict_mode) mentions: 
* `Options are st (standard user), adm (admin user), or ro (read-only user).` 

but the argument to the function is an [AccessRole](https://datadoghq.dev/datadog-api-client-python/datadog_api_client.v1.model.html#datadog_api_client.v1.model.access_role.AccessRole) which mentions:

* `If omitted defaults to “st”. Must be one of [“st”, “adm”, “ro”, “ERROR”].` 

Might be unrelated to this specific issue, but still interesting.

PR #4004, opened 2026-07-21
[datadog_organization_settings] Stop forcing saml_autocreate_access_role to st

Cesarsk:
Refs #1649.

## Problem

`settings.saml_autocreate_access_role` hardcoded `Default: "st"`. An org whose role was configured elsewhere (UI or another tool) and left unmanaged in Terraform was continually reset toward `st`, so every plan showed drift. The code carried a matching `// FIXME: leave it "unspecified" by default like the child org schema ?`.

## Change

- `saml_autocreate_access_role` is now `Optional + Computed` with no static default.
- It is only sent to the API when explicitly set in config (read via raw config, since a `Computed` attribute can't distinguish "unset" from a stored value in the resolved data). An omitted value is left unchanged.

## Clearing is intentionally not supported

Thanks to @fpighi for testing against the live API — clearing the role isn't possible:

- `null` → `400 'saml_autocreate_access_role' should be string`
- `""` → `400 should be any of these ['st', 'adm', 'ro', 'ERROR']`

The spec's `nullable: true` is response-only. `ERROR` is accepted as a write value but behaves as a no-op status, so it isn't special-cased here. (An earlier revision of this PR tried empty-string/`null` clearing; that has been removed.)

## Testing

- `make fmtcheck`, `make vet`, `go build ./...`, `make docs`: clean.
- Heads-up: `TestAccDatadogOrganizationSettings_Update` and `_SecurityContacts` need re-recording (`RECORD=true`) — an omitted role is no longer sent as `"st"`, so the request body changes. This no longer depends on clearing any state, so it should record reliably.

Suggested changelog label: `changelog/bugfix`.


fpighi (approved):

fpighi:
I don't have a lot of time today to go through this, but in trying to re-record there's an issue.

If I pass null I get:
```
{"errors":["'saml_autocreate_access_role' should be string"]}
HTTP 400
```

With empty string:
```
{"errors":["Value of parameter 'saml_autocreate_access_role' should be any of these ['st', 'adm', 'ro', 'ERROR']"]}
HTTP 400
```

This prevents creating a test that can reliably run multiple times, because there's no option to clear the state in our testing organization used for the terraform provider integration tests.

I'll try to check with the relevant product team to see if we have other options here.

fpighi (changes requested):
See my previous comment, as it is it won't work reliably.

Cesarsk:
Thanks for hitting the live API — that's decisive. Your results:

- `null` → `400 'saml_autocreate_access_role' should be string`
- `""` → `400 should be any of these ['st', 'adm', 'ro', 'ERROR']`

So the field can't be cleared via `null` or an empty string, and the spec's `nullable: true` is response-only, not a valid request value. My clearing approach here is invalid — I'll drop it. (Your second error also confirms the API accepts `ERROR` as a write value, so the original #1649 behavior is "accepted but no-op" rather than rejected.)

Since there's no clear mechanism, I've narrowed this PR to the part that still fixes the core of #1649 and is reliably testable: make `saml_autocreate_access_role` `Optional + Computed` and drop the hardcoded `st` default, so a role set in the UI and left unmanaged in Terraform is no longer forced back to `st` (set → change → omit-keeps-current — no state-clearing needed).

Let me know if you'd prefer to hold until you hear back from the product team on whether clearing is possible at all — happy to track that separately as an API-side request.

drichards-87 (approved):

fpighi:
I recorded cassettes but had to rework some tests because things have changed since 2024 and the testing org we use has gone through some incompatible changes over time. It's far from ideal, but should unblock most of this PR.

datadog/tests/resource_datadog_organization_settings_test.go
```diff
@@ -82,7 +82,6 @@ func TestAccDatadogOrganizationSettings_Update(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "name", organizationName),
 					resource.TestCheckResourceAttrSet(resourceName, "public_id"),
 					resource.TestCheckResourceAttrSet(resourceName, "settings.0.private_widget_share"),
-					resource.TestCheckResourceAttr(resourceName, "settings.0.private_widget_share", "true"),
```
fpighi, line 85: There's a API behavior getting in the way when re-recording cassettes for this, since for our testing org the value is always returned as `false`. To avoid having broken tests just for this field, I decided to get rid of it since there are other fields proving a working update.

datadog/tests/resource_datadog_organization_settings_test.go
```diff
@@ -43,11 +48,9 @@ func TestAccDatadogOrganizationSettings_Update(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "settings.0.saml_autocreate_users_domains.0.enabled", "false"),
 					resource.TestCheckResourceAttr(resourceName, "settings.0.saml_autocreate_users_domains.0.domains.#", "0"),
 					resource.TestCheckResourceAttrSet(resourceName, "settings.0.saml_can_be_enabled"),
-					resource.TestCheckResourceAttr(resourceName, "settings.0.saml_can_be_enabled", "false"),
```
fpighi, line 46: Our testing org has saml enabled already and can't be unset. Since this field depends on the testing org being used at point in time, checking the value would cause it to be flaky over time.

datadog/tests/resource_datadog_organization_settings_test.go
```diff
@@ -43,11 +48,9 @@ func TestAccDatadogOrganizationSettings_Update(t *testing.T) {
...
 					resource.TestCheckResourceAttrSet(resourceName, "settings.0.saml_idp_initiated_login.0.enabled"),
 					resource.TestCheckResourceAttr(resourceName, "settings.0.saml_idp_initiated_login.0.enabled", "false"),
 					resource.TestCheckResourceAttrSet(resourceName, "settings.0.saml_idp_metadata_uploaded"),
-					resource.TestCheckResourceAttr(resourceName, "settings.0.saml_idp_metadata_uploaded", "false"),
```
fpighi, line 50: Our testing org has saml enabled already and can't be unset. Since this field depends on the testing org being used at point in time, checking the value would cause it to be flaky over time.

fpighi (approved):

Cesarsk:
thanks for the improvements @fpighi !

minh-hoang-trinh (approved):

fpighi:
@Cesarsk sorry to ask again in this PR, can you please sign your commits so we can merge? Thank you.

Cesarsk:
> @Cesarsk sorry to ask again in this PR, can you please sign your commits so we can merge? Thank you.

done, thanks!

Cesarsk:
Hey @fpighi, let me know if something else is required here! 
Cheers!

fpighi:
> Hey @fpighi, let me know if something else is required here! Cheers!

@Cesarsk I still see an unverified commit (the one I co-authored). You should also re-sign that one (and feel free to omit co-authored by if you just want to squash everything in a single commit.

Cesarsk:
@fpighi done!

fpighi (approved):

PR #4005, opened 2026-07-21
[datadog_on_call_team_routing_rules] Validate catch-all last rule at plan time

Cesarsk:
## What this PR does

Adds plan-time validation (`ValidateConfig`) to `datadog_on_call_team_routing_rules`, enforcing the API's catch-all contract for the last rule. Today `terraform plan` passes on configs the API is guaranteed to reject, and the failure only shows up at apply:

```
400 Bad Request: attribute "rules.default_rule" the last rule must not use a query or a time restriction
400 Bad Request: attribute "rule.escalation_policy" escalation policy is required
```

Both are statically checkable from the config alone. This PR surfaces them at `terraform plan` / `terraform validate`:

- the last rule must not set `query`
- the last rule must not set `time_restrictions`
- the last rule must define an escalation policy, either via the rule-level `escalation_policy` attribute or via an `escalation_policy` action

## Implementation notes

- The resource already had model-level validation (`onCallTeamRoutingRulesModel.Validate()`), but it only ran inside Create/Update — i.e. at apply time. This PR implements `resource.ResourceWithValidateConfig` and runs the same `Validate()` there, so all existing shape checks (missing `time_zone`, empty actions, conflicting escalation-policy configuration, …) now also fail at plan time, plus the new catch-all checks.
- Presence checks tolerate unknown values: an `escalation_policy` referencing a not-yet-created policy (`datadog_on_call_escalation_policy.x.id`) is unknown at plan time and must not fail validation. The `query` check explicitly skips unknown/null/empty values; null-based presence checks pass on unknowns by construction. The existing `TestAccOnCallTeamRoutingRulesCreateAndUpdate` cassette test covers this: its last rule references an unapplied escalation policy.
- Only the **last** rule is validated for the catch-all contract. The recorded cassette of the existing acceptance test shows the API accepts non-last rules without any escalation policy (rule 0 has only a `trigger_workflow_automation` action), so flagging those would reject valid configs.
- The docs example was itself invalid per this contract (its last rule had a `query`; the catch-all rule sat in the middle) — reordered so the catch-all rule is last. `docs/resources/on_call_team_routing_rules.md` is on the docs-generation exclude list, so it was updated by hand to match.

## Test coverage

`TestAccOnCallTeamRoutingRulesCatchAllValidation` (validation-only, empty cassette): one `ExpectError` step per rule violation.

```
PASS datadog/tests.TestAccOnCallTeamRoutingRulesCatchAllValidation (0.99s)
PASS datadog/tests.TestAccOnCallTeamRoutingRulesCreateAndUpdate (2.20s)
```

`make fmtcheck`, `make test`, `make vet`, `make docs && make check-docs` all pass. `make errcheck` reports only pre-existing findings in legacy SDKv2 files untouched by this PR.

Suggested changelog label: `changelog/bugfix`.


daniel-m-dd (approved):

daniel-m-dd:
thanks for the improvement!

fpighi (approved):

fpighi:
@Cesarsk can you please amend commits with a signature. We require it for this repo.

Cesarsk:
> @Cesarsk can you please amend commits with a signature. We require it for this repo.

Done!

PR #4010, opened 2026-07-21
Tfgen Pipeline Fixes

j10czar:
### Motivation
Improve the tfgen split pipeline’s handling of aggregate generation PRs and generated child PRs. Aggregate PRs should act as temporary transport branches, with work continuing in traceable child PRs.

### Changes
- Skip generated-doc checks for aggregate `datadog-api-spec/generated/*` PRs because documentation is generated per child PR.
- Include individual untracked example files in the split allowlist check.
- Derive and link the originating `datadog-api-spec` PR from the aggregate branch name.
- Add OpenAPI source comments to generated child PRs.
- Reuse child PRs associated with the same OpenAPI PR and spec hash during reruns.
- Collect child PR successes and pipeline failures for reporting.
- Reuse existing blocked-retirement issues instead of creating duplicates.
- Add an aggregate PR summary containing child PRs, blocked-retirement issues, and failures.
- Automatically close the aggregate PR only after every split operation succeeds.
- Leave the aggregate PR open when any operation fails.

### Test
- Parsed the modified workflow as valid YAML.
- Validated the embedded Bash with `bash -n`.
- Ran `git diff --check`.
- No provider tests were changed or run.
- End-to-end pipeline behavior still needs validation on an aggregate generation PR.

### DIFF
--- a/.github/workflows/tfgen-split.yml
+++ b/.github/workflows/tfgen-split.yml
@@ -134,7 +134,6 @@ jobs:
 
           PLAN_PATH="$RUNNER_TEMP/tfgen-split-plan.json"
           SPLIT_DIR="$RUNNER_TEMP/tfgen-split"
-          GENERATION_REPORT="$GITHUB_WORKSPACE/generated/.tfgen-run-report.json"
           OAPI_REPOSITORY="DataDog/datadog-api-spec"
 
           git config user.name "github-actions[bot]"
@@ -157,7 +156,6 @@ jobs:
           fi
 
           oapi_pr_url="https://github.com/$OAPI_REPOSITORY/pull/$oapi_pr_number"
-          spec_hash="$(jq -r '.spec_hash // "unknown"' "$GENERATION_REPORT")"
 
           aggregate_pr_json="[]"
           for attempt in 1 2 3 4 5 6; do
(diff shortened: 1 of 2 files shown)


.github/workflows/tfgen-split.yml
```diff
@@ -195,6 +296,39 @@ jobs:
                 ;;
             esac
 
+            source_body_marker="<!-- tfgen-source oapi_pr=$oapi_pr_number spec_hash=$spec_hash -->"
```
Supam: IIUC, the "reuse child PR on rerun" logic keys off an exact `spec_hash` match, but `spec_hash` is a SHA-256 of the whole OpenAPI spec, not per-artifact. Any spec edit changes the hash, and the next push on the same OpenAPI PR will fail to match its child PRs.

It might be worth changing it to something artifact-specific rather than the full-spec hash. A few alternatives could be to use the artifact's own file hash or only matching by branch/OpenAPI-PR and artefact name nstead so dedup actually kicks in across reruns.

Supam (approved):
Left a comment but otherwise, it looks good

.github/workflows/tfgen-split.yml
j10czar: Good catch! Will start working on a fix

PR #4018, opened 2026-07-23
[datadog_security_monitoring_rule] Fix inconsistent result after apply when query uses metrics without metric

### LINKED ISSUE #3038: security_monitoring_rule - spurious change in tf plan for new_value detection rules

v3.62.0

### Terraform Version

v1.12.0

security_monitoring_rule

```hcl
resource "datadog_security_monitoring_rule" "example_rule" {
  case {
    notifications = []
    status        = "medium"
  }

  enabled            = "true"
  has_extended_title = "false"
  message            = "EXAMPLE"
  name               = "example rule"

  options {
    decrease_criticality_based_on_env = "false"
    detection_method                  = "new_value"
    evaluation_window                 = "0"
    keep_alive                        = "0"
    max_signal_duration               = "0"

    new_value_options {
      forget_after       = "14"
      learning_duration  = "7"
      learning_method    = "duration"
      learning_threshold = "0"
    }
  }

  query {
    aggregation     = "new_value"
    group_by_fields = ["@example_group_by_attr"]
    metrics         = ["@example_count_attr"]
    query           = "source:example_source"
  }

  tags     = []
  type     = "log_detection"
}
```

Terraform plan showing:

```
  # module.whatever.datadog_security_monitoring_rule.example_rule will be updated in-place
~   resource "datadog_security_monitoring_rule" "example_rule" {
        id                 = "..."
        name               = "example rule"
        # (4 unchanged attributes hidden)

~       query {
-           metric          = "@example_count_attr" -> null
            name            = null
        }

        # (2 unchanged blocks hidden)
    }
```

### Expected Behavior

No changes should be shown

### Actual Behavior

A change is shown to the `query.metric` field even though nothing has changed

### Steps to Reproduce

1. Create a New Value Detection Rule in Datadog
2. Somehow (unclear exactly why) the backend supports _two_ fields for the "count" attribute: `query.metric` and `query.metrics`
3. The `query.metrics` field is the one supported/set in terraform
4. It seems like the backend copies this to the field `query.metric`, so terraform shows a change and tries to set this field to `null`
5. The Datadog API takes the `null` and ignores it, so a change is constantly shown

### Important Factoids

Somehow this behavior seems inconsistent, not all new value detection rules show it? It might depend on the `type` of the rule. We use `log_detection` rules mainly.

The behavior can be worked around using:

```
  lifecycle {
    ignore_changes = [
      query[0].metric
    ]
  }
```

### References

_No response_

clementgbcn:
### What this PR does

Makes the deprecated `query.metric` attribute `Computed` and extends the plan modifier so that a `datadog_security_monitoring_rule` authored with the non-deprecated `metrics` list (and no `metric`) no longer fails with **`Error: Provider produced inconsistent result after apply`**.

### Why

The API always mirrors `metrics[0]` back into the deprecated `metric` field on every response (for `sum` / `max` / `geo_data` / `new_value` aggregations — anything that uses metrics). But `query.metric` was declared `Optional` only, not `Computed`. So a config that follows the deprecation guidance and sets only `metrics`:

1. plans `metric = null` (not configured, not computed),
2. applies and the API returns `metric = "<field>"`,
3. the framework's post-apply consistency check compares plan `null` vs new state `"<field>"` for a non-Computed attribute and errors.

This blocks impossible-travel (`geo_data`) rules in particular, since they require the field and can't just drop it. It is the same root cause as #3038 (which surfaced as a perpetual `metric -> null` plan diff on the old SDKv2 implementation); the framework migration tightened the consistency check and turned it into a hard apply error.

Reported via a customer ticket for an impossible-travel rule managed with Terraform/OpenTofu on v4.15.0.

### Fix

- `query.metric` is now `Optional + Computed`, so the plan can carry the value the API will return.
- `normalizeDeprecatedMetricPlan` now syncs **both** directions, mirroring the backend:
  - `metric` set, `metrics` unset → derive `metrics` from `metric` (existing behavior).
  - `metrics` set, `metric` unset → derive `metric` from `metrics[0]`, so the planned `metric` always equals what the API returns (on create *and* update). This is why `UseStateForUnknown` is intentionally **not** used — it would pin `metric` to a stale value when `metrics` changes on update and re-introduce the inconsistency.

Both `metric`-only and `metrics`-only authoring now round-trip cleanly. Docs are unchanged (an `Optional + Computed` attribute still renders under Optional).

### Testing

- New acceptance test `TestAccDatadogSecurityMonitoringRule_ImpossibleTravelRuleMetricsField` exercises the full geo_data create + update + update lifecycle authored with `metrics = ["@usr.handle"]`, and asserts both `query.0.metric` and `query.0.metrics.0` round-trip.
- Verified the new test **fails on the pre-fix schema** and passes with the fix.
- Full `TestAccDatadogSecurityMonitoringRule*` suite green in replay (20/20), so no regression to the existing `metric`-only / `sum` / `max` tests.

Fixes #3038


clementgbcn:
Closing in favor of #3802 (@alvizi-dd), which fixes the same bug with a lighter, more idiomatic read-side reconcile that keeps `metric` Optional and out of state. I've pushed the acceptance test from this PR onto that branch, so the coverage carries over.

PR #4019, opened 2026-07-23
[tfgen] Strengthen reconciliation tests and CI

Supam:
## Motivation
Some tests were red when running them locally.
Fixed the tests and make them run in CI whenever the code of the generator is changed

## Changes

- tighten cassette adoption detection to regular YAML/freeze files with exact or underscore-suffix test names
- expand reconciliation tests for missing sources, handwritten files, endpoint tags, check mode, and deterministic orphan reporting
- run the generator test target in CI for every change under .generator-v2

## Testing

- make tfgen-test
- make fmtcheck
- make tfgen-build
- git diff --check

### DIFF
--- a/.generator-v2/go.mod
+++ b/.generator-v2/go.mod
@@ -1,8 +1,6 @@
 module github.com/terraform-providers/terraform-provider-datadog/generator
 
-go 1.25.8
-
-toolchain go1.26.1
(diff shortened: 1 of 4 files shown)


j10czar (approved):
lgtm

Issue #3862, opened 2026-06-03
`datadog_on_call_team_routing_rules`: ValidateConfig fails with "Received unknown value" when rule blocks are unknown at plan time (regression in 4.12.0)


David-Pickwell-Hearst:
### Datadog Terraform Provider Version

4.12.0 (works in 4.11.0)

### Terraform Version

v1.15.1

### What resources or data sources are affected?

datadog_on_call_team_routing_rules

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    datadog = {
      source  = "DataDog/datadog"
      version = "4.12.0"
    }
  }
}

variable "routing_rules" {
  type = list(object({
    query             = string
    escalation_policy = string
  }))
  # No default → the value is unknown during `terraform validate`.
  # In real usage this is equally triggered by feeding `rule` blocks from any
  # value that isn't known until apply (e.g. another resource's computed output).
}

resource "datadog_on_call_team_routing_rules" "this" {
  id = "00000000-0000-0000-0000-000000000000" # team UUID

  dynamic "rule" {
    for_each = var.routing_rules
    content {
      query             = rule.value.query
      escalation_policy = rule.value.escalation_policy
    }
  }
}
```

### Relevant debug or panic output

https://gist.github.com/David-Pickwell-Hearst/b5f962f433abdb4e2e61f5dbfbf17183

### Expected Behavior

`terraform validate` (and `plan`) should succeed when the rule blocks are not yet known — exactly as it did in 4.11.0. A value that is unknown at plan time is valid Terraform behaviour and the provider's config validation should tolerate it (validating only once values are known, or skipping validation for unknown values).

### Actual Behavior

terraform validate fails immediately with a provider-level value conversion error (this is a handled diagnostic, not a Go panic):

```powershell
╷
│ Error: Value Conversion Error
│
│   with datadog_on_call_team_routing_rules.this,
│   on main.tf line 24, in resource "datadog_on_call_team_routing_rules" "this":
│   24: resource "datadog_on_call_team_routing_rules" "this" {
│
│ An unexpected error was encountered trying to build a value. This is always
│ an error in the provider. Please report the following to the provider developer:
│
│ Received unknown value, however the target type cannot handle unknown values.
│ Use the corresponding `types` package type or a custom type that handles
│ unknown values.
│
│ Path: rule
│ Target Type: []*fwprovider.teamRuleModel
│ Suggested Type: basetypes.ListValue
╵
``` 

[...]

PR #4050, opened 2026-07-27
[datadog_on_call_team_routing_rules] Fix plan failure when rule blocks are unknown

alychoi-dd:
`ValidateConfig` was re-added in #[4005](https://github.com/DataDog/terraform-provider-datadog/pull/4005), reintroducing the bug that fails at plan/validate. 

The unknown-value guards added in #4005 live inside the validation logic, which runs after the decode, so they were never reached. This is the same problem #3862 fixed by removing `ValidateConfig`.

Guard before the decode instead: skip plan-time validation when the raw config contains an unknown value at any collection or object node. Unknown scalars (an escalation policy ID referencing an unapplied resource) decode fine and are still validated, so #4005's catch-all checks are preserved for every config that worked before. `Create`/`Update` validate the resolved plan, so nothing is lost for the skipped configs; the checks just move to apply.

### Testing

New `TestAccOnCallTeamRoutingRulesUnknownBlocks`:

1. unknown `rule` list via `dynamic "rule"` 
2. unknown `action` list nested in a known rule
3. unknown scalar with known block structure (asserts validation still fires at plan time)


Bnjis:
@codex review

PR #4054, opened 2026-07-28
[datadog_service_level_objective] Validate metric query data_source and stop panicking on invalid values

### LINKED ISSUE #4052: [datadog_service_level_objective] Provider panics with SIGSEGV on apply when time_slice SLI data_source != "metrics"

4.15.0

### Terraform Version

1.11.5 (opentofu)

- datadog_service_level_objective (time_slice SLI path; the count SLI path has the same defect)

```hcl
resource "datadog_service_level_objective" "example" {
  name = "example time-slice SLO"
  type = "time_slice"

  sli_specification {
    time_slice {
      comparator             = "<="
      threshold              = 3000000000
      query_interval_seconds = 300
      query {
        formula { formula_expression = "query1" }
        query {
          metric_query {
            data_source = "rum" # <-- any value other than "metrics"
            name        = "query1"
            query       = "p75:@view.loading_time{@type:view}"
          }
        }
      }
    }
  }

  thresholds {
    timeframe = "7d"
    target    = 99.9
  }
  validate = false
}
```

https://gist.github.com/lpmi-13/51650c1dcb3d40f9b96cef662fd7f413

### Expected Behavior

plan (or apply) fails with a clear validation error stating the only supported data_source is metrics — e.g. invalid value 'rum' for FormulaAndFunctionMetricDataSource: valid values are [metrics].

### Actual Behavior

plan succeeds (the data_source schema field has no validation and defaults to "metrics"). At apply, the provider panics with a nil pointer dereference (SIGSEGV) and the whole plugin process dies — surfaced as Error: Plugin did not respond (and, with concurrency, one panic fails every in-flight resource in the same run).

### Steps to Reproduce

1. Apply the config above (any data_source other than "metrics").
2. plan passes; apply crashes the provider.

### Important Factoids

The invalid value passing plan and only crashing at apply is the core surprise — dangerous because it defeats plan-time review.

### References

_No response_

lpmi-13:
Fixes #4052

### What / why

An unsupported `data_source` (e.g. `"rum"`) in an SLO metric query crashes the
provider at apply time with a nil pointer dereference (SIGSEGV), taking down the
whole plugin process (surfaced to the user as `Error: Plugin did not respond`).

`NewFormulaAndFunctionMetricDataSourceFromValue` returns `(nil, err)` for any
value other than the only supported `"metrics"`; the error was discarded with
`_` and the nil pointer was then dereferenced
(`resource_datadog_service_level_objective.go:444`, and the same pattern in the
count SLI path). Because the `data_source` schema field had no validation, the
invalid value passed `plan`/`validate` cleanly and only failed — as a crash —
during `apply`. (`validate = false` does not help; the crash is at
request-build time, unrelated to that option.)

### Changes

- Add `ValidateDiagFunc: validators.ValidateEnumValue(datadogV1.NewFormulaAndFunctionMetricDataSourceFromValue)`
  to the `data_source` field of both the time-slice and count metric queries, so
  invalid values are rejected at plan time with a clear message
  (`invalid value '...' for FormulaAndFunctionMetricDataSource: valid values are [metrics]`).
- Propagate the error from `buildSLOTimeSliceQueryStruct` and `buildSLOCountSpec`
  instead of discarding it and dereferencing a nil pointer (defense in depth).
- Regenerate docs (`make docs`): the validator adds "Valid values are `metrics`."
  to the `data_source` description.
- Add unit tests covering the invalid-`data_source` regression.

### Testing

- `make fmtcheck`, `make vet`, `make test` pass.
- New unit tests: `TestBuildSLOTimeSliceQueryStruct_InvalidDataSource`,
  `TestBuildSLOTimeSliceQueryStruct_ValidDataSource`,
  `TestBuildSLOCountSpec_InvalidDataSource`.

### Changelog

Please apply the `changelog/bugfix` label.

### DIFF
--- a/datadog/resource_datadog_service_level_objective_unit_test.go
+++ b/datadog/resource_datadog_service_level_objective_unit_test.go
@@ -1,0 +1,76 @@
+package datadog
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// metricQueryRaw mirrors the schema-decoded shape of a single time-slice /
+// count `query` block containing one `metric_query`.
+func metricQueryRaw(dataSource string) []interface{} {
+	return []interface{}{
+		map[string]interface{}{
+			"formula": []interface{}{
+				map[string]interface{}{"formula_expression": "query1"},
+			},
+			"query": []interface{}{
+				map[string]interface{}{
+					"metric_query": []interface{}{
+						map[string]interface{}{
+							"data_source": dataSource,
+							"name":        "query1",
+							"query":       "avg:system.cpu.user{*}",
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// TestBuildSLOTimeSliceQueryStruct_InvalidDataSource guards against a
+// regression where an unsupported data_source (e.g. "rum") caused a nil
+// pointer dereference: NewFormulaAndFunctionMetricDataSourceFromValue returns
+// (nil, err) for anything other than "metrics", the error was discarded, and
+// dereferencing the nil pointer crashed the whole provider process at apply.
+func TestBuildSLOTimeSliceQueryStruct_InvalidDataSource(t *testing.T) {
+	require.NotPanics(t, func() {
+		_, err := buildSLOTimeSliceQueryStruct(metricQueryRaw("rum"))
+		assert.Error(t, err, "invalid data_source must return an error, not panic")
+	})
+}
+
+func TestBuildSLOTimeSliceQueryStruct_ValidDataSource(t *testing.T) {
+	q, err := buildSLOTimeSliceQueryStruct(metricQueryRaw("metrics"))
+	require.NoError(t, err)
+	require.NotNil(t, q)
+	require.Len(t, q.GetQueries(), 1)
+}
+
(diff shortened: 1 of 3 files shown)


datadog/resource_datadog_service_level_objective_unit_test.go
```diff
@@ -0,0 +1,76 @@
...
+func TestBuildSLOCountSpec_InvalidDataSource(t *testing.T) {
+	raw := []interface{}{
+		map[string]interface{}{
+			"good_events_formula":  "query1",
+			"total_events_formula": "query1",
+			"queries": []interface{}{
+				map[string]interface{}{
+					"metric_query": []interface{}{
+						map[string]interface{}{
+							"data_source": "rum",
```
lpmi-13, line 63: since it looks like [there are efforts](https://github.com/DataDog/terraform-provider-datadog/pull/3938) to start adding in rum data sources, this test could be updated to a different obviously wrong value like `cheese` if that would be preferable.

janine-c (approved):

datadog/resource_datadog_service_level_objective_unit_test.go
fpighi: The PR you linked is not related to metric query data sources, but to provide a terraform datasource.
So keeping this value here is fine.

fpighi (approved):

PR #4055, opened 2026-07-28
ci (APIR-3132): publish PR integration results as a check run

fpighi:
## Motivation

PR integration tests are selected automatically and run against the live API, but the result never appears on the pull request. GitHub attributes `workflow_run`-triggered runs to the default branch, so the runs report `head_branch: master` and their job statuses never attach to the PR that caused them. The only integration-looking check on a PR is the label-gated `Run Integration Tests`, which is skipped without `ci/integrations`, so contributors reasonably conclude that no integration tests ran.

This was found on #4053, a PR whose sole purpose was fixing a failing integration test. The selection was correct and the tests did run live — and two of them failed — but nothing surfaced, so the PR looked like it had only been exercised in cassette-replay mode.

Related: APIR-3132.

## Overview of changes

The integration workflow now publishes an explicit check run named `pr-integration-tests` against the PR head SHA: opened as in-progress when the workflow starts, and resolved on every path out of the workflow — tests passed, tests failed, nothing was affected, the selection run failed, or the run was cancelled. Leaving it unresolved on any path would block merges once it becomes required, so that is handled deliberately rather than incidentally.

The check has to be published from this workflow rather than from PR Test Selection. That one runs on `pull_request`, whose `GITHUB_TOKEN` is read-only for fork PRs no matter what permissions it declares; `workflow_run` always runs in trusted base-repo context. The job that opens the check runs in parallel with selection, so it adds no latency. The summary is likewise built in the reporting job from a base-branch checkout, with the raw results handed over as an artifact — the job running the tests has the PR head checked out, which on an older PR branch has no summarizer at all and whose copy of it would be PR-authored anyway.

The summary is built from the raw test2json stream rather than the job's pass/fail, so it names the tests that failed instead of only linking to logs. This also makes a test that failed and then passed on gotestsum's re-run visible as flaky rather than silently green — the situation that hid the failure on #4053. Pass/fail semantics are unchanged: `--rerun-fails` still decides the conclusion. Capturing that stream needs an opt-in variable on the `testacc` target; it is unset by default, so local runs are unaffected.

Making the check actually gate merges is a follow-up: it needs `pr-integration-tests` added to the required contexts on the protected-branch rulesets, which requires repo admin. That should wait until the check has resolved correctly on real PRs, since enabling it early would block every open PR at once.

## Validation

This was tested by producing a synthetic test run report and executing a dummy workflow to verify interactions with github.

`workflow_run` workflows always execute the copy of the file on the default branch, so this PR cannot exercise its own change and the new check will not appear here. `workflow_dispatch` has the same restriction. To avoid merging the check-publishing code unproven, it was verified with a temporary `push`-triggered workflow on this branch that ran the same steps verbatim, fed by fabricated test2json streams. It was removed again in e38f714e, but its run and logs remain:

**Workflow run: https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800**

Each job builds a fixture, then runs the real check-publishing steps against it. The 1050-test case is generated by scraping every `TestAcc*` name out of `datadog/tests/*_test.go` and emitting one synthetic `pass` event each — it reproduces production row widths and volume without running anything.

| Fixture | Check conclusion | Check title | Links |
| --- | --- | --- | --- |
| nothing selected | success | No affected integration tests | [job](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800/job/90333990190) · [check](https://github.com/DataDog/terraform-provider-datadog/runs/90334054379) |
| 1 failure, 1 re-run flake, 1 skip-listed, 1 pass | **failure** | 1 failed, 1 flaky, 1 passed, 1 not run | [job](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800/job/90333990255) · [check](https://github.com/DataDog/terraform-provider-datadog/runs/90334052010) |
| 1050 synthetic passes, sized to the real suite | success | 1050 passed — summary 62002 bytes, truncated to 766 rows, accepted | [job](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800/job/90333990315) · [check](https://github.com/DataDog/terraform-provider-datadog/runs/90334029130) |
| no results file | success | No test results | [job](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800/job/90333990227) · [check](https://github.com/DataDog/terraform-provider-datadog/runs/90334028971) |
| 4 synthetic passes | success | 4 passed | [job](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30376670800/job/90333990160) · [check](https://github.com/DataDog/terraform-provider-datadog/runs/90334017406) |

All five *jobs* report success because the job itself succeeded; the second row's *check run* is the one that concluded failure, which is the intended behaviour. The `no results file` row concluded success only because that fixture faked a passing job — in production the job result drives the conclusion, so a test job that dies without producing results yields failure with that title. That pairing is not covered here.

**Also verified locally**:

- `actionlint` and `pre-commit` pass; 22 unit tests for the summarizer pass.
- Confirmed against a real `gotestsum --rerun-fails` run — a throwaway two-test Go package, not this suite — that the re-run's `pass` event is appended to the jsonfile after the original `fail`. That ordering is what the flaky detection depends on.
- Replaying #4053's reported results produces `2 flaky, 1 passed, 1 not run`, correctly attributing the two tests that only passed on re-run and the one filtered out by `flaky_tests.yaml`.
- `make -n testacc` with the new variable unset is byte-identical to before.

The real testing can only happen after merge because it's a workflow_run target.


.github/workflows/test_pr_integration.yml
```diff
@@ -17,8 +17,46 @@ concurrency:
...
+  # always exists and always reaches a conclusion. A required check that is never
+  # posted blocks the PR forever.
+  report-start:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+    outputs:
+      check_run_id: ${{ steps.create.outputs.check_run_id }}
+    steps:
+      - name: Create in-progress check run
```
fpighi, line 43: Using the check api because it's the only way to create new checks in a PR when executing from a workflow_run

Supam (approved):

PR #4068, opened 2026-07-30
DASHB-1136: Prevent new dashboards from using the v1 resource

Jiawen23:
## Summary
Adds an opt-in provider guard for the internal Dashboard v2 rollout. When DD_TERRAFORM_DISALLOW_DASHBOARD_V1_CREATE=true, Terraform rejects new datadog_dashboard resources and layout_type replacements while continuing to allow existing v1 dashboards to be refreshed, updated, imported, and deleted.

The guard defaults off, so the public provider behavior is unchanged until internal Terraform runners enable it.

## Ticket
https://datadoghq.atlassian.net/browse/DASHB-1136

## Related Docs
https://datadoghq.atlassian.net/wiki/spaces/~712020eec892ae2dd84589a031320009d839ec/pages/6291161503/Terraform+Dashboard+v2+and+Export+as+Terraform+-+Rollout+Plan

## Test Plan
- [x] make fmtcheck
- [x] make test
- [x] make vet
- [x] make check-docs
- [ ] Enable the guard in a staging Terraform runner and verify a new v1 dashboard plan is rejected
- [ ] Verify an existing v1 dashboard still produces a clean no-change plan

Note: make errcheck currently reports pre-existing unchecked errors outside this PR's changed files.

🤖 Generated with Claude Code

### DIFF
--- a/datadog/internal/utils/utils.go
+++ b/datadog/internal/utils/utils.go
@@ -630,6 +630,10 @@ func UseMonitorFrameworkProvider() bool {
 	return getEnv("TERRAFORM_MONITOR_FRAMEWORK_PROVIDER", "false") == "true"
 }
 
+func DisallowDashboardV1Create() bool {
+	return getEnv("DD_TERRAFORM_DISALLOW_DASHBOARD_V1_CREATE", "false") == "true"
+}
+
 func IsDatabricksIntegrationEnabled() bool {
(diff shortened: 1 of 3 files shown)


Jiawen23:
Superseded by DataDog/terraform-config#55217. The replacement enforces the policy in internal terraform-config CI only, without changing behavior for external provider users.

PR #4073, opened 2026-07-31
[APIR-3165] Fix datadog_dataset integration tests — parallel tests shared one hardcoded RUM filter

fpighi:
## Motivation

`TestAccDatadogDataset_Update` and `TestAccDatadogDataset_InvalidInput` have failed on every nightly master integration run with:

```
400 Bad Request: {"errors":["Invalid request: [DatasetFilterInUse] the filter
"@application.id:ce9843b0-7a45-453c-a831-55dd15f85141" is already used in a different "rum" dataset"]}
```

All four dataset acceptance tests run in parallel against that one hardcoded RUM application id. The API scopes a filter exclusively to a single dataset per product, so whichever test created its dataset first won and the rest were rejected. In `_InvalidInput` the collision also masked the `PrincipalInvalid` error step 2 was asserting on, which is why it surfaced as "expected an error with pattern, no match on".

This is not stale org state — the org holds only two long-lived non-test datasets, and each test passes when run on its own.

Jira: [APIR-3165](https://datadoghq.atlassian.net/browse/APIR-3165)

## Overview of changes

Each test now derives its own RUM and APM filter values from its unique dataset name, so no two tests can claim the same filter. The values are deterministic under the frozen clock, so cassette replay stays stable; cassettes were re-recorded because the request bodies changed.

`TestAccDatadogDataset_Basic` and `TestAccDatadogDatasetImport` are fixed by the same change. They fail for the identical reason, and leaving them holding the shared filter is what breaks the other two.

`TestAccDatadogDataset_Basic` was in the flaky skip list attributed to rate limiting. The real cause was this collision, so the entry is removed.

## Validation

- All four tests pass together against the live API with `RECORD=none`; before the change three of four failed with `DatasetFilterInUse`.
- Cassette replay passes with `RECORD=false` after re-recording.
- No provider code changed — the API constraint is real and the provider handles it correctly.

[APIR-3165]: https://datadoghq.atlassian.net/browse/APIR-3165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fpighi:
Checked the integration run ([30623087209](https://github.com/DataDog/terraform-provider-datadog/actions/runs/30623087209/job/91131981820)): the failures are all in unrelated resources, none in datasets.

All four dataset tests pass, including the two this PR fixes:

```
PASS datadog/tests.TestAccDatadogDataset_Basic
PASS datadog/tests.TestAccDatadogDataset_Update
PASS datadog/tests.TestAccDatadogDataset_InvalidInput
PASS datadog/tests.TestAccDatadogDatasetImport
```

36 other tests failed. 22 of them also fail on master in the three most recent nightly runs, so they are pre-existing and unrelated (api_key quota, CSM threats 423s, tag pipeline rulesets 429s, openapi_api, and so on). The remaining 14 are in unrelated resources too — security findings ordering, security monitoring default rules, custom framework, incident notification rules — none of which this change can reach, since it only touches the dataset test file, its cassettes, and the dataset entry in the flaky skip list.

Removing `ci/integrations` so the suite is not re-run on every push for failures this PR cannot affect.

rzyxie (approved):

datadog/tests/resource_datadog_dataset_test.go
```diff
@@ -18,17 +20,31 @@ var (
...
+// The API rejects a dataset whose filter is already claimed by another dataset of
+// the same product (`DatasetFilterInUse`), so every test must use its own filter
+// values, otherwise the dataset tests collide with each other when run in parallel.
+// The values are derived from the unique dataset name so they stay stable under the
+// frozen clock used for cassette replay.
+func uniqueDatasetFilters(ctx context.Context, t *testing.T) (string, string) {
+	name := uniqueDatasetName(ctx, t)
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(name)))
+	applicationID := fmt.Sprintf("%s-%s-%s-%s-%s", hash[0:8], hash[8:12], hash[12:16], hash[16:20], hash[20:32])
+	return fmt.Sprintf("@application.id:%s", applicationID), fmt.Sprintf("service:%s", name)
```
amarshall-dd: Perhaps this would be better as two fns, one for each filter? Multiple return values with same opaque type is confusing to me. Thought, it’s not clear to me if the dataset name needs to be shared between them (I’d guess not, esp. as there is already another dataset name in most of the call sites).

amarshall-dd (approved):

PR #4083, opened 2026-08-03
[datadog_governance_control] Add resource for Governance Console controls

Cesarsk:
### What does this PR do?

Adds a new `datadog_governance_control` resource that manages the built-in Governance Console controls (Organization Settings → Governance) via the Governance Console API (`GET`/`PATCH /api/v2/governance/control/{detection_type}`, plus a separate notification-settings sub-API).

The resource configures, per control: `detection_parameters`, `mitigation_type`, `mitigation_parameters`, and `notification_settings` (one entry per event type, with a list of notification targets), keyed by `detection_type`. `name` is read-only.

```hcl
resource "datadog_governance_control" "unused_api_keys" {
  detection_type = "unused_api_keys"

  detection_parameters = jsonencode({
    api_key_threshold = 30
  })

  notification_settings = [
    {
      event_type = "new_detection"
      enabled    = true
      targets = [
        {
          type   = "slack"
          handle = "#platform-governance"
        }
      ]
    }
  ]
}
```

### Motivation

Tag hygiene and resource governance can currently only be enforced as code for monitors (`datadog_monitor_config_policy`). The Governance Console controls cover the surfaces no tag policy reaches (Dashboards Without Team Tags, Notebooks Without Team Tags, Unused API/Application Keys, etc.), but they can only be configured by clicking toggles in the UI — which can't be reviewed, versioned, or kept consistent across multiple Datadog orgs. This resource closes that gap for anyone managing several orgs from Terraform.

### Design notes

- Controls are built into Datadog: there are no create/delete endpoints. The resource follows the `datadog_team_permission_setting` pattern — Create and Update both `PATCH` the existing control, Delete emits a warning and only removes it from state, and import uses the `detection_type` as the ID.
- `detection_parameters`/`mitigation_parameters` are free-form maps with mixed-type values in the API, so they are modeled as JSON-encoded strings (`jsontypes.Normalized`), same as `datadog_org_group_policy.content`.
- `notification_settings` is a separate sub-API on the backend (`ControlNotificationSettings`), unlike the other `*_parameters` fields. It's modeled as a typed nested list attribute (`event_type`/`enabled`/`targets`) rather than a JSON string, since its shape is fixed and known (unlike the free-form parameter maps) — this gives users real schema validation and autocomplete instead of hand-written JSON. It's `Optional`+`Computed`, so a control's existing notification configuration is reflected even if the attribute is left unset, and stays untouched if it's later removed from config.
[...]

PR #4084, opened 2026-08-03
[datadog_webhook_oauth2_client_credentials] Add resource for webhook OAuth2 auth methods

### LINKED ISSUE #3177: Webhook OAuth 2.0

resource: datadog_webhook

### Feature Request

Today I'm not able to create the webhook Auth Method or link them to a webhook, please could this be added

### References

_No response_

Cesarsk:
Fixes #3177

### What this does
Adds a new framework resource `datadog_webhook_oauth2_client_credentials` managing webhook Auth Methods of type OAuth2 client credentials (**Integrations → Webhooks → Auth Methods**), backed by the public v2 endpoints under `/api/v2/integration/webhooks/configuration/auth-method/oauth2-client-credentials`. Until now these could only be created through the UI, so a `terraform import` of a webhook or a DR rebuild silently loses the auth wiring.

### Secret handling
`client_secret` is required on create, optional on update, and never returned by the API. The attribute is marked `Sensitive`, and `Read` keeps the value already in state — the same approach as `datadog_webhook_custom_variable` secret values. This means the secret can't be drift-detected or filled in on import (documented in the attribute description).

### Notes for reviewers
- `GetWebhooksIntegrationApiV2()` added to `api_instances_helper.go` (only the v1 instance existed).
- The acceptance test has no cassettes — I don't have an org I can record against with dummy OAuth2 credentials. Happy to adjust if you'd rather record them on your side (as offered on #4083).
- Related gap, out of scope here: the v1 `WebhooksIntegration` schema doesn't expose `auth_method_id` in the public OpenAPI spec (the live API does return/accept it), so `datadog_webhook` still can't *link* to an auth method. If that field could be added to the spec, a small follow-up PR would close the loop end to end.


docs/resources/webhook_oauth2_client_credentials.md
```diff
@@ -0,0 +1,55 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "datadog_webhook_oauth2_client_credentials Resource - terraform-provider-datadog"
+subcategory: ""
+description: |-
+  Provides a Datadog webhook OAuth2 client credentials auth method resource. This can be used to create and manage the auth methods available under Integrations -> Webhooks -> Auth Methods.
+---
+
+# datadog_webhook_oauth2_client_credentials (Resource)
+
+Provides a Datadog webhook OAuth2 client credentials auth method resource. This can be used to create and manage the auth methods available under Integrations -> Webhooks -> Auth Methods.
```
brett0000FF: I know these are generated but FYI we use > instead of -> for navigation paths.

brett0000FF (approved):

docs/resources/webhook_oauth2_client_credentials.md
Cesarsk: Good catch, thanks! Fixed in 91e8fc56 (updated the schema `Description`, which is where the doc text is generated from, plus the generated doc).

alex-sariy (approved):
LGTM!

fpighi (approved):

PR #4080, opened 2026-07-31
[tfgen] Bind scalar SDK arguments

j10czar:
## What does this PR do?

- Discover method signatures and optional-parameter setters from the pinned `datadogV2` SDK source.
- Bind OpenAPI path and query parameters to required SDK arguments in SDK call order.
- Generate Terraform inputs and SDK expressions for scalar strings, booleans, integers, numbers, named scalar types, enums, and UUIDs.
- Treat a terminal path parameter as the Terraform `id`, while exposing additional path or required-query arguments as required attributes.
- Support singleton SDK methods with no ID argument and optional query parameters through SDK `With*` setters.
- Include required and optional inputs in plural data-source identity hashes.

Collection-valued required SDK arguments remain intentionally unsupported in this scalar-first change.

## Why?

tfgen previously assumed every singular read accepted one string ID and every plural read accepted only optional parameters. That prevented generation for singleton endpoints, nested paths, UUID IDs, named scalar arguments, and required scalar query parameters.

## Validation

- `make tfgen-test` — pass with race detection
- `make fmtcheck` — pass
- `make test` — 235 tests pass
- `make vet` — pass
- Build-verified full-spec coverage sweep — all 565 candidates inspected
- Remaining `sdk-arg-binding` gap count — zero

On the combined Graphite stack, build-verified in-scope coverage is 95/157 singular data sources (60.5%) and 122/195 plural data sources (62.6%).

This PR changes generator behavior and tests only; it leaves generated provider artifacts untouched.

PR #4087, opened 2026-08-03
[tfgen] Normalize bounded allOf schemas

j10czar:
## Motivation

TFgen resolves pure OpenAPI `$ref` schemas before building its attribute tree, but it previously classified every `allOf` schema as unsupported. This also affected ordinary OpenAPI 3.0 `$ref` nodes with `description` or `example` siblings because libopenapi represents those siblings as a synthetic `allOf` overlay.

On top of #4080, that gap accounted for 28 blocked direct artifacts, including otherwise ordinary objects and arrays such as budget costs, ServiceNow results, notification rules, tables, and widgets.

This PR adds bounded `allOf` normalization for the shapes used by the Datadog API spec while deliberately avoiding general JSON Schema intersection semantics.

### Normalize the supported `allOf` subset

- Treat a single structural branch plus annotation-only branches as a metadata overlay.
- Merge multi-branch object compositions when their properties are disjoint.
- Union and sort required fields, preserve description precedence, and OR sensitive markers.
- Preserve response and item reference identities through single-branch overlays so SDK binding continues to use the correct generated types.

### Keep unsupported intersections artifact-local

- Reject duplicate properties, mixed structural kinds, incompatible outer types, and other unsupported intersections with deterministic reasons.
- Carry those reasons through the normalized schema and include them in the existing full-path `UnsupportedKindError`.
- Continue processing the rest of the spec so one unsupported composition fails only its affected artifact.

### Add regression coverage

- Cover synthetic OpenAPI 3.0 `$ref` siblings, authored overlays, two- and three-object composition, inline branches, nested composition, metadata precedence, sensitivity, depth limits, and rejection cases.
- Exercise representative JSON:API list and singular responses through artifact construction.

The direct `unsupported-schema-kind` population fell from 68 to 41 artifacts. Ten affected rows now pass, while seventeen advance to a later independent blocker such as UUID conversion or generated-model naming. No unexplained outcomes changed outside the affected resource footprint.

## QA Instructions

Automated validation:

- `make tfgen-test`
- `make tfgen-build`
- `make fmtcheck`
- `make test`
- `make vet`

Full-spec validation used `datadog-api-spec` commit `efa610ceed9a69e8e4319ef4b4f46f344be283e4` and SDK version `v2.62.1-0.20260724190103-95e5c326c65e`, running every SDK-bound singular, plural, and paired candidate.

`make errcheck` retains the repository's existing legacy unchecked-error failures; it reports no findings under `.generator-v2`.

## Blast Radius

This changes only generator-v2 schema normalization and diagnostics. It does not modify existing provider resources, checked-in generated data sources, Terraform schemas, or runtime API behavior.

The new behavior applies only when a source OpenAPI schema contains `allOf`. Unsupported or ambiguous intersections remain rejected per artifact rather than being guessed. The PR is a single commit and can be reverted independently; it is stacked on #4080 and should be reviewed and merged after that PR.

### DIFF
--- a/.generator-v2/internal/model/schema.go
+++ b/.generator-v2/internal/model/schema.go
@@ -22,12 +22,17 @@ const (
 // one rather than emitting a types.Dynamic escape hatch. (oneOf variants are
 // dropped, not errored.)
 type UnsupportedKindError struct {
-	Path string
-	Kind SchemaKind
+	Path   string
+	Kind   SchemaKind
+	Reason string
 }
 
 func (e *UnsupportedKindError) Error() string {
-	return fmt.Sprintf("model: cannot build attribute at %q: schema kind %q is not representable", e.Path, e.Kind)
+	msg := fmt.Sprintf("model: cannot build attribute at %q: schema kind %q is not representable", e.Path, e.Kind)
+	if e.Reason != "" {
+		msg += ": " + e.Reason
+	}
+	return msg
 }
 
 // BuildResponseTree converts a response-body schema into an AttributeTree,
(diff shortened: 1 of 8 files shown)


Supam (approved):

Supam:
losing as superseded this is on master via #4111

PR #4088, opened 2026-08-03
[tfgen] Handle UUID and absent response IDs

j10czar:
## Motivation

After bounded `allOf` normalization, UUID conversion is the next recurring generator-v2 blocker. The Terraform model exposes UUIDs as strings, while the pinned Datadog SDK uses `uuid.UUID` for some request arguments and response getters. The emitter previously rejected optional UUID filters and generated invalid string conversions for UUID response values.

A smaller adjacent blocker affected JSON:API responses whose data object has no `id` getter. Those responses can still populate useful state, but the emitter unconditionally generated `GetIdOk()` and failed to compile.

This PR handles both low-hanging cases at the emitter boundary, where the SDK types and Terraform model types are both known.

### Convert UUIDs at SDK boundaries

- Render UUID response values through `.String()` for singular IDs, scalar fields, nested fields, and plural results.
- Parse Terraform string filters with `uuid.Parse` only when the resolved SDK setter accepts `uuid.UUID`.
- Add a Terraform diagnostic and stop the read when a UUID filter is invalid.
- Keep string-backed setters unchanged even when the OpenAPI field carries `format: uuid`.

### Support responses without an ID getter

- Emit the response ID assignment only when `data.id` exists.
- Preserve the Terraform identity already selected from inputs or search results when the response omits an ID.
- Avoid generating an unused attributes preamble when the response has no flattened attribute leaves.

### Add regression coverage

- Cover UUID conversion across singular and plural output shapes.
- Cover optional UUID filters for both UUID-backed and string-backed SDK setters.
- Cover ID-less JSON:API responses and empty attribute envelopes.
- Exercise a production-shaped generated SDK binding through the CLI end-to-end test.

## QA Instructions

Automated validation:

- `make tfgen-test`
- `make tfgen-build`
- `make fmtcheck`
- `make test`
- `make vet`

Emitter package coverage is 86.8%. `make errcheck` retains the repository existing legacy unchecked-error failures and reports no generator-v2 findings.

## Blast Radius

This changes only generator-v2 data-source emission and its tests. It does not modify checked-in provider resources, Terraform schemas, generated documentation, or runtime behavior of existing resources.

UUID parsing is selected from the resolved pinned-SDK setter type rather than inferred from OpenAPI format alone. Responses without `data.id` retain the already established Terraform identity instead of inventing one. The PR is a single commit and can be reverted independently; it is stacked on #4087 and should be reviewed and merged after that PR.

### DIFF
--- a/.generator-v2/internal/cli/generate_test.go
+++ b/.generator-v2/internal/cli/generate_test.go
@@ -25,6 +25,56 @@ func runTfgen(args ...string) error {
 	return root.Execute()
 }
 
+// TestGenerateHandlesUUIDResponseAndOptionalFilter drives a production-shaped
+// incident-notification-template list operation through parsing, pinned-SDK
+// binding, model construction, and rendering. Its response data ID and optional
+// incident-type filter are both UUIDs, covering the two emitter boundaries
+// together.
+func TestGenerateHandlesUUIDResponseAndOptionalFilter(t *testing.T) {
+	spec, err := os.ReadFile(filepath.Join("..", "testdata", "mini-oas", "mini-datadog_incident_notification_template.yaml"))
+	if err != nil {
+		t.Fatalf("reading incident notification template spec: %v", err)
+	}
+	annotation := `      operationId: ListIncidentNotificationTemplates
+      x-datadog-tf-generator:
+        artifact_kind: data_source
+        artifact_name: incident_notification_templates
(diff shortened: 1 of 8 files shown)


Supam (approved):

Supam:
Closing as superseded this is on master via #4111 and #4118

PR #4101, opened 2026-08-04
[TFgen] Scope generated model names to their artifact and component

Supam:
## Motivation

Two differently-shaped objects reachable under the same property name produced one struct name, and the artifact could not compile.

## Changes

- Adds `Attribute.ModelRefName`: the OpenAPI component that supplied the *object* schema backing a generated model struct — the node's own schema for an object, its element schema for an array or map of objects. Empty when that schema was inline.
- Emit names a nested model after its component rather than after the property that happens to point at it, which is the same preference the SDK generator's `child_models()` applies and that oneOf envelope naming already applies via `OneOfSpec.Name`.
- Scopes every generated model name to its artifact, so two data sources cannot collide, and keeps nested models unexported as the hand-written data sources do.
- `dedupeModels` collapses one component reached twice into a single declaration both sites point at, and reports a real conflict (same name, different fields) rather than silently emitting a redeclaration.

## Testing

`go build ./...` and `go test ./...` green. New `modelname_test.go` covers the collision, scoping and dedup cases. Goldens regenerated — the diff is the artifact-scoped model names throughout, including the two oneOf goldens.

PR #4100, opened 2026-08-04
[TFgen] (oneOf) Emit the envelope's schema blocks and model structs

Supam:
## Motivation

Render the Terraform half of a oneOf envelope, replacing the blanket emit guard from #4096 so an artifact with a union generates instead of failing.

## Changes

- One nested block per alternative under the envelope's block, and a model struct per envelope holding one pointer field per variant, whose non-nil-ness selects the alternative. A value-wrapped alternative gets a single `value` child, since Terraform has no bare-scalar block.
- Envelopes are deduplicated by the parser's envelope identity and ordered by first use, so an envelope reached from two places is declared once and the output cannot depend on OpenAPI alternative order.
- The plural item element carries envelopes too, not just the singular record.

What this does **not** yet do is populate the envelope from the SDK wrapper — so the view carries a `Warnings` entry saying exactly that, and the run report surfaces it. Emitting a struct that silently stays empty without saying so would be worse than failing. #4051 populates it and drops the warning.

## Testing

`go build ./...` and `go test ./...` green. New `oneof_test.go` plus two golden data sources (`data_source_singular_oneof`, `..._oneof_list`).

PR #4102, opened 2026-08-04
[tfgen] Qualify generated nested model names

j10czar:
## Motivation

TFgen currently names nested Go model structs from only their final field name. Different paths such as `forecast.costs` and `budget.costs`, or different generated data sources that both use an SDK `User` model, can therefore declare the same package-level Go type. Those collisions prevent otherwise valid generated data sources from compiling.

This PR is stacked on #4088 and makes generated model names unique to both their Terraform artifact and complete ownership path. The names are private Go implementation details; Terraform attribute names, `tfsdk` tags, state shape, and API behavior remain unchanged.

## Changes

- Seed singular nested model names from the generated data-source name and append every parent path segment during recursive traversal.
- Scope plural item models to the Terraform artifact and SDK item type, then derive nested children from that qualified item path.
- Reject duplicate model type names before rendering an uncompilable artifact.
- Add regression coverage for repeated leaf names under different parents and multiple artifacts sharing the same SDK item type.
- Update golden files mechanically to reflect the new private Go type names.

Readable path-qualified names are longer than leaf-only names, but make generated diagnostics deterministic and understandable without introducing hashes or a global name registry.

## QA Instructions

- `make tfgen-test` — passed, including 140/140 emitter specs.
- `make tfgen-build` — passed.
- `make fmtcheck` — passed.
- `make test` — passed, 235 provider tests.
- `make vet` — passed.
- `make errcheck` — reports the existing repository-wide legacy unchecked-error backlog; no finding touches generator-v2 or this change.

An exhaustive build-verified sweep was run against all 565 full-spec report candidates using a temporary integration head containing every PR in the stack, including the latest API-accessor follow-up. The resulting stack coverage was:

| View | Pass | Coverage |
|---|---:|---:|
| Singular | 111/156 | 71.2% |
| Plural | 155/194 | 79.9% |
| Both | 63/96 | 65.6% |

All previous duplicate Go model declaration diagnostics disappeared. The collision-only set contributed four singular and nine plural build-verified passes. `GetStatusPage` and `ListStatusPages` now expose a separate state-mapping issue rather than a model-name collision.

## Blast Radius

For internal generator users, nested-model name collisions stop blocking generated data sources from compiling. The change is limited to generator-v2's private emitted Go identifiers and can be reverted without affecting existing Terraform configuration or state compatibility.

### DIFF
--- a/.generator-v2/internal/emit/builder.go
+++ b/.generator-v2/internal/emit/builder.go
@@ -104,7 +104,8 @@ func BuildDataSourceView(a *model.Artifact) (DataSourceView, error) {
 		}
 	}
 
-	rootStruct := dsGoName(a.Name) + "DataSourceModel"
+	goName := dsGoName(a.Name)
+	rootStruct := goName + "DataSourceModel"
 	env := b.flattenEnvelope(topLevel, idStrategy, rootExpr)
 
 	if len(b.unsupported) > 0 {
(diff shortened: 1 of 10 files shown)


Supam (approved):

Supam:
Closing as superseded, this is on master via #4100, with the naming rules also covered by #4101

PR #4103, opened 2026-08-04
[tfgen] Support typed nested collections

j10czar:
## Motivation

TFgen currently replaces an array or map nested inside another collection with an unsupported-schema sentinel. Terraform Plugin Framework can represent recursively typed collections natively, so valid SDK shapes such as `[][]string`, `map[string][]string`, and `map[string]map[string][]string` are blocked unnecessarily.

This PR is stacked on #4102. It preserves typed collection chains through parsing, models them as recursive Framework element types, and emits their state without encoding values as JSON strings.

## Changes

- Preserve nested array and map schemas instead of downgrading them during normalization.
- Recursively construct `types.ListType` and `types.MapType` expressions for collection chains that terminate in a Terraform primitive.
- Emit `schema.ListAttribute` and `schema.MapAttribute` leaves with matching `types.ListValueFrom` or `types.MapValueFrom` state conversion.
- Support typed maps at the response root, on plural result items, and inside generated object-list models.
- Add parser, model, emitter, and rendered-source coverage for nested lists, nested maps, mixed list/map chains, and unsupported terminal values.

Collections ending in arbitrary JSON or an object remain fail-closed. Supporting those shapes would require separate dynamic-value or nested-object semantics and is intentionally outside this PR.

## QA Instructions

- `make tfgen-test` — passed on this branch and on the fully integrated stack.
- `make tfgen-build` — passed.
- `make fmtcheck` — passed.
- `make test` — passed, 235 provider tests.
- `make vet` — passed.

Focused generation/build probes passed for the affected full-spec operations. An exhaustive build-verified sweep was also run against all 565 candidates using a temporary integration head containing every PR in the stack, including the latest API-accessor and `allOf` follow-ups.

| View | Before | After | Gain |
|---|---:|---:|---:|
| Singular | 111/156 (71.2%) | 115/156 (73.7%) | +4 |
| Plural | 155/194 (79.9%) | 161/194 (83.0%) | +6 |
| Combined | 266/350 (76.0%) | 276/350 (78.9%) | +10 |

No previously passing candidate regressed. New build-verified coverage includes ServiceNow templates, CSM threat policies, resource-evaluation filters, aggregated connections, signal investigation queries and suggested actions, and tag-pipeline rulesets.

## Blast Radius

The change is limited to generator-v2 schema normalization, generated Framework schema types, and generated response-to-state conversion. It does not modify handwritten data sources or existing Terraform state. Generated artifacts may now expose typed attributes that previously failed generation. Arbitrary JSON and object-terminal collections remain blocked rather than being represented ambiguously. The PR can be reverted without a state migration.

### DIFF
--- a/.generator-v2/internal/emit/builder.go
+++ b/.generator-v2/internal/emit/builder.go
@@ -392,9 +392,8 @@ func (b *dataSourceBuilder) flattenEnvelope(topLevel []*model.Attribute, idStrat
 		return nil
 	}
 
-	// Hoist the attribute children to top-level paths: scalar leaves, array nodes
-	// (list-of-primitive / list-of-object), and bare nested objects are in scope;
-	// a map is not.
+	// Hoist the attribute children to top-level paths: scalar leaves, typed list/map
+	// attributes, list-of-object blocks, and bare nested objects are in scope.
 	leaves := make([]*model.Attribute, 0, len(attributes.Children))
 	for _, child := range attributes.Children {
 		if isAuditField(tfNameOf(child.Path)) {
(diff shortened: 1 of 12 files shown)


Supam (approved):

Supam:
Closing as superseded, this is on master via #4111 
`collectionElement` and `isMapType` are both present, along with the `framework_types.go` typed-collection changes, and all 34 of this PR's test descriptions exist on master.

PR #4105, opened 2026-08-05
[APIR-3056] Fail CI when a generated file is hand-edited

Supam:
## Motivation

Jira: https://datadoghq.atlassian.net/browse/APIR-3056

Files produced by tfgen carry the standard Go "Code generated ... DO NOT EDIT." header, but nothing currently stops a human from editing them directly, which silently diverges the checked-in file from what the generator would produce.

## Changes

- Added a `no-generated-file-edits` job to `.github/workflows/test.yml` that scans the PR diff for any changed file carrying the "Code generated ... DO NOT EDIT." marker and fails the build if found.
- Exempted the `generate/*` and `retire/*` branches (created by the `tfgen-split` workflow) and PRs opened by `dd-octo-sts[bot]`, since those are the legitimate paths for generated file changes.
- Wired the new job into the `all-checks` gate.

## Testing

Validated the workflow YAML parses correctly.

PR #4107, opened 2026-08-06
[APIR-3056] (CI) Exclude testdata fixtures from generated-file check

Supam:
## Motivation

Jira: https://datadoghq.atlassian.net/browse/APIR-3056

The `no-generated-file-edits` gate (#4105) flags any changed file carrying the `// Code generated ... DO NOT EDIT.` header as a hand-edited generated file. Generator golden fixtures under `testdata/` (e.g. `.generator-v2/internal/testdata/emit/*.golden`) legitimately contain that header as fixture content — they are expected generator output snapshots, not files produced directly by tfgen — so the gate was failing CI on PRs that update those fixtures.

## Changes

- Skip any changed file under a `testdata/` directory when scanning for the generated-file header in `.github/workflows/test.yml`.

## Testing

- Not directly testable outside of CI; relying on this PR's own CI run of the `no-generated-file-edits` job to confirm testdata changes no longer trip the gate.


Supam:
* **#4107** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

.github/workflows/test.yml
```diff
@@ -72,6 +72,11 @@ jobs:
           files=$(git diff --name-only --diff-filter=d "${{ github.event.pull_request.base.sha }}...${{ github.event.pull_request.head.sha }}")
           generated=""
           for f in $files; do
+            # testdata fixtures (e.g. generator golden files) may legitimately
+            # carry this header as fixture content; they aren't tfgen output.
+            case "$f" in
+              */testdata/*) continue ;;
+            esac
             if grep -qE '^// Code generated .* DO NOT EDIT\.$' "$f" 2>/dev/null; then
```
fpighi: Instead of using a generic regex, can we match lines added by tfgen by appending e.g. `// source: tfgen` to the auto-generated comment? That would mean we can just match one string and ignore where the file lives.

fpighi: Actually we already have `// Code generated by tfgen; DO NOT EDIT.` so we can just match that right?

fpighi (changes requested):
We should update the regex.

.github/workflows/test.yml
Supam: we do match this, however the golden files also have this header.
When we add golden files, or chnage them, the CI fails

fpighi (approved):

PR #4026, opened 2026-07-23
[datadog_incident_postmortem_template] Add incident postmortem template resource

rakshith91:
## Summary
- New resource `datadog_incident_postmortem_template` for the incidents postmortem-templates config API (`/api/v2/incidents/config/postmortem-templates`).
- Schema derived from the incidents-rapid-api model (source of truth) via the necessity gate: `name` (required), `content`, `is_default` (bool ↔ the API's default-timestamp), `location` (`datadog_notebooks` | `confluence` | `google_docs`), `incident_type` (required, **ForceNew** — immutable), and the `confluence_postmortem_settings` / `google_docs_postmortem_settings` nested blocks. Computed: `id`, `created`, `modified`, `last_modified_by_user`.
- Plan-time `ValidateConfig` guard rejects a settings block that doesn't match `location` (and requires the matching block), mirroring the API's create/update validation.
- Excluded `created_by_user` (necessity gate: web-ui never sends or displays it).

## Prerequisites (blocked until both land)
1. **Spec:** [DataDog/datadog-api-spec#6178](https://github.com/DataDog/datadog-api-spec/pull/6178) must merge. Generated client: [DataDog/datadog-api-client-go#4362](https://github.com/DataDog/datadog-api-client-go/pull/4362).
2. **Client sync:** `go.mod` is bumped to that generated client (`v2.62.1-0.20260723174644-c8b69f9f30ce`) so the `PostmortemTemplate*` types resolve. That client also renamed `ObservabilityPipelineTls` → `ObservabilityPipelineClientTls`, which TF master hasn't adopted — so a whole-repo build currently fails **only** in the unrelated `observability_pipeline` package. This resource type-checks clean; the obs-pipeline breakage resolves on the provider's normal client-sync.

## Terraform usage
```hcl
resource "datadog_incident_type" "outage" {
  name = "Service Outage"
}

resource "datadog_incident_postmortem_template" "standard" {
  name          = "Standard Postmortem"
  content       = "# Overview\n\n# Timeline\n\n# Action Items\n"
  is_default    = true
  incident_type = datadog_incident_type.outage.id
}
```

## Test plan
- [x] `go build ./datadog/...` — zero errors in the new resource/test (only the unrelated obs-pipeline client-skew fails; see prerequisite 2)
- [ ] Rebase after the provider client-sync, confirm whole-repo build
[...]

PR #4115, opened 2026-08-07
[datadog_incident_postmortem_template] enable unstable operations for incident postmortem templates

sonya:
The `datadog_incident_postmortem_template` resource was added in #4026 and registered in the framework provider, but the corresponding unstable operations were never enabled. Every operation on the resource fails:

    Could not create incident postmortem template, unexpected error:
    Unstable operation 'v2.CreateIncidentPostmortemTemplate' is disabled.
    HTTP Response: <nil>

This makes the resource unusable in all released versions that contain it (v4.17.0 onward). Enable the four operations the resource calls, in both the provider and the acceptance test client config, matching the existing `IncidentNotificationTemplate` block.


rakshith91 (approved):

fpighi (approved):

fpighi:
@sonya we require commits to be signed in this repo, could you please ensure that happens and then I'll be happy to merge.

sonya:
> @sonya we require commits to be signed in this repo, could you please ensure that happens and then I'll be happy to merge.

Thanks for the quick review @fpighi! I've just updated the commit, let me know if there's anything else I should do.

PR #4117, opened 2026-08-10
[datadog_saml_idp_metadata] Add resource to upload SAML IdP metadata

### LINKED ISSUE #1211: Feature Request: Add support to upload the SAML IdP metadata
Hi,

It would be useful to have a resource to update the SAML IdP metadata.
We can configure everything around SAML with datadog_organization_settings, except the IdP metadata.

This is the API that can be used:
https://docs.datadoghq.com/api/latest/organizations/#upload-idp-metadata

Thanks.

supermarine1377:
Closes #1211

## What this does

Adds a new framework resource `datadog_saml_idp_metadata` that uploads or replaces the organization's SAML IdP metadata via [`POST /api/v2/saml_configurations/idp_metadata`](https://docs.datadoghq.com/api/latest/organizations/#upload-idp-metadata). This is the one piece of SAML configuration not covered by `datadog_organization_settings`, and users are currently working around it with `local-exec` + curl (see issue thread).

```hcl
resource "datadog_saml_idp_metadata" "example" {
  idp_metadata = file("${path.module}/idp_metadata.xml")
}
```

## Design notes

- **Create/Update** both call `UploadIdPMetadata` (the endpoint's semantics are "upload or replace"), then resolve the resource ID and computed attributes (`entity_id`, `sso_url`, `expires_at`) from `ListSAMLConfigurations` — an organization has at most one SAML configuration.
- **BOM handling**: a leading UTF-8 BOM is stripped before upload. Azure AD metadata downloads include one and the API rejects it (called out explicitly in the issue thread).
- **Read** uses `GetSAMLConfiguration` and removes the resource from state on 404. The raw metadata XML cannot be read back from the API, so out-of-band changes to the content are not detected (documented in the resource description).
- **Delete** is state-only with a warning, as the API has no delete endpoint (same pattern as `datadog_team_permission_setting`).
- **Import** is intentionally not supported since the required `idp_metadata` content cannot be populated from the API.

## Testing

- `make fmtcheck`, `make vet`, `make test` pass locally (`make errcheck` failure in `resource_datadog_synthetics_test_.go` is pre-existing on master).
- Acceptance test `TestAccSamlIdpMetadataBasic` covers create + in-place update (SSO URL change), using metadata built around a long-lived self-signed certificate.

⚠️ **Cassettes are not yet recorded** — this is why the PR is a draft. Recording replaces the real SAML IdP metadata of the org the credentials point to, which could break SSO login for that org, so it needs to run against a sandbox org where that is acceptable. Guidance from maintainers on how you'd like this recorded would be appreciated.


datadog/fwprovider/resource_datadog_saml_idp_metadata.go
```diff
@@ -0,0 +1,174 @@
...
+	response.Schema = schema.Schema{
+		Description: "Provides a Datadog SAML IdP metadata resource. This can be used to upload or replace the identity provider (IdP) metadata used for the organization's SAML login configuration. An organization has at most one SAML configuration. Note: the uploaded metadata XML cannot be read back from the Datadog API, so changes made outside of Terraform to the metadata content are not detected. Destroying this resource removes it from Terraform state but does not remove the IdP metadata from Datadog.",
+		Attributes: map[string]schema.Attribute{
+			"idp_metadata": schema.StringAttribute{
+				Required:    true,
+				Description: "The content of the IdP metadata XML file, for example loaded with the `file()` function. A leading UTF-8 byte order mark (BOM) is stripped automatically before upload.",
+			},
+			"entity_id": schema.StringAttribute{
+				Computed:    true,
+				Description: "The IdP entity ID of the SAML configuration.",
```
fpighi: This is not the uploaded URL as the tests below seem to assert, but the one coming from the backend which is going to be `https://app.datadoghq.com/account/saml/metadata.xml`. You can try fetching your config via `https://api.datadoghq.com/api/v2/saml_configurations` to see what's returned for you.
Same for the `sso_url` below which is a org-specific url. So the tests asserting on these can't pass as written today.

Additionally, might be worth exposing `assertion_consumer_service` as well if you think that's useful.

supermarine1377: Thanks for catching this — confirmed against a live org (GET /api/v2/saml_configurations): entity_id comes back as  https://<region>.datadoghq.com/account/saml/metadata.xml and sso_url as the org-specific login URL, exactly as you described. 

Addressed in 8fbb1b3:
- Updated the entity_id / sso_url descriptions to match the SP-side semantics (aligned with the API client's field docs).
- The tests no longer assert equality with the uploaded values; they only check the attributes are set.
- Exposed assertion_consumer_service as a computed list — agreed it's useful, since it's exactly what users need to configure their IdP.

fpighi:
@supermarine1377 Cassettes are uploaded.
As an extra step for you, we also require signed commits in this repo. Please ensure they are signed.

supermarine1377:
@fpighi 
Thx for your help.
All comits are rebased to be signed so I force-pushed them.
Now I make this PR to be ready...

clreaume (approved):
approving docs/resources/saml_idp_metadata.md on behalf of docs

supermarine1377:
@clreaume @fpighi 
Hi there! I'm excited to have my first opportunity to contribute to this repository.
It looks like the CI tests require maintainer approval, so could you please trigger them when you get a chance?

fpighi (approved):

PR #4121, opened 2026-08-10
[datadog_custom_allocation_rule] Add based_on_timeseries passthrough support

denozor-dog:
## Summary
 
- Implement opaque JSON passthrough for `strategy.based_on_timeseries` on `datadog_custom_allocation_rule`, fixing silent drops of timeseries allocation payloads on create/update/read. Previously the schema was an empty block and the serializer always sent `nil`.
- Add structural validation for the formulas-and-functions request without enumerating `data_source` values or requiring a top-level `query` field, so compound sources such as `aggregate_augmented_query` (which carry `base_query`/`augment_query` instead) round-trip intact.
- Add a `stringvalidator.OneOf` on `strategy.method` covering all six API values, and give Create/Update distinct error strings — all three call sites previously reported `error retrieving DatadogCustomAllocationRule`, so a CREATE failure surfaced as a retrieval error.
- Add acceptance tests and cassettes for simple metrics queries, aggregate augmented queries, and rename-only updates; update resource/data source docs and example HCL.
## Usage
 
```hcl
strategy {
  method      = "proportional_timeseries"
  granularity = "daily"
 
  based_on_timeseries {
    json = jsonencode({
      response_format = "timeseries"
      queries  = [{ name = "query1", data_source = "metrics", query = "avg:system.cpu.user{*} by {host}" }]
      formulas = [{ formula = "query1" }]
    })
  }
}
```
 
## Why a nested block rather than a plain string attribute
 
The payload is modelled as `based_on_timeseries { json = ... }` rather than `based_on_timeseries = "..."` specifically for backward compatibility. Changing the existing empty block into a string attribute breaks existing users two independent ways, both reproduced against a real rule:
 
1. **Config parse** — `based_on_timeseries {}` becomes `Error: Unsupported block type ... Blocks of type "based_on_timeseries" are not expected here.` This happens before state is read, so it cannot be worked around with a state upgrader.
2. **State decode** — state written by the current released provider holds `{}` at that path, producing `Unable to Read Previously Saved State for UpgradeResourceState ... AttributeName("strategy").AttributeName("based_on_timeseries"): unsupported type json.Delim sent as tftypes.String`.
Keeping it a block avoids both: the path's type is unchanged and adding an attribute to it is tolerated by existing state.
 
## Backward compatibility
 
No action is required from existing users.
 
- Existing `based_on_timeseries {}` blocks continue to parse.
- State written by the current released provider decodes without a schema version bump or `UpgradeState`.
- `json` is `Optional + Computed`, so on first refresh it is populated from the API and reported as informational drift. Plans stay clean.
- Because the value is carried in state, an upgraded provider preserves the payload on writes even if the config has not been updated yet — this closes the existing silent-wipe path without any config change.
- Adding `json = jsonencode(...)` later is purely additive.
Verified end to end: imported a timeseries rule with the released provider (`4.17.0`), swapped in this build with the config unchanged, and confirmed `terraform plan` reports no changes and `terraform apply -refresh-only` populates `json` with the exact stored payload.
 
## Test plan
 
- [x] `make test TESTARGS="-run TestBasedOnTimeseries"` — 21 validator cases plus a round-trip test asserting nothing is dropped, including `is_negated: false`
- [x] `go build ./...` and `go vet ./datadog/fwprovider/`
- [x] Staging rehearsal: applied 10 production-derived `proportional_timeseries` rules on ddstaging; second plan empty; stored payloads byte-identical to source
- [x] `RECORD=false TESTARGS="-run TestAccDatadogCustomAllocationRuleTimeseries" make testacc`
- [x] `RECORD=false TESTARGS="-run TestAccDatadogCustomAllocationRuleTimeseriesAggregateQuery" make testacc`
- [x] `RECORD=false TESTARGS="-run TestAccDatadogCustomAllocationRuleTimeseriesPreservedOnRename" make testacc`
- [x] Backward-compatibility check described above, against a rule created by the released provider
## Related
 
- dd-source companion PR (server-side validation for `based_on_timeseries`): https://github.com/ddoghq/dd-source/pull/49568


rtrieu (approved):

fpighi (approved):

datadog/fwprovider/data_source_datadog_custom_allocation_rule.go
```diff
@@ -143,7 +146,7 @@ func (d *datadogCustomAllocationRuleDataSource) Schema(_ context.Context, _ data
 					},
 					"method": schema.StringAttribute{
 						Computed:    true,
-						Description: "The allocation method. Valid values are `even`, `proportional`, `proportional_timeseries`, or `percent`.",
+						Description: "The allocation method. Valid values are `even`, `even_timeseries`, `percent`, `proportional`, `proportional_timeseries`, or `usage_metric`.",
```
ashleyyip: Even though we might theoretically support `usage_metric` and `even_timeseries` on the backend, they're not really populated as strategies on the UI (and I think might be remnants when we were designing it). Is it possible to keep them removed?

datadog/fwprovider/data_source_datadog_custom_allocation_rule.go
```diff
@@ -253,7 +256,14 @@ func (d *datadogCustomAllocationRuleDataSource) Schema(_ context.Context, _ data
 						},
 					},
 					"based_on_timeseries": schema.SingleNestedBlock{
-						Attributes: map[string]schema.Attribute{},
+						Attributes: map[string]schema.Attribute{
+							"json": schema.StringAttribute{
+								Computed:   true,
+								CustomType: jsontypes.NormalizedType{},
+								Description: "The timeseries query that determines the allocation proportions, encoded as a JSON object. " +
+									"Set when `method` is `proportional_timeseries` or `even_timeseries`.",
```
ashleyyip: same comment, not sure we really support even_timeseries

ashleyyip (commented):
my overall comment is that there are some rule strategies that we don't officially support (and may have undefined behaviour if someone actually tries to implement them) so I think we should just leave them out, especially as they don't exist on the UI. looks good otherwise

ashleyyip (approved):
ccm files look good to me, there are some unrelated changes now in the PR though just for you to confirm

PR #4130, opened 2026-08-11
[tfgen] Coalesce lookup and returned fields

j10czar:
## Summary

Coalesce lookup inputs and returned fields that resolve to the same Terraform attribute instead of emitting duplicate model tags.

## Motivation

By-ID and filtered singular endpoints commonly return the field used for lookup. Treating input and output as separate declarations caused otherwise valid artifacts to fail generation.

## Coverage impact

- Singular: 133/156 → 138/156 (+5)
- Plural: unchanged at 183/194
- Paired: 82/96 → 86/96 (+4)

## Testing

- `make tfgen-test`
- `make fmtcheck`
- `make test`
- Fresh full-spec build-verified sweep

## Stack

- Base: `jason.tenczar/generator-v2/numeric-ids`
- Next: `jason.tenczar/generator-v2/flat-data`

## Reviewer notes

- Required lookups remain required; optional search inputs become optional and computed when also returned by the API.
- Incompatible duplicate shapes still fail with a targeted diagnostic.

### DIFF
--- a/.generator-v2/internal/emit/builder.go
+++ b/.generator-v2/internal/emit/builder.go
@@ -343,6 +343,8 @@ func BuildDataSourceView(a *model.Artifact) (DataSourceView, error) {
 	leafFields := b.models[0].Fields
 
 	inputAttrs, inputFields := buildInputViews(inputLeaves)
+	inputAttrs, inputFields, recordAttrs, leafFields, collisions := coalesceLookupOutputs(inputAttrs, inputFields, recordAttrs, leafFields)
+	b.unsupported = append(b.unsupported, collisions...)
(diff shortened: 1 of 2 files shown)


Supam:
For a coalesced Required field, assignments still overwrites `state.X` with the API-returned value on read, but the field stays `non-Computed`.
If the API ever echoes the lookup key back reformatted (casing, padding, precision), Terraform will flag "provider produced inconsistent result" at apply & read time.  The generator's mocked-schema tests can't catch it. 

Does the API always returns identical `Required` lookup keys? or should these stay non-coalesced (or become `Computed`) ?

Supam (approved):

PR #4158, opened 2026-08-19
[datadog_rum_exclusion_filter] Add resource for RUM exclusion filters

Noham-Martin:
## Summary
- Adds a new `datadog_rum_exclusion_filter` resource for managing RUM exclusion filters (create/read/update/delete/import).
- The underlying `/api/v2/rum/applications/{app_id}/retention_filters/exclusion` endpoint is currently private (`public: false`) in the API spec pending GA, and has no generated SDK bindings yet, so requests are issued via the raw-HTTP `utils.SendRequest` escape hatch (same pattern as `resource_datadog_cloud_inventory_sync_config.go`).
- Known limitation: this resource does not special-case the built-in `error_tracking_exclusion_filter` (which cannot be deleted and only supports updating `enabled`) — that filter isn't manageable via this resource yet.

Manually verified against the `datad0g.com` sandbox with a locally built dev-override provider: full create → read (no diff) → update → destroy cycle succeeded end-to-end.

## Test plan
- [x] `go build ./...` / `go vet ./...`
- [x] Manual create/read/update/destroy cycle against real sandbox API via `terraform apply`/`terraform destroy` with a dev-build override
- [x] `go tool tfplugindocs generate` produced only the expected new doc file, no unintended diffs
- [ ] Acceptance test (`resource_datadog_rum_exclusion_filter_test.go`) — not yet added, can follow up once the API is public and cassette recording is straightforward


maycmlee (approved):

katieed (approved):
lgtm

Supam (approved):

PR #4165, opened 2026-08-20
[datadog_action_execution_policy] Add resource

DMatheo:
## Summary

Adds a new `datadog_action_execution_policy` resource to manage Action Platform execution policies (allow/deny rules for which actions can run against infrastructure, scoped by Kubernetes namespace, script name, or remote shell path, and optionally by Agent-tag targets).

API SPEC : https://github.com/DataDog/datadog-api-spec/pull/6376

Note: This PR currently pins datadog-api-client-go to an unreleased commit (0335221a, post-v2.64.0) — it will need to be re-pinned to a tagged release in a follow-up.


DMatheo:
@codex review

DMatheo:
@codex review

DMatheo:
@codex review

fpighi (approved):

OliviaShoup (approved):

datadog/fwprovider/resource_datadog_action_execution_policy.go
```diff
@@ -0,0 +1,696 @@
...
+	state.Version = types.Int32Value(data.Attributes.Version)
+	state.CreatedAt = types.StringValue(data.Attributes.CreatedAt.Format(time.RFC3339))
+	state.CreatedBy = types.StringValue(data.Attributes.CreatedBy)
+	state.UpdatedAt = types.StringValue(data.Attributes.UpdatedAt.Format(time.RFC3339))
+	state.UpdatedBy = types.StringValue(data.Attributes.UpdatedBy)
+}
+
+// updateExecutionPolicyStateFromResponse rebuilds the whole model from an API response. Used
+// by Read, so that drift and imports both reflect what the API actually holds.
+func updateExecutionPolicyStateFromResponse(ctx context.Context, state *actionExecutionPolicyModel, data *datadogV2.ExecutionPolicyResponseData, diags *diag.Diagnostics) {
```
irving-santiago, line 515: `updateExecutionPolicyStateFromResponse`'s `stateHasExplicitEmptyScope` check infers "user meant scope {}" from prior Terraform state. On terraform import, prior state is always empty, so this guess is wrong: import a scope-less policy, declare scope {}, and you get a spurious diff and apply (version/updated_by bump with no real change).

| Scope in state | Scope declared in config | Result today |
|---|---|---|
| Never set (omitted) | Omitted | No diff |
| Never set (e.g. fresh import) | `scope {}` | Diff shown (phantom) |
| Explicit `scope {}` (created via this resource) | `scope {}` | No diff |
| Explicit `scope {}` | Omitted | Diff shown (phantom, likely) |
| Populated (e.g. `kubernetes` rule) | Omitted or `scope {}` | Diff shown (real change) |

irving-santiago: We could:
- Add a planmodifier.Object on scope. If both plan and state mean "no scope" (null or all-variants-null), force plan to be equal to the state.
- Drop the read-time guess (`stateHasExplicitEmptyScope`), and instead report the literal API value.
- Unit test the modifier (null vs empty, empty vs empty, real scope vs empty as negative case)
- Add a test for import + scope {}

irving-santiago:
Create/Update writes `action_pattern`, `scope`, `target` straight from the plan.

The backend canonicalizes target.agent_tags (sorts, dedupes, reorders) on write. That means that the state right after `apply` already disagrees with what's stored in the BE.

I suggest deriving state from backend data after write, same as `Read()` does. 
We (datadog) already do this for restriction policies. `resource_datadog_restriction_policy.go` uses `UpdateState(ctx, state, resp)` on Read, Create, and Update.

datadog/fwprovider/resource_datadog_action_execution_policy.go
DMatheo: Plan modifiers cannot change a SingleNestedBlock between present and absent. I was thinking about being more restrictive on scope and require it to be non-empty when present, API/imported state with empty scope would be treated and stored with an absent scope. What do you think ?

DMatheo:
@copilot resolve the merge conflicts in this pull request

irving-santiago (approved):

fpighi (approved):

ross-m-young (approved):

Issue #4090, opened 2026-08-03
datadog_team_notification_rule  Provider produced inconsistent result after apply


yshollander-selerity:
### Datadog Terraform Provider Version

4.17.0

### Terraform Version

1.7.5

### What resources or data sources are affected?

datadog_team_notification_rule

### Terraform Configuration Files

```hcl
resource "datadog_team" "team" {
    description = "Team"
    handle = "team"
    name = "Team"
}

locals {
    ddusers = [
        "user1@example.com",
        "user2@example.com",
        "user3@example.com"
    ]
    teams_webhook_url = "https://long_webhook_url"
}

data "datadog_user" "users" {
    for_each = toset(local.ddusers)
    filter = each.value
}

resource "datadog_team_membership" "team_membership" {
    for_each = data.datadog_user.users
    team_id = datadog_team.team.id
    user_id = each.value.id
}

resource "datadog_team_notification_rule" "notification_rule" {
    team_id = datadog_team.team.id
    ms_teams {
        connector_name = local.teams_webhook_url
    }
}
```

### Relevant debug or panic output

datadog_team_notification_rule.selerity_notification_rule: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ datadog_team_notification_rule.notification_rule, provider
│ "provider[\"registry.terraform.io/datadog/datadog\"]" produced an
│ unexpected new value: .email: was absent, but now present.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵

### Expected Behavior

Rule to be created

### Actual Behavior

Terraform reported a bug

### Steps to Reproduce

Try applying the terraform above

jcharry:
Fixed in #4170 
It will go in in the next release.

PR #4170, opened 2026-08-21
[datadog_team_notification_rule] Exclude team notification rule email if not provided in the resource definition

jcharry:
## Motivation

Addresses #4090 

When a team notification rule is created, and the `email` field is omitted, like this:
```
resource "datadog_team_notification_rule" "notification_rule" {
    team_id = datadog_team.team.id
    ms_teams {
        connector_name = local.teams_webhook_url
    }
}
```

The API backing this resource response by injecting `email.enabled = false` into the response, causing a failure.

## Description of changes

The fix is to strip the `email` field when it's not present on the resource, otherwise retain it.


javierciccarelli (approved):
LGTM!

PR #4172, opened 2026-08-25
[datadog_reference_table] Fix inconsistent result after apply when description is empty

### LINKED ISSUE #3786: `datadog_reference_table`: provider returns `null` for `description` when config sets it to `""`, causing "Provider produced inconsistent result after apply"

v4.5.0

### Terraform Version

v1.15.3

datadog_reference_table

```hcl
terraform {
  required_providers {
    datadog = {
      source = "datadog/datadog"
    }
  }
}

resource "datadog_reference_table" "repro" {
  description = ""           # <-- explicit empty string
  source      = "S3"
  table_name  = "repro_table"
  tags        = ["example"]

  file_metadata {
    sync_enabled = true
    access_details {
      aws_detail {
        aws_account_id  = "111111111111"
        aws_bucket_name = "example-bucket"
        file_path       = "example/path.csv"
      }
    }
  }

  schema {
    primary_keys = ["key"]
    fields {
      name = "key"
      type = "STRING"
    }
    fields {
      name = "value"
      type = "STRING"
    }
  }
}
```

```
╷
│
│ When applying changes to datadog_reference_table.repro,
│ provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced
│ an unexpected new value: .description: was cty.StringVal(""), but now null.
│
│ This is a bug in the provider, which should be reported in the
│ provider's own issue tracker.
╵
```

### Expected Behavior

Setting `description = ""` in config should either:
1. Result in a state value of `""` (matching the config and satisfying Terraform's apply-consistency contract), **or**
2. Be rejected at plan time with a validation error if the API does not accept empty strings.

### Actual Behavior

`terraform plan` succeeds with `description = ""` and `terraform apply` then fails with `Provider produced inconsistent result after apply` because the provider's `Read`/`Apply` returns `null` for `.description` while the planned value is `cty.StringVal("")`.

This makes the workspace unappliable until every affected resource's `description = ""` is either:
- removed from config (which then triggers a diff every plan, because state still differs), or
- replaced with a non-empty placeholder.

### Steps to Reproduce

1. `terraform init`
2. `terraform apply` against the MCVE above.

### Important Factoids

_No response_

### References

https://github.com/DataDog/terraform-provider-datadog/pull/3658

dee-see:
Setting `description = ""` on `datadog_reference_table` fails the apply with "Provider produced inconsistent result after apply". The read path turns the API's empty string back into null, so the state doesn't match what's in the config.

The fix is to only fall back to null when the attribute was actually left out, by checking `state.Description.IsNull()` as well. An explicit `""` now stays `""`, and leaving it out still gives you null with no drift.

I added `TestAccReferenceTable_EmptyDescription` for both cases (create with empty, then update to a real value). It still needs a cassette recorded with `RECORD=true`, which I can't do without an org, so let me know how you want to handle that.

Fixes #3786


fpighi:
Thanks for this PR. Just a heads up that we are working on being able to re-record VCRs for this particular case in our testing org. Once that's available I'll re-record and push here.

fpighi (approved):

fpighi (approved):

karanbokil (approved):

PR #4178, opened 2026-08-26
[datadog_workflow_automation] Add run as configuration

GabeWM:
### What does this PR do?

Adds public Workflow Automation `run_as` support for owner, service-account, and initiator execution. The resource preserves server-managed identity when configuration is omitted, and read/import plus the data source reconstruct state from `runAsUserMode` and the `runAs` relationship.

### Motivation

Workflow execution identity is available through the public API but was not configurable through the Terraform provider.

### Reviewer notes

- Explicit `owner` clears service-account mode; omitting `run_as` leaves the server-managed value unchanged.
- Legacy `internal_user` responses are not configurable and are normalized to absent Terraform state.
- Pins the generated client from https://github.com/DataDog/datadog-api-client-go/pull/4506; merge this PR after that client PR.
- Initiator availability remains controlled by `WFRunAsInitiator`.

### Review checklist

- [x] Added the appropriate `changelog/` label
- [x] Title prefixed with the resource name
- [x] Schema attributes all carry `Description` fields
- [x] Docs regenerated with `make docs`
- [x] Acceptance tests and cassettes committed


datadog/fwprovider/data_source_datadog_workflow_automation.go
```diff
@@ -79,6 +80,20 @@ func (d *workflowAutomationDatasource) Schema(_ context.Context, request datasou
 				Description: "The spec defines what the workflow does.",
 				CustomType:  jsontypes.NormalizedType{},
 			},
+			"run_as": schema.SingleNestedAttribute{
```
GabeWM, line 83: Yes I know, this is a weird way to do the definition but there is unfortunately no better way. The problem is that we need this field to respect that it is optional, only sent when it changes, and service account includes an ID. This is the best way to do so. 

Luckily the only change to this in the foreseeable future is managed identities which would just be a type field nothing extra.

datadog/fwprovider/resource_datadog_workflow_automation.go
```diff
@@ -102,6 +141,58 @@ func (r *workflowAutomationResource) ImportState(ctx context.Context, request re
 	resource.ImportStatePassthroughID(ctx, frameworkPath.Root("id"), request, response)
 }
 
+func (r *workflowAutomationResource) ValidateConfig(ctx context.Context, request resource.ValidateConfigRequest, response *resource.ValidateConfigResponse) {
```
GabeWM, line 144: ValidateConfig is an optional Terraform Plugin Framework interface for resource-wide, imperative configuration validation.

Basically it allows for stronger validation we previously did not need.

joepeeples (approved):

liubenjamin (approved):

mathispesch (approved):
lgtm for on-call

OctaviPascual (approved):

datadog/fwprovider/resource_datadog_workflow_automation.go
```diff
@@ -323,6 +526,70 @@ func apiResponseToWorkflowAutomationResourceModel(workflow *datadogV2.GetWorkflo
...
+
+func apiWorkflowRunAsToModel(runAsUserMode *datadogV2.WorkflowRunAsUserMode, relationships *datadogV2.WorkflowDataRelationships) (types.Object, error) {
+	if runAsUserMode == nil {
+		return types.ObjectNull(workflowAutomationRunAsAttributeTypes), fmt.Errorf("workflow response is missing runAsUserMode")
+	}
+
+	runAsType := string(*runAsUserMode)
+	if runAsType != "owner" && runAsType != "service_account" && runAsType != "initiator" {
+		return types.ObjectNull(workflowAutomationRunAsAttributeTypes), fmt.Errorf("workflow response has unsupported runAsUserMode %q", runAsType)
+	}
```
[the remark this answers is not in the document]
GabeWM: `internal_user` is only supported for a handful of org 2 workflows none of which are managed in terraform. This will not change.

amarshall-dd (approved):

PR #2869, opened 2025-02-25
[datadog_dashboard] Add `clustering_pattern_field_path` and `group_by` for ListStream Widget

ecdatadog:
Fixes #2759

PR #4183, opened 2026-08-30
[datadog_dashboard] Fix list_stream clustering_pattern_field_path and group_by not read into state

### LINKED ISSUE #4149: [datadog_dashboard] list_stream clustering_pattern_field_path and group_by are never read into state, causing a perpetual diff
`datadog_dashboard` declares `clustering_pattern_field_path` and `group_by` on
`list_stream_definition.request.query` and sends both on write, but
`buildTerraformListStreamWidgetRequests` never maps either back into state. State records
them as absent while the API holds the correct values, so every plan re-adds them and every
apply is a no-op that never converges.

Introduced by #2869 (v3.57.0), which added the schema and `buildDatadogListStreamRequests`
but not the read path. Still present on `master`.

### Evidence

In `datadog/resource_datadog_dashboard.go` the string `clustering_pattern_field_path` occurs
exactly twice — the schema and the write path — with no third occurrence:

- v3.85.0: schema L5942, write L6037
- `master`: schema L6404, write L6502

The read function (`buildTerraformListStreamWidgetRequests`, L9387 on v3.85.0 / L10022 on
`master`) maps only `response_format`, `columns`, `indexes`, `query_string`, `storage`,
`data_source`, `event_size` and `sort`.

The API round-trips correctly — a `GET` on the dashboard returns both fields:

```json
"query": {
  "clustering_pattern_field_path": "message",
  "data_source": "logs_pattern_stream",
  "group_by": [{"facet": "service"}, {"facet": "status"}],
  "query_string": "..."
}
```

So this is purely the provider's read path, not the API.

### Reproduce

1. Apply a `datadog_dashboard` with a `list_stream_definition` request whose query sets
   `data_source = "logs_pattern_stream"`, `clustering_pattern_field_path`, and one or more
   `group_by` blocks.
2. Apply succeeds.
3. Plan again — the same three attributes are re-added. Indefinitely.

**Expected:** an empty plan after apply.
**Actual:** `1 to change` on every plan, forever. Applying it changes nothing server-side.

### Versions

- Provider 3.85.0, and `master` (4.18.0)
- Terraform 1.15.8

### Suggested fix

Two blocks in `buildTerraformListStreamWidgetRequests`, after the `event_size` block and
before `sort`:

```go
if clusteringPatternFieldPath, ok := q.GetClusteringPatternFieldPathOk(); ok {
      queryRequest["clustering_pattern_field_path"] = clusteringPatternFieldPath
}
if v, ok := q.GetGroupByOk(); ok {
      terraformGroupBys := make([]map[string]interface{}, len(q.GetGroupBy()))
      for i, groupBy := range *v {
              terraformGroupBys[i] = map[string]interface{}{
                      "facet": groupBy.GetFacet(),
              }
      }
      queryRequest["group_by"] = terraformGroupBys
}
```

The pointer assignment mirrors how `query_string` and `storage` are handled a few lines
above, and the group-by loop mirrors the existing flatten at ~L11197 in the same file.
`ListStreamGroupByItems` carries only a facet, so nothing further needs mapping.

Happy to open a PR for this.

lemon0333:
### What does this PR do?

Fixes the `datadog_dashboard` `list_stream` read path so `clustering_pattern_field_path` and `group_by` converge after apply.

Both attributes are declared on `list_stream_definition.request.query` and sent on write, but `buildTerraformListStreamWidgetRequests` (the read/flatten path) never mapped either back into state. State recorded them as absent while the API held the correct values, so every `terraform plan` re-added them and every apply was a no-op that never converged.

This adds the two missing flatten mappings, placed after `event_size` / before `sort` to mirror the adjacent fields in the same function. The write path is unchanged.

### Description of the Change

In `buildTerraformListStreamWidgetRequests` (`datadog/resource_datadog_dashboard.go`):

```go
if clusteringPatternFieldPath, ok := q.GetClusteringPatternFieldPathOk(); ok {
    queryRequest["clustering_pattern_field_path"] = clusteringPatternFieldPath
}
if groupBy, ok := q.GetGroupByOk(); ok && groupBy != nil && len(*groupBy) > 0 {
    terraformGroupBys := make([]map[string]interface{}, len(*groupBy))
    for i, g := range *groupBy {
        terraformGroupBys[i] = map[string]interface{}{
            "facet": g.GetFacet(),
        }
    }
    queryRequest["group_by"] = terraformGroupBys
}
```

This was introduced by #2869 (v3.57.0), which added the schema and `buildDatadogListStreamRequests` (write) but not the read path.

### Test coverage

Added `datadog/resource_datadog_dashboard_list_stream_unit_test.go`, a unit test of the `buildTerraformListStreamWidgetRequests` flatten round-trip (no cassette / no live API needed):

- `TestBuildTerraformListStreamWidgetRequests` — a `logs_pattern_stream` query with `clustering_pattern_field_path` + multiple `group_by` items flattens both back into the state map.
- `TestBuildTerraformListStreamWidgetRequestsOmitsUnsetFields` — an `event_stream` query that sets neither leaves both keys absent (no spurious diff in the other direction).

Locally: `go build ./...`, `go vet ./datadog/...`, and `gofmt -l` are clean; both unit tests pass. The existing cassette-based `list_stream` acceptance tests were not modified; they can't be recorded offline, so no cassette re-record is required for this change.

### Related issues

Fixes #4149

### DIFF
--- a/datadog/resource_datadog_dashboard.go
+++ b/datadog/resource_datadog_dashboard.go
@@ -10051,6 +10051,18 @@ func buildTerraformListStreamWidgetRequests(datadogListStreamRequests *[]datadog
 		if eventSize, ok := q.GetEventSizeOk(); ok && q.GetDataSource() == datadogV1.LISTSTREAMSOURCE_EVENT_STREAM {
 			queryRequest["event_size"] = eventSize
 		}
+		if clusteringPatternFieldPath, ok := q.GetClusteringPatternFieldPathOk(); ok {
+			queryRequest["clustering_pattern_field_path"] = clusteringPatternFieldPath
+		}
+		if groupBy, ok := q.GetGroupByOk(); ok && groupBy != nil && len(*groupBy) > 0 {
+			terraformGroupBys := make([]map[string]interface{}, len(*groupBy))
+			for i, g := range *groupBy {
+				terraformGroupBys[i] = map[string]interface{}{
+					"facet": g.GetFacet(),
+				}
+			}
+			queryRequest["group_by"] = terraformGroupBys
+		}
 		if v, ok := q.GetSortOk(); ok {
 			terraformSort := map[string]interface{}{}
 			if v, ok := v.GetColumnOk(); ok {
--- a/datadog/resource_datadog_dashboard_list_stream_unit_test.go
+++ b/datadog/resource_datadog_dashboard_list_stream_unit_test.go
@@ -1,0 +1,74 @@
+package datadog
+
+import (
+	"testing"
+
+	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV1"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestBuildTerraformListStreamWidgetRequests verifies that the list_stream read
+// path flattens clustering_pattern_field_path and group_by from the API response
+// back into Terraform state. Both fields have schema and write paths but were
+// never mapped back on read, producing a perpetual diff.
+// Regression test for https://github.com/DataDog/terraform-provider-datadog/issues/4149.
+func TestBuildTerraformListStreamWidgetRequests(t *testing.T) {
+	query := datadogV1.NewListStreamQuery(datadogV1.LISTSTREAMSOURCE_LOGS_PATTERN_STREAM, "service:foo")
+	query.SetClusteringPatternFieldPath("message")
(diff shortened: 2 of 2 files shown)


Jiawen23 (approved):

fpighi:
@lemon0333 could you please sign commits in this PR? Happy to merge afterwards.

PR #4186, opened 2026-08-31
[datadog_datastore] Fix inconsistent result when primary_key_generation_strategy is omitted

### LINKED ISSUE #3610: datadog_datastore: Inconsistent result with omitted `primary_key_generation_strategy`

v4.1.0

### Terraform Version

OpenTofu 1.10.2

datadog_datastore

```hcl
resource "datadog_datastore" "ConfigDatastore" {
  name = "[TableFlip] Configured Tables"
  primary_column_name = "table"
  description = <<-EOT
    Configured reference tables for TableFlip to process.
    * Primary key `table` is the name of the source reference table.
    * `target_cols` is an array of secondary column name(s) to invert key-space onto.
  EOT
}
```

```console
datadog_datastore.ConfigDatastore: Creating...
╷
│ 
│ When applying changes to datadog_datastore.ConfigDatastore, provider
│ "provider[\"registry.opentofu.org/datadog/datadog\"]" produced an
│ unexpected new value: .primary_key_generation_strategy: was null, but now
│ cty.StringVal("none").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
```

### Expected Behavior

Resource was created with `primary_key_generation_strategy: none`, or `primary_key_generation_strategy` is marked required if indeed required by the provider

### Actual Behavior

Provider error on application (see above)

### Steps to Reproduce

`tofu apply` configuration containing aforementioned resource block

### Important Factoids

Explicitly defining `primary_key_generation_strategy = "none"` on the resource appears to work around this issue, but from the docs it sounds like this shouldn't be necessary.

### References

_No response_

lemon0333:
### Fixes #3610

When `primary_key_generation_strategy` is omitted from a `datadog_datastore` resource, `tofu apply` / `terraform apply` fails with:

```
.primary_key_generation_strategy: was null, but now cty.StringVal("none")
```

#### Root cause

The attribute is `Optional` with a documented default of `none`, but nothing supplies that default at plan time. The plan therefore holds `null`, while apply reads the value back from the API as `none` — Terraform flags the mismatch as an inconsistent result.

#### Fix

Mark the attribute `Optional` + `Computed` and add `Default: stringdefault.StaticString("none")`. The plugin framework now resolves the planned value to `none` when the config omits it, so the plan matches the post-apply state. Users can still set `uuid` (or any value) explicitly. `updateState` already maps the API's `primary_key_generation_strategy` back consistently, so no read-path change is needed.

#### Testing

- Added an offline unit test (`datadog/fwprovider/resource_datadog_datastore_unit_test.go`) asserting the schema attribute is Optional+Computed and defaults to `none`.
- `go build ./...`, `go vet ./datadog/...`, and `gofmt -l` are clean.
- Acceptance tests for this resource run against recorded cassettes and are unchanged.

Docs are unchanged: an Optional+Computed attribute continues to render under the "Optional" section.

### DIFF
--- a/datadog/fwprovider/resource_datadog_datastore.go
+++ b/datadog/fwprovider/resource_datadog_datastore.go
@@ -8,6 +8,7 @@ import (
 	frameworkPath "github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
@@ -74,6 +75,8 @@ func (r *datastoreResource) Schema(_ context.Context, _ resource.SchemaRequest, 
 			},
 			"primary_key_generation_strategy": schema.StringAttribute{
 				Optional:    true,
+				Computed:    true,
+				Default:     stringdefault.StaticString("none"),
 				Description: "Can be set to `uuid` to automatically generate primary keys when new items are added. Default value is `none`, which requires you to supply a primary key for each new item.",
 			},
 			"created_at": schema.StringAttribute{
--- a/datadog/fwprovider/resource_datadog_datastore_unit_test.go
+++ b/datadog/fwprovider/resource_datadog_datastore_unit_test.go
@@ -1,0 +1,45 @@
+package fwprovider
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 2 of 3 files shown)


hestonhoffman (approved):

lemon0333:
Pushed `4d110875` regenerating `docs/resources/datastore.md` — `tfplugindocs` now appends `Defaults to "none".` for `primary_key_generation_strategy` once the `stringdefault.StaticString("none")` default was added, and CI's `make check-docs` was failing on that uncommitted diff. `linter-checks`/`all-checks` should be green on the re-run.

The remaining `changelog` check is red only because it needs a `changelog/*` label, which external contributors can't apply — could a maintainer add one (e.g. `changelog/note`) when you get a chance? The PR title already matches the required `^\[datadog_\w+\] .+` format. Thanks!

fpighi (approved):

fpighi:
Thanks for your PR @lemon0333. We require signed commits in this repo, could you please sign your commits? Happy to merge afterwards.

PR #4192, opened 2026-09-01
datadog_datastore_item: fix the ignored item key filter and JSON column rendering

vhphan-dd:
Two bugs in `datadog_datastore_item`. Each is independent and has its own commit.

### 1. The item listing was never filtered

The data source and the resource both filtered with `ItemKey`, which the generated client sends as the query parameter `item_key`. The API reads `itemKey` and ignores unknown parameters, so no filter was applied — the call returned the whole datastore and the code took `items[0]`.

Measured against a 30-item datastore:

```
GET .../items?itemKey=apm    ->  1 row  (apm)
GET .../items?item_key=apm   -> 30 rows (items[0] = aaa)
```

So the data source resolved every `item_key` to the first row in the store, with no error and a clean plan. The resource's `Read` never detected an out-of-band deletion, because some other item was always there to find.

Fixed by using `filter`, which the API reads and which matches on the primary key, and by selecting the item whose primary key column equals the requested key instead of trusting the position in the response.

### 2. A JSON column was unreadable

A column of the datastore's JSON type arrives as a Go map or slice. `updateState` rendered every column with `fmt "%v"`, so it reached Terraform as Go map syntax:

```
map[tags:[a b] threshold:10 window:map[seconds:300]]
```

No Terraform function can parse that — `jsondecode` fails with `invalid character 'm' looking for beginning of value` — so the only workaround was to store the column as a JSON string.

Fixed by rendering composite columns with `encoding/json`. Strings still pass through unquoted and scalars keep their `fmt` rendering, so no existing configuration changes.

### Test

`make test`: 337 pass, 0 fail. New unit tests cover key selection (including a non-string key and a key not in the listing) and column rendering.

Verified against the live API with a locally built provider, reading two real datastores:

```
a_apm_row_key           = "apm"      # was "aaa"
f_rum_row_key           = "rum"      # was "aaa"
d_nested_object_column  = "{\"tags\":[\"a\",\"b\"],\"threshold\":10,\"window\":{\"seconds\":300}}"
e_nested_object_decoded = 300        # jsondecode of that column, previously an error
```

`make docs` regenerated only the `value` description.

### DIFF
--- a/datadog/fwprovider/data_source_datadog_datastore_item.go
+++ b/datadog/fwprovider/data_source_datadog_datastore_item.go
@@ -107,9 +107,10 @@ func (d *datastoreItemDataSource) Read(ctx context.Context, request datasource.R
 	datastoreID := state.DatastoreID.ValueString()
 	itemKey := state.ItemKey.ValueString()
 
-	// Use ListDatastoreItems with item_key query parameter to get specific item
-	optionalParams := datadogV2.NewListDatastoreItemsOptionalParameters()
-	optionalParams.ItemKey = &itemKey
+	// Filter with `filter`, not `ItemKey`: the client sends the latter as
+	// `item_key` while the API reads `itemKey`, so it is ignored and the
+	// response is the whole datastore.
+	optionalParams := datadogV2.NewListDatastoreItemsOptionalParameters().WithFilter(itemKey)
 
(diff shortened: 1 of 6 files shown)


vhphan-dd:
Split into two independent PRs, one per bug: #4193 (ignored item key filter) and #4194 (JSON column rendering).

PR #4195, opened 2026-09-01
[TF Generator] (v2) Resolve the resource CRUD lifecycle

Supam:
Implements T046 and T052: a tracked resource operation now builds an Artifact
carrying its four CRUD SDK calls.

- buildArtifact dispatches on ArtifactKind before Cardinality. Cardinality
  describes how a data source resolves records and says nothing about a
  resource, which always maps to exactly one.
- New internal/model/lifecycle.go resolves Create/Read/Update/Delete. bodyCall
  carries GoRequestType from Operation.RequestRefName for the two roles that
  send a body; Read and Delete use the shared sdkCall, renamed from readCall
  now that it serves all four roles.
- requireResolvedRoles rejects a resource whose group does not resolve Create,
  Read or Delete, and any role the annotation declares but that matches no
  operation. It runs before any sub-builder, because Create and Read are
  preconditions of the schema merge rather than of the lifecycle. The
  pre-existing dangling group.search check in buildSingularBothArtifact now
  reports through the same typed MissingRoleError.
- An undeclared update sets LifecycleBindings.UpdateUnsupported and records a
  warning; a declared-but-dangling update fails instead, since treating a typo
  as an absence would silently make every later change destructive.

Artifact.Schema is deliberately nil for a resource until the request/response
merge lands (FR-034, T120-T124). The CLI still skips kind=resource, so nothing
consumes a resource artifact yet.

Generated output and run reports are byte-identical to the parent commit for
both annotated corpus slices.

PR #4187, opened 2026-08-31
[TF Generator] (v2) Resolve tracking-group operationIds to operations

Supam:
### Changes 
Implements an OperationGroup decoder that wires a tracking group's create/read/search/update/delete operationIds to the *model.Operation.
- internal/parser/group.go: ResolveOperationGroups, a whole-spec pass run by LoadSpec before NormalizeSchemas 
- model.ResolvedGroup replaces the single-purpose Operation.SearchOp field, which becomes the nil-safe SearchOp() accessor over it. Group wiring now has one source of truth instead of an index rebuilt per pass.
- A dangling operationId leaves its role nil and lands in Unresolved; BuildArtifact turns each into a per-artifact warning rather than failing the run. 
- NormalizeSchemas walks ResolvedGroup.Operations() instead of re-indexing operationIds, and now also normalizes the annotated operation itself, closing a latent gap where a group naming only other operations left the annotated one's own request/response trees nil.

PR #4196, opened 2026-09-01
[TF Generator] (v2) Derive the SDK API receiver with the Go generator's rule

Supam:
## Motivation

The Datadog Go SDK derives an API receiver name with `utils.class_name` from the generator bundled inside the pinned module:

```python
def class_name(value):
    value = re.sub(r'[^a-zA-Z0-9]', '', value)
    return value + "Api"
```

It **strips** non-alphanumerics and does **not** re-capitalize — it relies on OpenAPI tags already being PascalCase. tfgen was instead using the *TypeScript* generator's `tag_to_class_name` rule: replace non-alphanumerics with a space, then capitalize each word.

The two agree on every Title Case Datadog tag, which is why this survived unnoticed. They diverge the moment a tag word starts lower-case, where the old rule named a Go type the SDK never generated — and nothing in the pipeline would have caught it, because no test compiles generated output.

## Changes

Before: `tagToClassName` word-broke on non-alphanumerics and capitalized each word, and `SDKCall.GoApiStruct`'s doc comment documented that (wrong) rule. A wrong derivation was invisible until someone compiled the generated provider.

Now:

- **`model.SdkClassName`** replaces `tagToClassName`, in `internal/model/identifier.go` beside the other four ports of the same two Python modules (`SnakeCase`, `UpperFirst`, `SdkName`, `EscapeReservedKeyword`). It is expressed as `patternNonAlphanumeric.ReplaceAllString(tag, "") + "Api"` — the regexp literal *is* the Python character class — following that file's existing one-var-per-upstream-pattern convention. It folds in the `"Api"` suffix, as `class_name` does, so a caller cannot omit it.
- **`SDKCall.GoApiStruct`'s doc comment** now describes the rule the code follows, in the struct's own 3-line house shape rather than restating the function's doc.
- **`SnakeCase`** gained the upstream citation its three neighbours already had.
- **Kept correct, not just corrected.** `sdkbinding.Load` already computed the pinned receiver name and used it solely as a `HasSuffix` filter before discarding it. `methodSignature` now retains it, and `corroborate` compares it against `SdkClassName(op.Tag)` on **every** generate — so this divergence class surfaces as a per-operation warning instead of waiting on a human re-reading `utils.py` (FR-005a step 2). It warns rather than fails, since the pinned SDK may legitimately lag the spec.

Two things found on the way, worth a reviewer's attention:

- Nothing rejects an empty `Operation.Tag` — `parser.firstTag` returns `""` — despite the spec's validation table recording it as parser-enforced. An empty tag derives the useless receiver `Api` and the package `datadog`. The new corroboration deliberately **skips** an untagged operation rather than report the derived name when the real problem is the missing tag. The guard itself is not added here.
- The recorded SDK pin in the spec's plan was two versions behind what `go list -m` resolves. That matters here because every identifier translator is a port of the generator *inside* that module, so the version and the naming rules move together.

## Testing

- Test tables repinned to `class_name` output, with the divergent cases (`org groups` → `orggroupsApi`, `cloud-cost.management` → `cloudcostmanagementApi`) as regression pins.
- **Differentially verified against the upstream Python**: extracted `PATTERN_*`, `snake_case`, `upperfirst`, `camel_case` and `class_name` from `$(go list -m -f '{{.Dir}}' github.com/DataDog/datadog-api-client-go/v2)/.generator/src/generator/utils.py` and re-ran them over every expectation. All 15 `SdkName` cases and all 7 `SdkClassName` cases agree exactly.
- Corroborated that the emitted receiver exists in the pinned SDK: `type CloudCostManagementApi datadog.Service`.
- Full generator suite green; `gofmt` and `go vet` clean.
- **Generated output and normalized run reports are byte-identical to the parent commit** for both annotated corpus slices — the meaningful check, since this changes a naming rule. No new corroboration warning fires, i.e. the derived receiver already matched the pinned SDK everywhere.

<details>
<summary>Corpus verification output</summary>

```
IDENTICAL: generated .go/.tf unchanged
=== any new corroboration warning? ===
none — derived receiver matches the pinned SDK on every corpus operation
  aws_cur_config: report IDENTICAL
  integration_elastic_cloud_account: report IDENTICAL
```

</details>

### DIFF
--- a/.generator-v2/internal/model/artifact.go
+++ b/.generator-v2/internal/model/artifact.go
@@ -3,7 +3,6 @@ package model
 import (
 	"fmt"
 	"strings"
-	"unicode"
 )
 
 // BuildArtifact wraps a tracked Operation into an *Artifact and resolves its SDK
@@ -369,7 +368,7 @@ func SDKPackageForPath(path string) string {
 func sdkCall(op *Operation, aliasTerminalID bool) *SDKCall {
 	call := &SDKCall{
 		GoPackage:      SDKPackageForPath(op.Path),
-		GoApiStruct:    tagToClassName(op.Tag) + "Api",
+		GoApiStruct:    SdkClassName(op.Tag),
 		GoMethod:       op.OperationId,
 		GoResponseType: op.ResponseRefName,
 	}
@@ -561,21 +560,4 @@ func versionSegment(path string) string {
 		}
 	}
 	return ""
-}
-
-// tagToClassName converts an OpenAPI tag into the datadog-api-client-go API
-// struct base name: non-alphanumeric runs become word breaks, each word is
-// capitalized on its first rune, and in-word casing is preserved. So "org
-// groups" → "OrgGroups" and "APM" → "APM". This deliberately differs from
-// SdkName, which lower-cases acronyms ("APM" → "Apm").
-func tagToClassName(tag string) string {
-	var b strings.Builder
-	for _, word := range strings.FieldsFunc(tag, func(r rune) bool {
-		return !unicode.IsLetter(r) && !unicode.IsNumber(r)
-	}) {
-		runes := []rune(word)
-		b.WriteRune(unicode.ToUpper(runes[0]))
-		b.WriteString(string(runes[1:]))
-	}
-	return b.String()
 }
--- a/.generator-v2/internal/model/artifact_test.go
+++ b/.generator-v2/internal/model/artifact_test.go
@@ -8,14 +8,6 @@ import (
 )
 
 var _ = Describe("BuildArtifact", func() {
-	DescribeTable("tagToClassName capitalizes each word and preserves in-word casing",
-		func(tag, want string) { Expect(tagToClassName(tag)).To(Equal(want)) },
-		Entry("a single word passes through", "Incidents", "Incidents"),
-		Entry("a space-separated tag joins into PascalCase", "org groups", "OrgGroups"),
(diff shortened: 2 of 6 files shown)


Supam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4196?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4203** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4196** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4196?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4195** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4195?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4187** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4187?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

PR #4201, opened 2026-09-02
Reduce calls to the SDS API

vhourdel:
### Description

This PR introduces a local cache in front of the Sensitive Data Scanner config API. Each SDS resource fetches the whole config to read its current state, leading to a very high number of calls made towards the API depending on the number of SDS resources.

This cache allows using a single call as the source of truth for the current remote state of many groups/rules, instead of fetching the same data over and over again. It is invalidated on create/update/delete so that the provider refetches the data in case of changes.

For a config of around 1500 rules, this reduces the refresh time from ~48 seconds to ~3 seconds.


Supam (approved):

datadog/resource_datadog_sensitive_data_scanner_rule.go
```diff
@@ -507,10 +509,12 @@ func resourceDatadogSensitiveDataScannerRuleDelete(ctx context.Context, d *schem
 	if err != nil {
 		// API returns 404 when the specific rule id doesn't exist through DELETE request.
 		if httpResp != nil && httpResp.StatusCode == 404 {
+			apiInstances.InvalidateSensitiveDataScannerConfigCache()
```
isabella-garza-datadog, line 512: I’m a little worried that as these functions change and gain new exit paths, it’ll be easy to forget to call `InvalidateSensitiveDataScannerConfigCache()` and accidentally return stale data.
What do you think about adding `defer apiInstances.InvalidateSensitiveDataScannerConfigCache()` near the top of each function (maybe after locking the mutex)? We might make an unnecessary GET after a failed request but I think it will help to make the invalidation harder to miss.

isabella-garza-datadog (commented):
I noticed the group order resource (defined [here](https://github.com/DataDog/terraform-provider-datadog/blob/4d2579202c9bf5a39f4b95777be9966cd5e30a12/datadog/fwprovider/resource_datadog_sensitive_data_scanner_group_order.go)) and data source (defined [here](https://github.com/DataDog/terraform-provider-datadog/blob/4d2579202c9bf5a39f4b95777be9966cd5e30a12/datadog/fwprovider/data_source_datadog_sensitive_data_scanner_group_order.go)) are still calling `ListScanningGroups` directly, so they won’t use the cached snapshot with the other SDS resources. Also, `ReorderScanningGroups` updates the configuration but doesn’t currently invalidate the cache.

What do you think about using `ListSensitiveDataScannerGroups` for these reads and invalidating the cache after reordering?

datadog/resource_datadog_sensitive_data_scanner_rule.go
vhourdel: Yes that seems like a very good idea

vhourdel:
> I noticed the group order resource (defined [here](https://github.com/DataDog/terraform-provider-datadog/blob/4d2579202c9bf5a39f4b95777be9966cd5e30a12/datadog/fwprovider/resource_datadog_sensitive_data_scanner_group_order.go)) and data source (defined [here](https://github.com/DataDog/terraform-provider-datadog/blob/4d2579202c9bf5a39f4b95777be9966cd5e30a12/datadog/fwprovider/data_source_datadog_sensitive_data_scanner_group_order.go)) are still calling `ListScanningGroups` directly, so they won’t use the cached snapshot with the other SDS resources. Also, `ReorderScanningGroups` updates the configuration but doesn’t currently invalidate the cache.
> 
> What do you think about using `ListSensitiveDataScannerGroups` for these reads and invalidating the cache after reordering?

@isabella-garza-datadog So I've looked into this, and the issue is that those are not using the same provider instance as the other resources, meaning we can't easily reuse what has been done in this PR without changing a lot of generic code used throughout the codebase.

As far as I can tell, we should be okay keeping those on the non-cached part, and not invalidating the cache there:
- Rule and Group resources only write to rule/group-specific endpoints, not the whole config
- Read operations on these resources is order-independent, we use the ID to find which JSON object corresponds to our local resource

In such a case, the staleness of the order of groups in the config should not matter.

Another thing is the group order data source/resource should not be used more than once per config, so its impact on the number of calls is very limited.

What do you think?

datadog/resource_datadog_sensitive_data_scanner_rule.go
vhourdel: Addressed in followup commits for rules and groups

isabella-garza-datadog:
> @isabella-garza-datadog So I've looked into this, and the issue is that those are not using the same provider instance as the other resources, meaning we can't easily reuse what has been done in this PR without changing a lot of generic code used throughout the codebase.
> 
> As far as I can tell, we should be okay keeping those on the non-cached part, and not invalidating the cache there:
> 
> * Rule and Group resources only write to rule/group-specific endpoints, not the whole config
> * Read operations on these resources is order-independent, we use the ID to find which JSON object corresponds to our local resource
> 
> In such a case, the staleness of the order of groups in the config should not matter.
> 
> Another thing is the group order data source/resource should not be used more than once per config, so its impact on the number of calls is very limited.
> 
> What do you think?

@vhourdel I see, yeah that does seem more complicated. I think your approach makes sense. Thanks for looking into it!

isabella-garza-datadog (approved):

PR #1483, opened 2022-07-05
[datadog_security_monitoring_rule] Dynamic Criticality Support

pietrodll:
# What does this PR do?
Add support for dynamic criticality via the `decrease_criticality_based_on_env` rule option, which is translated to `decreaseCriticalityBasedOnEnv` in the API payload.

# Update process
After changing code and tests, I recorded cassettes and generated documentation with `make docs`. It changed files not related to this feature, so I reverted the changes with git.
I didn't have to update the Go API client as it is a recent version.

PR #4205, opened 2026-09-02
[datadog_security_monitoring_rule] Only read decrease_criticality_based_on_env for log_detection rules

### LINKED ISSUE #4204: [datadog_security_monitoring_rule] decrease_criticality_based_on_env is read but never written for non-log_detection rules, making them unmanageable

v4.19.0. The same code path is present on `master` (8e7f65a8).

### Terraform Version

v1.9.6

- `datadog_security_monitoring_rule`

### Description

`options.decrease_criticality_based_on_env` is written to the API only for `log_detection` rules, but it is read back into state for every rule type. Any rule that is not `log_detection` and has the option enabled in Datadog therefore plans a change that can never be applied, and the resource cannot be managed at all.

In `datadog/fwprovider/resource_datadog_security_monitoring_rule.go`, the write path is gated on the rule type:

```go
if !tfOptions.DecreaseCriticalityBasedOnEnv.IsNull() && !tfOptions.DecreaseCriticalityBasedOnEnv.IsUnknown() &&
    ruleType == string(datadogV2.SECURITYMONITORINGRULETYPECREATE_LOG_DETECTION) {
    payloadOptions.SetDecreaseCriticalityBasedOnEnv(tfOptions.DecreaseCriticalityBasedOnEnv.ValueBool())
}
```

The read path in `extractTfOptions` is not:

```go
if decreaseCriticalityBasedOnEnv, ok := options.GetDecreaseCriticalityBasedOnEnvOk(); ok {
    tfOptions.DecreaseCriticalityBasedOnEnv = types.BoolValue(*decreaseCriticalityBasedOnEnv)
}
```

The attribute is `Optional + Computed` with `Default: booldefault.StaticBool(false)`, so an omitted attribute plans as a known `false` rather than unknown. For an `application_security` rule whose stored value is `true`, Terraform plans `false`, sends nothing, reads back `true`, and fails.

The gate has been on the write path since #1483 and carried over to the plugin framework migration in #3716. The read path has never matched it.

### Expected Behavior

Either the rule is manageable, or the attribute is left alone for rule types the provider does not write it for. A practitioner should not end up with a resource that fails on every apply.

### Actual Behavior

```

When applying changes to
module.registration_velocity_single_ip[0].datadog_security_monitoring_rule.this,
provider "provider[\"registry.terraform.io/datadog/datadog\"]" produced an
unexpected new value: .options[0].decrease_criticality_based_on_env: was
cty.False, but now cty.True.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

The plan is `true -> false` on the attribute and nothing else. It repeats on every run. There is no configuration that resolves it, because `false` is also the schema default. The only workaround is to turn the option off outside Terraform.

### Steps to Reproduce

1. Create an `application_security` rule in the Datadog UI with "decrease criticality based on env" enabled.
2. Adopt it with an `import` block into a `datadog_security_monitoring_rule` resource that does not set `decrease_criticality_based_on_env`.
3. `terraform apply`. The import succeeds and the apply fails with the error above.

### Important Factoids

Datadog stores this option for `application_security` rules but never returns it as an explicit `false`. Across 3142 rules in our org, `application_security` rules hold either no value or `true`, while `log_detection` rules are the only type that hold an explicit `false`. That matches the write gate: rules the provider creates never carry the option, and rules created in the UI can.

The same asymmetry exists in the SDKv2 helper at `datadog/resource_datadog_security_monitoring_rule.go`, which now backs the `datadog_security_monitoring_rules` data source. A data source reporting the real API value is correct, so that one looks fine as is.

### References

- #1483 added the attribute and the write gate
- #3716 migrated the resource to the plugin framework

dgdelahera:
Fixes #4204.

`options.decrease_criticality_based_on_env` is sent to the API for `log_detection` rules only, but it was read back into state for every rule type. A rule of any other type that has the option enabled in Datadog plans a `true -> false` change, sends nothing, reads back `true`, and fails:

```
  .options[0].decrease_criticality_based_on_env: was cty.False, but now cty.True.
```

The plan repeats on every run and no configuration resolves it, since `false` is also the schema default. Adopting such a rule with an `import` block leaves it unmanageable.

The write gate has been there since #1483 and carried over to the framework migration in #3716. This change gates the read on the same rule type, and for the types the provider does not write the option for it keeps the configured value rather than the API value. That covers both directions: a configured `false` survives a remote `true`, and a configured `true` survives the API dropping the option on create.

The attribute description already documents the option as "Only available when the rule type is `log_detection`", so this makes the read path match both the write path and the documented behaviour.

### Testing

Added `TestSecurityMonitoringRuleDecreaseCriticalityReadMatchesWrite` covering the four cases: `log_detection` follows the API, and `application_security` keeps a configured `false`, defaults to `false` with no prior state, and keeps a configured `true`. The three `application_security` subtests fail without the fix.

- `make test` passes (337 tests)
- `make fmtcheck` and `make vet` pass
- `make docs && make check-docs` reports no generated documentation changes

No cassettes were re-recorded: the change is in the response-to-state mapping and I do not have sandbox credentials for the acceptance suite. Happy to add acceptance coverage if you would prefer it.

### Note for maintainers

I cannot set labels on this PR. It needs `changelog/bugfix` per the contributing guide.

### DIFF
--- a/datadog/fwprovider/resource_datadog_security_monitoring_rule_test.go
+++ b/datadog/fwprovider/resource_datadog_security_monitoring_rule_test.go
@@ -1,0 +1,78 @@
+package fwprovider
+
+import (
+	"context"
+	"testing"
+
+	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV2"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// newRuleOptionsWithDecreaseCriticality builds rule options carrying
+// decreaseCriticalityBasedOnEnv, mimicking a rule that has the option enabled
+// in Datadog.
+func newRuleOptionsWithDecreaseCriticality(decrease bool) datadogV2.SecurityMonitoringRuleOptions {
+	options := datadogV2.NewSecurityMonitoringRuleOptions()
+	options.SetDecreaseCriticalityBasedOnEnv(decrease)
+	return *options
+}
+
+// TestSecurityMonitoringRuleDecreaseCriticalityReadMatchesWrite guards the fix
+// for the "inconsistent result after apply" error on rules that are not
+// log_detection. buildPayloadOptions only sends
+// decrease_criticality_based_on_env for log_detection rules, so reading the API
+// value back for any other type plans a change the provider can never apply.
+func TestSecurityMonitoringRuleDecreaseCriticalityReadMatchesWrite(t *testing.T) {
+	ctx := context.Background()
+	logDetection := string(datadogV2.SECURITYMONITORINGRULETYPECREATE_LOG_DETECTION)
+	applicationSecurity := string(datadogV2.SECURITYMONITORINGRULETYPECREATE_APPLICATION_SECURITY)
+
+	// log_detection is the one type the provider writes the option for, so the
+	// API value is authoritative.
+	t.Run("log_detection follows the api", func(t *testing.T) {
+		got, diags := extractTfOptions(ctx, newRuleOptionsWithDecreaseCriticality(true), nil, logDetection)
+		if diags.HasError() {
+			t.Fatalf("unexpected diagnostics: %v", diags)
+		}
+		if !got[0].DecreaseCriticalityBasedOnEnv.ValueBool() {
+			t.Fatal("expected decrease_criticality_based_on_env to follow the API value true")
+		}
+	})
+
+	// Importing an application_security rule that has the option enabled must not
+	// overwrite the configured false, which is a value the provider cannot change.
+	t.Run("application_security keeps configured false", func(t *testing.T) {
+		prior := []ruleOptionsModel{{DecreaseCriticalityBasedOnEnv: types.BoolValue(false)}}
+		got, diags := extractTfOptions(ctx, newRuleOptionsWithDecreaseCriticality(true), prior, applicationSecurity)
+		if diags.HasError() {
+			t.Fatalf("unexpected diagnostics: %v", diags)
+		}
+		if got[0].DecreaseCriticalityBasedOnEnv.ValueBool() {
+			t.Fatal("expected decrease_criticality_based_on_env to stay false, the API value is not manageable for this rule type")
+		}
+	})
+
+	// A fresh import has no prior state, so the schema default applies.
+	t.Run("application_security without prior state defaults to false", func(t *testing.T) {
+		got, diags := extractTfOptions(ctx, newRuleOptionsWithDecreaseCriticality(true), nil, applicationSecurity)
+		if diags.HasError() {
+			t.Fatalf("unexpected diagnostics: %v", diags)
+		}
+		if got[0].DecreaseCriticalityBasedOnEnv.ValueBool() {
+			t.Fatal("expected decrease_criticality_based_on_env to default to false")
+		}
+	})
+
+	// The same holds in the other direction: the API drops the option when it
+	// creates an application_security rule, and the configured value must survive.
+	t.Run("application_security keeps configured true", func(t *testing.T) {
+		prior := []ruleOptionsModel{{DecreaseCriticalityBasedOnEnv: types.BoolValue(true)}}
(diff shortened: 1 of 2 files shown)


datadog/fwprovider/resource_datadog_security_monitoring_rule_test.go
```diff
@@ -0,0 +1,78 @@
...
+			t.Fatalf("unexpected diagnostics: %v", diags)
+		}
+		if got[0].DecreaseCriticalityBasedOnEnv.ValueBool() {
+			t.Fatal("expected decrease_criticality_based_on_env to default to false")
+		}
+	})
+
+	// The same holds in the other direction: the API drops the option when it
+	// creates an application_security rule, and the configured value must survive.
+	t.Run("application_security keeps configured true", func(t *testing.T) {
```
maldogant, line 68: I'm not sure if this should be the intended behaviour as only `log_detection` should currently support `decrease_criticality_based_on_env`.

Maybe it should be changed to:

```
  func TestDecreaseCriticalityRequiresLogDetection(t *testing.T) {
  	model := securityMonitoringRuleResourceModel{
  		Type: types.StringValue("application_security"),
  		Options: []ruleOptionsModel{{
  			DecreaseCriticalityBasedOnEnv: types.BoolValue(true),
  		}},
  	}

  	diags := validateDecreaseCriticalityRuleType(&model)

  	if !diags.HasError() {
  		t.Fatal(
  			"expected decrease_criticality_based_on_env=true " +
  				"to be rejected for application_security",
  		)
  	}
  }
```

maldogant (commented):
Thanks for the fix proposal!

Small comment about the intended behaviour of non-log based rules.

datadog/fwprovider/resource_datadog_security_monitoring_rule_test.go
dgdelahera: Thanks for the review! I looked into this and I do not think the option is `log_detection-only`, so I would like to push back on the premise before reworking the tests.

Four things point the other way:
- `decreaseCriticalityBasedOnEnv` is defined on the shared `SecurityMonitoringRuleOptions` model in the API client, not on a log_detection-specific options type, and the spec puts no rule type restriction on it.
- `PUT /api/v2/security_monitoring/rules/{rule_id}` accepts and persists an explicit false on application_security rules. I tested this on three rules and each came back false with the version incremented.
- The UI exposes the setting for span event rules. spans and app_sec_spans are dataSource values on a standard rule query rather than rule types, so the option sits at the options level, not the type level.
- The write gate arrived with #1483 in July 2022, and that PR gives no API constraint as the reason. It reads like the scope of the feature at the time.

If that holds, this PR fixes the asymmetry in the wrong direction. It hardens a 2022 limitation into the read path and leaves the option unmanageable for every other rule type. I would rather do the opposite, and I am happy to rework the branch:

1. Remove the `log_detection` condition on the write path, so the option is sent for any rule type.
2. Drop `Default: booldefault.StaticBool(false)` from the schema and add `boolplanmodifier.UseStateForUnknown()`. This is the actual root cause. The static default is why an omitted attribute plans as a known false instead of unknown, which is what produces "inconsistent result after apply". Without it the attribute takes the API value and the error is unreachable for any rule type. It also matches `evaluation_window` and `max_signal_duration`, which are already Optional + Computed with no default and `UseStateForUnknown()`.
3. Revert the read gate, so` extractTfOptions` passes the API value through as before and loses the ruleType parameter.
4. Fix the attribute description, which still says "Only available when the rule type is log_detection".

That ends up a smaller diff than what is here now, and the option becomes manageable everywhere instead of silently preserved.

Two questions before I push:

- Does any rule type actually reject the field server side? `workload_security` and `workload_activity` are the ones I cannot test.
- Are you comfortable with removing the schema default? It is user visible: a rule whose remote value is true while the config omits the attribute will move from false to true in state on the next refresh. I think that is the correct value, but it is a plan change for existing users.

PR #4210, opened 2026-09-02
Support sensitivity on data quality monitor options

pkalmakis:
## What

Exposes `sensitivity` on the `data_quality_query.monitor_options` block of `datadog_monitor`.

`MonitorFormulaAndFunctionDataQualityMonitorOptions` has carried a `Sensitivity *float64` field since `datadog-api-client-go` v2.64.1, which this provider already depends on. The provider just never surfaced it, so the width of the anomaly detection bounds was not configurable from Terraform even though the API accepted it.

**No client bump needed** — the field is already in the pinned client.

## Changes

- `sensitivity` (optional, `TypeFloat`) on `data_quality_query.monitor_options`
- `getOptionalFloat` helper, and `buildDataQualityMonitorOptions` sets the field
- `buildTerraformDataQualityMonitorVariables` reads it back on refresh
- Docs regenerated

Zero is treated as unset, mirroring how `getOptionalString` treats `""`. Sensitivity is a multiplier on bound width, must be positive, and defaults to `3.0` server side, so zero is unambiguously "not configured" rather than a meaningful value being dropped.

## Testing

New `datadog/resource_datadog_monitor_unit_test.go` — the package had no unit coverage for these builders. It covers `buildDataQualityMonitorOptions` (unset, empty-value, sensitivity-only, all-fields) and a build → read round trip through `buildMonitorFormulaAndFunctionDataQualityQuery`. Both tests fail without the change and pass with it.

`go test ./datadog/` passes, and the five existing `TestAccDatadogMonitor_DataQuality_*` acceptance tests still replay green against their current cassettes.

No acceptance test exercises `sensitivity` itself, and this is a deliberate gap rather than an oversight.

Cassette matching compares request bodies, so adding `sensitivity` to the test config invalidates `TestAccDatadogMonitor_DataQuality_WithMonitorOptions.yaml` (`requested interaction not found`) until it is re-recorded. Nothing in CI records provider cassettes: `test.yml` replays with `RECORD: false` and `test_pr_integration.yml` runs live with `RECORD: "none"`. Recording is a local task needing test-org credentials, which I could not obtain. Committing the test change without a fresh cassette would just make this PR red, so it is left out.

**For whoever can record**, the whole job is these two edits plus `RECORD=true` on that one test:

```hcl
# testAccCheckDatadogDataQualityMonitorWithOptions, inside monitor_options
        model_type_override = "freshness"
+       sensitivity         = 2.5
```

```go
// TestAccDatadogMonitor_DataQuality_WithMonitorOptions, in the Check block
+					resource.TestCheckResourceAttr(
+						"datadog_monitor.data_quality_with_options", "variables.0.data_quality_query.0.monitor_options.0.sensitivity", "2.5"),
```

Worth knowing while you are in there: the five existing `TestAccDatadogMonitor_DataQuality_*` cassettes were recorded against **org 2 (us1 prod)**, not the frog test org that `isTestOrg()` defaults to. Across the whole cassette directory it is 620 interactions on org 321813 versus 22 on org 2, and those 22 are these five files plus two Datastore tests. Re-recording against frog would fix that drift.

Also unrelated to this change but noticed while running them: these tests are flaky under `t.Parallel()`. `Basic` and `EmptyMonitorOptions` failed together in a full-suite run and both pass in isolation, as did `EmptyGroupBy` on an earlier run. Looks like contention over shared cassette state.

## Related

The remaining two `monitor_options` fields the backend supports, `source_to_target_config` and `model_configuration`, are not in the public API spec yet. Spec PR: https://github.com/DataDog/datadog-api-spec/pull/6591. The provider side of those is stacked on this branch and stays in draft until the generated client ships.

### DIFF
--- a/datadog/resource_datadog_monitor.go
+++ b/datadog/resource_datadog_monitor.go
@@ -1181,10 +1181,18 @@ func getMonitorFormulaQuerySchema() *schema.Schema {
 											Optional:    true,
 											Description: "Crontab expression to override the default schedule.",
 										},
+										// Deliberately not validated against the client enum: the API
+										// accepts values this provider version may not know about yet,
+										// and rejects bad ones with a clear error.
 										"model_type_override": {
 											Type:        schema.TypeString,
 											Optional:    true,
 											Description: "Override for the model type. Valid values are `freshness`, `percentage`, `any`.",
+										},
+										"sensitivity": {
+											Type:        schema.TypeFloat,
+											Optional:    true,
+											Description: "Sensitivity of the anomaly detection model, expressed as a multiplier on the width of the predicted bounds. Higher values widen the bounds and produce fewer alerts; lower values tighten them and produce more alerts. Defaults to `3.0`.",
 										},
 									},
 								},
@@ -1887,6 +1895,13 @@ func getOptionalString(data map[string]interface{}, fieldName string) (string, b
 	return val, ok && len(val) > 0
 }
 
+// getOptionalFloat safely extracts an optional float field from a map. A zero value is
+// treated as unset, mirroring how getOptionalString treats the empty string.
+func getOptionalFloat(data map[string]interface{}, fieldName string) (float64, bool) {
+	val, ok := data[fieldName].(float64)
+	return val, ok && val != 0
+}
+
 // getOptionalStringSlice safely extracts an optional string slice from a map, filtering out nil and empty values
 func getOptionalStringSlice(data map[string]interface{}, fieldName string) []string {
 	list, ok := data[fieldName].([]interface{})
@@ -1931,6 +1946,10 @@ func buildDataQualityMonitorOptions(data map[string]interface{}) *datadogV1.Moni
 	}
 	if v, ok := getOptionalString(data, "model_type_override"); ok {
 		opts.SetModelTypeOverride(datadogV1.MonitorFormulaAndFunctionDataQualityModelTypeOverride(v))
+		hasAnyOption = true
+	}
+	if v, ok := getOptionalFloat(data, "sensitivity"); ok {
+		opts.SetSensitivity(v)
 		hasAnyOption = true
 	}
 
@@ -2708,6 +2727,9 @@ func buildTerraformDataQualityMonitorVariables(datadogVariables []datadogV1.Moni
 				}
 				if modelTypeOverride, ok := monitorOptions.GetModelTypeOverrideOk(); ok {
 					terraformMonitorOptions["model_type_override"] = modelTypeOverride
+				}
+				if sensitivity, ok := monitorOptions.GetSensitivityOk(); ok {
+					terraformMonitorOptions["sensitivity"] = sensitivity
 				}
 				terraformQuery["monitor_options"] = []map[string]interface{}{terraformMonitorOptions}
 			}
--- a/docs/resources/monitor.md
+++ b/docs/resources/monitor.md
@@ -717,6 +717,7 @@ Optional:
 - `custom_where` (String) Custom WHERE clause for the query.
 - `group_by_columns` (List of String) Columns to group results by.
 - `model_type_override` (String) Override for the model type. Valid values are `freshness`, `percentage`, `any`.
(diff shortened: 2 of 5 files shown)


datadog/resource_datadog_monitor.go
```diff
@@ -1181,11 +1181,19 @@ func getMonitorFormulaQuerySchema() *schema.Schema {
...
 										},
+										// Deliberately not validated against the client enum: the API
+										// accepts values this provider version may not know about yet,
+										// and rejects bad ones with a clear error.
 										"model_type_override": {
 											Type:        schema.TypeString,
 											Optional:    true,
 											Description: "Override for the model type. Valid values are `freshness`, `percentage`, `any`.",
 										},
+										"sensitivity": {
```
kevinpombo-datadog, line 1192: Can you also make sure this is applied to https://github.com/DataDog/terraform-provider-datadog/blob/master/datadog/fwprovider/resource_datadog_monitor.go as well?

pkalmakis: Good catch, done in baf60c6f.

`datadog/fwprovider/resource_datadog_monitor.go` has its own copy of the data quality schema, so it needed the same field in four places: the `DataQualityMonitorOptions` model struct, the `monitor_options` schema block, the build path, and the read-back path.

`fwutils` had no float helpers (only string/int32/int64/bool), so I added `SetOptFloat64` and `ToTerraformFloat64` following the existing pattern.

One intentional difference between the two implementations worth flagging: the SDKv2 path reads from an untyped `map[string]interface{}` and so has to treat `0` as "unset", the same convention `getOptionalString` already uses for `""`. The framework path has real null semantics, so it uses `IsNull()` and an explicit `0` stays distinguishable from an omitted attribute. That matches how the surrounding fields already behave in each implementation rather than forcing one convention across both.

Two things I could not verify, flagging rather than leaving implied:

- **The framework path has no acceptance coverage.** Setting `TERRAFORM_MONITOR_FRAMEWORK_PROVIDER=true` and replaying the `TestAccDatadogMonitor_DataQuality_*` tests fails with `Error: Invalid resource type` — the harness doesn't register the framework resource from that env var alone. This reproduces on the branch without my changes, so it's pre-existing, but it does mean the framework code here is compile-checked only.
- **Docs are unchanged** because they generate from the SDKv2 schema and both descriptions are now identical.

Separately, while running these I noticed `TestAccDatadogMonitor_DataQuality_*` is flaky under `t.Parallel()` — `Basic` and `EmptyMonitorOptions` failed in a full-suite run and both pass in isolation, and I saw `EmptyGroupBy` do the same earlier. Looks like contention on shared cassette state rather than anything in this change. Happy to open a separate issue if that's useful.

docs/resources/monitor.md
```diff
@@ -717,6 +717,7 @@ Optional:
 - `custom_where` (String) Custom WHERE clause for the query.
 - `group_by_columns` (List of String) Columns to group results by.
 - `model_type_override` (String) Override for the model type. Valid values are `freshness`, `percentage`, `any`.
+- `sensitivity` (Number) Sensitivity of the anomaly detection model, expressed as a multiplier on the width of the predicted bounds. Higher values widen the bounds and produce fewer alerts; lower values tighten them and produce more alerts. Defaults to `3.0`.
```
buraizu: While the semicolon is grammatically fine here, these are generally used sparingly in favor of two shorter sentences — which scans faster for a two-way behavior contrast like this one.

buraizu (commented):
Thanks for the PR, just left a minor suggestion

docs/resources/monitor.md
buraizu:
```suggestion
- `sensitivity` (Number) Sensitivity of the anomaly detection model, expressed as a multiplier on the width of the predicted bounds. Higher values widen the bounds and produce fewer alerts. Lower values tighten them and produce more alerts. Defaults to `3.0`.
```

buraizu (approved):

pkalmakis:
Closing in favour of #4211, which now carries this change plus the two fields that were waiting on the spec.

This PR was split off only because `source_to_target_config` and `model_configuration` did not exist in the public API spec yet, so `sensitivity` could ship independently. [datadog-api-spec#6591](https://github.com/DataDog/datadog-api-spec/pull/6591) has since merged and the generated client is on master, so the split no longer buys anything and one review is easier than two stacked ones.

@kevinpombo-datadog your fwprovider comment is addressed there, and extended to cover the two new fields as well so they do not drift either. Sorry for the thread move.

PR #4211, opened 2026-09-02
[datadog_monitor] Support sensitivity, source_to_target_config and model_configuration on data quality queries

pkalmakis:
## What

Exposes all three remaining `monitor_options` fields that the data quality backend supports on `datadog_monitor`'s `data_quality_query` block:

| Field | Description |
| --- | --- |
| `sensitivity` | Multiplier on the width of the predicted bounds. Higher widens and alerts less; lower tightens and alerts more. Defaults to `3.0`. |
| `source_to_target_config` | Compares the same measure across two data entities and alerts on the difference. Holds `source` and `target` entity measure configs, plus `diff_type` and `entity_type`. |
| `model_configuration` | Tunes the anomaly detection model: `auto_resolve_days`, `enable_flatline_detection`, `function`, `min_lower_bound_size`, `min_upper_bound_size`, `model_bounds_override`. |

All three are implemented in both the SDKv2 resource and the plugin-framework port, so the two do not diverge as the migration proceeds.

`sensitivity` already existed in the pinned client and was simply never surfaced. The other two did not exist in the public API spec at all, so that had to land first:

- Spec: https://github.com/DataDog/datadog-api-spec/pull/6591 (merged)
- Generated client: https://github.com/DataDog/datadog-api-client-go/commit/406c957e6fba, *"Regenerate client from commit 914e068 of spec repo"*

The client bump this needed has since landed on `master` independently, so this PR no longer touches `go.mod` or `go.sum` and does not wait on a tagged release.

## Design notes

**`enable_flatline_detection` differs between the two implementations, deliberately.** SDKv2 reads from an untyped `map[string]interface{}` where an omitted bool and an explicit `false` are both `false`. Since `false` is a meaningful setting, the attribute is declared `Default: true`, matching the server default: omitting it behaves as before, and an explicit `false` still reaches the API. The framework can represent null, so it needs no default. An omitted attribute is not sent, and `false` is sent as `false`.

**Zero means unset for the numeric fields in SDKv2**, following the existing `getOptionalString`/`""` convention in that file. `sensitivity` must be positive, and `auto_resolve_days` is bounded `minimum: 1` in the spec precisely because the backend treats `0` as "no configuration". The framework path uses real null checks.

**`auto_resolve_days` is range-checked before narrowing to the `int32` the API expects**, since Terraform `TypeInt` is a platform int and a large value would otherwise wrap to a negative day count. The schema rejects one at plan time as well. The framework provider types the attribute as `Int32` and gets both for free.

## Testing

`go build ./...` and `go test ./datadog/` pass.

Unit tests in `datadog/resource_datadog_monitor_unit_test.go`, which the package previously had none of for these builders:

- `TestBuildDataQualityMonitorOptions`: unset, empty-value, sensitivity-only, all-fields
- `TestBuildDataQualitySourceToTargetConfig`: nested source/target, optionals absent when unset
- `TestBuildDataQualityModelConfiguration`: all fields, explicit `false` flatline detection, zero-valued numerics as unset
- Two build to read round trips through `buildMonitorFormulaAndFunctionDataQualityQuery`

Framework-path unit coverage in `datadog/fwprovider/resource_datadog_monitor_test.go`, following the in-package convention the other `fwprovider/*_test.go` files use:

- `TestBuildDataQualityQueryStruct_Sensitivity`: value sent; null omitted rather than sent as zero
- `TestBuildDataQualityQueryStruct_SourceToTargetConfig`: nested source/target, optionals absent when unset
- `TestBuildDataQualityQueryStruct_ModelConfiguration`: all six fields
- `TestBuildDataQualityQueryStruct_FlatlineDetectionTriState`: omitted, explicit `false`, explicit `true`
- `TestBuildDataQualityQueryState_MonitorOptions`: read path
- `TestDataQualityMonitorOptionsRoundTrip`: build to read, no permanent diff

Those tests found two read-path defects, fixed here. `fwutils.ToTerraformStr` type-switches on `*string`, but enum getters return their own named pointer type, so `diff_type`, `function` and `model_bounds_override` fell through to `StringNull` and read back empty, a permanent diff for anyone setting them. And an absent `group_by_columns` was left as a zero-value `types.List` with a nil element type, which is not a valid null and the framework rejects when writing state.

### Acceptance coverage

`sensitivity` is now covered end to end. `testAccCheckDatadogDataQualityMonitorWithOptions` sets `sensitivity = 2.5`, and `TestAccDatadogMonitor_DataQuality_WithMonitorOptions` asserts `monitor_options.0.sensitivity` reads back as `2.5`.

All five `TestAccDatadogMonitor_DataQuality_*` cassettes were re-recorded for this, against the frog test org (`fasjyydbcgwwc2uc`, 321813) that `isTestOrg()` defaults to. Each new cassette carries `"org_id":321813`, and the `WithMonitorOptions` cassette contains `"sensitivity":2.5`, so the value genuinely round-trips through the API rather than only through the builders. Replay passes for all five with no credentials in the environment, confirming they work offline.

### Remaining gaps, stated plainly

**`source_to_target_config` and `model_configuration` have no acceptance coverage**, on either implementation. Both are unit-tested. Cassette matching compares request bodies, so covering them means extending the same test config and re-recording that one cassette, and nothing in CI records provider cassettes (`test.yml` replays with `RECORD: false`, `test_pr_integration.yml` runs live with `RECORD: "none"`). That needs test-org credentials, so it is a reasonable follow-up rather than a blocker.

**The framework path has no acceptance coverage at all**, for the same cassette reason, and relies on the unit tests above.

## Pre-existing issue noticed along the way

**The `TestAccDatadogMonitor_DataQuality_*` tests are flaky under `t.Parallel()`.** Running the five together, `EmptyGroupBy` and `EmptyMonitorOptions` failed with `requested interaction not found`, and both pass in isolation. This looks like contention over shared cassette state. It is not caused by this change and re-recording did not fix it, so the cassettes above were recorded one test at a time.

Separately, `TestAccDatastoreBasic` and `TestAccDatastoreItemBasic` are still recorded against org 2 (8 interactions). They are the last cassettes in the directory that are, now that the data quality five have moved to frog, and they are out of scope here.

## History

Supersedes https://github.com/DataDog/terraform-provider-datadog/pull/4210, which covered `sensitivity` alone while the spec change was still in flight. The framework-provider parity requested in review there is included here.


docs/resources/monitor.md
```diff
@@ -716,8 +716,64 @@ Optional:
...
 - `sensitivity` (Number) Sensitivity of the anomaly detection model, expressed as a multiplier on the width of the predicted bounds. Higher values widen the bounds and produce fewer alerts; lower values tighten them and produce more alerts. Defaults to `3.0`.
+- `source_to_target_config` (Block List, Max: 1) Compare the same measure across two data entities and alert on the difference between them. (see [below for nested schema](#nestedblock--variables--data_quality_query--monitor_options--source_to_target_config))
+
+<a id="nestedblock--variables--data_quality_query--monitor_options--model_configuration"></a>
+### Nested Schema for `variables.data_quality_query.monitor_options.model_configuration`
+
+Optional:
+
+- `auto_resolve_days` (Number) Number of days after which an open alert is automatically resolved. When unset, alerts stay open until the measure returns within bounds.
+- `enable_flatline_detection` (Boolean) Whether to alert when the measure stops changing entirely. Defaults to `true`. Defaults to `true`.
```
estherk15, line 730:
```suggestion
- `enable_flatline_detection` (Boolean) Whether to alert when the measure stops changing entirely. Defaults to `true`.
```

estherk15 (approved):

datadog/fwprovider/resource_datadog_monitor.go
```diff
@@ -264,11 +264,69 @@ type DataQualityQuery struct {
 }
 
 type DataQualityMonitorOptions struct {
```
kevinpombo-datadog, line 266: Can we add test coverage here as well before merging? there are separate tests for each provider

kevinpombo-datadog (approved):

datadog/fwprovider/resource_datadog_monitor.go
pkalmakis: Added in f44ae08a — unit tests in `datadog/fwprovider/resource_datadog_monitor_test.go` covering the data quality monitor options.

They caught two read-path bugs, fixed in the same commit: enum fields (`diff_type`, `function`, `model_bounds_override`) were reading back empty because `ToTerraformStr` only handles `*string`, and an absent `group_by_columns` was an untyped null. Both would have caused permanent diffs.

fpighi (approved):

PR #4212, opened 2026-09-02
[datadog_org_group_policy] Add role policy_type support

saraslee:
## What does this PR do

Adds `policy_type = "role"` to `datadog_org_group_policy`, so Terraform can create and manage shared-role policies (a permission set provisioned into every org in a group), alongside the existing `org_config` policies.

This PR adds:
- `policy_type` validator widened to accept `role`
- Plan-time guard rejecting `enforcement_tier = OVERRIDE_ALLOWED` for `role` policies
- In-place rename support for `policy_name`, restricted to `role` policies (see Design Decisions)
- `Delete()` handling for role policies, which are never hard-deleted server-side
- `role` example, docs, and behavior notes

## Motivation
OMT-589: Terraform CRUD support for shared roles, building on OMT-27's org-group resources. Server-side support landed via datadog-api-spec#6152.

## Impact

Low risk — `role` is a new opt-in `policy_type` value; `org_config` behavior is unchanged (confirmed by re-recording `TestAccDatadogOrgGroupPolicy_Basic`'s cassette against the same staging org and diffing behavior). `role` policies are further gated by the `org_groups_shared_roles` feature flag, staging-only for now.

## Testing
- `go build`/`go vet` pass
- Unit tests pass (`ValidateConfig`, `Delete` guards)
- `TestAccDatadogOrgGroupPolicy_Role` and `TestAccDatadogOrgGroupPolicy_Basic` pass in both record (against a staging org) and replay mode

## Design Decisions
- **Role policies can't be deleted, only disabled.** The API never hard-deletes a `role` policy (`DELETE` 405s). `Delete()` only succeeds once `enforcement_tier = "DELEGATE"` (disabled); otherwise it errors and tells the user to disable it first. On success the resource is dropped from state, but the policy itself isn't removed server-side — this is a deliberate mismatch between Terraform's delete semantics and the API's, matching how disabling works.
- **Disabling is one-way.** Once `DELEGATE`, the API rejects any transition back to `GROUP_MANAGED`/`OVERRIDE_ALLOWED`. `Update()` catches this before calling the API (using the already-fetched prior state) so users get a clear error instead of a raw 400.
- **`policy_name` rename is asymmetric.** The API only supports in-place rename for `role` policies; renaming an `org_config` policy 400s, so it still forces a replace there. The initial spec doc suggested both types supported rename in place — that turned out to be wrong when recording against the live API.
- Reused the existing `datadog_org_group_policy` resource rather than adding a new one, matching the API design where `role` is just another `policy_type` on the same CRUD endpoints.


datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -210,13 +244,29 @@ func (r *OrgGroupPolicyResource) Update(ctx context.Context, request resource.Up
...
 		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "org group policy ID must be a valid UUID"))
 		return
 	}
 
+	if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
+		priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
+		state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
+		response.Diagnostics.AddAttributeError(
+			frameworkPath.Root("enforcement_tier"),
+			"Cannot re-enable a disabled role policy",
```
pgoldrbx: **[Important]** The one-way DELEGATE guard's third condition compares `state.EnforcementTier.ValueString()` against `"DELEGATE"` without first checking for null/unknown. In the Plugin Framework, `ValueString()` returns `""` when the value is null or unknown, so `"" != "DELEGATE"` is `true` and the guard fires even when the user is not trying to change `enforcement_tier`. A content-only update of a DELEGATE role policy (where `enforcement_tier` is null/unknown in the plan) would be incorrectly blocked with "Cannot re-enable a disabled role policy."

The same function already guards against this a few lines down (`if !state.EnforcementTier.IsNull() && !state.EnforcementTier.IsUnknown()`), confirming the attribute can legitimately be null/unknown here. Mirror that check:

```go
// Before:
if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
    priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
    state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {

// After:
if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
    priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
    !state.EnforcementTier.IsNull() && !state.EnforcementTier.IsUnknown() &&
    state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
```

Also consider adding an acceptance-test step that updates only `content` while the policy is in DELEGATE (no `enforcement_tier` change) to verify the guard does not fire:

```go
{
    // Content can still be updated while disabled (DELEGATE).
    Config: testAccCheckDatadogOrgGroupPolicyRoleConfig(orgGroupName, renamedPolicyName,
        fmt.Sprintf(`{"permissions":[%q]}`, permissionID), "DELEGATE"),
    Check: resource.ComposeTestCheckFunc(
        testAccCheckDatadogOrgGroupPolicyExists(providers.frameworkProvider, resourceName),
        resource.TestCheckResourceAttr(resourceName, "enforcement_tier", "DELEGATE"),
    ),
},
```

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -210,13 +244,29 @@ func (r *OrgGroupPolicyResource) Update(ctx context.Context, request resource.Up
...
 	id, err := uuid.Parse(state.ID.ValueString())
 	if err != nil {
 		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "org group policy ID must be a valid UUID"))
 		return
 	}
 
+	if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
+		priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
+		state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
+		response.Diagnostics.AddAttributeError(
```
pgoldrbx: **[Important]** The DELEGATE → non-DELEGATE transition is only caught here in `Update()` (apply time), not at plan time. `ValidateConfig` only sees the proposed config (no prior state), so a user changing `enforcement_tier` from `DELEGATE` back to `GROUP_MANAGED` sees a clean `terraform plan` with an in-place update and no error — the error only surfaces at `terraform apply`. The Plugin Framework provides `ResourceWithModifyPlan` exactly for surfacing state-dependent errors at plan time.

```go
// Add to the interface declarations:
var (
    _ resource.ResourceWithConfigure      = &OrgGroupPolicyResource{}
    _ resource.ResourceWithImportState    = &OrgGroupPolicyResource{}
    _ resource.ResourceWithValidateConfig = &OrgGroupPolicyResource{}
    _ resource.ResourceWithModifyPlan     = &OrgGroupPolicyResource{}
)

// Add the method:
func (r *OrgGroupPolicyResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
    if req.State.Raw.IsNull() { // skip on create
        return
    }
    var priorState OrgGroupPolicyModel
    resp.Diagnostics.Append(req.State.Get(ctx, &priorState)...)
    if resp.Diagnostics.HasError() {
        return
    }
    if priorState.PolicyType.ValueString() != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) {
        return
    }
    if priorState.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        return
    }
    var plan OrgGroupPolicyModel
    resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
    if resp.Diagnostics.HasError() {
        return
    }
    if !plan.EnforcementTier.IsNull() && !plan.EnforcementTier.IsUnknown() &&
        plan.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        resp.Diagnostics.AddAttributeError(
            frameworkPath.Root("enforcement_tier"),
            "Cannot re-enable a disabled role policy",
            "This role policy was disabled (enforcement_tier = \"DELEGATE\") and cannot be transitioned back to GROUP_MANAGED or OVERRIDE_ALLOWED.",
        )
    }
}
```

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

datadog/tests/resource_datadog_org_group_policy_test.go
```diff
@@ -121,14 +220,24 @@ func testAccCheckDatadogOrgGroupPolicyDestroy(accProvider *fwprovider.FrameworkP
...
 				if httpResp != nil && httpResp.StatusCode == 404 {
 					continue
 				}
 				return fmt.Errorf("received an error retrieving org group policy: %w", err)
 			}
 
+			// role policies are never hard-deleted server-side; Delete() only succeeds
+			// once the policy is already disabled (enforcement_tier = DELEGATE), so
+			// "destroyed" means disabled, not 404.
+			if r.Primary.Attributes["policy_type"] == "role" {
```
pgoldrbx: **[Important]** This role-specific DELEGATE branch in `CheckDestroy` is unreachable in the acceptance test. The cassette confirms the cascading-delete sequence: `DELETE /org_groups/... → 204` then `GET /org_group_policies/... → 404` ("org group ... not found"). Since the Go API client returns an error for non-2xx, the `httpResp.StatusCode == 404 → continue` branch above fires first, so this DELEGATE branch (which only runs when `err == nil`) is never reached. The test passes via the 404 path, not the DELEGATE path, so the intended destroy semantics are not actually verified.

The unit test `TestOrgGroupPolicyDelete_RoleAllowsWhenAlreadyDisabled` already verifies `Delete()` succeeds for a DELEGATE policy without an API call, so the lower-effort fix is to remove the dead branch and document the reliance on the cascade:

```go
// Remove the dead branch:
// if r.Primary.Attributes["policy_type"] == "role" {
//     if resp.Data.Attributes.GetEnforcementTier() == "DELEGATE" {
//         continue
//     }
//     return fmt.Errorf("role policy was not disabled on destroy")
// }

// And document:
// Note: for role policies, the org group's cascading delete removes the
// policy server-side, so CheckDestroy sees a 404 and continues. The
// DELEGATE-persistence semantics are verified by the unit test
// TestOrgGroupPolicyDelete_RoleAllowsWhenAlreadyDisabled, not here.
```

Alternatively, add a test step that removes only the policy resource (keeping the org group) and verifies the policy still exists server-side with `enforcement_tier = DELEGATE` — that actually exercises this branch.

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy_test.go
```diff
@@ -0,0 +1,153 @@
...
+		resp := &resource.ValidateConfigResponse{}
+		r.ValidateConfig(context.Background(), req, resp)
+
+		if resp.Diagnostics.HasError() {
+			t.Errorf("did not expect an error for policy_type=role with enforcement_tier=%s: %v", tier, resp.Diagnostics)
+		}
+	}
+}
+
+func TestOrgGroupPolicyValidateConfig_OrgConfigAllowsOverrideAllowed(t *testing.T) {
```
pgoldrbx: **[Important]** The new unit tests cover `Delete` (2) and `ValidateConfig` (3), but three key code paths have no isolated unit test — only acceptance-test coverage that is gated behind `isRecording()`/`isReplaying()` + `skipIfNoCassette(t)` and may be skipped in CI:

1. **Update one-way DELEGATE guard** (returns before any API call, so `r.API` can be nil):

```go
func TestOrgGroupPolicyUpdate_RoleRejectsReEnableFromDelegate(t *testing.T) {
    s := orgGroupPolicyValidateConfigSchema()
    r := &OrgGroupPolicyResource{}
    priorStateRaw := orgGroupPolicyStateValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE),
        string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE))
    planRaw := orgGroupPolicyConfigValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED")
    req := resource.UpdateRequest{
        State: tfsdk.State{Raw: priorStateRaw, Schema: s},
        Plan:  tfsdk.Plan{Raw: planRaw, Schema: s},
    }
    resp := &resource.UpdateResponse{}
    r.Update(context.Background(), req, resp)
    if !resp.Diagnostics.HasError() {
        t.Fatal("expected an error re-enabling a disabled role policy (DELEGATE -> GROUP_MANAGED)")
    }
}
```

2. **`RequiresReplaceIf` plan modifier** — extract the closure to a package-level function and test both branches (role → no replace, org_config → replace):

```go
func TestOrgGroupPolicyNameRequiresReplace_RoleDoesNotForceReplace(t *testing.T) {
    s := orgGroupPolicyValidateConfigSchema()
    stateRaw := orgGroupPolicyStateValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED")
    req := planmodifier.StringRequest{State: tfsdk.State{Raw: stateRaw, Schema: s}}
    resp := &planmodifier.StringRequiresReplaceIfFuncResponse{}
    orgGroupPolicyNameRequiresReplaceIf(context.Background(), req, resp)
    if resp.RequiresReplace {
        t.Fatal("role policy rename should NOT force replace")
    }
}
```

3. **`ValidateConfig` with null `enforcement_tier`** — confirm no false diagnostic when the Optional attribute is absent (add a `orgGroupPolicyConfigValueNullTier` helper that sets `enforcement_tier` to `tftypes.NewValue(tftypes.String, nil)`).

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

datadog/tests/cassettes/TestAccDatadogOrgGroupPolicy_Role.yaml
```diff
@@ -0,0 +1,976 @@
...
+        content_length: 438
+        uncompressed: false
+        body: '{"data":{"id":"01a0682b-786c-70c0-84f9-fd1899c51900","type":"org_group_policies","attributes":{"content":{"permissions":["d99415a4-dc4d-11e8-b4b1-4f9e475593a0","d99443a8-dc4d-11e8-b4b2-9f5f4b0b2f5a"]},"enforcement_tier":"DELEGATE","modified_at":"2026-09-03T16:48:04.933659Z","policy_name":"finance_read_write","policy_type":"role"},"relationships":{"org_group":{"data":{"id":"01a0682b-77c6-7515-95b5-8a4d628a2339","type":"org_groups"}}}}}'
+        headers:
+            Content-Type:
+                - application/vnd.api+json
+        status: 200 OK
+        code: 200
+        duration: 133.895583ms
+    - id: 25
```
pgoldrbx: **[Minor]** Cassette interaction id 25 (this `PATCH → 400 "disabled role policies cannot be re-enabled"`) is stale: the provider's `Update()` guard now returns a diagnostic before any API call, so this PATCH is never sent during replay. go-vcr won't fail on the unused entry (it only fails on unmatched requests), so there's no replay breakage — but the interaction is a maintenance hazard suggesting the provider makes an API call here when it doesn't. Consider removing interaction 25 or re-recording the cassette after the guard was added.

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

docs/resources/org_group_policy.md
```diff
@@ -42,6 +53,14 @@ Auto-created overrides show up in the [`datadog_org_group_policy_overrides`](htt
 
 Changing `enforcement_tier` to `"GROUP_MANAGED"` automatically deletes every override associated with this policy server-side. Any `datadog_org_group_policy_override` resources pointing at this policy must be removed from configuration in the same commit. Otherwise, Terraform's next apply will try to recreate the server-deleted overrides and fail with a `FailedPrecondition` error.
```
pgoldrbx: **[Minor]** The new `### role policies` section documents the enforcement_tier constraints, rename, and delete semantics, but doesn't mention that `role` policies require the `org_groups_shared_roles` feature flag (staging-only as of writing). A production user setting `policy_type = "role"` will hit an opaque API error with no warning from the provider docs. The test file notes this gating in a comment — surface it in the user-facing docs too:

```markdown
### `role` policies

> **Note:** `role` policies require the `org_groups_shared_roles` feature flag,
> which is currently staging-only. Using `policy_type = "role"` against a
> production org will fail at the API level.

`role` policies only support `GROUP_MANAGED` and `DELEGATE` for `enforcement_tier`; ...
```

(The same note should go in `templates/resources/org_group_policy.md.tmpl`.)

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy.go
pgoldrbx: **[Important]** The one-way DELEGATE guard's third condition compares `state.EnforcementTier.ValueString()` against `"DELEGATE"` without first checking for null/unknown.

In the Plugin Framework, `ValueString()` returns `""` when the value is null or unknown, so `"" != "DELEGATE"` is `true` and the guard fires even when the user is not trying to change `enforcement_tier`. A content-only update of a DELEGATE role policy (where `enforcement_tier` is null/unknown in the plan) would be incorrectly blocked with "Cannot re-enable a disabled role policy."

The same function already guards against this a few lines down (`if !state.EnforcementTier.IsNull() && !state.EnforcementTier.IsUnknown()`), confirming the attribute can legitimately be null/unknown here. Mirror that check:

```go
// Before:
if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
    priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
    state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {

// After:
if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
    priorState.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) &&
    !state.EnforcementTier.IsNull() && !state.EnforcementTier.IsUnknown() &&
    state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
```

Also consider adding an acceptance-test step that updates only `content` while the policy is in DELEGATE (no `enforcement_tier` change) to verify the guard does not fire:

```go
{
    // Content can still be updated while disabled (DELEGATE).
    Config: testAccCheckDatadogOrgGroupPolicyRoleConfig(orgGroupName, renamedPolicyName,
        fmt.Sprintf(`{"permissions":[%q]}`, permissionID), "DELEGATE"),
    Check: resource.ComposeTestCheckFunc(
        testAccCheckDatadogOrgGroupPolicyExists(providers.frameworkProvider, resourceName),
        resource.TestCheckResourceAttr(resourceName, "enforcement_tier", "DELEGATE"),
    ),
},
```

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [x] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy.go
pgoldrbx: **[Important]** The DELEGATE → non-DELEGATE transition is only caught here in `Update()` (apply time), not at plan time.

`ValidateConfig` only sees the proposed config (no prior state), so a user changing `enforcement_tier` from `DELEGATE` back to `GROUP_MANAGED` sees a clean `terraform plan` with an in-place update and no error — the error only surfaces at `terraform apply`. The Plugin Framework provides `ResourceWithModifyPlan` exactly for surfacing state-dependent errors at plan time.

```go
// Add to the interface declarations:
var (
    _ resource.ResourceWithConfigure      = &OrgGroupPolicyResource{}
    _ resource.ResourceWithImportState    = &OrgGroupPolicyResource{}
    _ resource.ResourceWithValidateConfig = &OrgGroupPolicyResource{}
    _ resource.ResourceWithModifyPlan     = &OrgGroupPolicyResource{}
)

// Add the method:
func (r *OrgGroupPolicyResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
    if req.State.Raw.IsNull() { // skip on create
        return
    }
    var priorState OrgGroupPolicyModel
    resp.Diagnostics.Append(req.State.Get(ctx, &priorState)...)
    if resp.Diagnostics.HasError() {
        return
    }
    if priorState.PolicyType.ValueString() != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) {
        return
    }
    if priorState.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        return
    }
    var plan OrgGroupPolicyModel
    resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
    if resp.Diagnostics.HasError() {
        return
    }
    if !plan.EnforcementTier.IsNull() && !plan.EnforcementTier.IsUnknown() &&
        plan.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        resp.Diagnostics.AddAttributeError(
            frameworkPath.Root("enforcement_tier"),
            "Cannot re-enable a disabled role policy",
            "This role policy was disabled (enforcement_tier = \"DELEGATE\") and cannot be transitioned back to GROUP_MANAGED or OVERRIDE_ALLOWED.",
        )
    }
}
```

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [x] _reviewed by @phil-gold_ddog_

This sounds like a good idea but I don't know that it's a blocker. It feels like it would make a better experience for terraform users but could be a follow-up.

datadog/tests/resource_datadog_org_group_policy_test.go
pgoldrbx, line 280: **[Important]** This DELEGATE branch in `CheckDestroy` is not exercised by the acceptance test.

The branch handles the case where a user removes only the policy block from config while keeping the org group: `Delete()` returns early (role+DELEGATE, no API call), the policy persists server-side as disabled, and `CheckDestroy` then `GET /org_group_policies/{id}` → 200 with `DELEGATE` → the branch fires and passes. That is the intended "disabled means destroyed" semantics, and the branch should stay.

The current test destroys *both* resources, so Terraform destroys in reverse order (policy, then org group), and the org group's cascading delete removes the policy server-side. By the time `CheckDestroy` runs, `GET /org_group_policies/{id}` returns 404 ("org group ... not found", cassette interaction 28), so the `404 → continue` branch above fires first and this DELEGATE branch is never reached. The test passes via the 404 path, not the DELEGATE path — so the branch's actual contract is unverified.

Add a test step that removes only the policy resource (keeping the org group) and asserts the policy still exists server-side with `enforcement_tier = DELEGATE`:

```go
// Step that removes only the policy, keeping the org group, to exercise
// the DELEGATE branch in CheckDestroy.
Config: fmt.Sprintf(`
resource "datadog_org_group" "role" {
  name = "%s"
}
`, orgGroupName),
Check: resource.ComposeTestCheckFunc(
    // Verify the role policy still exists server-side and is DELEGATE.
    func(s *terraform.State) error {
        // ... GetOrgGroupPolicy(auth, id) and assert
        // resp.Data.Attributes.GetEnforcementTier() == "DELEGATE"
        return nil
    },
),
```

This actually drives `CheckDestroy` down the DELEGATE branch (org group still exists → no cascade → 200 → branch fires). The cassette needs a corresponding `GET /org_group_policies/{id} → 200` interaction with `enforcement_tier: DELEGATE`.

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [x] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy_test.go
pgoldrbx, line 145: **[~Important~ Minor]** The new unit tests cover `Delete` (2) and `ValidateConfig` (3), but three key code paths have no isolated unit test — only acceptance-test coverage that is gated behind `isRecording()`/`isReplaying()` + `skipIfNoCassette(t)` and may be skipped in CI:

1. **Update one-way DELEGATE guard** (returns before any API call, so `r.API` can be nil):

```go
func TestOrgGroupPolicyUpdate_RoleRejectsReEnableFromDelegate(t *testing.T) {
    s := orgGroupPolicyValidateConfigSchema()
    r := &OrgGroupPolicyResource{}
    priorStateRaw := orgGroupPolicyStateValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE),
        string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE))
    planRaw := orgGroupPolicyConfigValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED")
    req := resource.UpdateRequest{
        State: tfsdk.State{Raw: priorStateRaw, Schema: s},
        Plan:  tfsdk.Plan{Raw: planRaw, Schema: s},
    }
    resp := &resource.UpdateResponse{}
    r.Update(context.Background(), req, resp)
    if !resp.Diagnostics.HasError() {
        t.Fatal("expected an error re-enabling a disabled role policy (DELEGATE -> GROUP_MANAGED)")
    }
}
```

2. **`RequiresReplaceIf` plan modifier** — extract the closure to a package-level function and test both branches (role → no replace, org_config → replace):

```go
func TestOrgGroupPolicyNameRequiresReplace_RoleDoesNotForceReplace(t *testing.T) {
    s := orgGroupPolicyValidateConfigSchema()
    stateRaw := orgGroupPolicyStateValue(
        string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED")
    req := planmodifier.StringRequest{State: tfsdk.State{Raw: stateRaw, Schema: s}}
    resp := &planmodifier.StringRequiresReplaceIfFuncResponse{}
    orgGroupPolicyNameRequiresReplaceIf(context.Background(), req, resp)
    if resp.RequiresReplace {
        t.Fatal("role policy rename should NOT force replace")
    }
}
```

3. **`ValidateConfig` with null `enforcement_tier`** — confirm no false diagnostic when the Optional attribute is absent (add a `orgGroupPolicyConfigValueNullTier` helper that sets `enforcement_tier` to `tftypes.NewValue(tftypes.String, nil)`).

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking)** Downgraded to minor. I don't think this is a blocker at all. The acceptance tests are what really matter.

datadog/tests/cassettes/TestAccDatadogOrgGroupPolicy_Role.yaml
pgoldrbx, line 844: **[Minor]** Cassette interaction id 25 (this `PATCH → 400 "disabled role policies cannot be re-enabled"`) is stale: the provider's `Update()` guard now returns a diagnostic before any API call, so this PATCH is never sent during replay. go-vcr won't fail on the unused entry (it only fails on unmatched requests), so there's no replay breakage — but the interaction is a maintenance hazard suggesting the provider makes an API call here when it doesn't. Consider removing interaction 25 or re-recording the cassette after the guard was added.

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [x] _reviewed by @phil-gold_ddog_

docs/resources/org_group_policy.md
pgoldrbx: **[Minor]** The new `### role policies` section documents the enforcement_tier constraints, rename, and delete semantics, but doesn't mention that `role` policies require the `org_groups_shared_roles` feature flag (staging-only as of writing). A production user setting `policy_type = "role"` will hit an opaque API error with no warning from the provider docs. The test file notes this gating in a comment — surface it in the user-facing docs too:

```markdown
### `role` policies

> **Note:** `role` policies require the `org_groups_shared_roles` feature flag,
> which is currently staging-only. Using `policy_type = "role"` against a
> production org will fail at the API level.

`role` policies only support `GROUP_MANAGED` and `DELEGATE` for `enforcement_tier`; ...
```

(The same note should go in `templates/resources/org_group_policy.md.tmpl`.)

---

🤖 Written by GLM-5.2 (ai-gw-baseten)

- [ ] _reviewed by @phil-gold_ddog_

pgoldrbx (commented):
General clean addition of `policy_type = "role"` — the schema validation, one-way DELEGATE guard, and soft-delete semantics are well-structured and the docs are thorough.

I've left a few blocking issues (a false-positive in the Update DELEGATE guard, a plan-time validation gap, an unexercised CheckDestroy branch, and missing unit tests for two key code paths) plus a couple of minor cleanups.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Good catch — `ValueString()` returns `""` for both null and unknown, so the guard was firing incorrectly whenever `enforcement_tier` wasn't a known concrete value. Added `!IsNull() && !IsUnknown()` checks before the comparison in `Update()`. Fixed in f427e9c6.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Agreed this is worth doing given the API only 400s at apply. Added `ModifyPlan()`: when prior state is a `role` policy at `DELEGATE` and the new plan tries to move `enforcement_tier` to anything else, it now errors at `terraform plan` instead of failing at apply. Fixed in f427e9c6.

datadog/tests/resource_datadog_org_group_policy_test.go
saraslee: Added a step to `TestAccDatadogOrgGroupPolicy_Role` that removes the policy from config (keeping the org group), which exercises `Delete()` succeeding without an API call. Since the resource drops out of state after that step, captured its ID via closure during the prior DELEGATE step and used it to directly call `GetOrgGroupPolicy` afterward, confirming the policy is still `DELEGATE` server-side. Re-recorded the cassette against staging and verified in replay. Fixed in f427e9c6.

datadog/fwprovider/resource_datadog_org_group_policy_test.go
saraslee: Added all 3 requested unit tests: reject re-enable from DELEGATE, `policy_name` rename does not force replace for `role` (and does for `org_config`), and a null `enforcement_tier` produces no validation diagnostic. All passing. Fixed in f427e9c6.

datadog/tests/cassettes/TestAccDatadogOrgGroupPolicy_Role.yaml
saraslee: That interaction was recording the now-dead PATCH→400 call from the old (buggy) guard in the first comment — since the guard now short-circuits before any API call, it was unreachable. Removed it, and confirmed go-vcr doesn't match on interaction `id` so no renumbering was needed. Cassette was re-recorded fresh for the new DELEGATE-branch test coverage and contains no stale re-enable interaction. Fixed in f427e9c6.

hestonhoffman (approved):

datadog/tests/resource_datadog_org_group_policy_test.go
```diff
@@ -121,14 +266,24 @@ func testAccCheckDatadogOrgGroupPolicyDestroy(accProvider *fwprovider.FrameworkP
...
```
pgoldrbx, line 280: **[Important]** Dead branch: the `policy_type == "role"` case in `CheckDestroy` never executes.

The last `Steps` entry swaps in `testAccCheckDatadogOrgGroupPolicyRoleOnlyGroupConfig`, which drops `datadog_org_group_policy.role` from config. That step's own apply already calls `Delete()` and removes the resource from state — so by the time `CheckDestroy` takes its snapshot, there's nothing left to iterate over. This contradicts the "Fixed in f427e9c6" claim; the invariant is only actually verified today via the separate `testAccCheckDatadogOrgGroupPolicyStillDisabled` direct-API check.

**Fix:** keep the DELEGATE-tier policy in config through the last step, move `testAccCheckDatadogOrgGroupPolicyStillDisabled` into that step's `Check`, and let the framework's own implicit end-of-run destroy remove it *after* `CheckDestroy` runs. Delete the now-unused `testAccCheckDatadogOrgGroupPolicyRoleOnlyGroupConfig` helper.

```go
{
    // DELEGATE disables the shared role.
    Config: testAccCheckDatadogOrgGroupPolicyRoleConfig(orgGroupName, renamedPolicyName, fmt.Sprintf(`{"permissions":[%q,%q]}`, permissionID, otherPermissionID), "DELEGATE"),
    Check: resource.ComposeTestCheckFunc(
        testAccCheckDatadogOrgGroupPolicyExists(providers.frameworkProvider, resourceName),
        resource.TestCheckResourceAttr(resourceName, "enforcement_tier", "DELEGATE"),
        testAccCheckDatadogOrgGroupPolicyStillDisabled(providers.frameworkProvider, &roleResourceID),
    ),
},
...
{
    ResourceName:      resourceName,
    ImportState:       true,
    ImportStateVerify: true,
},
// no further steps — policy stays in state so CheckDestroy's
// role/DELEGATE branch fires before the framework's final teardown.
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -117,6 +124,70 @@ func (r *OrgGroupPolicyResource) ImportState(ctx context.Context, request resour
...
+		priorState.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
+		return
+	}
+
+	var plan OrgGroupPolicyModel
+	response.Diagnostics.Append(request.Plan.Get(ctx, &plan)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+	if !plan.EnforcementTier.IsNull() && !plan.EnforcementTier.IsUnknown() &&
```
pgoldrbx: **[Important]** This guard doesn't check `response.RequiresReplace` before firing, so it can misfire on a legitimate replace. `policy_name` and `policy_type` are both plan-modified attributes; a config edit that changes `policy_type` away from `"role"` on an existing role+DELEGATE resource is a legitimate destroy/create, but `plan.EnforcementTier` here is populated from the *new* (replacement) resource's config, not carried forward — so a fresh `enforcement_tier = "GROUP_MANAGED"` on the new resource incorrectly trips this "cannot re-enable" error.

```go
// before
var plan OrgGroupPolicyModel
response.Diagnostics.Append(request.Plan.Get(ctx, &plan)...)
if response.Diagnostics.HasError() {
    return
}
if !plan.EnforcementTier.IsNull() && !plan.EnforcementTier.IsUnknown() &&
    plan.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
    response.Diagnostics.AddAttributeError(...)
}

// after
if len(response.RequiresReplace) > 0 {
    // Attribute-level plan modifiers already forced a destroy/create for
    // this plan; the old role policy is being removed, not "re-enabled".
    return
}
var plan OrgGroupPolicyModel
response.Diagnostics.Append(request.Plan.Get(ctx, &plan)...)
if response.Diagnostics.HasError() {
    return
}
if !plan.EnforcementTier.IsNull() && !plan.EnforcementTier.IsUnknown() &&
    plan.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
    response.Diagnostics.AddAttributeError(...)
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -210,13 +281,30 @@ func (r *OrgGroupPolicyResource) Update(ctx context.Context, request resource.Up
...
+	var priorState OrgGroupPolicyModel
+	response.Diagnostics.Append(request.State.Get(ctx, &priorState)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
 
 	id, err := uuid.Parse(state.ID.ValueString())
 	if err != nil {
 		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "org group policy ID must be a valid UUID"))
 		return
```
pgoldrbx, line 336: **[Minor] (Pre-existing)** The `policy_type == "role"` + `enforcement_tier == OVERRIDE_ALLOWED` invariant is only checked in `ValidateConfig`. That check is skipped by `terraform plan -out=file; terraform apply file`, which applies a saved plan directly without re-running `ValidateConfig`. `Update()` already has `state.PolicyType`/`state.EnforcementTier` in scope here (and `Create()` has the equivalent gap), so the same check can be added cheaply as a backstop, ideally extracted into a helper shared with `ValidateConfig`. Not a correctness risk — the API itself 400s on the combination — but it turns a raw API error into a proper Terraform diagnostic for the saved-plan workflow.

```go
if state.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
    state.EnforcementTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_OVERRIDE_ALLOWED) {
    response.Diagnostics.AddAttributeError(
        frameworkPath.Root("enforcement_tier"),
        "Invalid enforcement_tier for policy_type \"role\"",
        "role policies only support GROUP_MANAGED and DELEGATE.",
    )
    return
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) recommendation:** save this for future cleanup

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -102,6 +108,7 @@ func (r *OrgGroupPolicyResource) Schema(_ context.Context, _ resource.SchemaRequ
 				Validators: []validator.String{
 					stringvalidator.OneOf(
 						string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ORG_CONFIG),
+						string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE),
 					),
 				},
 				PlanModifiers: []planmodifier.String{
```
pgoldrbx, line 115: **[Minor] (Pre-existing)** `policy_type` already carries `stringplanmodifier.RequiresReplace()` here (pre-existing, unchanged by this PR), so no production fix is needed — but no test asserts that changing `policy_type` on an existing resource forces a replace. Given this PR doubles `policy_type`'s value space (adding `"role"`), that gap is worth closing.

```go
// acceptance step, mirroring the existing policy_name rename-forces-replace pattern
{
    Config: testAccCheckDatadogOrgGroupPolicyConfig(orgGroupName, policyName, content, enforcementTier), // policy_type changed
    ConfigPlanChecks: resource.ConfigPlanChecks{
        PreApply: []plancheck.PlanCheck{
            plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionReplace),
        },
    },
},
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) recommendation:** consider addressing this in a follow-up?

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -263,6 +354,19 @@ func (r *OrgGroupPolicyResource) Delete(ctx context.Context, request resource.De
 		return
 	}
 
+	// role policies are never hard-deleted server-side (API returns 405). If it's already
+	// disabled (enforcement_tier = DELEGATE), there's nothing left for the API to do; just
+	// drop it from state. Otherwise reject and tell the user to disable it first.
+	if state.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) {
+		if state.EnforcementTier.ValueString() != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
+			response.Diagnostics.AddError(
```
pgoldrbx: **[Minor]** `Delete()`'s role-policy error uses `AddError` with no attribute path, unlike `ModifyPlan`/`Update()`, which both use `AddAttributeError(frameworkPath.Root("enforcement_tier"), ...)` for the same error class. Scoping it the same way gives a more precise diagnostic in `terraform destroy` output.

```go
// before
response.Diagnostics.AddError(
    "role policies cannot be deleted",
    `Set enforcement_tier = "DELEGATE" to disable this role policy before removing it from your configuration.`,
)

// after
response.Diagnostics.AddAttributeError(
    frameworkPath.Root("enforcement_tier"),
    "role policies cannot be deleted",
    `Set enforcement_tier = "DELEGATE" to disable this role policy before removing it from your configuration.`,
)
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) possible improvement?**: as I understand it, adding the attribute scope helps make it clear that the delete fails because of the `enforcement_tier` constraint.

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -210,13 +281,30 @@ func (r *OrgGroupPolicyResource) Update(ctx context.Context, request resource.Up
...
+		return
+	}
 
 	id, err := uuid.Parse(state.ID.ValueString())
 	if err != nil {
 		response.Diagnostics.Append(utils.FrameworkErrorDiag(err, "org group policy ID must be a valid UUID"))
 		return
 	}
 
+	if priorState.PolicyType.ValueString() == string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) &&
```
pgoldrbx: **[Minor]** This DELEGATE re-enable guard (predicate + error text) is duplicated verbatim from `ModifyPlan`'s copy at line 161. They're already diverging in required behavior (see the `ModifyPlan`/`RequiresReplace` comment above), which is exactly the drift risk that motivates extracting a shared helper now.

```go
// roleDelegateReEnableDiagnostic returns a non-nil diagnostic if transitioning a
// disabled (DELEGATE) role policy's enforcement_tier away from DELEGATE is being
// attempted; both ModifyPlan and Update must reject this one-way transition.
func roleDelegateReEnableDiagnostic(priorPolicyType, priorTier string, newTier types.String) diag.Diagnostic {
    if priorPolicyType != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) ||
        priorTier != string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        return nil
    }
    if newTier.IsNull() || newTier.IsUnknown() ||
        newTier.ValueString() == string(datadogV2.ORGGROUPPOLICYENFORCEMENTTIER_DELEGATE) {
        return nil
    }
    return diag.NewAttributeErrorDiagnostic(
        frameworkPath.Root("enforcement_tier"),
        "Cannot re-enable a disabled role policy",
        "This role policy was disabled (enforcement_tier = \"DELEGATE\") and cannot be transitioned back to GROUP_MANAGED or OVERRIDE_ALLOWED.",
    )
}

// Update call site:
if d := roleDelegateReEnableDiagnostic(priorState.PolicyType.ValueString(), priorState.EnforcementTier.ValueString(), state.EnforcementTier); d != nil {
    response.Diagnostics.Append(d)
    return
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) Opinion:** this could be handled in a future cleanup/follow-up PR with other polish tasks.

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -117,6 +124,70 @@ func (r *OrgGroupPolicyResource) ImportState(ctx context.Context, request resour
...
+
+// ValidateConfig catches role + OVERRIDE_ALLOWED at plan time; the API only 400s.
+func (r *OrgGroupPolicyResource) ValidateConfig(ctx context.Context, request resource.ValidateConfigRequest, response *resource.ValidateConfigResponse) {
+	var config OrgGroupPolicyModel
+	response.Diagnostics.Append(request.Config.Get(ctx, &config)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	if config.PolicyType.ValueString() != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) {
```
pgoldrbx: **[~Minor~ Nit]** `config.PolicyType.ValueString()` collapses null/unknown to `""`, so if `policy_type` is unknown at config time (e.g. interpolated from another resource's computed output), this OVERRIDE_ALLOWED-for-role check is silently skipped. Low-probability in practice since `policy_type` is normally a plain literal, so this is optional hardening rather than a required fix.

```go
// after
if config.PolicyType.IsUnknown() {
    return // can't validate yet; re-evaluated once policy_type resolves
}
if config.PolicyType.ValueString() != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE) {
    return
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) thoughts:** I don't know that this is worth the time honestly. Your call.

datadog/fwprovider/resource_datadog_org_group_policy.go
```diff
@@ -117,6 +124,70 @@ func (r *OrgGroupPolicyResource) ImportState(ctx context.Context, request resour
...
+	if diags.HasError() {
+		return
+	}
+	resp.RequiresReplace = policyType.ValueString() != string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE)
+}
+
+// ModifyPlan catches a DELEGATE -> non-DELEGATE transition on a role policy at
+// plan time; without this, Update() is the only place that rejects it, so the
+// error only surfaces at apply.
+func (r *OrgGroupPolicyResource) ModifyPlan(ctx context.Context, request resource.ModifyPlanRequest, response *resource.ModifyPlanResponse) {
```
pgoldrbx, line 195: **[Important]** No unit test calls `ModifyPlan` directly — only `Update()`'s equivalent guard and `orgGroupPolicyNameRequiresReplaceIf` are covered. That gap is why the `RequiresReplace` bug flagged above shipped: `Update()` is never invoked when replacement is required, so it can't exercise this path at all.

```go
func TestOrgGroupPolicyModifyPlan_RoleRejectsReEnableFromDelegate(t *testing.T) {
    r := &OrgGroupPolicyResource{}
    ctx := context.Background()
    s := orgGroupPolicyValidateConfigSchema()

    req := resource.ModifyPlanRequest{
        State: tfsdk.State{Raw: orgGroupPolicyStateValue(string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "DELEGATE"), Schema: s},
        Plan:  tfsdk.Plan{Raw: orgGroupPolicyConfigValue(string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED"), Schema: s},
    }
    resp := &resource.ModifyPlanResponse{Plan: req.Plan}
    r.ModifyPlan(ctx, req, resp)

    if !resp.Diagnostics.HasError() {
        t.Fatal("expected error rejecting re-enable from DELEGATE")
    }
}

// Regression test: a legitimate replace must NOT be blocked by the re-enable
// guard just because the prior state was a disabled role policy.
func TestOrgGroupPolicyModifyPlan_ReplaceNotBlockedByPriorDelegateState(t *testing.T) {
    r := &OrgGroupPolicyResource{}
    ctx := context.Background()
    s := orgGroupPolicyValidateConfigSchema()

    req := resource.ModifyPlanRequest{
        State: tfsdk.State{Raw: orgGroupPolicyStateValue(string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "DELEGATE"), Schema: s},
        Plan:  tfsdk.Plan{Raw: orgGroupPolicyConfigValue("org_config", "OVERRIDE_ALLOWED"), Schema: s},
    }
    resp := &resource.ModifyPlanResponse{
        Plan:            req.Plan,
        RequiresReplace: []path.Path{path.Root("policy_type")},
    }
    r.ModifyPlan(ctx, req, resp)

    if resp.Diagnostics.HasError() {
        t.Fatalf("legitimate replace incorrectly blocked: %v", resp.Diagnostics)
    }
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

datadog/fwprovider/resource_datadog_org_group_policy_test.go
```diff
@@ -0,0 +1,250 @@
...
+		},
+		Plan: tfsdk.Plan{
+			Raw:    orgGroupPolicyStateValue(string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE), "GROUP_MANAGED"),
+			Schema: s,
+		},
+	}
+	resp := &resource.UpdateResponse{}
+	r.Update(context.Background(), req, resp)
+
+	if !resp.Diagnostics.HasError() {
```
pgoldrbx: **[Minor]** This only asserts `resp.Diagnostics.HasError()`, not which diagnostic fired — same pattern at lines 78 and 114. That's currently safe given the hardcoded, schema-conformant fixtures used, but it's cheap defense against a future refactor silently changing which guard trips.

```go
if !resp.Diagnostics.HasError() {
    t.Fatal("expected an error re-enabling a disabled role policy (DELEGATE -> GROUP_MANAGED)")
}
if got := resp.Diagnostics[0].Summary(); got != "Cannot re-enable a disabled role policy" {
    t.Errorf("unexpected diagnostic summary: got %q", got)
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) opinion:** this is optional hardening that could be included in future cleanup/polish if you think it's valuable.

datadog/tests/resource_datadog_org_group_policy_test.go
```diff
@@ -69,6 +71,149 @@ func TestAccDatadogOrgGroupPolicy_Basic(t *testing.T) {
...
+						roleResourceID = rs.Primary.ID
+						return nil
+					},
+				),
+			},
+			{
+				// Disabling is one-way: the provider rejects transitioning back to
+				// GROUP_MANAGED before ever calling the API.
+				Config:      testAccCheckDatadogOrgGroupPolicyRoleConfig(orgGroupName, renamedPolicyName, fmt.Sprintf(`{"permissions":[%q,%q]}`, permissionID, otherPermissionID), "GROUP_MANAGED"),
+				ExpectError: regexp.MustCompile(`(?is)cannot be\s+transitioned back`),
```
pgoldrbx: **[Minor]** `(?is)cannot be\s+transitioned back` is looser than the fixed, single-line error text needs. "transitioned back" doesn't collide with any other error text in this file, so this is low-risk, but tightening it costs nothing and matches the pattern already used one step earlier (line 134).

```go
ExpectError: regexp.MustCompile(`cannot be transitioned back to GROUP_MANAGED or OVERRIDE_ALLOWED`),
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking) opinion**: I think it's simpler and easier to use literal values than regex, especially if we _know_ the expected value. but this is more 'polish'

datadog/fwprovider/resource_datadog_org_group_policy_test.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+}
+
+func TestOrgGroupPolicyValidateConfig_OrgConfigAllowsOverrideAllowed(t *testing.T) {
+	s := orgGroupPolicyValidateConfigSchema()
+	r := &OrgGroupPolicyResource{}
+
+	req := resource.ValidateConfigRequest{
+		Config: tfsdk.Config{
+			Raw:    orgGroupPolicyConfigValue("org_config", "OVERRIDE_ALLOWED"),
```
pgoldrbx: **[~Minor~ Nit]** Hardcoded `"org_config"` string literal here (and at line 222) instead of the generated constant — every `"role"` counterpart in this file consistently uses `string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ROLE)`.

```go
Raw: orgGroupPolicyConfigValue(string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ORG_CONFIG), "OVERRIDE_ALLOWED"),
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking)**: downgraded to nit

datadog/fwprovider/resource_datadog_org_group_policy_test.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+func orgGroupPolicyValidateConfigSchema() schema.Schema {
+	r := &OrgGroupPolicyResource{}
+	resp := &resource.SchemaResponse{}
+	r.Schema(context.Background(), resource.SchemaRequest{}, resp)
+	return resp.Schema
+}
+
+func orgGroupPolicyConfigValue(policyType, enforcementTier string) tftypes.Value {
```
pgoldrbx: **[~Minor~ Nit]** `orgGroupPolicyConfigValue` here, `orgGroupPolicyStateValue` (line 44), and `orgGroupPolicyConfigValueNullTier` (line 157) declare byte-for-byte identical `tftypes.Object` schemas by hand instead of deriving from `orgGroupPolicyValidateConfigSchema()`. Low risk since drift would surface as a loud `HasError()` failure, but worth extracting.

```go
func orgGroupPolicyObjectType(ctx context.Context) tftypes.Object {
    s := orgGroupPolicyValidateConfigSchema()
    return s.Type().TerraformType(ctx).(tftypes.Object)
}
```

---

🤖 Written by claude-sonnet-5 (anthropic)

- [x] _reviewed by @phil-gold_ddog_

**(non-blocking)**: another optional cleanup/polish if you see value in it. I wouldn't bother right now

pgoldrbx (commented):
One thing needs a re-look before merge: the `CheckDestroy` fix in the last acceptance-test step (see the inline comment on `resource_datadog_org_group_policy_test.go`) doesn't actually make the new `role`/`DELEGATE` branch reachable — the policy is dropped from config, and thus from state, before `CheckDestroy` runs, so the fix commit's claim doesn't hold.

The rest are mostly minor consistency/test-hygiene items plus one real `ModifyPlan` correctness gap.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Real bug, fixed differently. `response.RequiresReplace` is always empty inside `ModifyPlan` — `fwserver/server_planresourcechange.go:331` constructs it as `path.Paths{}` and only merges back at 353, so that check can never fire. Guard now infers the replace from prior state vs. plan (`replacingAttrChanged`), covering `org_group_id` too — unknown counts as changed, since it has no `UseStateForUnknown`. Three `ModifyPlan` tests, each mutation-verified.

datadog/tests/resource_datadog_org_group_policy_test.go
saraslee: Applied, but the branch still never executes and no step arrangement fixes it: cassette 35 is `DELETE /org_groups -> 204`, 36 is `GET .../{id} -> 404` — deleting the group cascades the policy, so the 404 `continue` always wins. Kept as defensive code with a comment stating that evidence. The invariant is asserted in the DELEGATE step by `testAccCheckDatadogOrgGroupPolicyStillDisabled`.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Optional half covered: new step updates `content` while `DELEGATE`. Recorded interaction 27 is `PATCH -> 200`, permissions 2->1, tier still `DELEGATE`. Added `ExpectResourceAction(..., ResourceActionUpdate)` so a replace regression fails loudly.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Extracted to `roleDelegateReEnableDiagnostic`, called from `ModifyPlan` and `Update`; tier check to `roleEnforcementTierDiagnostic`.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Added — `roleEnforcementTierDiagnostic` now runs in `Create`/`Update`, since `terraform apply <planfile>` skips `ValidateConfig`. Ordered before the re-enable guard so an invalid tier reports the invalid value; `TestOrgGroupPolicyUpdate_RoleInvalidTierReportedBeforeReEnable` pins it.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Switched to `AddAttributeError` on `path.Root("enforcement_tier")`.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Four tests now call `ModifyPlan` directly: the DELEGATE->GROUP_MANAGED rejection plus three replace/unknown skip cases.

datadog/fwprovider/resource_datadog_org_group_policy_test.go
saraslee: Added `assertDiagnosticSummary`; all six error-path assertions pin the summary now.

datadog/fwprovider/resource_datadog_org_group_policy_test.go
saraslee: Replaced with `string(datadogV2.ORGGROUPPOLICYPOLICYTYPE_ORG_CONFIG)`; no bare literals left.

datadog/fwprovider/resource_datadog_org_group_policy_test.go
saraslee: Fixtures derive their object type from the schema (`s.Type().TerraformType(ctx)`), so a schema change breaks them at the type level.

datadog/tests/resource_datadog_org_group_policy_test.go
saraslee: Not applied. Kept `\s+`. Literal single-line string fails on replay: Terraform hard-wraps the diagnostic detail before it reaches `ExpectError`, inserting a newline mid-sentence.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: No acceptance step. Covered at unit level via the `policy_type`-change path in `TestOrgGroupPolicyModifyPlan_SkipsReEnableGuardWhenReplaceRequired`. Skipped the plancheck: a `role`->`org_config` step needs a fresh org group and another re-record for what schema `RequiresReplace` already guarantees. Say the word if you want it.

datadog/fwprovider/resource_datadog_org_group_policy.go
saraslee: Skipped per your note, leaving as-is. The re-enable path does handle null/unknown explicitly (`replacingAttrChanged` treats unknown as changed; tier check bails on null/unknown).

pgoldrbx (approved):
blockers all addressed

PR #4227, opened 2026-09-09
[TfGen] Resource generation

Supam:
### Motivation
Generator v2 needs resource generation in addition to its existing data source support. Resources require coordinated CRUD operations and reconciliation of differing create, update, and read schemas. Generated code must also handle nested structures, unions, path parameters, sensitive fields, and jsonapi request envelopes safely. End to end validation is also needed to ensure generated resources compile against the provider and its pinned Datadog SDK.

### Changes
 - Added tracking group (`x-datadog-tf-generator`) resolution and SDK bindings for resource create, read, update, and delete operations.
 - Added schema merging with Terraform presence rules and plan modifiers.
 - Added resource schema, model, CRUD method, request envelope, path parameters, collection, and `oneOf` code generation.
 - Integrated generated resources with the CLI and provider registration, including overwrite reconciliation and unstable operation tracking.
 - Improved schema normalization, cycle isolation, SDK type derivation, generated-file synchronization, and developer Make targets.

### Tests
 - Added parser and model unit tests for operation groups, lifecycle validation, schema merging, provenance, cycles, and `oneOf` handling.
 - Added emitter and CLI tests covering nested requests, collections, jsonapi envelopes, path parameters, registration, overwrite behavior, and idempotency.
 - Added golden snapshot coverage for generated data sources and resources.
 - Added SDK binding corroboration and provider module discovery tests.
 - Added an integration compile gate and a real `incident_type` fixture proving generated resource code compiles against the provider's pinned dependencies.

PR #4232, opened 2026-09-11
[TF Gen] Add write-only field support

Supam:
### Motivation

Secret values belong in a request body, never in Terraform state. The OpenAPI specs already mark those properties `writeOnly`, but the generator treated them like any other string and emitted an attribute that persists the secret. This PR teaches every tfgen layer to recognize `writeOnly` and project it into the framework's write-only attribute interface, and wires up the CI job that proves the generated code actually honors that contract.

The generated resources this unblocks (`datadog_integration_elastic_cloud`, `datadog_integration_twilio_account`), their docs/examples, and the acceptance-test cassette are **not** part of this PR — they land in the next PR in the stack.

### Changes

- parser carries the OpenAPI `writeOnly` keyword onto the normalized schema.
- model reconciles that fact across the Create, Update and Read bodies, and now answers separately whether a secret is required on create and on update. Every merge function takes an `updateRequired` parameter to carry the second answer.
- model applies the framework containment rule once the tree is projected: a request-settable ancestor of a secret keeps its input flag but loses `Computed` and `UseStateForUnknown`.
- emit expands one `writeOnly` property into a pair of attributes, `<name>_wo` and `<name>_wo_version`, and routes the request through a handler that reads the value from configuration and sets the SDK field only when a value should be sent.
- emit normalizes computed unknowns to null, so a plan stays consistent where a secret backs no readable state.
- The resource and shared templates render that pair in the schema and the guarded setter in the request body.
- fwutils gains a write-only mode that drops the plaintext attribute, a `SecretResult`, and a handler with separate create and update retrieval, so an API accepting partial updates omits an unchanged secret while one demanding it on every update always sends it.
- fwprovider enables the generator-owned unstable-operation registry on the client config through `EnableGeneratedUnstableOperations`, so a generated artifact calling a beta endpoint no longer compiles and registers only to fail at apply time.
- Adds a CI job (`integration-test` in `.github/workflows/tfgen.yml`) that runs `make tfgen-test-integration` — the integration-tagged compile-and-schema-validate gate that previously only ran locally. It proves a generated write-only resource compiles against the provider module and that its framework schema keeps the plaintext attribute out while marking `_wo` write-only and `_wo_version` as a required stateful trigger.

### Tests

- Unit tests per layer: parser normalization of the keyword, the three-way merge of write-only facts and the required-on-create versus required-on-update split, and the tree projection including the ancestor containment rule.
- emit tests over the generated schema, the model fields and the guarded request mapping.
- fwutils tests cover the handler directly: retrieval on create, the partial-update path where an unchanged version trigger sends nothing, and the path where the API requires the secret on every update.
- New CI job runs the integration-tagged tests (`tfgen-test-integration`), which previously only ran locally: the CLI-driven compile gate that builds a generated write-only resource against the provider module and validates its framework schema.

### DIFF
--- a/.generator-v2/internal/cli/generate_writeonly_test.go
+++ b/.generator-v2/internal/cli/generate_writeonly_test.go
@@ -1,0 +1,284 @@
+//go:build integration
+
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/terraform-providers/terraform-provider-datadog/generator/internal/model"
+	testinfra "github.com/terraform-providers/terraform-provider-datadog/generator/internal/testdata"
+)
+
+type writeOnlyResourceFixture struct {
+	name     string
+	spec     string
+	resource string
+}
+
+var writeOnlyResourceFixtures = []writeOnlyResourceFixture{
+	{
+		name:     "Twilio",
+		spec:     "mini-datadog_integration_twilio_account.yaml",
+		resource: "resource_datadog_integration_twilio_account.go",
+	},
+	{
+		name:     "Elastic Cloud",
+		spec:     "mini-datadog_integration_elastic_cloud_account.yaml",
+		resource: "resource_datadog_integration_elastic_cloud.go",
+	},
+}
+
+// TestGenerateWriteOnlyResources is the checked-in, offline full-pipeline gate
+// for the canonical generated write-only resources. It deliberately starts at
+// the CLI so parser, model, SDK binding, rendering, writing and reporting are
+// all covered before the generated code is compiled and its framework schema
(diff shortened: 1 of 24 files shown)


Supam:
* **#4238** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4232** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4232?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4232?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4227** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

fpighi (approved):

PR #4233, opened 2026-09-11
[datadog_observability_pipelines] Add prometheus remote_write source and destination

petere-datadog:


datadog/fwprovider/observability_pipeline/prometheus_remote_write_source.go
```diff
@@ -0,0 +1,238 @@
...
+					Validators: []validator.String{
+						stringvalidator.OneOf("none", "plain"),
+					},
+				},
+				"address_key": schema.StringAttribute{
+					Optional:    true,
+					Description: "Name of the environment variable or secret that holds the listen address for the Prometheus Remote Write endpoint.",
+				},
+				"path": schema.StringAttribute{
+					Optional:    true,
```
tessneau: if we keep the [default in the openapi spec ](https://github.com/DataDog/datadog-api-spec/pull/6687/changes#diff-0a18ed25c29b5ceabd70014eeb3dca2e7f457578554a19ab255d964fd23e6fe8R3210)we should add `Computed: true` here to indicate to TF that this field can show up populated if omitted by the user

datadog/fwprovider/observability_pipeline/prometheus_remote_write_source.go
```diff
@@ -0,0 +1,238 @@
...
+								Validators: []validator.List{
+									listvalidator.SizeAtMost(1),
+								},
+								NestedObject: schema.NestedBlockObject{
+									Attributes: map[string]schema.Attribute{
+										"location": schema.StringAttribute{
+											Optional:    true,
+											Description: "Built-in token location on the incoming HTTP request. One of `path`, `address`.",
+											Validators: []validator.String{
+												stringvalidator.OneOf("path", "address"),
```
tessneau, line 218: nit: right now if a user includes both `location` and `header`, TF accept that input but under the hood discards `location` and just sends `header` over. This should restrict users to just set one of them
```suggestion
												stringvalidator.OneOf("path", "address"),
												stringvalidator.ExactlyOneOf(
                           path.MatchRelative().AtParent().AtName("header"),
          ),
```

tessneau (approved):
few non-blocking comments

amarshall-dd (approved):

hestonhoffman (approved):

PR #4234, opened 2026-09-11
[datadog_aws_wif_identity_mapping] [WIF-91] Add AWS WIF identity mappings

### LINKED ISSUE #3893: Support for Workload Identity Federation resource mapping

- datadog_service_account
- datadog_role

### Feature Request

We're excited about the new Workplace Identity Federation, but it seems like there's no terraform resource support for creating a mapping between an ARN <> Service Account. We can terraform the role, the service account, but not the mapping.

It [looks like the plumbing is already there](https://docs.datadoghq.com/api/latest/cloud-authentication/create-an-aws-cloud-authentication-persona-mapping/), are we open to implementing this in `terraform-provider-datadog`?

Happy to contribute to this if so.

### References

_No response_

devand123:
Adds `datadog_aws_wif_identity_mapping` so Terraform can manage the mapping from an AWS identity to a Datadog user or service account. This lets teams define the WIF mapping alongside their IAM roles and Datadog service accounts.

Fixes #3893. Tracks [WIF-91](https://datadoghq.atlassian.net/browse/WIF-91).

## Behavior

- Supports create, read, delete, and import. Changing either input replaces the mapping; the API has no update operation.
- Accepts supported caller ARNs and trailing `/*` wildcards in the `aws` partition. Role mappings use STS assumed-role ARNs.
- Recommends `datadog_service_account.id` for stable creation and import. Email identifiers work, but importing returns the canonical handle and can force replacement.
- Retries AWS integration propagation errors and transient mapping 404s.

Requires an integrated AWS account and bootstrap credentials with WIF write permission and all permissions assigned to the target identity. Uses the public beta Cloud Authentication API. Includes a service-account example and import instructions.

## Validation

- 452 unit tests pass, including ARN validation, state preservation, and read retries. Repeated cancellation checks pass under the race detector.
- Formatting, vet, generated-doc checks, and the existing acceptance cassette pass. `errcheck` reports pre-existing violations in files unchanged by this PR.
- [Live CI passed](https://github.com/DataDog/terraform-provider-datadog/actions/runs/35271096886) both caller and service-account mapping tests: creation, UUID resolution, import, and deletion.

Manual staging validation passed: Terraform created a Read Only service account and mapping; normal and post-import plans were unchanged; AWS-only authentication resolved to that exact account and org. After deletion propagated, fresh authentication was rejected. Temporary resources and credentials were cleaned up. The backend currently requires an unmapped caller for creation, so the test used a separate bootstrap service account.

The service-account acceptance test skips cassette replay until a recording is available.

[WIF-91]: https://datadoghq.atlassian.net/browse/WIF-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/resources/datadog_aws_wif_identity_mapping/resource.tf
```diff
@@ -0,0 +1,16 @@
...
+}
+
+resource "datadog_service_account" "terraform" {
+  email = "terraform-service-account@example.com"
+  name  = "Terraform service account"
+  roles = [data.datadog_role.read_only.id]
+}
+
+resource "datadog_aws_wif_persona_mapping" "terraform" {
+  account_identifier = datadog_service_account.terraform.id
```
devand123, line 17: **⚠️ RETRACTED — this comment is wrong. `datadog_service_account.id` works correctly. My apologies.**

I sampled `TestAccDatadogServiceAccountDatasourceError.yaml` to establish what `datadog_service_account.id` resolves to, and that cassette's user is `"service_account":false` — a regular user, created deliberately so the datasource errors. I generalized from the wrong record.

For an actual service account, **the handle *is* the UUID**, by construction. `dd-source` `domains/aaa/apps/terry/internal/services/user_service.go:397`:

```go
func (us *UserService) CreateServiceAccountUser(...) {
	userUuid, _ := us.newUuid()
	user := &userdomain.User{
		UUID:           userUuid.String(),
		Handle:         userUuid.String(),   // handle == uuid
		Email:          email,
		ServiceAccount: true,
```

Confirmed in three independent cassettes spanning 2022–2025 — every `"service_account":true` record has `handle == id`:

| cassette | handle | id | matches |
|---|---|---|---|
| `...DatasourceMatchFilter` | `770a0f3a-790a-11f0-958a-4e9ff18b21fa` | same | yes |
| `...DatasourceMatchID` | `ffa92ba9-0628-11ee-9cfd-9e791e69b7e1` | same | yes |
| `TestServiceAccountCreate` | `008954f3-8c5a-11ed-837b-6ebf3eefa6a4` | same | yes |

So the handle lookup succeeds on the **first** try — `req.Handles = [uuid]` matches the stored handle. The email fallback is never reached, and there is no UUID-shape special-casing in the handler.

Better still, `validateUserMappingConfig` normalizes via `config.SetUserHandle(user.Handle)`, so for a service account the normalized value equals the value you sent. **`.id` is a fixed point under normalization** — no drift on create *or* import. That is exactly the property the schema description claims.

Concretely: my suggestion to switch to `.email` would have *introduced* the drift I flagged separately, since an email normalizes to the UUID handle and the two then differ in state. This example is right as written.

---

*Original (incorrect) comment below.*

> **This example cannot work — `account_identifier` will 400.**
> 
> `datadog_service_account.id` is the user **UUID**, not a handle or email. Confirmed in `datadog/tests/cassettes/TestAccDatadogServiceAccount*.yaml`: `"id":"0d17c71b-790b-11f0-8ef8-6a020aa62ff8","type":"users"`, while `handle` is the email.
> 
> The backend cannot resolve a UUID from this field. Tracing `dd-source@origin/main`:
> 
> 1. `validateUserMappingConfig` → `getUserByIdentifierOrEmail(...)` (`cloud-auth-config/internal/handler_user_mapping.go:118`)
> 2. → `GetUserByIdentifier(org, ouiclient.WithUserHandle(identifier))`
> 3. → `toUsersByIdentifiersOptions` (`ouiclient/user.go:256`) places the value in **`req.Handles`** only. `req.Uuids` is a separate gRPC field that is never populated here — `ouiclient.WithUserUuid` exists and is not used.
> 4. Handle lookup fails → fallback is `SearchUsers(UserFilter{Emails: [...]})`, which also fails for a UUID.
> 5. → `400: unable to find user with handle or email: <uuid>`
> 
> So an `apply` following this example fails. This file is rendered verbatim into `docs/resources/aws_wif_persona_mapping.md` and ships to the public Terraform registry.
> 
> Suggest using the handle (or email) instead, e.g. `datadog_service_account.terraform.email`, and dropping the UUID recommendation in the schema description.
> 
> Secondary: the example uses `data "aws_caller_identity"` without a `required_providers` block for `hashicorp/aws`, so copy-pasting it from the registry fails `terraform init`.

datadog/fwprovider/resource_datadog_aws_wif_persona_mapping.go
```diff
@@ -0,0 +1,288 @@
...
+	response.Schema = schema.Schema{
+		Description: "Provides an AWS Workload Identity Federation (WIF) persona mapping. The mapping allows an AWS IAM principal matching `arn_pattern` to authenticate as the Datadog user or service account identified by `account_identifier`. The AWS account in the ARN must already be integrated with Datadog. The identity creating the mapping must have every permission assigned to the target identity. Creating the initial mapping requires API and application credentials with the Workload Identity Federation write permission; a provider already using WIF cannot bootstrap its own mapping. This resource uses a public beta API and is subject to change.",
+		Attributes: map[string]schema.Attribute{
+			"id": utils.ResourceIDAttribute(),
+			"account_identifier": schema.StringAttribute{
+				Description: "The email or handle of the Datadog user or service account that the AWS principal authenticates as. For a Terraform-managed service account, prefer the stable UUID exported by `datadog_service_account.id`; Datadog accepts it as the service account identifier.",
```
devand123: **⚠️ RETRACTED — this description is accurate. See my retraction on `examples/.../resource.tf:14` for the evidence.**

Service-account handles are UUIDs by construction (`terry/internal/services/user_service.go:397` sets `Handle: userUuid.String()`), so `datadog_service_account.id` resolves via the handle path and is stable under the backend's normalization. The wording here is correct.

One thing genuinely worth adding, though: the resource depends on the undocumented invariant `handle == uuid` for service accounts. It holds by construction today and has for years, but it is not part of the public API contract — if service accounts ever got human-readable handles, this example breaks. A one-line comment recording that dependency would be worth it.

---

*Original (incorrect) comment below.*

> This description states *"prefer the stable UUID exported by `datadog_service_account.id`; Datadog accepts it as the service account identifier."*
> 
> Per the trace in my comment on `examples/.../resource.tf`, the API resolves this field by **handle, then email** — there is no UUID path. This sentence is the source of the broken example and it ships into the registry docs.
> 
> Worth noting the appeal of a stable UUID is real (a handle change would force replacement). If UUID support is wanted, it needs `ouiclient.WithUserUuid` wired into `getUserByIdentifierOrEmail` backend-side first.

datadog/fwprovider/resource_datadog_aws_wif_identity_mapping.go
```diff
@@ -0,0 +1,288 @@
...
+	data := apiResponse.GetData()
+	attributes := data.GetAttributes()
+
+	state.ID = types.StringValue(data.GetId())
+	state.AccountUUID = types.StringValue(attributes.GetAccountUuid())
+	state.ArnPattern = types.StringValue(attributes.GetArnPattern())
+
+	// The API normalizes an email identifier to the resolved handle. Preserve the
+	// configured value to prevent a perpetual replacement; imports start with no value.
+	if state.AccountIdentifier.IsNull() || state.AccountIdentifier.IsUnknown() {
```
devand123, line 302: **Scope correction:** this is real but narrower than I first wrote, and it does *not* affect the documented example.

With `account_identifier = datadog_service_account.x.id`, the value is already the canonical handle (service-account handles are UUIDs — see my retraction on `examples/.../resource.tf:14`), so it is a fixed point under `config.SetUserHandle(user.Handle)` and imports cleanly.

The replace only triggers when the configured identifier is **not** already the canonical handle — most realistically a service account's **email**, which normalizes to the UUID handle. Then imported state holds the UUID, config holds the email, and the next plan sees a diff on a `RequiresReplace()` attribute → destroy + create.

So it's a footgun on a path the docs steer away from, not a defect in the recommended one. Still worth handling, since `.email` is an obvious thing to reach for: either resolve the configured value to its canonical handle before comparing, or note in the schema description that the email form is not import-stable.

---

*Original comment below.*

> **`terraform import` will destroy and recreate the mapping on the next plan.**
> 
> The backend normalizes the identifier to the handle — `dd-source` `aws_user_mapping_test.go:647`:
> ```go
> // The config should have been normalized to use the handle instead of the email
> require.Equal(t, "service-account-handle", cloudAuthConfig.DDAccountIdentifier)
> ```
> 
> This guard is correct and load-bearing for the create/read path. But on import, state is null, so the branch takes the API's canonical **handle**. If the config holds an email that normalizes to a different handle, the next plan sees config != state on an attribute carrying `RequiresReplace()` (line 79) → **destroy + create**.
> 
> For a WIF persona mapping that's an authentication gap, and per the PR description a WIF-authenticated provider can't recreate its own mapping — so the recreate can fail and leave the principal locked out.
> 
> This contradicts *"Supports import and handles API normalization without Terraform drift"* in the PR body.
> 
> Options: resolve the configured value to its canonical handle before comparing (a plan modifier or `CustomizeDiff`-equivalent), or make `account_identifier` non-authoritative and expose the canonical handle as a separate computed attribute.

datadog/tests/resource_datadog_aws_wif_persona_mapping_test.go
```diff
@@ -0,0 +1,128 @@
...
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: accProviders,
+		CheckDestroy:             testAccCheckAwsWifPersonaMappingDestroy(providers.frameworkProvider),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAwsWifPersonaMappingConfig(accountID),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsWifPersonaMappingExists(providers.frameworkProvider, resourceName),
+					resource.TestCheckResourceAttr(resourceName, "arn_pattern", fmt.Sprintf("arn:aws:sts::%s:assumed-role/terraform-runner/*", accountID)),
+					resource.TestCheckResourceAttrSet(resourceName, "account_identifier"),
```
devand123, line 44: **⚠️ RETRACTED — both halves of this were wrong. I had missed the colocated unit test entirely.**

`datadog/fwprovider/resource_datadog_aws_wif_persona_mapping_test.go` already covers exactly the case I said was untested, in both directions, with a fixture where `handle != email`:

```go
apiResponse := newAwsWifPersonaMappingResponse(
    "mapping-id", "service-account-handle", "account-uuid", "arn:aws:sts::…/session-name")

t.Run("preserves configured email when API returns handle", …)  // configured "terraform-service-account@example.com" survives
t.Run("uses API handle during import", …)                        // null → "service-account-handle"
```

So the guard at `resource_..._mapping.go:285` is *not* indistinguishable from an unconditional overwrite — the second subtest fails if you delete the condition, and the first fails if you make it unconditional. I judged coverage from the acceptance test alone and never opened the unit test file sitting next to the resource.

**And the `TestCheckResourceAttr` suggestion was wrong too.** The expected value legitimately differs by RECORD mode — live/record resolve the caller through `data.datadog_current_user`, replay uses the recorded literal — so there is no single correct value to assert. `TestCheckResourceAttrSet` is the right call here. I changed it to `TestCheckResourceAttr` and had to revert that in `f542c75` along with the rest.

**What's actually left** is smaller than either version of this comment: there's no *acceptance-level* test of a service-account-backed mapping, so the end-to-end import-produces-a-replacement behaviour for `.email` is unverified against the live API. The state-transition logic underneath it is covered by the unit test above. Optional, not blocking.

<details><summary>Previous versions (both wrong — kept for the record)</summary>

> **Correction:** the "bugs above" framing is wrong — the UUID finding was mine and I've retracted it. `datadog_service_account.id` works.
>
> The coverage gap itself still stands, just more narrowly. The test only exercises a human user, where `handle == email` makes normalization a no-op, so the guard at `resource_..._mapping.go:285` is indistinguishable from an unconditional overwrite and `ImportStateVerify: true` passes trivially. A service-account case — where `handle` is a UUID and `email` is not — is the one that actually exercises normalization, and it is untested in both directions:
>
> - `account_identifier` = service account **`.id`** → should round-trip unchanged (the documented path, currently unverified)
> - `account_identifier` = service account **`.email`** → normalizes to the UUID handle, which is the import-replace case
>
> Also still worth asserting the exact value with `TestCheckResourceAttr` rather than `TestCheckResourceAttrSet`.

---

> **This test cannot catch either of the two bugs above, which is why CI is green.**
>
> `TestCheckResourceAttrSet` passes for any non-empty string. And the identifier in play is a *human user's handle*, which at Datadog equals their email — so the API echoes it back verbatim and normalization is a no-op. That makes the `IsNull() || IsUnknown()` branch at `resource_..._persona_mapping.go:285` indistinguishable from an unconditional overwrite, and makes `ImportStateVerify: true` (line 43) pass trivially.
>
> The two cases that matter are untested:
> - a **service account UUID** (the documented recommendation) — should 400
> - an **email that normalizes to a different handle** — should expose the import replace
>
> Suggest asserting the exact configured value with `TestCheckResourceAttr` and adding a step backed by a service account, where handle != email.

</details>

datadog/tests/resource_datadog_aws_wif_persona_mapping_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/fwprovider"
+	"github.com/terraform-providers/terraform-provider-datadog/datadog/internal/utils"
+)
+
+const awsWifReplayAccountIdentifier = "tf-testacccurrentuserdatasource-local@example.com"
```
devand123, line 26: **⚠️ RETRACTED as a defect — the `isReplaying()` fork below is load-bearing, and I proved it the hard way by removing it and breaking CI.**

I pushed a "fix" that deleted the fork so record and replay would share one config (`677e63f`). `pr-integration-tests` went red:

```
400 Bad Request: "unable to find user with handle error: rpc error: code = NotFound
desc = failed to get users by identifiers: ... 321813), email lookup error: <nil>"
```

`.github/workflows/test_pr_integration.yml` runs `make testacc` with `RECORD: "none"` — integration tests hit the **live** API, not the cassette. `05843fa` (fork present) was green at 21 passed; `677e63f` (fork removed) failed. Reverted in `f542c75`.

Both sides are required, for different reasons:
- **live / record** must resolve the caller dynamically, because the recorded user doesn't exist in other orgs *and* the API requires the caller's permissions to be a superset of the mapped account's — satisfied by mapping the caller to itself.
- **replay** must use the literal, because the cassette predates the data source and holds no `current_user` interaction.

**And my "different resource graphs" framing was wrong.** The two paths are value-identical. The cassette's `account_uuid` is `0bd557d2-6ed0-423e-9654-c0e1cd376abc`, which is exactly the `id` that `data.datadog_current_user` returns in `data_source_datadog_current_user_test.go:28-30`, whose `handle` is this same literal. Same user, same org (`DD Integration Tests (321813)`). So the literal *is* what the data source would have returned at record time — the fork differs by one data-source node, not by what gets tested. It also isn't "borrowed from another test's fixture": it's the recording org's own caller, which both tests happen to run as.

**What actually survives:** the literal has to stay in sync with the cassette. `f542c75` documents that, and why the fork exists, in a comment here — the missing explanation is what led me to misread it as sloppiness in the first place.

The re-record divergence I raised is harmless: `initRecorderWithName` (`datadog/tests/provider_test.go:690-726`) sets `ModeReplayOnly` + `matchInteraction` and asserts nothing about full playback, so an extra `current_user` interaction in a future cassette is simply never requested on replay.

<details><summary>Original comment (wrong — kept for the record)</summary>

> Two issues with this constant.
>
> **It is borrowed from another test's fixture.** `tf-testacccurrentuserdatasource-local@example.com` is the hardcoded expectation in `data_source_datadog_current_user_test.go:29-30`.
>
> **The cassette is no longer reproducible.** Since `4366a6e` added `data "datadog_current_user" "test" {}` on the record path (line 52), record mode now issues a `GET /api/v2/current_user`. The committed cassette contains **zero** `current_user` interactions — all 11 are `integration/aws/accounts` or `cloud_auth/aws/persona_mapping`. So `RECORD=true make testacc` produces a structurally different cassette, and record vs. replay exercise different resource graphs.
>
> The cassette is genuine for the path it was recorded on, so this isn't fabricated evidence — but "Acceptance test cassette replay" in the PR body only attests to the human-caller happy path, and the next person to re-record gets a conflict.

</details>

datadog/fwprovider/resource_datadog_aws_wif_identity_mapping.go
```diff
@@ -0,0 +1,288 @@
...
+)
+
+const (
+	awsWifPersonaMappingCreateTimeout     = 2 * time.Minute
+	awsWifPersonaMappingVisibilityTimeout = 30 * time.Second
+
+	// The authoritative backend parser is cloudconfig.ParseARNPattern in
+	// domains/aaa/external_authn/internal/libs/cloudconfig/arn_parser.go. It
+	// currently supports only the aws partition and this resource character set.
+	awsWifArnAccountIDPattern   = `[0-9]{12}`
```
devand123, line 35: **⚠️ CORRECTION — I had this backwards. The validator is *right*, and tighter than the backend for a good reason.**

`arn_pattern` is matched against an STS `GetCallerIdentity` ARN — `external-token-servicer/internal/awsHandler/user_handler.go:341`:

```go
if cloudconfig.MatchARNPattern(configItem.ARNPattern, callerArn) {
```

where `callerArn` comes from `GetCallerIdentityResult[0].Arn`. `GetCallerIdentity` can only ever return three shapes: `iam::…:user/…`, `sts::…:assumed-role/…/…`, and `sts::…:federated-user/…`. That is **exactly** the set this regex accepts.

Every one of the 7 "rejections" I listed is a shape `GetCallerIdentity` can never emit. `ParseARNPattern` accepts `role`, `group`, `instance-profile`, `policy` and `*-provider`, but a mapping using one of those would be stored happily and then **never match any caller** — a silently dead auth grant with no feedback. Rejecting `iam::…:role/…` at plan time is the single most valuable thing this validator does, since an IAM role ARN is the most natural thing for a user to reach for.

Re-ran the diff scoring each case by whether `GetCallerIdentity` can actually produce it:

| ARN | real caller? | backend | provider |
|---|---|---|---|
| `sts::…:assumed-role/MyRole/session-abc` | yes | ok | ok |
| `sts::…:assumed-role/MyRole/*` | yes | ok | ok |
| `iam::…:user/bob` | yes | ok | ok |
| `iam::…:user/division/subdivision/bob` | yes | ok | ok |
| `sts::…:federated-user/bob` | yes | ok | ok |
| `iam::…:role/MyRole` | **no** | ok | **rejects — correct** |
| `iam::…:group/Admins` | **no** | ok | **rejects — correct** |
| `iam::…:instance-profile/web` | **no** | ok | **rejects — correct** |
| `iam::…:policy/MyPolicy` | **no** | ok | **rejects — correct** |
| `iam::…:oidc-provider/example.com` | **no** | ok | **rejects — correct** |
| `sts::…:assumed-role/My:Role/sess` | **no** | ok | **rejects — correct** |
| `sts::…1234:assumed-role/MyRole/sess` | **no** | ok | **rejects — correct** |

**Zero cases where the provider blocks a real caller the backend would accept.** Also `[0-9]{12}` is stricter than the backend's `[a-z0-9]+` and correct — AWS account IDs are always 12 digits. No sign-off needed; this is good as-is.

Two **backend** gaps the diff did surface, rejected by *both* layers and so not this PR's problem, but worth tracking separately:

- **`+`, `=`, `,` in names.** AWS permits these in IAM user names, role names and role session names. Both charsets (`[A-Za-z0-9_.@-]` here, `[a-zA-Z0-9_.\-:@]` in `ParseARNPattern`) exclude them, so an affected principal cannot be mapped at all. Role session names in particular are often caller-supplied.
- **Non-`aws` partitions.** `^arn:aws:` in `arn_parser.go` means `aws-us-gov` and `aws-cn` are unmappable. The test here asserting GovCloud is unsupported is an accurate encoding of current behavior.

Both would need the backend regex widened first; the provider is faithfully mirroring it.

---

*Original (incorrectly framed) comment below.*

> **This client-side regex is narrower than the API it guards, and it can't be bypassed.**
> 
> I diffed it mechanically against `cloudconfig.ParseARNPattern` on `dd-source@origin/main`. The provider rejects these, which the backend accepts:
> 
> | ARN | backend | provider |
> |---|---|---|
> | `arn:aws:iam::123456789012:role/MyRole` | accepts | **rejects** |
> | `arn:aws:iam::123456789012:role/MyRole/*` | accepts | **rejects** |
> | `arn:aws:iam::123456789012:group/Admins` | accepts | **rejects** |
> | `arn:aws:iam::123456789012:instance-profile/web` | accepts | **rejects** |
> | `arn:aws:iam::123456789012:policy/MyPolicy` | accepts | **rejects** |
> | `arn:aws:iam::123456789012:oidc-provider/example.com` | accepts | **rejects** |
> | `arn:aws:sts::123456789012:assumed-role/My:Role/session` | accepts | **rejects** |
> 
> No false-accepts in the other direction, so the narrowing is at least safe. Rejecting bare `iam::…:role/…` is also genuinely useful — it catches the most common user mistake, which the schema description calls out.
> 
> The concern is coupling: a plan-time validator runs before any API call, so there is no escape hatch for a user whose valid config is rejected, and every backend broadening needs a provider release plus a user upgrade. GovCloud is explicitly rejected (`TestAwsWifArnPattern/GovCloud_partition_is_not_supported_by_the_API`) while still being an open question for WIF.
> 
> Not blocking — loosening a validator later isn't a breaking change. But it's a deliberate product decision rather than an implementation detail, so worth a maintainer's explicit sign-off. A looser shape check plus letting the API's 400 be authoritative is the lower-maintenance alternative.

datadog/tests/resource_datadog_aws_wif_persona_mapping_test.go
```diff
@@ -0,0 +1,128 @@
...
+  # Using the test caller guarantees the API's required permission-subset relation.
+  account_identifier = %[3]s
+  arn_pattern         = "arn:aws:sts::${datadog_integration_aws_account.test.aws_account_id}:assumed-role/terraform-runner/*"
+}
+`, accountID, currentUserConfig, accountIdentifier)
+}
+
+func testAccCheckAwsWifPersonaMappingExists(provider *fwprovider.FrameworkProvider, resourceName string) resource.TestCheckFunc {
+	return func(state *terraform.State) error {
+		id := state.RootModule().Resources[resourceName].Primary.ID
```
devand123: Nil dereference if the resource is missing from state — the map index returns a nil `*terraform.ResourceState` and `.Primary` panics, taking down the parallel test binary and hiding the real failure.

Repo convention checks the lookup, e.g. `data_source_datadog_csm_threats_policies_test.go:52`:
```go
res, ok := state.RootModule().Resources[dataSourceName]
```

datadog/fwprovider/resource_datadog_aws_wif_persona_mapping.go
```diff
@@ -0,0 +1,288 @@
...
+func (r *awsWifPersonaMappingResource) Read(ctx context.Context, request resource.ReadRequest, response *resource.ReadResponse) {
+	var state awsWifPersonaMappingModel
+	response.Diagnostics.Append(request.State.Get(ctx, &state)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	apiResponse, httpResponse, err := r.Api.GetAWSCloudAuthPersonaMapping(r.Auth, state.ID.ValueString())
+	if err != nil {
+		if httpResponse != nil && httpResponse.StatusCode == 404 {
```
devand123: Minor: bare `404` here and at line 266, but `http.StatusNotFound` at line 200 for the same check in the same file. `net/http` is already imported — worth making these consistent.

devand123 (commented):
Reviewed this end to end against repo conventions and the `dd-source` backend, then re-verified every finding. **Net result: no correctness defects in the resource logic.** Two of my own findings were wrong and are retracted inline — detail below.

**Verified green:** `go build ./...`, `make test` (409 pass, including all 31 new ones), `make fmtcheck`, `make vet`, `make check-docs` ("No generated documentation changes detected"), `RECORD=false make testacc -run TestAccAwsWifPersonaMapping`. All 28 CI jobs pass.

### Retracted (my errors)

- **`datadog_service_account.id` does not work** — wrong. Service-account handles *are* UUIDs by construction (`terry/internal/services/user_service.go:397` sets `Handle: userUuid.String()`), so `.id` resolves on the first handle lookup and is a fixed point under `config.SetUserHandle(user.Handle)`. It round-trips through create *and* import without drift. I'd sampled a cassette whose user was `"service_account":false`. The example and schema description are correct as written.
- **The ARN validator is too strict** — wrong, and backwards. `arn_pattern` is matched against a `GetCallerIdentity` ARN (`user_handler.go:341`), which emits exactly three shapes — precisely the set this regex accepts. All 7 ARN classes I listed as "rejected" are shapes a caller identity can never have; `ParseARNPattern` accepting them means a mapping can be stored that silently never matches. Rejecting `iam::…:role/…` at plan time is the most useful thing the validator does. Zero cases where the provider blocks a real caller.

I also chased a suspicion that the `updateState` guard suppresses out-of-band drift. It doesn't — the API exposes only create/get/delete/list, so a mapping's `account_identifier` is immutable server-side and there is nothing to drift to. The guard is correct and necessary.

### What's genuinely good

Provider wiring is idiomatic (`GetCloudAuthenticationApiV2` mirrors the existing `ApiInstances` pattern; `SetUnstableOperationEnabled` matches `Dataset`; the erroring `Update` matches six existing immutable resources). Docs are genuinely `tfplugindocs`-generated. The retry-on-400 is evidence-backed — the cassette holds a real recorded `400 "AWS Account Id is not integrated with this Datadog account"` from integration propagation lag. Preserving state on an unparsed create response is better than the repo norm and well-judged for a beta API where a leaked mapping is a dangling auth grant. `[0-9]{12}` for account IDs is stricter than the backend and correct. And it carries colocated unit tests, which only 6 of 127 resources here do.

### Actually outstanding

1. **Merge conflict** in `datadog/tests/framework_provider_test.go` — `mergeable: CONFLICTING`. Just the unstable-ops list. Mechanical, but blocks merge.
2. **Cassette is not reproducible in record mode.** The record path builds `data "datadog_current_user" "test" {}` (test line 50), but the committed cassette has **zero** `current_user` interactions — all of them are `integration/aws/accounts` or `cloud_auth/aws/persona_mapping`. `RECORD=true` produces a structurally different recording, so replay and record exercise different resource graphs. Next person to re-record gets a conflict.
3. **Test coverage gap.** Only a human user is exercised, where `handle == email` makes normalization a no-op — so the guard is indistinguishable from an unconditional overwrite and `ImportStateVerify: true` passes trivially. A service-account case (handle is a UUID, email is not) exercises both directions, and would have caught my mistaken reading of the intent here.
4. **`.email` is not import-stable.** `.id` imports cleanly; a service account's `.email` normalizes to the UUID handle, so imported state holds the UUID while config holds the email → diff on a `RequiresReplace()` attribute → destroy + create. A footgun on a path the docs steer away from, not a defect in the recommended one, but `.email` is an obvious thing to reach for. Worth a line in the schema description at minimum.
5. Nits: bare `404` at lines 235/266 vs `http.StatusNotFound` at line 200; unchecked map index in the test check function (26 other places in `datadog/tests` do the same, so minority style rather than a convention break).

Worth recording somewhere that this resource depends on the undocumented invariant `handle == uuid` for service accounts. True by construction and stable for years, but not part of the public API contract.

Two real **backend** gaps the ARN diff surfaced, rejected by both layers and out of scope here: `+`/`=`/`,` are legal in IAM user, role and session names but absent from both charsets; and `^arn:aws:` makes `aws-us-gov`/`aws-cn` unmappable.

Line counts in perspective: of 1202 additions, 288 are resource logic, 449 tests, 376 a recorded cassette, 72 docs/examples, 17 wiring.

brett0000FF (approved):

PR #4238, opened 2026-09-15
[TFGen] Added the option to omit fields with defaults

Supam:
### Motivation

When an OpenAPI request property declares a `default`, the generator previously discarded that keyword entirely: the field always rendered as `Required` (or plain `Optional`), so a user had to repeat the API's default in every config, and the SDK's own `New<Model>WithDefaults()` was left to apply it silently on the wire. That made it impossible to omit a field like `auth_type` from a resource config even when the API already defaults it, and left generated code with no way to render a stable, framework native default. This PR teaches the generator to recognize scalar `default` declarations and turn them into a proper Terraform Plugin Framework default, so a request field with a server side default becomes an omittable, `Optional+Computed` attribute instead of a `Required` one.

### Changes

- parser (`schema.go`) decodes the OpenAPI `default` for string, boolean, integer and number scalars against the field's final resolved type and enum, including through `$ref` and `allOf` composition; an unusable declaration (wrong YAML type, disallowed enum value, non finite number, overflow, non scalar) is retained as a `Problem` rather than discarded, so a later resource only validation pass can decide whether it matters.
- model (`merge.go`, `types.go`) adds `ScalarDefault`/`SchemaDefault` and a `Union` composition rule (equal repeats are harmless, conflicting composed defaults become a `Problem`), plus a lifecycle policy: Create's default is authoritative, Update may omit or repeat it but not contradict it, and secret and write only fields are excluded from validation entirely. A new `SchemaDefaultError` reports the artifact, path, role, and reason for an invalid default.
- model (`schema.go`) limits native defaults to configurable, non secret primitive request fields (`resourceDefaultEligible`) and changes `applyPresence` so a defaulted field becomes `Optional+Computed` with `UseStateForUnknown`, instead of `Required`, while keeping its existing enum validators and other plan modifiers.
- emit (`builder.go`, `view.go`, templates) renders the framework's static default constructor (e.g. `stringdefault.StaticString("basic")`) on the schema attribute and imports the matching `resource/schema/<type>default` package, while still sending the field's actual configured or planned value, not the schema default, through both the Create and Update request bodies, so an explicitly configured value is never silently overwritten.
- Adds a parser fixture (`schema_normalize_defaults.yaml`) and extends the write only generated resource contract test to also assert the `auth_type` default contract on the two write only fixtures.

### Tests

- Parser: table tests over direct scalar defaults (including falsy values like `""`, `false`, `0`), defaults reached through `$ref`/`allOf`, and every rejection case (wrong type, invalid enum member, integer overflow, non finite number, null, array and object defaults, response only invalid defaults).
- Model merge: tests for composing and conflicting defaults across `allOf`/`oneOf` branches, the Create authoritative Update agreement policy, artifact/path/role/reason error content, and confirmation that sensitive and write only defaults are never validated or disclosed.
- Model schema tree: tests that a defaulted required field becomes Optional+Computed with the correct literal and plan modifiers, that enum validators are preserved, that nested/child defaults do not leak onto parents, and that null, response only, secret, write only and compound defaults never surface on the tree.
- Emit: a resource builder test rendering all four scalar default kinds with their import and setter counts, plus a regression test proving an explicitly configured value (not the schema default) is the one sent to the SDK in both Create and Update.
- The CLI full pipeline write only fixture test now also asserts the `auth_type` static default contract end to end on the two generated write only resources.

### DIFF
--- a/.generator-v2/internal/cli/generate_writeonly_test.go
+++ b/.generator-v2/internal/cli/generate_writeonly_test.go
@@ -1,0 +1,284 @@
+//go:build integration
+
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/terraform-providers/terraform-provider-datadog/generator/internal/model"
+	testinfra "github.com/terraform-providers/terraform-provider-datadog/generator/internal/testdata"
+)
+
+type writeOnlyResourceFixture struct {
+	name     string
+	spec     string
+	resource string
+}
+
+var writeOnlyResourceFixtures = []writeOnlyResourceFixture{
+	{
+		name:     "Twilio",
+		spec:     "mini-datadog_integration_twilio_account.yaml",
+		resource: "resource_datadog_integration_twilio_account.go",
+	},
+	{
+		name:     "Elastic Cloud",
+		spec:     "mini-datadog_integration_elastic_cloud_account.yaml",
+		resource: "resource_datadog_integration_elastic_cloud.go",
+	},
+}
+
+// TestGenerateWriteOnlyResources is the checked-in, offline full-pipeline gate
+// for the canonical generated write-only resources. It deliberately starts at
+// the CLI so parser, model, SDK binding, rendering, writing and reporting are
+// all covered before the generated code is compiled and its framework schema
+// is validated in the provider module.
+func TestGenerateWriteOnlyResources(t *testing.T) {
+	providerRoot := filepath.Clean(filepath.Join("..", "..", ".."))
+	tempRoot := t.TempDir()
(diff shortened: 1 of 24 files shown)


Supam:
* **#4238** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4238?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4232** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4227** <a href="https://app.graphite.com/github/pr/DataDog/terraform-provider-datadog/4227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

fpighi (approved):

PR #4244, opened 2026-09-16
[datadog_rum_team_ownership] Add RUM team ownership resource

britofelipe:
## What does this PR do?

Adds the datadog_rum_team_ownership Terraform resource for managing RUM Ownership of Views rules.

The resource supports create, read, delete, and import. It also enables the three required unstable SDK operations in both provider and acceptance-test clients.

## Additional changes

- Supports application-scoped, service-scoped, and global mappings. Omitting both application_id and service creates a rule across all applications and services.
- Handles API 404 responses during refresh and delete using Terraform conventions.
- Adds browser, mobile, and global acceptance coverage, including import, replacement, defaults/no drift, and out-of-band deletion.
- Adds a recorded cassette, an example, and generated documentation.

## Testing

- make build
- make test (378 tests passed)
- make vet
- Formatting and go vet checks passed
- Acceptance test replay passed locally
- Live acceptance test passed against staging, including the global mapping

## API prerequisites

- DataDog/datadog-api-spec#6396
- DataDog/datadog-api-client-go#4445
- DataDog/dd-source#90700


AxelGeist (commented):
LGTM! Since this is resource specific, I will approve when the resource owning team [rum-backend](https://github.com/orgs/DataDog/teams/rum-backend) approves it.

joepeeples (approved):

hemeryar (commented):
Starting with a few AI findings, since the automated reviews don't run on this repo it seems. Some of them seem important, and are probably worth documenting even if you're not going to fix them. Some are linked to the API itself or the UI and might require a bit of work.

```
**Should address before merge**

1. **Global mappings (neither `application_id` nor `service`)**: only single create accepts them. The batch endpoint rejects them (`teamsownership_validation.go:155`), and the UI writes only through the batch endpoint, so Terraform would be the only way to create them. We need a product decision before documenting this: either support it and align the batch validation, or drop it from the resource and add the check to single create.
2. **Removing an optional field is a silent no-op**: `application_id`, `match_type` and `service` are Optional+Computed with `UseStateForUnknown` and no default (`resource_datadog_rum_team_ownership.go:60,70,87`). Deleting `match_type = "prefix"` keeps a prefix rule; deleting `application_id` doesn't make the mapping global. Please add defaults matching what the server stores (`"exact"`, the nil UUID, `""`), as the other RUM resources do, and a test step that removes `match_type`. The spec already declares `default: exact` on `TeamsOwnershipMatchType`, but tfgen didn't emit it, so the fix probably belongs in the generator.
3. **Remove `org_id`**: we don't want to expose org IDs in Terraform resources.
4. **Duplicates and import**: the table has no unique constraint, so declaring a mapping that already exists in the UI silently creates a second row. Import is the only clean way to adopt existing mappings, but there's no `import.sh`, so the docs have no Import section (every other RUM resource has one). Please add it. The unique index is on us.
5. **Prefix is flag-gated**: creating a prefix rule returns 400 unless `rum_teams_ownership_prefix_matching` is enabled for the org, yet the example uses `match_type = "prefix"`. Unless the flag is GA, switch the example to `exact` or document the gating.
6. **Missing `utils.CheckForUnparsed`**: every SDK-based RUM resource calls it after Create/Read. Without it, a `match_type` value this provider version doesn't know gets written to state as `""` and forces an unexplained replacement instead of a clear error.

**Smaller / docs**

7. **UI edits vs Terraform**: there's no update endpoint, so a UI edit is a delete plus a create with a new ID. If someone swaps team-a to team-b in the UI on a Terraform-managed mapping, the next apply recreates team-a as a plain `+ create`, and both teams end up as owners. Worth a doc note: manage a view's ownership in Terraform or in the UI, not both.
8. **Import ID validation**: IDs are integers. A non-numeric ID returns a 500 with the Postgres error text. The easy mistake is the 64-character rule `id` from `/rules`, when the importable ID is `teams[].mapping_id`. Please validate on import and say which ID to use in `import.sh`. Returning 404/400 instead of 500 is on us.
9. **Preview endpoints**: all three operations are marked Preview in the spec. That's fine by provider convention, but once released, breaking changes (e.g. the validation alignment in 1, or the unique index in 4) break Terraform users, so we should settle those first.
10. **Naming**: the UI calls these "Ownership Rules", and one rule can have several teams. One instance here is a single rule-and-team pair, which the API calls a mapping. Resource names are hard to change later, so worth deciding now (e.g. `datadog_rum_team_ownership_mapping`).
11. **Rate limits (rum-backend, not blocking)**: `rum_teams_ownership_read`/`_write` are still 200/100 per minute per org, and each managed mapping costs one GET per plan. We'll raise them in line with the other Terraform-driven RUM endpoints before announcing.
```

PR #3848, opened 2026-06-02
[datadog_integration_databricks_account] Add Databricks Resource under env var

guttymora:
## Context
Implements [SIMI-630](https://datadoghq.atlassian.net/browse/SIMI-630). Re-adds the Databricks AMS integration resource, reverted in #3825, but this time is behind an opt-in env var for a private beta cohort.
the resource is only registered when `DD_TERRAFORM_DATABRICKS_INTEGRATION_ENABLED=true`. With the env var unset (default), the resource type is invisible to `terraform plan` and excluded from the Terraform Registry docs/examples. 
> Only customers in the beta cohort receive the env var name and usage instructions.

**No docs, no CHANGELOG, no examples — by design.**
Publishing them would index the resource on the public registry and leak it into release notes, defeating the env-var hide. Cohort customers receive schema docs directly from the SIMI PM. On GA, we flip the helper default to `true`, remove the gate, and ship docs in that same release.

## Changes
- New helper `utils.IsDatabricksIntegrationEnabled()`
- Conditional registration in `FrameworkProvider.Resources()`
- Resource code, PAT + OAuth acceptance tests, and cassettes restored from https://github.com/DataDog/terraform-provider-datadog/pull/3814
- **Intentionally omitted**: `docs/resources/integration_databricks_account.md`, `examples/resources/datadog_integration_databricks_account/`, and the CHANGELOG entry

## Validation
- `make fmtcheck`, `make build`, `make vet`, `make lint-new`: clean
- `make testacc` (env set): `TestAccIntegrationDatabricksAccountPat` + `TestAccIntegrationDatabricksAccountOAuth` PASS
- Provider registration check: env unset -> 100 resources (no Databricks); env=true -> 101 resources (Databricks registered)

## Related
- API spec: [DataDog/datadog-api-spec#5626](https://github.com/DataDog/datadog-api-spec/pull/5626)
- Go SDK: [DataDog/datadog-api-client-go#4041](https://github.com/DataDog/datadog-api-client-go/pull/4041)

[SIMI-630]: https://datadoghq.atlassian.net/browse/SIMI-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3797, opened 2026-05-18
[datadog_integration_databricks_account] Add Terraform resource for Databricks AMS integration

guttymora:
## Context
Implements [SIMI-630](https://datadoghq.atlassian.net/browse/SIMI-630).

Datadog supports Databricks workspaces as a SaaS integration via the AMS Web Integrations API (`/api/v2/web-integrations/databricks/accounts`). Customers managing Databricks workspaces at scale — Carl Zeiss (70+), Mailgun, OpenAI, SoFi, SendGrid — have asked for Terraform support to handle SOX audit trails, credential rotation, and IaC workflows. The generic `datadog_web_integration_account` resource (#3625, Cristian) is still on the table for long-tail integrations, but Databricks and Snowflake are being prioritized as typed per-integration resources ahead of the generic approach due to customer demand.

This PR follows the foundation laid in #3787, which bumped `datadog-api-client-go` to expose `WebIntegrationsApi` and wired the 5 unstable opt-ins.

## Related
- API spec: [DataDog/datadog-api-spec#5626](https://github.com/DataDog/datadog-api-spec/pull/5626) (merged)
- Go SDK: [DataDog/datadog-api-client-go#4041](https://github.com/DataDog/datadog-api-client-go/pull/4041) (merged)
- TF foundation: #3787 (merged)

[SIMI-630]: https://datadoghq.atlassian.net/browse/SIMI-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4247, opened 2026-09-17
[datadog_integration_databricks_account] Remove env-gated resource ahead of generated replacement

rossicris:
### What does this PR do?

Removes `datadog_integration_databricks_account` — the resource, its acceptance test and cassettes, the `DD_TERRAFORM_DATABRICKS_INTEGRATION_ENABLED` flag that gated it, and the wiring that existed only to support it.

### Why

The resource never shipped. It is registered only inside `if utils.IsDatabricksIntegrationEnabled()`, which reads an env var defaulting to `false`; it has no `docs/resources/` page, and `DEVELOPMENT.md` explicitly said gated resources are "intentionally excluded from public registry documentation". Its history reflects that it was always provisional: added in #3797, reverted a week later in #3825 "pending AMS GA", then re-added in #3848 behind the env var for a private beta, with nothing since.

It is being replaced by a resource generated from the Integration Accounts v2 spec, which targets `/api/v2/integration-interfaces/databricks/accounts` rather than the legacy `/api/v2/web-integrations/databricks/accounts` endpoint this one uses.

The replacement cannot land while this file exists. `tfgen` writes each resource to `datadog/fwprovider/resource_datadog_<artifact_name>.go`, and `WriteArtifactSource` refuses a path already occupied by a source carrying no generated marker:

```go
if existing != nil && !authorizedTakeover && !IsGeneratedSource(existing) {
	return model.ArtifactStatusFailed, &HandwrittenCollisionError{Path: path}
}
```

That failure is quiet in the spec pipeline: `failed` is not one of the statuses summed into the "changed artifacts" total, so the generation job reports `0 changed artifacts`, pushes no branch, and the check passes green while producing nothing.

The alternative to deleting it would be the `overwrites` annotation, which authorizes an in-place takeover. That doesn't apply here: `RemoveHandwrittenResource` retires the superseded constructor from the plain `Resources` slice, and this one isn't in that slice — it's in the env-gated block — so the wiring step would fail with `overwrites target ... not found in the framework Resources slice`. Deleting the never-released resource is both simpler and more honest about what happened.

### User impact

None for the released provider. The resource is unreachable unless an operator sets an undocumented env var, and it appears in no registry documentation. Only private-beta participants who opted in are affected, and for them the generated resource is the intended replacement.

### What's removed

| File | Change |
|---|---|
| `datadog/fwprovider/resource_datadog_integration_databricks_account.go` | deleted |
| `datadog/tests/resource_datadog_integration_databricks_account_test.go` | deleted |
| `datadog/tests/cassettes/TestAccIntegrationDatabricksAccount{OAuth,Pat}.{yaml,freeze}` | deleted |
| `datadog/fwprovider/framework_provider.go` | gated registration dropped; capacity hint `+2` → `+1`; stale comment corrected |
| `datadog/internal/utils/utils.go` | `IsDatabricksIntegrationEnabled` removed |
| `datadog/tests/main_test.go` | env var no longer exported for the suite |
| `datadog/tests/provider_test.go` | `testFiles2EndpointTags` entry removed |
| `datadog/tests/framework_provider_test.go` | web-integrations unstable operations removed |
| `.github/CODEOWNERS`, `.github/workflows/test.yml` | entry and CI env var removed |
| `DEVELOPMENT.md` | "Env-gated resources" section removed — Databricks was its only entry |

Two judgement calls worth a reviewer's attention. The five `v2.*WebIntegrationAccount*` unstable operations are dropped from both `framework_provider.go` and `framework_provider_test.go`, because this resource was their only caller — `GetWebIntegrationsApiV2` is otherwise referenced solely by the generic accessor helper. And the whole "Env-gated resources" section of `DEVELOPMENT.md` goes rather than just its table row, since Databricks was the only gated resource and the section was introduced alongside it in #3848. Happy to keep either if you'd rather preserve the documented pattern.

`CHANGELOG.md` is left untouched, as its entries are historical release notes.

### Verification

`gofmt` is clean across `datadog/`. I could not compile locally: this module needs Go 1.25.8 with the `go1.26.1` toolchain, and the toolchain download goes through `binaries.ddbuild.io`. Instead I checked statically that nothing dangles — no remaining references to `NewIntegrationDatabricksAccountResource`, `IsDatabricksIntegrationEnabled`, or `DD_TERRAFORM_DATABRICKS_INTEGRATION_ENABLED` anywhere outside `CHANGELOG.md` history; every symbol the deleted test defined was local to it; and both `utils` in `framework_provider.go` and `os` in `main_test.go` keep other callers, so no import is orphaned. CI compiles and runs the suite.


guttymora (approved):

DEVELOPMENT.md
```diff
@@ -27,22 +27,6 @@ For generated resources, only an OpenAPI property with `writeOnly: true` uses Te
 
 Existing hand-written resources can use the compatibility-preserving helpers in [`datadog/internal/fwutils`](./datadog/internal/fwutils/README.md#write-only-secret-helpers-writeonly_helpersgo). Their existing three-attribute pattern (`<attr>`, `<attr>_wo`, `<attr>_wo_version`) is retained for already-shipped schemas; do not use that legacy shape for new generated resources.
 
-## Env-gated resources
```
fpighi, line 30: We should keep this section and the mechanism in place despite removing the resource.

rossicris: I had removed that because this was single introduced by our original PR - https://github.com/DataDog/terraform-provider-datadog/pull/3848/changes#diff-fc2a310fcfedfefb0046def697f932def80cbb1e78c689bc0af3c8eab3e33ece

Do you want to keep this section just because customers might be on a legacy tf provider client version?

fpighi: I had a second look at that code, and it's ok to remove it. There's nothing overly generic in there and we can do the same trick in the future if needed.

fpighi (approved):

